Issue #4, opened 2017-05-19
[High Prio] New Blocks


greshake:
Ordered by importance:

- [ ]  NetworkManager with Dbus
- [x] Brightness
- [x]  Temperature
- [x]  Sound
- [x]  Scriptable Block to execute shell commands etc. (work in progress)
- [x]  Pacman updates
- [x]  Disk Space (work in progress)
- [x]  CPU utilization
- [x]  Load
- [x]  Memory
- [x]  Battery
- [x]  Network 

Any more ideas and custom modules are appreciated!

shioju:
Hey I'm pretty new to rust and i3, but I'd like to help. Will take a look at your code first.

XYunknown:
Sure, I'd love you to! If you have questions, just ask here.

NullVoxPopuli:
This may help for temperature:

```
nvidia-smi --query-gpu=temperature.gpu --format=csv,noheader
```

not sure if there is a catchall for any gpu/driver, though.

greshake:
That command is very specific and depends on nvidia-smi, I'd  favor any approach to read the temperatures directly from the kernel, or using a program that is readily available and works for more components than a specific GPU.

greshake:
This would be my go-to approach: https://wiki.archlinux.org/index.php/lm_sensors

svmnotn:
On the network one, I just made a quick one using speedtest-cli that show ping, upload, and download. It still needs some work to make it nice and error free enough to use but that might be a thing. (I didn't know what else to use to measure all 3)

greshake:
@svmnotn Hey, what happened to the block you mentioned?

svmnotn:
it seems to be working fine, i'll add a few config options and then submit a pr.

svmnotn:
any thoughts on what that last module left should look like / do? or any off the top of the head knowledge about the needed docs to make it? Probably useful to put here ;)

greshake:
So, I was imagining some block that displays current network connectivity (e.g. Ethernet/Wifi) and updates through dbus signaling instead of periodic updates... But I'm not sure I cheecked if NetworkManager actually provides those signals.

etrombly:
looks like you'd probably want to monitor the active connection https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.Connection.Active.html

etrombly:
[...]

Issue #11, opened 2017-05-19
Memory block doesn't filter click events


greshake:
The click function is called on **every** click on any block. The memory module needs to assign a name/id to the displaying widget and match on that id when the event comes in. Otherwise, it doesn't matter where you click and the memory block switches states.

robojeb:
I didn't know if I should file another issue but it looks like the Memory block also remains clickable even if you specify that clickable should be false.

XYunknown:
@jheyens promised to fix the memory block today, I'll ensure the clickable thing is fixed too

jheyens:
Fixed in https://github.com/greshake/i3status-rust/commit/b849e72c00d76250ade8194d133d9d9c82011b2d

PR #13, opened 2017-05-20
Add custom scripts block

ftriquet:
Adds support for custom scripts blocks in the bar.
Part of #4 

Features that needs to be added:
- [x] Use a JSON array instead of string for the command fore more flexibility
- [x] Handle click events (call the command with a special env variable?)
- [ ] Block state
- [ ] Custom colors


greshake:
Thanks for your help. 
1. We should support entering a shell command directly into the config, instead of requiring a separate script file
2. Please remove the absolute path with your home directory from the example.conf
3. Remove the example shell script, because it doesn't really add much use :)
4. Clicks could be supported by different modes:
- OnClick- Enter a command in the config to be executed when the block is clicked, maybe with modifiers to specify mouse buttons
- Cycle- Allow for an array of different output commands, and clicking cycles through them

ftriquet:
I added support for string array and strings. If the command is an array, the first element will be used as the command and the rest of the array as arguments.
If the command is a string it will be split on whitespaces.

ftriquet:
I wonder what should be the format of the config for click events.
Maybe we could accept 2:
```json
{
  "block": "script",
  "command": ["uname", "-s"],
  "onClick": ["uname", "-a"]
}
```
or
```json
{
  "block": "script",
  "cycle": [["uname", "-s"], ["uname", "-a"]]
}
```
What do you think?

greshake:
Couple more things: We should just use string commands and forwarrd them to a shell environment, to allow things like pipes, e.g. 
```
{
  "block": "script",
  "command": "cat /proc/cpuinfo | grep whatever"
}
```
The rust code could execute them by calling sh -c '<command>'

2. Should we rename the block "Custom" instead of "Script"?

greshake:
About the click events: The format you suggested is exactly what I imagined, except we should stick to the naming conventions (Rust-style) and call onClick on_click. Just make "cycle" and "on_click" mutually exclusive

ftriquet:
Yes it sounds nice.
I agree about the block's name.

Should the "cycle" key complete "command" or make it obsolete?
Maybe we could avoid the "cycle" key and consider an array of strings provided in "command" as the cycle.

greshake:
Like the idea, but then people might not even realize the feature exists.... I think I prefer making "command" and "cycle" mutually exclusive.

greshake:
Oh, and what do you mean by 'Variables expansions in the command'?

ftriquet:
> what do you mean by 'Variables expansions in the command'?
I mean expand environments variables to allow this kinds of input:
`{"block": "custom", "command": "$HOME/bin/script.sh"}`
But it should not be necessary since we will use sh for execution

greshake:
Also, we should add some functionality to control the state of the block, which also defines it's color based on the theme (State are Critical, Warning, Info, etc.). Custom color could also be a feature, only need to add it to the text and button widgets. Currently you can only set a state and the color is determined based on state and theme.

ftriquet:
Text widgets don't have name for now so I can't check if an event is coming from the script widget.
Should I assume that this will be fixed somewhere else (#11) or should I add name to text widgets?

greshake:
Just use a ButtonWidget. It is identical except it takes an ID in the constructor. The widget system is kind of a big clusterfuck right now, but it'll be cleaned up soon. It's supposed to abstract away from i3 specific details of the view.

ftriquet:
Alright thanks

ftriquet:
What do you think about on_click? It should replace the main command until the next update or until the next click on the button?

greshake:
I think it should do neither, the output of on_click shouldn't be used. Just trigger another update of the usual output command after executing on_click. 

Example: I want to create a toggle button for whatever. 
```javascript
{"block": "custom", "cycle": ["echo 'ON'", "echo 'OFF'"], "on_click": "toggle stuff here"}
```

ftriquet:
Ok it makes sense. I thought you opposed "command" + "on_click" vs "cycle".

greshake:
Maybe we should add the option to use static text instead of commands. Then we don't have to spawn a shell just to get the ON/OFF text. Something like a prepended string could be used, e.g. 
```javascript
{"block": "custom", "cycle": ["TXT: ON", "TXT: OFF"], "on_click": "toggle stuff here"}
```
and use "content" instead of "command":
```javascript
{"block": "custom", "content": "CMD: cat /proc/cpuinfo | grep something", "on_click": "something"}
```

greshake:
Like your work, will merge as soon as I have some time to test... What do you think about my last suggestion?

ftriquet:
Yeah you're right. It's done

src/blocks/custom.rs
```diff
@@ -0,0 +1,117 @@
...
+    output: ButtonWidget,
+    command: Option<String>,
+    on_click: Option<String>,
+    cycle: Option<Peekable<Cycle<vec::IntoIter<String>>>>,
+    tx_update_request: Sender<Task>,
+}
+
+impl Custom {
+    pub fn new(config: Value, tx: Sender<Task>, theme: Value) -> Custom {
+        let mut custom = Custom {
```
JonasCir: I think using  [Field Init Shorthand](https://doc.rust-lang.org/nightly/book/second-edition/appendix-07-newest-features.html) would be better here (stable since 1.17). No need to make custom mutable for initialization.

src/blocks/custom.rs
```diff
@@ -0,0 +1,117 @@
...
+            update_interval: Duration::new(get_u64_default!(config, "interval", 10), 0),
+            output: ButtonWidget::new(theme.clone(), ""),
+            command: None,
+            on_click: None,
+            cycle: None,
+            tx_update_request: tx,
+        };
+        custom.output = ButtonWidget::new(theme, &custom.id);
+
+        if let Some(on_click) = config["on_click"].as_str() {
```
JonasCir: I think is a little bit cleaner:
```rust
let on_click = if let Some(on_click) = config["on_click"].as_str() {
          Some(on_click.to_string())  
      } else {
          None 
      };
....
Custom{
....
on_click: on_click
....
}

src/blocks/custom.rs
JonasCir: Just initialize Custom directly. See comment below.

JonasCir:
@ftriquet: Thanks for the PR. See inline comments for more details. Otherwise lgtm.

PR #14, opened 2017-05-20
Update main.rs

tjikini:
Add newline to print
Dont work in swaybar without it


greshake:
Weird, this should work as to i3bar protocol specification. I'll let this PR through, but you should PR sway to fix the bug in swaybar.

PR #15, opened 2017-05-20
Pacman Block

keeslinp:
I wrote up a block that checks the pending pacman updates. I'm not sure experienced in Rust so I'm hoping that this is all idiomatic. If there is something that should be different or could be done better please let me know :). I'm just trying to get a stronger handle on things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,17 @@ Key | Values | Required | Default
 interval | Update interval in seconds | No | 10
 device | Which BAT device in /sys/class/power_supply/ to read from. | No | 0
 
+## Pacman
+Creates a block which displays the pending updates available on pacman.
+
+**Example**
+```javascript
+{"block": "pacman"},
+```
+
+**Options**
+There are no options available yet. If you need a specific option, file an issue.
+
 # How to write a Block
 
 ## Step 1: Create the file
--- a/README.md
+++ b/README.md
@@ -162,11 +162,14 @@ Creates a block which displays the pending updates available on pacman.
 
 **Example**
 ```javascript
-{"block": "pacman"},
+{"block": "pacman", "interval": 10},
 ```
 
 **Options**
-There are no options available yet. If you need a specific option, file an issue.
+
+Key | Values | Required | Default
+----|--------|----------|--------
+interval | Update interval in seconds | No | 1
 
 # How to write a Block
 
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -13,8 +13,6 @@ pub struct Pacman {
     text: TextWidget,
     id: String,
     update_interval: Duration,
-
-    theme: Value,
 }
 
 impl Pacman {
@@ -24,8 +22,6 @@ impl Pacman {
                 id: Uuid::new_v4().simple().to_string(),
                 update_interval: Duration::new(get_u64_default!(config, "interval", 5), 0),
                 text: TextWidget::new(theme.clone()).with_text(""),
-                tx_update_request: tx,
-                theme: theme,
             }
         }
         
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -1,0 +1,58 @@
+use std::time::Duration;
+use std::sync::mpsc::Sender;
+use std::process::Command;
+
+use block::Block;
+use widgets::text::TextWidget;
+use widget::I3BarWidget;
+use input::I3barEvent;
+use scheduler::Task;
+
+use serde_json::Value;
+use uuid::Uuid;
+
+
+pub struct Pacman {
+    text: TextWidget,
+    id: String,
+    update_interval: Duration,
+
+    //useful, but optional
+    #[allow(dead_code)]
+    theme: Value,
+    #[allow(dead_code)]
+    tx_update_request: Sender<Task>,
+}
+
+impl Pacman {
+    pub fn new(config: Value, tx: Sender<Task>, theme: Value) -> Pacman {
+        {
+            Pacman {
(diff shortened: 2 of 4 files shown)


src/blocks/pacman.rs
```diff
@@ -0,0 +1,53 @@
...
+    id: String,
+    update_interval: Duration,
+}
+
+impl Pacman {
+    pub fn new(config: Value, theme: Value) -> Pacman {
+        {
+            Pacman {
+                id: Uuid::new_v4().simple().to_string(),
+                update_interval: Duration::new(get_u64_default!(config, "interval", 5), 0),
```
greshake: Retrieving pacman updates is a quite resource intensive task, so put the default interval to something more sensible, like every 10 minutes.

src/blocks/pacman.rs
```diff
@@ -0,0 +1,53 @@
...
+                text: TextWidget::new(theme.clone()).with_text(""),
```
greshake: If you don't add text right away, just omit the .with_text. Also, you should add an Icon :) Add it from the awesome fonts icon set to the icons.rs file, don't forget to put an empty default value into the no icons JSON, for people that don't plan to use an icon font. Then use your icon by calling .with_icon("<name of your icon>")

src/blocks/pacman.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+impl Block for Pacman
+{
+    fn update(&mut self) -> Option<Duration> {
+        let output = String::from_utf8(
+            Command::new("sh")
+                .arg("-c")
+                .arg("pacman -Sup 2>/dev/null")
```
greshake: The command you use to retrieve number of updates requires root, that's a No-go. But there is a way! Look at the guys from bumblebee-status: https://github.com/tobi-wan-kenobi/bumblebee-status/blob/master/bumblebee/modules/pacman.py

They use a bash script and fakeroot:
#!/usr/bin/bash

if ! type -P fakeroot >/dev/null; then
	error 'Cannot find the fakeroot binary.'
	exit 1
fi

if [[ -z $CHECKUPDATES_DB ]]; then
	CHECKUPDATES_DB="${TMPDIR:-/tmp}/checkup-db-${USER}/"
fi

trap 'rm -f $CHECKUPDATES_DB/db.lck' INT TERM EXIT

DBPath="${DBPath:-/var/lib/pacman/}"
eval $(awk -F' *= *' '$1 ~ /DBPath/ { print $1 "=" $2 }' /etc/pacman.conf)

mkdir -p "$CHECKUPDATES_DB"
ln -s "${DBPath}/local" "$CHECKUPDATES_DB" &> /dev/null
fakeroot -- pacman -Sy --dbpath "$CHECKUPDATES_DB" --logfile /dev/null &> /dev/null
fakeroot pacman -Su -p --dbpath "$CHECKUPDATES_DB"

exit 0

src/blocks/pacman.rs
```diff
@@ -0,0 +1,53 @@
...
+                .output()
+                .expect("You need to have pacman set up")
+                .stdout
+            ).expect("Something is wrong with the pacman output.");
+        let count = output.lines().count() - 1;
+        self.text.set_text(format!("{}", count));
+        self.text.set_icon("update");
```
greshake: Oh yea, I see the icon now :D Working through your code from top to bottom here. Don't call set_icon, use with_icon above once.

greshake:
Oh, and document your block in the readme. Cheers!

src/blocks/pacman.rs
keeslinp: Hmm, that's weird. It doesn't prompt me for sudo like it does for a regular pacman call. I think that if you use the p flag on there it doesn't require root. When I run `pacman -Su` is does tell me  I need to be root but when I do `pacman -Sup` it doesn't complain. I can implement the other way if you prefer that though.

src/blocks/pacman.rs
keeslinp: Sounds good :).

src/blocks/pacman.rs
keeslinp: Should I make that a parameter for the config? Or is 10 minutes a safe bet for all cases?

src/blocks/pacman.rs
greshake: Weird, I thought it ougth to work but when I checked the command wanted root, maybe I switched soemthing up. Anyway, seems to work now, so I guess all good :)

src/blocks/pacman.rs
greshake: This macro: get_u64_default!(config, "interval", 5) actuelly looks for "interval" in the config, and if not specified defaults to 5. The Macro just wraps some nicer error handling to avoid boilerplater, it's in utils.rs.

README.md
```diff
@@ -157,6 +157,17 @@ Key | Values | Required | Default
...
+## Pacman
+Creates a block which displays the pending updates available on pacman.
+
+**Example**
+```javascript
+{"block": "pacman"},
+```
+
+**Options**
+There are no options available yet. If you need a specific option, file an issue.
```
greshake: You actually already provide the "interval" option :)

keeslinp: Woops, you're totally correct! I'll fix that.

keeslinp:
It should all be updated and good to go now :).

greshake:
One more thing: you should set the state of your output widget, to colorize the block depending on the number of updates available. Look for example into cpu.rs, to see how to use states. Set something like idle for none, info for 0-30, warning for over 30 etc.. Or whatever thresholds you deem sensible.

PR #19, opened 2017-05-23
Fixing pacman package updates

### LINKED ISSUE #18: Pacman Block not updating package lists
Seems like pacman -Sup isn't enough after all, we need to use the method I suggested earlier to @keeslinp:
Look at the guys from bumblebee-status: https://github.com/tobi-wan-kenobi/bumblebee-status/blob/master/bumblebee/modules/pacman.py

They use a bash script and fakeroot:
```
#!/usr/bin/bash

if ! type -P fakeroot >/dev/null; then
error 'Cannot find the fakeroot binary.'
exit 1
fi

if [[ -z $CHECKUPDATES_DB ]]; then
CHECKUPDATES_DB="${TMPDIR:-/tmp}/checkup-db-${USER}/"
fi

trap 'rm -f $CHECKUPDATES_DB/db.lck' INT TERM EXIT

DBPath="${DBPath:-/var/lib/pacman/}"
eval $(awk -F' *= *' '$1 ~ /DBPath/ { print $1 "=" $2 }' /etc/pacman.conf)

mkdir -p "$CHECKUPDATES_DB"
ln -s "${DBPath}/local" "$CHECKUPDATES_DB" &> /dev/null
fakeroot -- pacman -Sy --dbpath "$CHECKUPDATES_DB" --logfile /dev/null &> /dev/null
fakeroot pacman -Su -p --dbpath "$CHECKUPDATES_DB"

exit 0
```

keeslinp:
This technically works but it is missing some of the functionality of the bash script. I wasn't sure of the proper way to attempt to grab the system variables before giving them a default value.

In javascript it would be something like:
`let db = get_sys_vars("CHECKUPDATES_DB") || default_val;`

but I'm not quite sure the idiomatic way to do that in rust.

This fixes #18


greshake:
https://doc.rust-lang.org/std/env/

On a side note:
Scary to see that when a dev thinks of getting an environment variable, he thinks of javascript. It's for the web, people! NodeJS is poisioning devs...

keeslinp:
Ah, so the best way to to always wrap it in an option if you don't know for sure that it's going to be the value you want? I guess that makes sense, having a `None` is much clearer than checking for `""`. 

Don't ya know, javascript is gonna take over the world? /s. I'm just a web developer at work so that's my first thought :). I would never try to write this in node haha.

greshake:
Yea, Option describes a *possible* value, Result describes the result cof a process which might fail. That's how Rust does error handling and gets around the whole null pointer thing.

Btw:
Unfortunately, people actually do write this kind of software in JS these days. I mean hell, they use electron! There is a tool which only shows your clipboard history (as an electron app) which ships with a full version of chromium which is loaded each time to execute 100 lines of javascript... Only because the web devs don't bother with anything else these days because computing power and storage are cheap, right? :D Could go on ranting for days. Anywhere, here are the replacements for i3status written for Node (only the ones I could dig up quickly):
https://github.com/stigok/statusbar-js
https://github.com/Bubujka/my-i3status
https://github.com/projectivemotion/i3status-taskwarrior (Only a single block, but that makes it even worse)
https://github.com/clehner/wpa_i3bar

Tadaaaa

greshake:
Tell me when you're ready to be merged

keeslinp:
I had a good laugh showing your comment to my boss. I'm always complaining about electron at work because it's a nightmare. So I really appreciate your comment! I'll try to get an MR done tonight but I dunno for sure.

greshake:
Thanks, maybe I should finally start a blog. But I think there is already enough ranting out there. Then again, maybe it helps me to mentally survive until webassembly eradicates most of this JS nonsense.

keeslinp:
I made the changes you mentioned. I definitely like it a lot better using the `Option` as opposed to just checking the string. A lot more reliable. Thanks for explaining that to me, I feel like I'm getting a much better handle on thinking in terms of Options and Results.

greshake:
Possibly merged prematurely.... Doesn't show my available updates, only 0!

greshake:
There is definitely something wrong here. It keeps my pacman db locked, either not removing db.lock or keeping the fakeroot going.... You definitely need to give it another look

keeslinp:
hmmm, if you want to revert the merge I can take a look at it tonight. For me it was displaying my updates correctly. I wonder if it has something to do with my fakeroot check. That would be the most plausible because that is supposed to return 0 if it fails.

Issue #21, opened 2017-05-24
Allow for themes and icons to be defined in a file


svmnotn:
This way we can avoid having to recompile every single time we want to test color or icon changes.
Changing the themes in a file seems more useful than the icons, but the code shouldn't be too different

greshake:
Alright, hit me with a pull request. I'm busy for the next week, but 'll help you review the PR..

greshake:
Fixed in https://github.com/greshake/i3status-rust/pull/22

PR #22, opened 2017-05-25
allow for theme and icon files

svmnotn:
addresses #21, however the implementation could be better.

For example we could have a struct for the icons and theme files which would allow serde to enforce that all the values have been filled out (or at least give us the option to provide sane defaults) 

However I made this PR be as minimal a change as possible. Whilst including the example files.


greshake:
We could do that, altough I think we should rather just document what a Theme has to look like. The user being bombarded with serde parsing errors won't help him very much... And because there aren't many things to adjust in the first place, we can probably get away with loosely typed parsing and documentation.

greshake:
My original motivation for not adding them files was to encourage people to submit their themes by forcing them to alter the source code. But I totally see that it's crap for debugging, especially while the rust compiler stll takes his time.

svmnotn:
If anything, the README can always say something about sending a PR with good looking themes/icons after making them.

PR #24, opened 2017-05-25
fix bar crash if a temperature sensor gives a negative value

svmnotn:
One of my sensors seems to not be working and gives -128 as the input.. that is obviously wrong but it crashes the program. So, I made it ignore any issues with the number parsing in the temperature. Maybe an error should be raised somewhere for awareness but it shouldn't kill everything.

### DIFF
--- a/src/blocks/temperature.rs
+++ b/src/blocks/temperature.rs
@@ -31,7 +31,7 @@ impl Temperature {
                 id,
             }
         }
-        
+
     }
 }
 
@@ -56,7 +56,9 @@ impl Block for Temperature
                     .collect::<Vec<_>>();
 
                 if rest[1].starts_with("input") {
-                    temperatures.push(rest[2].parse::<u64>().unwrap());
+                    if let Some(t) = rest[2].parse::<u64>().ok() {
+                        temperatures.push(t);
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


greshake:
I agree, but I'd like a different approach here, nothing major. Because in some cases negative temperatures could be valid (rarely, but still) we should parse the values as i64, ignoring other parsing errors the way your commit already does, and enforce a range of -100 to 150 °C, does that sound good?We should still ignore the errors, but write an error message to stderr, which shouldn't affect i3bar.

svmnotn:
ok, will do the changes

greshake:
All right, I'm just not sold on the println! macro that prints to stderr. On the one hand, it's good because noone can accidentally print stuff to stdout, but I don't want to have "magic" overrides that are non-obvious. Also, it might tempt future contributors to leave print statements in the release version where they are of no use. I suggest we add a log! macro, which uses conditional compilation to only compile with the debug profile and prints to stderr.

svmnotn:
That macro was simply something so that it would compile on stable since only nightly has `eprintln!` which is basically println to stderr, I probably should have just named it the same and then let the compiler crash once stable got the `eprintln!` macro. Does that seem like an ok change? otherwise we might want to change over to use something like the log crate.

greshake:
Yep, got it now. Really didn't notice the **e** just until now... Shows that eprintln really is a better name :D Just change that small thing in the online editor, merging. Thanks for your help!

PR #25, opened 2017-05-26
Weather block

Soft:
This commit includes a simple weather block for displaying outside
temperature. The data is provided by darksky.net and requires an API
key.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,3 +17,8 @@ nix = "0.8.1"
 # for profiling blocks
 cpuprofiler = "0.0.3"
 progress = "0.2"
+darksky = { version = "0.5.0", optional = true }
+
+[features]
+default = ["weather"]
+weather = ["darksky"]
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -1,0 +1,169 @@
+use darksky::{self, Unit, Icon};
+use serde_json::{self, Value};
+use std::boxed::Box;
+use std::error::Error;
+use std::sync::mpsc::Sender;
+use std::sync::{Arc, Mutex};
+use std::thread;
+use std::time::{Duration, Instant};
+use uuid::Uuid;
+
+use widget::I3BarWidget;
+use widgets::text::TextWidget;
+use scheduler::Task;
+use block::Block;
+
+const DEFAULT_UPDATE_INTERVAL: u64 = 30;
+const DEFAULT_UNIT: InputUnit = InputUnit::Si;
+
+pub struct Weather {
+    id: String,
+    widget: TextWidget,
+    config: Config,
+    forecast: Arc<Mutex<Option<darksky::Forecast>>>
+}
+
+#[derive(Clone, Deserialize)]
+enum InputUnit {
+    #[serde(rename = "SI")]
+    Si,
+    #[serde(rename = "US")]
+    Us
+}
+
+impl InputUnit {
+    fn to_str(&self) -> &'static str {
+        match self {
+            &InputUnit::Si => "℃",
+            &InputUnit::Us => "℉"
+        }
+    }
+}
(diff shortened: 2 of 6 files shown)


src/icons.rs
```diff
@@ -37,12 +37,29 @@ fn no_icons() -> Value {
...
+
+    if cfg!(feature = "weather") {
+        let mut icons = icons.as_object_mut().unwrap();
+        let weather = [
+            ("weather_sun", " OUT "),
+            ("weather_snow", " OUT "),
+            ("weather_thunder", " OUT "),
+            ("weather_clouds", " OUT "),
+            ("weather_rain", " OUT "),
+            ("weather_default", " OUT ")
```
greshake: Are you sure about the OUT description? Could just output Sun, Snow, Thunder, etc in no_icons mode.

Cargo.toml
```diff
@@ -12,4 +12,9 @@ clap = "2.23.2"
 uuid = { version = "0.4", features = ["v4"] }
 dbus = "0.5.2"
 regex = "0.2"
-nix = "0.8.1"
\ No newline at end of file
+nix = "0.8.1"
+darksky = { version = "0.5.0", optional = true }
```
greshake: I think it's fine to pull this dependency without conditional compilation. If weather isn't used it ends up only as a few extra MB on the output binary, not impacting performance. If there is another reason for having this as an optional dependency, tell me and you should document in the README how to disable the weather feature when compiling.

src/blocks/weather.rs
```diff
@@ -0,0 +1,169 @@
...
+                    darksky::get_forecast_with_options(config.token.clone(),
+                                                       config.latitude,
+                                                       config.longitude,
+                                                       set_units) {
+                    let mut forecast = forecast.lock().unwrap();
+                    *forecast = Some(data);
+                    send.send(Task {
+                        id: id.clone(),
+                        update_time: Instant::now()
+                    }).unwrap();
```
greshake: I agree this should be done in a separate thread. Usually however I would discourage this pattern if the update() doesn't download information from the internet, because the scheduler can then group updates of blocks that have the same intervals, that way avaoiding redrawing the bar twice in a very short period of time.

greshake:
Offtopic: Did you actually write this block in the 6 minutes between your fork and your PR? :D

greshake:
I looked at the terms of service of darksky and unfortunately I can't accept them.
https://darksky.net/dev/docs/terms#attribution

They require a text like 'powered by darksky' possibly as a link. I'm not willing to incorporate something like that, and I don't think it's possible with i3bar. Please just switch to a free weather API :)

Soft:
I agree that the attribution requirement is a little problematic considering the nature of the user interface. As we do not provide the API keys and include "powered by darksky.net" string in the readme, I think that any user would have to be aware where the data is coming from in order to setup the block. However, I am not a lawyer and I can understand how one could view that as not enough to comply.

About the display format, I agree that the OUT strings could be changed to something more meaningful.

greshake:
I think they are very explicit in their demands, and I wouldn't see a problem in any other UI environment. But in this case we are clearly not able to comply, so darksky is just not a suitable provider in our case. If you are willing to publish your block nontheless (I don't think there is much risk of repercussions) you need to do it in your fork or change the provider. I am not very happy about this situation but darksky just doesn't work for us :/

oschoudhury:
How about OpenWeatherMap? From a quick glance, they seem to provide a similar json api interface, which means that most of the code can be re-used. Their attribution is similar without the explicit clickable logo, and it doesn't tell that you need to place it next to their weather information, which I interpret that putting it in the README is fine (http://openweathermap.org/terms).

atheriel:
I think this can probably be closed, given the block introduced in #113.

PR #28, opened 2017-05-27
Fix calculation of CPU usage for the case where it is 100%.

adeschamps:
In that case, the idle time will not increase between samples, so
using && would result in displaying 0% CPU usage.


greshake:
Thanks! I only changed it from || to <= :)

PR #33, opened 2017-05-30
xrandr block

Drogglbecher:
Hi, I created a block for screen information based on xrandr showing name, brightness and resolution. With a click you can toggle through the active screens. At the moment it maybe does not make that much sense but for the future (when there is the possibility to react on scrolling) then one can scroll up and down to set the screens brightness. I wrote such an module for python i3-bar once and I found it very helpful ;)


src/blocks/xrandr.rs
```diff
@@ -0,0 +1,191 @@
...
+                    }
+                }
+            }
+        }
+        Some(self.update_interval.clone())
+    }
+    fn view(&self) -> Vec<&I3BarWidget> {
+        vec![&self.text]
+    }
+    fn click(&mut self, _: &I3barEvent) {
```
greshake: You currently need to use a ButtonWidget instead of a TextWidget, use the block ID as name and match on your blocks ID.... Otherwise you trigger on every Click on any Block. Look at the other blocks for more info. Will be cleaned and refactored in the future.

greshake:
Not really sure what is going on, but there is a huge delay between clicking and the switching... Also, thanks for your help!

src/blocks/xrandr.rs
Drogglbecher: Ah, alright, changed it ^^

Drogglbecher:
The duration was caused by the display got reload after the `update_interval` and not immediately. Fixed it now :)

greshake:
Alright, I am going to merge now, any changes you want to make? Otherwise you will need another PR

Drogglbecher:
I think some minor code styling things. I will change tomorrow and then you can merge. I think this won't be my last PR here. Tanks for the quick reactions and the cool project 😎

Drogglbecher:
So, ready to merge. Thanks 👍

greshake:
Good job, looking forward to the next PR :)

PR #35, opened 2017-06-02
Fix some minor typos

michelboaventura:
Just some small typos on the doc. @greshake, what is your opinion about break all the lines at every 80 columns? It is hard to edit text with a single line and at every word you change git can't help you showing only the word.

Cheers.


greshake:
Sure. I have a big cleanup planned anyway, will use rustfmt then, that should also limit line width.

PR #36, opened 2017-06-03
Make click and scroll events more user-friendly

Drogglbecher:
Since checking for the number of event.button was necessary to differentiate between different kinds of click and scroll actions and these were only documented inside a code-comment I think it would be good to provide a more specific way to react on different mouse actions :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,7 +389,11 @@ fn id(&self) -> &str {
 ```
 
 
-### `fn click(&mut self, event: &I3barEvent)` (Optional)
+### `fn click_left(&mut self, event: &I3barEvent)` (Optional)
+### `fn click_center(&mut self, event: &I3barEvent)` (Optional)
+### `fn click_right(&mut self, event: &I3barEvent)` (Optional)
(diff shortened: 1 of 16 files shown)


greshake:
Hm.... I'm not sure this is a good change. I was going to deprecate the on_click function anyway. I plan to replace the different widgets with only one kind of widget, which is going to provide an on_click() event handler. That way, there will be no more need for filtering by name. Secondly the changes you made introduce a lot of redundancy, considering all te function args are the same, the functions only differ in context, which is generally not a great thing. If you want to make this issue more easy to tackle, you could introduce an enum like this, replacing the integer for buttonclicks with an instance of this enum:
```rust
pub enum MouseButton {
    Left, Right, Middle, WheelUp, WheelDown
}

pub struct I3barEvent {
    ...
    pub button: MouseButton,
}
```

Only thing is you need to define a custom serde deserializer for MouseButton from an u64.

Drogglbecher:
Yeah, you're right. The enum is captured with #39, thus this can be declined and closed I think ;)

PR #41, opened 2017-06-06
Make consolidated configuration

pitkley:
This consolidates the potentially three configurations into one JSON file, which in this proposal look like follows:

```json
{
    "blocks": {
         "disk_space": {"path": "/", "alias": "/", "type": "available", "unit": "GB", "interval": 20},
         "memory": {"type":"memory", "format_mem":"{Mup}%", "format_swap":"{SUp}%"},
         "cpu": {"interval": 1},
         "load": {"interval": 1, "format": "{1m}"},
         "time": {"interval": 60, "format": "%a %d/%m %R"}
    },
    "theme": "solarized-dark",
    "icons": "awesome"
}
```

The keys `theme` and `icons` can both be full mappings, see [`example_theme.json`](https://github.com/pitkley/i3status-rust/blob/9a7501b15f377dd8d84918b5d07c78881eeb6001/example_theme.json) and [`example_icon.json`](https://github.com/pitkley/i3status-rust/blob/9a7501b15f377dd8d84918b5d07c78881eeb6001/example_icon.json).

The command-line parameters `--theme` and `--icons` have been removed (although they could be added back with a bit of effort if desired).

-----

Even though this branch works, I still very much see it as a work-in-progress and hope to get some feedback and have some discussions about this and potential further design-changes in `i3status`.

Even if this PR does not fit as-is, I personally am very much in favour of having one consolidated configuration backed by Serde (which gives us so much for free, e.g. extremely flexible deserialization while retaining superb errors, both at build and runtime). (I would also love if the block configurations would be `struct`s, but I have not thought of a design yet.)


greshake:
I'm on board with this change, altough I need to properly test and review it first. Also, because this is a breaking change for existing setups, please update the README and add a special section explaining how to convert from the old version.

pitkley:
Will do 👍

pitkley:
I have updated the README to reflect the changes and added a section on breaking changes, which we can extend as they occur. LMK if this is in line with what you were thinking of.

svmnotn:
@pitkley hey, I just quickly looked through the code, it looks like this would allow for icons to be deserialized within the `theme` block of json. Have you checked if there would be an issue if the icons are declared inside the theme but not outside of it?

Other than that, :+1: I also thought of having the theme/blocks/icons be made from a struct as that would make some things with regards to serde much nicer to work with. :D

pitkley:
At least the way the deserialization is set up right now, I can not inline `icons` into `theme`, since `theme` can deserialize both as a single string or as a JSON-object. While including `icons` in the JSON-object would probably be easy, I don't see how one could deserialize a single string into a theme with icons (at least as long as icons are a separate entity from a theme, which I understand they should be).

greshake:
So, I'm doing my review and share my thoughts as I go through. 
First, instead of choosing between a fixed set of icons/themes or copy-pasting any theme and modifiying it, we should always take one theme/icon set as "base" and let the user override anything from the config, e.g. instead of having to do this to change one icon:
```json
{
    "blocks": {
         "disk_space": {"path": "/", "alias": "/", "type": "available", "unit": "GB", "interval": 20},
         "memory": {"type":"memory", "format_mem":"{Mup}%", "format_swap":"{SUp}%"},
         "cpu": {"interval": 1},
         "load": {"interval": 1, "format": "{1m}"},
         "time": {"interval": 60, "format": "%a %d/%m %R"}
    },
    "theme": "solarized-dark",
    "icons": {
        "": "",
        "time": " \uf017 ",
        "music": " \uf001 ",
        "music_play": "  \uf04b  ",
        "music_pause": "  \uf04c  ",
        "music_next": " \uf061 ",
        "music_prev": " \uf060 ",
        "cogs": " \uf085 ",
        "memory_mem": " \uf2db ",
        "memory_swap": " \uf0a0 ",
        "cpu": " \uf0e4 ",
        "bat": " \uf242 ",
        "bat_full": " \uf240 ",
        "bat_charging": " \uf1e6 ",
        "bat_discharging": " \uf242 ",
        "update": " \uf062 ",
        "toggle_off": " \uf204 ",
        "toggle_on": " \uf205 ",
        "volume_full": " \uf028 ",
        "volume_half": " \uf027 ",
        "volume_empty": " \uf026 ",
        "volume_muted": "\uf00d",
        "thermometer": " \uf2c8 "
    }
}
```
Allow this:
```json
{
    "blocks": {
         "disk_space": {"path": "/", "alias": "/", "type": "available", "unit": "GB", "interval": 20},
         "memory": {"type":"memory", "format_mem":"{Mup}%", "format_swap":"{SUp}%"},
         "cpu": {"interval": 1},
         "load": {"interval": 1, "format": "{1m}"},
         "time": {"interval": 60, "format": "%a %d/%m %R"}
    },
    "theme": "solarized-dark",
    "icons": ["awesome", {
        "time": " \uf017 "
    }]
}
```
So, instead of expecting a string or a map, expect either:
- A single string which specifies a fixed theme/set
- An array which has a "base" theme at index 0, and a map of "overrides" at index 1

greshake:
Also, be prepared for some commits by me today which include refactorings. Could cause a painstaking merge conflict ;) I'll try to stay away from your changes though

pitkley:
I have updated the PR to include the latest suggestion of @greshake to allow simple overriding of theme and icon related values. The following is now legal:

```json
{
    "blocks": {
         "disk_space": {"path": "/", "alias": "/", "type": "available", "unit": "GB", "interval": 20},
         "memory": {"type":"memory", "format_mem":"{Mup}%", "format_swap":"{SUp}%"},
         "cpu": {"interval": 1},
         "load": {"interval": 1, "format": "{1m}"},
         "time": {"interval": 60, "format": "%a %d/%m %R"}
    },
    "theme": ["solarized-dark", {
        "idle_bg": "#123456",
        "idle_fg": "#abcdef"
    }],
    "icons": ["awesome", {
        "bat": " | | ",
        "bat_full": " |X| ",
        "bat_charging": " |^| ",
        "bat_discharging": " |v| "
    }]
}
```

In fact, the `theme` and `icons` lists are not limited to two items, and every item can either be a simple string referencing a predefined layout or an object setting and potentially overriding previous values. This means the following is legal, too, and identical to the above:

```json
...
    "theme": [
        "solarized-dark",
        { "idle_bg": "#123456" },
        { "idle_fg": "#abcdef" }
    ],
    "icons": [
        "awesome",
        { "bat": " | | ", "bat_full": " |X| " },
        { "bat_charging": " |^| ", "bat_discharging": " |v| "}
    ]
...
```

Multiple predefined layouts can be defined as well, in case we ever see the need for layouts that extend each other:

```json
...
    "theme": ["base-theme", "extending-theme"],
    "icons": ["icon-set-1", "icon-set-2"]
...
```

### Unknown fields

* If a user defines unknown/unused fields in their configuration for the **`icons`** key, the key will be added with the value the user associated with it.

    This will have almost no effect during runtime.

* If a user defines unknown fields in their configuration for the **`theme`** key, the key will currently be ignored, since it cannot be deserialized into the `struct Theme`.

    This behaviour can be changed to output an error message stating that the configuration failed to parse because of unknown fields.

    **Pro denying unknown fields:**

    * Users will be notified of a potentially mistyped configuration key.

    **Con denying unknown fields:**

    * The user will have the burden of correcting his configuration, potentially not knowing what the correct field-name is.

-----

Except for this last discussion point (should unknown fields be denied?) I think that this PR should be ready to merge 👍.

greshake:
I definitely think they should lead to a fatal error... Otherwise it might mislead the user thinking his input was correct, and assume an implementation error. Otherwise, great work!

pitkley:
Oh, I [didn't remove](https://github.com/greshake/i3status-rust/pull/41/files#diff-732935ba71423092461edc532c7a60ecR39) the deny anyway. 😄

If you don't have any additional points, this should be good to merge. Feel free to look at and merge #45 first, I'll rebase once that is done! :blush:

pitkley:
Rebased, CI is green. ([`nightly` fail](https://travis-ci.org/greshake/i3status-rust/jobs/240515410) is not due to this PR, but definitely should be looked at.)

greshake:
I just noticed a major leftover issue here as I tried porting my current config.. The `blocks` section must not be a map! There may be blocks used multiple times (e.g. I use two disk_space blocks for my home and root partitions) but the map currently only allows for one.

pitkley:
Oh wow, very good point. Will fix that tomorrow!

pitkley:
Updated, the following is now legal:

```json
...
    "blocks": {
        "disk_space": [
            {"path": "/", "alias": "/", "type": "available", "unit": "GB", "interval": 20},
            {"path": "/mnt/media", "alias": "/mnt/media", "type": "available", "unit": "GB", "interval": 20},
        ],
        "memory": {"type":"memory", "format_mem":"{Mup}%", "format_swap":"{SUp}%"},
        "cpu": {"interval": 1},
        "load": {"interval": 1, "format": "{1m}"},
        "time": {"interval": 60, "format": "%a %d/%m %R"}
    },
...
```

@greshake let me know if you are okay with still having a map, but the value being either a single JSON-object or a list of JSON-objects.

jheyens:
I would still suggest to have a list of JSON objects, making sure that we have a specific order of elements. Mappings tend to not have this, and Serde may or may not shuffle the blocks (and even may change its current behaviour in the future).

pitkley:
Oh, order is also a very good point. Let me look into that, but making that deterministic is definitely a must!

greshake:
What about yml? Should we just jumnp from json to yml completely while we're at it? I used json only because it was an obvious choice as I had to use it anyway for the i3bar protocol and didn't want to bother with a different deserialization lib... Also makes it much easier to configure by hand.. json is a very cumbersome format.

pitkley:
I have reintroduced deterministically ordered blocks, allowing for multiple of the same. I have tested it successfully with my configuration.

-----

Regarding switching input format: good idea. I was thinking about suggesting to at least support [Hjson](https://hjson.org/) to make it less tedious, but switching altogether could be nice too.

One thing I think is important if we think about switching, is that we get rid of `serde_json::Value` (or whatever the replacement for a target-format would be) such that we can switch the input-format without having to update the backend (potentially even supporting multiple input formats simultaneously).

This will probably require every block to have their configuration as a `struct`, which fully defines what a block expects, but I see that as a very good thing, since error handling about missing fields or just plain non-existant ones would come for free -- and it would get rid of all those `get_` macros.

I have already thought about doing this and how it might work in the deserialization process, but I'm not sure yet if it is (easily) possible.

In regards on how to continue overall: I think that the implementation as it is right now works, and I would want to use it as a basis to build e.g. the block-config-structs on top of it. If it turns out that that will not work (at least not without a lot of work), I'd then think about switching to a different format, and move from `serde_json::Value` to `serde_yaml::Value` or whatever the new format would be.

If we have to decide on one format to switch to, I feel like I have to at least bring TOML up. Since the Rust eco-system pretty much decided on it, we should probably evaluate it too.

## Potential TOML configuration

```toml
theme = "solarized-dark"
icons = "awesome"

[[blocks]]
block = "sound"

[[blocks]]
block = "disk_space"
path = "/"
alias = "/"
type = "available"
unit = "GB"
interval = 20

[[blocks]]
block = "disk_space"
path = "/mnt/media"
alias = "/mnt/media"
type = "available"
unit = "GB"
interval = 20
```

Potential cause for issue: `theme` and `icons` have to come before the first `[[blocks]]` block so that they don't get matched to a preceeding block. Alternatively they could be seperated down into a separate category, e.g. `[general]` which can come whenever:

```toml
[[blocks]]
...

[general]
theme = "solarized-dark"
icons = "awesome"

[[blocks]]
...
```

If somebody could provide what a YAML configuration could look like I'd appreciate it, I'm not that firm with YAML.

svmnotn:
Definitly a :+1: on using toml, for the theme/icons we could just have them as a separate thing , like:
```toml
[theme]
name = "solarised-dark"
info-fg = "#00ac5a"
[icons]
name = "awesome"
# blocks
```
or something like that so that we can still use the whole override certain bit if we want thing, maybe allow for name to be missing and just take the default. Pehaps taking the default if the whole block is missing.

greshake:
Looked deeper into the TOML/YAML controversy and I agree TOML seems to be a good choice :)

pitkley:
Okay, I'll look into how exactly a TOML configuration would look like, such that it covers the features we have currently, and update this PR accordingly.

I was thinking about potentially splitting it up, i.e. merging this PR as it is and making the switch to TOML a separate step, but that could require users to update their configurations twice, so I think doing it in one go is probably preferable.

pitkley:
The following is a potential configuration I have thought about that should cover our use-case.

## The simplest form

Both `icons` and `theme` are simple strings, referencing predefined configurations. 

```toml
[general]
theme = "solarized-dark"
icons = "awesome"

[[blocks]]
block = "disk_space"
path = "/"
alias = "/"
type = "available"
unit = "GB"
interval = 20
...
```
## Extending a theme, icons

`icons` and `theme` can also be tables, allowing for overriding of themes and icons:

```toml
[general] # this line is not required

[general.theme]
name = "solarized-dark"
idle_bg = "#123456"
idle_fg = "#abcdef"

[general.icons]
name = "awesome"
bat = " | | "
bat_full = " |X| "
bat_charging = " |^| "
bat_discharging = " |v| "

[[blocks]]
block = "disk_space"
path = "/"
alias = "/"
type = "available"
unit = "GB"
interval = 20
...
```

If we want to support multiple themes/icon-sets, the following could additionally be supported too:

```toml
[general.themes]
name = ["base-theme", "extended-theme"]
...

[general.icons]
name = ["iconset-1", "iconset-2"]
...
```

-----

The `general` section can be dropped, but this requires that in the simplest case both `icons` and `theme` **have to** appear before any `blocks` definition. I don't know which is favourable: forcing a user to type `general`, or having them potentially confused as to why their configuration is not parsing.

So I guess there are three main questions:

1. Do those configuration files look okay to you?
2. Is everything covered? Did I miss some use- or edge-case?
3. Do we want the `general` section?

pitkley:
(Just to give an update on this, cc @greshake @svmnotn @jheyens.)

I have replaced the JSON configuration by a TOML configuration as shown in [my previous comment](https://github.com/greshake/i3status-rust/pull/41#issuecomment-307378637), and updated the README and examples to reflect this. See [`example_config.toml`](https://github.com/pitkley/i3status-rust/blob/config-struct/example_config.toml),
 [`example_theme.toml`](https://github.com/pitkley/i3status-rust/blob/config-struct/example_theme.toml) and [`example_icon.toml`](https://github.com/pitkley/i3status-rust/blob/config-struct/example_icon.toml).

The build failure is currently due to a missing `impl` in `toml-rs`, see [`toml-rs#188`](https://github.com/alexcrichton/toml-rs/pull/188). Once that PR is merged and released, CI should be green.

As for the questions I posed in the last question, they still stand. This implementation includes the `general` section, and fully works for my configuration, but some more input would be appreciated.

-----

Lastly, a little sneak peek. I have implemented block-configurations through structs: [`config-struct-blocks`](https://github.com/pitkley/i3status-rust/tree/config-struct-blocks). This allows all blocks to have simple, **type-safe** access to their configuration values. See [`MemoryConfig`](https://github.com/pitkley/i3status-rust/blob/0ec4c7b5e00260854fb99f65033929d81df31346/src/blocks/memory.rs#L214-L255) as an example.

(One example on how the type-safety can be pushed even further in the future using block-level config-structs is to move the enum-deserialization through to Serde, rather than [doing it manually](https://github.com/pitkley/i3status-rust/blob/0ec4c7b5e00260854fb99f65033929d81df31346/src/blocks/memory.rs#L394-L398)).

svmnotn:
## 3
I like the setup, my thought on the general section is that we could just make them [separate sections](https://github.com/greshake/i3status-rust/pull/41#issuecomment-307196915) but then if you don't modify the icons/themes then it is more to write than using `[general]` (a few characters but :man_shrugging:). However if we expect the main use case to change icons/themes then having them as separate sections would be less typing. So thinking of what @greshake has mentioned before, I think the use case desired would require less typing using the `[general]` setup instead of independent blocks.

If you can have it so that it supports either, so something like this:
```toml
[general]
theme = "solarized-dark"

[icons]
name = "awesome"
bat = " | | "
bat_full = " |X| "
bat_charging = " |^| "
bat_discharging = " |v| "

[[blocks]]
block = "disk_space"
path = "/"
alias = "/"
type = "available"
unit = "GB"
interval = 20
...
```
Then I think that would be the best of both worlds, however it is perhaps unnecessary added complexity.

## 2
Based solely on what I know of toml, that should cover everything, but I have not tested it (mostly due to my setup changing a few things)

## 1
Yes, they look great! (well, I would like less typing for icons, but I should also decide on a decent color scheme and make a pr so that I don't have to override it all the time...).

pitkley:
Thanks for the reply @svmnotn!

> my thought on the general section is that we could just make them [separate sections](https://github.com/greshake/i3status-rust/pull/41#issuecomment-307196915)

Having the general layer exist but still allow `theme` and `icons` to exist separately as tables might be possible, but I don't think it would be pretty implementation-wise, and might be relatively hard to maintain.

A decision on this probably boils down to what I initially mentioned: if we remove `[general]`, `icons` and `themes` *have to* come before any blocks. And I personally think that an error message telling the user that `theme` or `icons` is an unknown key in `[[blocks]]` probably causes more confusion and thus costs more time than requiring typing `[general]`. And if the end-user copies the `example_config.toml`, he will get it pre-typed anyway.

So for me, weighing the implementation and especially maintenance cost against what we might get, I'm **for** keeping general. As I understand you @svmnotn, you'd love to see it gone, but are not hard against keeping it, is that correct?

Overall I think we probably need a decision from @greshake on this.

> Based solely on what I know of toml, that should cover everything, but I have not tested it (mostly due to my setup changing a few things)

The changes in your setup that you mentioned: would you require an additional use-case covered by this configuration? Or are you speaking about a more advanced configuration which simply would take some time to port over?

> well, I would like less typing for icons/theme modifications

With "less typing", are you only refering to `[general]`, or is there anything else you could think of that we might be able to leave out?

greshake:
Okay, phew. First off, I greatly appreciate the effort that has already poured into this and other PRs. Because I've been absent two days and now also only have limited time, expect to see more input from me soon, for now I have these points:
1. Totally on board with changes adding to type safety/strictness or to let serde do more of the work.
2. Just wondering: There is some grammatical weirdness/dissonance introduced through TOML. Is this circumventable? What I refer to is this: In the config, we use `[[blocks]]` to configure a block (singular). This obviously makes sense because we deserialize into a Vec `blocks` of maps. But it feels weird when writing the config.
3. We should not need a `general` section. Global items can just be configured using a top-level assignment, and as long as we can miss on the multiple-theme-combing features (overkill for this imho) the config could look like this:
```toml
theme = "solarized-dark" # If not specified, defaults to plain theme
icons = "awesome" # If not specified, defaults to text icon set
# some other general config options may be added here later

# Override values here
[custom-icons] # Or nested as icons.custom?
bat = " | | "
bat_full = " |X| "
bat_charging = " |^| "
bat_discharging = " |v| "

# To reiterate, [[block]] would make more grammatical sense here...
# Maybe we can rename it when deserializing into a field called blocks again?
# Maybe plural is also best-practice in TOML, the official examples do the same
[[blocks]]
block = "disk_space"
path = "/"
alias = "/"
type = "available"
unit = "GB"
interval = 20
```
I'm sure people will not have a problem figuring out that general configuration has to occur before the blocks. We can make it clearer by adding a comment explicitly separating them in the example conf.

svmnotn:
@pitkley my setup changes a bit of the source code of other blocks and introduces a new block that I have not made a pr for as I'm still testing it. From what I can see, the config would work just fine, but require me to rebase and I would rather do that after the pr is accepted.

pitkley:
@greshake 

> 1. Totally on board with changes adding to type safety/strictness or to let serde do more of the work.

Great to hear! 👍

> 2. Just wondering: There is some grammatical weirdness/dissonance introduced through TOML. Is this circumventable?

Yes, this is circumventable: we can rename the field in the config to `block` while deserializing it into the `blocks` variable in code.

> 3. We should not need a general section. Global items can just be configured using a top-level assignment, and as long as we can miss on the multiple-theme-combing features (overkill for this imho) the config could look like this:

If we simply remove the general section from how it is now, a configuration could look like this:

```toml
icons = "awesome"

[theme]
name = "solarized-dark"
idle_bg = "#123456"
idle_fg = "#abcdef"

[[block]]
...
```

Introducing overrides as sub-keys or additional keys (instead of using the `name` field as above) would probably make deserialization quite a bit harder, again both to implement initially and to maintain at a later point. In fact, a sub-key is not possible: `theme` gets defined as a string through the first assignment, and strings can not be sub-keyed.

I can look into how hard deserializing two separate keys into effectively one target is, but I think I'd personally prefer to keep the `name` field.

-----

@svmnotn 

> From what I can see, the config would work just fine

Cool, just making sure that we cover as much as possible 👍!

greshake:
The thing that I'm not happy with is that the `name = ` key and all of the icon/theme overrides are on the same level. I'd prefer a scenario where we deserialize into a structure like this (JSON-equivalent):
```json
"icons": {
   "name": "awesome",
   "overrides": {
      "battery": "..."
   }
}
```
instead of this:
```json
"icons": {
   "name": "awesome",
   "battery": "..."
}
```
Not sure how to put this into TOML elegantly, I assumed nested tables would be a good way, but I have limited understanding of TOML currently. Maybe something like this can work?
```toml
# the icons/theme blocks would not be optional, 
# and the global key/value pairs skipped
[icons]
name = "awesome"

[theme]
name = "solarized-dark"
[theme.overrides]
idle_bg = "#123456"
idle_fg = "#abcdef"

[[block]]
...
```

pitkley:
Ah, thanks for laying out your last suggestion, that is actually pretty easy to do. I can make it work exactly as you proposed:

```toml
[icons]
name = "awesome"

[theme]
name = "solarized-dark"
[theme.overrides]
idle_bg = "#123456"
idle_fg = "#abcdef"

[[block]]
...
```

And in addition, the following will still work:

```toml
icons = "awesome"
theme = "solarized-dark"

[[block]]
...
```

I personally think we should keep the short-hand above, since it keeps the configuration very concise if you don't need overrides.

I'll update this PR later today to reflect these changes.

greshake:
Great, I'm happy with the new style. Not sure about the shorthand version (the behaviour when specifiying conflicting values is non-obvious) but I'd okay with it.

pitkley:
I've implemented the style as shown in [my previous comment](https://github.com/greshake/i3status-rust/pull/41#issuecomment-308027924). Incidentally, this changed part of the deserialization such that we do not have to wait for a `toml-rs` release, thus CI is green.

I quickly skimmed through the comments for any remaining open points and couldn't find anything that remains, so this PR should be at a very advanced stage now.

@greshake, @svmnotn, @jheyens: could you give this a final check and look for any omissions or deficiencies?

greshake:
I'll do my check and wait for feedback from @svmnotn and @jheyens as requested.

greshake:
First bug (according to README the memory block does not *require* any args. With all args, it doesn't crash):
When using this config:
```toml
[[block]]
block = "memory"
```
Program crashes, backtrace:
```
thread 'main' panicked at 'index not found', /checkout/src/libcore/option.rs:785
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:355
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:371
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:549
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:511
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:495
   7: rust_begin_unwind
             at /checkout/src/libstd/panicking.rs:471
   8: core::panicking::panic_fmt
             at /checkout/src/libcore/panicking.rs:69
   9: core::option::expect_failed
             at /checkout/src/libcore/option.rs:785
  10: i3status_rs::blocks::create_block
  11: i3status_rs::main
  12: std::panicking::try::do_call
             at /checkout/src/libstd/panicking.rs:454
  13: __rust_maybe_catch_panic
             at /checkout/src/libpanic_unwind/lib.rs:98
  14: std::rt::lang_start
             at /checkout/src/libstd/panicking.rs:433
             at /checkout/src/libstd/panic.rs:361
             at /checkout/src/libstd/rt.rs:57
  15: __libc_start_main
  16: _start
```

pitkley:
[Fixed](https://github.com/greshake/i3status-rust/pull/41/commits/fcef192cdd2e3eaf0be9df48badc0621fa859b8a).

greshake:
Same kind of bug also needs to be fixed in custom.rs (3 instances) and music.rs (1 instance). Give me push rights on your next PR, then I can fix that kind of stuff myself :)
Another behavioural difference is the parsing of unicode chars... `\uf011` works in JSON, but when used in TOML, the parser outputs:
```
'failed to parse JSON from file contents: Error { inner: ErrorInner { kind: InvalidHexEscape('f'), line: Some(59), col: 19, message: "", key: [] } }'
```
Weird that it says JSON. Must be a bug in toml-rs. But it doesn't matter very much because the error is fixed by using an uppercase F, not sure if scalar unicode values are defined only with capital hex letters or it is an implementation error...

README.md
```diff
@@ -17,18 +17,18 @@ i3, rustc and cargo. Only tested on Arch Linux. If you want to use the font icon
 # How to use it
 1. Clone the repository: `git clone https://github.com/XYunknown/i3status-rust.git`
 2. run `cd i3status-rust && cargo build --release`
-3. Edit example_config.json to your liking and put it to a sensible place (e.g. ~/.config/i3/status.json)
+3. Edit `example_config.json` to your liking and put it to a sensible place (e.g. `~/.config/i3/status.json`)
```
svmnotn: this should be `example_config.toml` and `../status.toml`

pitkley: Thanks, fixed.

pitkley:
> custom.rs (3 instances) and music.rs (1 instance)

Fixed.

> Weird that it says JSON. Must be a bug in toml-rs.

It wasn't, that was an unchanged `.expect` on my end.

pitkley:
> not sure if scalar unicode values are defined only with capital hex letters or it is an implementation error...

`toml-rs` [implements it](https://github.com/alexcrichton/toml-rs/blob/2ae8c79ab8f1ac22c53bd200f7131879fe097d80/src/tokens.rs#L359) using only uppercase characters, I'm not sure what the spec says on this.

svmnotn:
@pitkley seems to work just fine with my setup (excluding my code changes)

greshake:
No feedback from @heyens, I'll merge now.

PR #44, opened 2017-06-06
Adjust screen brightness with wheel up/down

Drogglbecher:
I extended the xrandr module with the possibility to adjust the selected screens brightness with mouse wheel scrolling. `step_width` can be adjusted from 0...50.


greshake:
Thanks! :)

Issue #42, opened 2017-06-06
Error handling


pitkley:
Currently, there is almost no error-handling, except for some `expect`s. I would like to introduce some form of error-handling to make working with `Result`s easier by using the `?` operator.

Functions like this:

```rust
pub fn get_file(name: &str) -> String {
    let mut file_contents = String::new();
    let mut file = File::open(name).expect(&format!("Unable to open {}", name));
    file.read_to_string(&mut file_contents).expect(&format!("Unable to read {}", name));
    file_contents
}
```

could be reduced to:

```rust
pub fn get_file(name: &str) -> Result<String> {
    let mut file_contents = String::new();
    let mut file = File::open(name)?;
    file.read_to_string(&mut file_contents)?;
    file_contents
}
```

Additionally, we could allow `Block`s to return e.g. `Result<Duration, SomeError>` and handle the error-case outside of the block, rather than forcing the blocks to `panic!` if they can't continue causing the whole of `i3status-rs` to crash.

-----

I did not want to go ahead with an implementation quite yet, rather I'd love some input on this. Personally, I would probably pick [`error-chain`](https://github.com/brson/error-chain) to take the burden of writing all the error-conversions needed for easy error-handling, especially since I had a good experiences in previous projects I used it in (at least after some initial learning and understanding).

Writing the conversion manually is also an option, see for example [how `hyper` does it](https://github.com/hyperium/hyper/blob/acd62cda446e4c647716a2d595342360dc24a080/src/error.rs). What has been written in `hyper` is (roughly) comparable to what `error-chain` generates.

Do you have any other experiences, opinions or ideas?

greshake:
Hmm... Without a doubt I can agree that error handeling isn't very good right now.
First, a few observations:
1. Users almost never get to see the stdout of the program, therefore usually seeing any errors/panics as "status program exited with code X" or something like that from i3bar only. 
2. Errors in the block implementations usually indicate terminal failure, and we should not take emerging errors lightly or continue running as usual.
3. Error handeling in the case of this program should be very light. There are basically no cases where we should recover from errors, because the application is stateless, easily restartable and all errors should be configuration errors or be cause by one specific block. 

My ideas now focus on indicating critical failure to the user, and terminating gracefully. My plan:
1. Use the result type for any block update and initialization, and provide a type of Error that has two properties: 1. Shorthand error message to be displayed in the statusbar, 2. Debug-level explanation that gets printed to stderr
2. Handle top-level configuration errors the same way.
[...]

Issue #5, opened 2017-05-19
[Low Prio] Performance and Interface Optimization


greshake:
This codebase isn't perfect- there is still a lot of stuff I'm not happy with, but I'm not sure how to handle them more elegantly. For example, we should allow people to register click handler during the construction of widgets, like this: `TextWidget::new().with_click(|event| <handle>)`. For that, we need to update a map of click handlers every time we update our view.

- [ ] Minimize copying and string comparisons
- [ ] Add  more elegant click handlers

ammgws:
Closing this due to inactivity.

> Minimize copying

With respect to this https://github.com/greshake/i3status-rust/pull/1033 is a step in the right direction. Other things will come up naturally.

MaxVerevkin:
@ammgws 
>Other things will come up naturally.

Right now I also have some concerns about copying the entire config structure to every block (or even many times per block).

ammgws:
See https://github.com/greshake/i3status-rust/issues/51

Re-open or create a new issue if you wish

MaxVerevkin:
OK, I'll create a new issue.

PR #46, opened 2017-06-07
Implement error handling

pitkley:
(Regarding #42, cc @greshake @jheyens.)

So, this turned out to be a pretty big PR, given that almost everything had to be touched to support the `Result` type and the associated error-handling.

I'll try to summary the biggest changes/additions to maybe ease the burden a bit.

## Summary (by files)

* `src/errors.rs`

    * Added supporting code for error handling. This includes:

        * An `enum Error` containing two types of errors: `BlockError` and `InternalError`
        * A custom `Result` type which allows omitting of our `Error`
        * Two trait extensions (`ResultExt`, `OptionExt`) which add:
        
            ```rust
            fn block_error(self, block: &str, message: &str) -> Result<T>;
            fn internal_error(self, block: &str, message: &str) -> Result<T>;
            ```

            to both `Option` and `Result`, allowing easy error handling.

        * An implementation for `std::fmt::Display` for conversion to string (used in `main.rs` to actually output an error).

* `src/main.rs`

    * Renamed `main` to `run`
    * Added another `main` method which executes `run` and catches errors which might occur directly in `main`.
    * Moved CLI matching to the new `main`
    * Added CLI flag `--exit-on-error` to exit if an error occurs, rather than outputting to i3bar.
    * Replaced every relevant `.unwrap` and `.expect` by `.internal_error` to have full error handling

* `src/block.rs`

    * Updated `trait Block` to include the `Result` type

* `src/blocks/*.rs`

    * Updated all blocks to match changed `trait Block`
    * Replaced every relevant `.unwrap` and `.expect` by `.block_error` to have full error handling

* `README.md`

    * Updated to reflect the new requirements regarding having a `Result` type

* `src/{icons,themes}.rs`

    * Updated to use `Result` type

* `src/scheduler.rs`

   * Updated to use `Result`
   * Implemented `std::fmt::Display` for `Task` to allow for additional error printing.

* `src/util.rs`

   * Updated to use `Result`

* `src/widgets/*.rs`

   * Updated to use `Result`

## Open tasks, points for discussion

* [x]  Currently both error-types, internal and block, get handled the same way: output to i3bar, then idle until the process gets killed.

    We should determine if we want to handle internal errors differently in anyway to block errors. If not, I'd still vote to keep the distinction between the two for future-proofing, in case we ever want to be able to differentiate between an error that has occured in a block and one that didn't.

    ~~**Note:** Right now the causing error is not retained, for neither internal nor block errors. I would like to implement this to keep the cause for extended debug information, but I am currently at a bit of a roadblock.~~

* [x]  I have not implemented a click-handler yet to allow automatic exporting/posting/... of an error that has occured.

    I think we should first decide on what exactly we want to do, such that it is not too complicated for both the user and the implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,14 +463,14 @@ Example:
 impl ConfigBlock for Template {
     type Config = TemplateConfig;
 
-    fn new(block_config: Self::Config, config: Config, tx_update_request: Sender<Task>) -> Self {
-        Template {
+    fn new(block_config: Self::Config, config: Config, tx_update_request: Sender<Task>) -> Result<Self> {
+        Ok(Template {
             id: Uuid::new_v4().simple().to_string(),
             update_interval: block_config.interval,
-            text: TextWidget::new(config.clone()).with_text("Template"),
+            text: TextWidget::new(config.clone()).with_text("Template")?,
             tx_update_request: tx_update_request,
             config: config,
-        }
+        })
     }
 }
 ```
@@ -481,13 +481,13 @@ This is required in addition to the `ConfigBlock` trait and is used to interact 
 
 This trait defines the following features:
 
-### `fn update(&mut self) -> Option<Duration>` (Required if you don't want a static block)
+### `fn update(&mut self) -> Result<Option<Duration>>` (Required if you don't want a static block)
 
 Use this function to update the internal state of your block, for example during periodic updates. Return the duration until your block wants to be updated next. For example, a clock could request only to be updated every 60 seconds by returning Some(Duration::new(60, 0)) every time. If you return None, this function will not be called again automatically.
 
 Example:
 ```rust
-fn update(&mut self) -> Option<Duration> {
+fn update(&mut self) -> Result<Option<Duration>> {
       self.time.set_text(format!("{}", Local::now().format(&self.format)));
       Some(self.update_interval.clone())
 }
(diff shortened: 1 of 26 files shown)


pitkley:
(Oh yeah, besides the two mentioned discussion points, I'd like to wait for #41 to be merged so I can rebase this on the configuration struct rather than the other way around.)

svmnotn:
@pitkley Anything going on here? To address the points:

> We should determine if we want to handle internal errors differently in anyway to block errors. If not, I'd still vote to keep the distinction between the two for future-proofing, in case we ever want to be able to differentiate between an error that has occured in a block and one that didn't.

Well, internal errors should be fairly rare considering that most of the code that does things is in the blocks, so we could just display internal errors with a message saying to check for an open issue or make one? While block errors could potentially be missing a program or something else on the user's end.

> I think we should first decide on what exactly we want to do, such that it is not too complicated for both the user and the implementation.

I think other than the previous point, the exporting should be simplest if we just write whatever we would send out to stderr to a file in some place (Maybe a config option, or maybe just `/tmp/`). Once the user has clicked on the error like @jheyens suggested.

pitkley:
> Well, internal errors should be fairly rare considering that most of the code that does things is in the blocks [...]

Good point. What would have to be changed though is that right now configuration errors are classified as internal errors too, and we wouldn't want a user to open an issue for a wrong configuration as an internal error.

> I think other than the previous point, the exporting should be simplest if we just write whatever we would send out to stderr to a file in some place (Maybe a config option, or maybe just `/tmp/`). Once the user has clicked on the error like @jheyens suggested.

What I thought of that might make it relatively easy for the end-users to open an issue is to implement a click-handler, which will try to `xdg-open` to the issue tracker with the title and body prepopulated:

* https://github.com/greshake/i3status-rust/issues/new?title=Internal%20error:%20yada&body=Log%20of%20internal%20error

If the call to `xdg-open` fails (either because it doesn't exist or exited with a non-zero exit code) we could fall back to a different method.

Implementing a click-handler without the need to create a full-fledged block would require something like #5 though.

Overall, I would suggest to split this second point (supporting the user to report an error) into a separate PR. This would mean merging this PR to lay the groundwork, with the error-reporting right now being a single, non-interactive block telling the user what happened. Cc @greshake, what do you think?

pitkley:
Small update: I added the `ConfigurationError` I mentioned in my [last comment](https://github.com/greshake/i3status-rust/pull/46#issuecomment-309101839). This PR should be okay to merge after review, if the split I mentioned is okay with everybody.

greshake:
I'll merge this now. I will add backtraces and a couple other improvements when it's on master. As always great work :+1:

PR #47, opened 2017-06-07
Format all the code to some standard

svmnotn:
Right now I simply used all the defaults but we could have it be the Rust Style RFC standard or whatever @greshake thinks looks best, however a change of this magnitude does affect the entire codebase making other prs have merge issues.
Thus deciding on this should be after errors and maybe after the config changes.


pitkley:
Good stuff, I was tempted to do the same already. I personally am 100 % for the RFC-style, since it is supposed to be *the* standard, similar to what `gofmt` is for Go.

We should maybe wait with doing this until [rustfmt 0.9 lands](http://www.ncameron.org/blog/rustfmt-changes/), although I'm not sure if the plans mentioned in the blogpost have changed.

svmnotn:
I'm already using 0.9 but looking at the codebase we don't seem to be using any of the new lang features that require rustfmt 0.9

svmnotn:
As a note [here](https://gist.github.com/svmnotn/678277cc639efc48fd7b8648cc9b0695) are all the possible config options. And [here](https://github.com/rust-lang-nursery/rustfmt/blob/master/rfc-rustfmt.toml) is the RFC rustfmt.toml

pitkley:
> I'm already using 0.9

Did I miss something? It has not been released yet, has it?

pitkley:
Oh, the `libsyntax` branch? Isn't that nightly only right now? Have not tested it yet.

svmnotn:
@pitkley I use nightly by default so I can intall and test it from source ;)

pitkley:
@svmnotn okay, cool, thanks for letting me know! 👍

svmnotn:
No problem, I also quite prefer the RFC version over the normal rustfmt defaults but it is up to @greshake to pick the style

greshake:
Yea I'd also go with the RFC style. I will add a format check to the CI, such that all PR's or commits with unformatted code fail the CI checks. Already added compiler warnings to the things Travis considers failed.

svmnotn:
@greshake Ok, then I'll wait for #41 and #46 to be merged before cleaning this one up

pitkley:
@greshake I used `cargo fmt -- --write-mode diff` with success in CI runs. Has a non-zero exit code if it there is code to reformat and prints the changes it would do to the log, so one can check in the CI log what exactly is different.

svmnotn:
@greshake I updated it to the latest master.

I also did a few small changes to fix lints that clippy brought up. I'm not sure if you want it to be on a separate PR that just fixes most of them, if so I can redo it without them.

Also I did a quick fix to the errors in the Temperature block since the temperature being outside the range is not something that should raise a block error but rather just be logged and ignored due to it not being fixable, again if u want that as a separate PR, I can redo this one and remove that fix.

Issue #17, opened 2017-05-21
Proposal: Change update interval consistently to milliseconds


jheyens:
I would like to discuss a proposed change of consistently switching to milliseconds from seconds in our update interval configurations.
I know that any block implementation can handle this on its own, but I would prefer to have it consistently for all blocks.
Some blocks would benefit from a more fine-grained update interval, f.e. I would CPU utilization to 250 or 500 ms. 
Opinions?

greshake:
We could just parse the interval in the config as floating point numbers. That way existing configs don't need to be changed, and we set everything from seconds to nanoseconds. Just add a macro that translates f64 to a Duration.

greshake:
@jheyens is on vacation, but I think #48 covers it :)

PR #48, opened 2017-06-15
Deserialize blocks configs into structs

pitkley:
## Summary

Building on what I mentioned in #41, this PR implements deserialization of block configs into type-safe structs.

## Main changes

* [Added](https://github.com/pitkley/i3status-rust/blob/d9165b1f4ab6861ef22f3511cc012871dc69fc7c/src/block.rs#L25-L29) `trait ConfigBlock: Block` which holds a blocks config in its associated type `Config`
* Added a configuration-struct to every block
* [Deserialize](https://github.com/pitkley/i3status-rust/blob/d9165b1f4ab6861ef22f3511cc012871dc69fc7c/src/blocks/mod.rs#L43-L48) a `toml::Value` into a block's associated config type on creation
* [Added](https://github.com/pitkley/i3status-rust/blob/d9165b1f4ab6861ef22f3511cc012871dc69fc7c/src/de.rs#L11-L54) deserialization helper for durations (used for intervals).

    This deserializes from either an integer, a float or a map:

    ```toml
    interval = 1
    interval = 1.5
    interval = { secs = 1, nanos = 5000000 }
    ```

    Both the integer and float are in seconds, i.e. `interval = 1` equals 1 second, `interval = 1.5` equals 1500 ms. This probably covers the request from #17.

* [Removed](https://github.com/greshake/i3status-rust/compare/master...pitkley:config-struct-blocks?expand=1#diff-d25ef168d08ddcfeaca8453639097a59) no longer needed macros from `src/util.rs`

### DIFF
--- a/README.md
+++ b/README.md
@@ -453,26 +453,33 @@ Create a block by copying the template: `cp src/blocks/template.rs src/blocks/<b
 
 Your block needs a struct to store it's state. First, replace all the occurrences of 'Template' in the file with the name of your block. Then edit the struct and add all Fields which you may need to store either options from the block config or state values (e.g. free disk space or current load). Use Widgets to display something in the i3Bar, you can have multiple Text or Button widgets on a Block. These have to be returned in the view() function and they need to be updated from the update() function. They also handle icons and theming for you.
 
-## Step 3: Implement the constructor
+## Step 3: Implement the `ConfigBlock` trait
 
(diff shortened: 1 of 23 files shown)


greshake:
I'd be okay with making this mandatory, e.g. just moving the associated type etc. into the block trait.

greshake:
Pls extend the changes to the README to reflect all field name changes, e.g. everywhere we moved from `-` to `_` or changed `type` to `info_type`, I think there are a few missing.

pitkley:
@greshake thanks for reminding me, updated:

* README and examples to reflect changed field names
* the "How to write a block" section in the README

greshake:
Great work!

pitkley:
Thanks for the quick merge! 🎉

PR #53, opened 2017-06-26
Decide between rustfmt and rustfmt-nightly

svmnotn:
When I did the rustfmt pr, I used the defaults from rustfmt-nightly. When I ran it now with rustfmt-0.9 defaults, there where differences. So, @greshake should we use those defaults or use rustfmt-nightly defaults? I can do another commit so that you can see the difference.

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -59,21 +59,13 @@ impl ConfigBlock for Battery {
 }
 
 fn read_file(path: &str) -> Result<String> {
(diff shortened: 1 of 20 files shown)


greshake:
I suggest we keep the nightly version. If we decide to go with nightly as the default compiler because we need some features, the stable version might not work. However, backwards-compatibility from the nightly fmt to stable should be good, so I think we should just go with the nightly format.

PR #58, opened 2017-07-04
Add new theme: Slim which uses material colors and transparency

draoncc:
Also replaces glyphs in source with utf notation.


annata83:
Hi there, How can I test this Theme?

ammgws:
Have you seen themes.md? It should hopefully explain how to use this theme. By the way the theme name is "slick", not "slim" as this title says

PR #59, opened 2017-07-05
Format according to the defaults in the latest rustfmt version

svmnotn:
Version: 0.1.8-nightly (cfec7ad2017-07-07)
As decided on #53 here is the update using the latest defaults for formatting.


greshake:
Thanks!

PR #60, opened 2017-07-05
Update Dependencies

svmnotn:
`chrono` updated to 0.4.0
`clap` updated to 2.25.0
`uuid` updated to 0.5.1
`dbus` updated to 0.5.3

Only `chrono` had a breaking change for us, and it was a 1 line change of a use statement.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,15 +4,15 @@ version = "0.1.0"
 authors = ["XYunknown <developement@kai-greshake.de>"]
 
 [dependencies]
-chrono = "0.2"
-lazy_static = "^0.2"
-serde = "^1.0"
-serde_derive = "^1.0"
-serde_json = "^1.0"
-toml = "^0.4"
-clap = "2.23.2"
-uuid = { version = "0.4", features = ["v4"] }
-dbus = "0.5.2"
+chrono = "0.4"
+lazy_static = "0.2"
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = "1.0"
+toml = "0.4"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -4,14 +4,14 @@ version = "0.1.0"
...
 lazy_static = "^0.2"
 serde = "^1.0"
 serde_derive = "^1.0"
 serde_json = "^1.0"
 toml = "^0.4"
-clap = "2.23.2"
-uuid = { version = "0.4", features = ["v4"] }
+clap = "^2.25"
+uuid = { version = "^0.5", features = ["v4"] }
 dbus = "0.5.2"
```
pitkley: Have you checked the other dependencies for updates? I think we should prefix them with `^` even if they don't have updates.

svmnotn: I changed them all to `*` and then ran `cargo update` these were the only ones that changed, but I'll double check quickly.

svmnotn: Actually on the latter point, I'm not sure if [this](http://doc.crates.io/specifying-dependencies.html#specifying-dependencies-from-cratesio) is still accurate but if it is then `^` is not required in any of them.

svmnotn: Looking into the `Cargo.lock` I just noticed that it also updated `dbus` to `0.5.3` even though the version specified is `0.5.2`

greshake:
Thanks!

greshake:
I was quite new to Rust at the start, so I didn't think twice about the Cargo.lock. Thanks for putting it in!

PR #62, opened 2017-07-06
Graph widget

etrombly:
Created a graph widget. Was testing with the cpu block, but didn't include those changes because I didn't make that portion configurable yet.

### DIFF
--- a/src/widgets/graph.rs
+++ b/src/widgets/graph.rs
@@ -45,7 +45,7 @@ impl GraphWidget {
         self
     }
 
-    pub fn set_values<'a, T>(&mut self, content: &'a [T], min: Option<T>, max: Option<T>)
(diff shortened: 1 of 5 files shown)


src/widgets/graph.rs
```diff
@@ -0,0 +1,105 @@
...
+        self
+    }
+
+    pub fn with_state(mut self, state: State) -> Self {
+        self.state = state;
+        self.update();
+        self
+    }
+
+    pub fn set_values<'a, T>(&mut self, content: &'a [T], min: Option<T>, max: Option<T>)
```
svmnotn: Is rustc unable to elide the `'a` in this case?

etrombly: I swear there was a reason I had that in there, removed it and it was able to elide it fine though. Made another minor change also. should be good now.

greshake:
Looking good I think! Can't really check right now, but I'll trust you and the CI :)

Issue #55, opened 2017-06-29
Net block


etrombly:
I started working on a net block (just rx and tx). Before I get much farther was wondering if people wanted it configurable to show a specific output, like MBps, or have it scale with a label that changes. I have the output labeled "⬆" and "⬇", is that something people would want configurable as well, or would those defaults be ok?

Also related, I think it would be nice to have a graph output similar to i3pystatus.

greshake:
Hey, welcome and thanks for your help! 
You can find examples of how to implement different units elegantly in the disk usage block. Units should be configurable (Kbps, Mbps, MB/s, etc.). You can use icons for the up/down label, and define them as text UP and DOWN for when the user selects the text icons. Could you elaborate on the graph thing? I don't know how we can display a proper graph with text only...

etrombly:
I have the units autoscale currently, I can make it configurable. I had the arrows hard coded, I'll change it to a definition like you recommended. 

Current graph trait is in https://github.com/etrombly/i3status-rust/blob/net/src/blocks/net.rs with an implementation for Vec<u64>. Need to add fields for defined min and max, for example cpu usage would be 0 min and 100 max. Seems to be working well though, so it should probably be somewhere that's available to all blocks, not just in the net block. I basically just ported the make_graph function from i3pystatus.

Currently I just show rx and tx speed, not sure if I should add IP address, wifi ssid, etc or just have those be different blocks.

greshake:
Okay, I like the idea of using the ["_","▁","▂","▃","▄","▅","▆","▇","█"] symbols. I thought maybe I had overlooked a feature of i3bar making this easier. Will have to try before I judge how well this does though :+1:

greshake:
If this works well, please make the graph trait thing a separate PR and add it to utils or make it a separate widget (my recommendation), e.g. add a GraphWidget analogous to the Text and ButtonWidgets. Then you also have greater control over caching/rendering etc.

etrombly:
I'll have to play around with it a little, I guess it depends on if you're going to typically want to display a graph alongside text, or standalone. I'll try both. 

How good it looks is a little dependent on font. Font Awesome seems to have some character size issues, so the output jumps around a little. For now I've settled on Bitstream Vera Sans from the powerline patched nerd fonts (on Arch).

greshake:
[...]

PR #63, opened 2017-07-06
Net Block

etrombly:
Implementation for issue #55 
Not sure how you want documentation done for new blocks and widgets. Here's an example config for now so you can test:

[[block]]
block = "net"
device = "eno1"
interval = 5
graph = true

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -8,7 +8,6 @@ use errors::*;
 use widgets::text::TextWidget;
 use widgets::graph::GraphWidget;
 use widget::I3BarWidget;
-use input::I3BarEvent;
 use scheduler::Task;
 use std::fs::OpenOptions;
 use std::io::prelude::*;
@@ -133,10 +132,6 @@ impl Block for Net {
             return vec![&self.output_tx, &self.graph_tx, &self.output_rx, &self.graph_rx]
         }
         vec![&self.output_tx, &self.output_rx]
-    }
-
-    fn click(&mut self, _: &I3BarEvent) -> Result<()> {
-        Ok(())
     }
 
     fn id(&self) -> &str {
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -52,7 +52,7 @@ impl NetConfig {
 impl ConfigBlock for Net {
     type Config = NetConfig;
 
-    fn new(block_config: Self::Config, config: Config, tx_update_request: Sender<Task>) -> Result<Self> {
+    fn new(block_config: Self::Config, config: Config, _tx_update_request: Sender<Task>) -> Result<Self> {
         Ok(Net {
             id: Uuid::new_v4().simple().to_string(),
             update_interval: block_config.interval,
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -90,7 +90,7 @@ fn convert_speed(speed: u64) -> (f64, &'static str) {
     (speed, unit)
 }
 
-impl Graphable for Vec<u64>{
+impl<'a> Graphable for &'a [u64]{
     fn make_graph(&self) -> String {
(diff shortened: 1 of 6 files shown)


src/blocks/net.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    fn view(&self) -> Vec<&I3BarWidget> {
+        if self.graph {
+            return vec![&self.output_tx, &self.graph_tx, &self.output_rx, &self.graph_rx]
+        }
+        vec![&self.output_tx, &self.output_rx]
+    }
+
+    fn click(&mut self, _: &I3BarEvent) -> Result<()> {
```
svmnotn: if you are not using click, you don't need to have it.

etrombly: removed it

greshake:
Sorry for the long feedback latency, really busy right now... Will review in a couple days at most, promised! Thanks for the PR!

greshake:
Okay, first bug to be fixed: The update interval may not be an integer number of seconds! The deserializer for duration also accepts fractions, such as `0.1`. This assumption causes a division by zero in line 103 onward:
```rust
        let rx_bytes = (current_rx - self.rx_bytes) / self.update_interval.as_secs();
        let (rx_speed, rx_unit) = convert_speed(rx_bytes);
        self.rx_bytes = current_rx;

        let current_tx = read_file(&format!("{}tx_bytes", self.device_path))?
            .parse::<u64>()
            .block_error("net", "failed to parse tx_bytes")?;
        let tx_bytes = (current_tx - self.tx_bytes) / self.update_interval.as_secs();
```

greshake:
Next, please add documentation to the README for your block. Remember to mention what units you are using (e.g. Bits or Bytes? Mbps or MB/s etc.).

greshake:
Also, please think about using the widget state property to change colors when downloading/uploading. Maybe you need to record somehting like highest seen bandwith to actually have an absolute value with which you can measure what is 
a "high" traffic rate, depending on the connection, s.t. you dont set the download widget to critical when only downloading with 1Kbps.

greshake:
Next bug: There is an error in your Graph widget... After running for a minute or so, I get this error:
```rust
thread 'main' panicked at 'index out of bounds: the len is 9 but the index is 9223372036854775808', src/widgets/graph.rs:62
```

greshake:
Code review will follow when the issues above are resolved. Happy debugging! :D

etrombly:
I haven't been able to reproduce the error on the graph widget, what block were you using it with, and what were your settings (update interval and if you had min and max manually set)? Thanks

greshake:
It happened with this configuration (your net block is the only one that currently uses the graph widget):
```toml
[[block]]
block = "net"
device = "wlp2s0"
interval = 1
graph = true
```

greshake:
By the way, 9223372036854775808 happens to be the max. signed 64 bit Integer, so you may be looking at an underflow issue.

etrombly:
looks like it was probably a divide by zero error. Please try again with the current fixes.

greshake:
Okay, so the issue didn't reoccur for me. But there is another issue- in the documentation you write that the units are in (prefix) bits per second, but my short testing suggests it is in (prefix) bytes per second... E.g. when I do a download test with 30Mbps, it displays roughly 3-4M, which is consistent with bytes/sec, e.g. 30(Megabit/s or Mbps) / 8 bit = 3.75 (Megabytes/s or MB/s). Also, please resolve the new merge conflict.. I assume it is only there because you didn't `cargo fmt` after the last changes. Good work though :+1:

etrombly:
updated the documentation, forgot that the base units were in bytes not bits. The conflict is from the fix I made to remove the divide by zero error. You can merge the version of graph.rs from this commit.

greshake:
Keep the nice work up! :+1:

PR #67, opened 2017-07-15
Add speedtest block

svmnotn:
Added speedtest block, organised the ReadMe so that all blocks would have the same formatting, ran the latest rustfmt, and did a minor dep upgrade, nothing is broken :D

If you want me to split off any of the changes into separate PRs feel free to tell me so.

Also my change to the `match_range` macro was to avoid using a temp variable to store the value of `vals[0]` as that is not a valid identifier.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,11 +10,11 @@ dependencies = [
  "i3ipc 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "nix 0.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "num 0.1.39 (registry+https://github.com/rust-lang/crates.io-index)",
+ "num 0.1.40 (registry+https://github.com/rust-lang/crates.io-index)",
  "progress 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.9 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_derive 1.0.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_derive 1.0.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "toml 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "uuid 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -39,7 +39,7 @@ version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "kernel32-sys 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.24 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.26 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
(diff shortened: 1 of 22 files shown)


svmnotn:
Right now I'm using `net-down` and `net-up` for the download/upload speed icons, show I perhaps use different icons as they don't quite show the same thing?

greshake:
I'll leave the icons to your judgement... But we should add a click-to-update feature

svmnotn:
I've been thinking about adding that for a while, but i would basically need to call update when clicked, which means i need to add the return value to the scheduler so the method actually does something. I havent quite figured out the way to do this

greshake:
Yea that's tricky. But you should do the long update of the value (the speedtest) in a separate thread anyway. Use a Mutex to wrap result values in your block, and update them in a separate thread. Then enqueue an update request when it's done and just call set_text etc. with the new values in update() on the main thread. Look into the weather block PR for an example how to implement this.

svmnotn:
@greshake I was going to go down the separate thread option at first, but `try_wait` is basically the non blocking version of `wait` so I didn't have to stop the thread to wait for the result. I still don't see how threads would solve the clicking issue however. Or was that a separate issue? In which case, I think `try_wait` makes the code less complex than using a separate thread, and I'm not sure if there would be any gain to using another thread.

greshake:
Not sure if I understand what you mean. Why would you need or want to block/wait on either thread? the main thread only starts an update on a separate thread, which then tells main when it's finished by passing an update request. You should not need to block either thread... For better understanding, let's just stick to the "user clicks, we want to update" scenario. `click()` should return immediately after starting another thread. The update scheduler will then be informed when the update is done and trigger a block update, giving the block the chance to incorporate the new data into the view.

svmnotn:
The blocking is because unlike the rest of the programs that are spawned, speedtest-cli takes around 25 seconds to produce output, so if someone were to naively copy what other blocks have, then it would block the main thread for 25 seconds until it can read the output. By using `try_wait` you can read the output when it is done instead. However that is not the point... From what I saw, I will try this soon, is doing this:
```rust
fn click() {
  sender.send(Task {
                         id: id.clone(),
                         update_time: Instant::now()
                     }).unwrap();
}
```
Ok? It seems like it would probably be the easier thing to do, but I don't know if it is possible as of yet.

greshake:
The thing is, the update function is executed on the main thread, which means that really long tasks block other blocks from updating.. While this recheck thing works, it's not very elegant and imposes additional computational overhead, because the bar is repainted each time an update is executed, even if it doesn't update anything. Just update in a separate thread, which can just block until execution is finished, and then notify main once the results are ready.

svmnotn:
Sure, I'll go ahead and do that

svmnotn:
Commits have been squashed into just the 3, and the `speedtest` block now uses a thread.

greshake:
Thanks for doing this, sorry if I annoyed you with my request...

You just have a small bug in your synchronization. If you have a boolean `updated` to signal the state of a different variable (`values`), but those are not synchronized together (e.g. by the same lock) you might run into logical errors. e.g. one thread seeing old data etc.. While this should still work here, it can lead to some really nasty bugs. You can go 2 ways here:
1. Omit the updated value, and start another thread sending periodic ticks to trigger updates (can be done easily with `chan::tick`). Then just copy the data into the view on update without checking. 
2. Use a lock to synchronize access to both `updated` and `values` simultaneously. Should be the simplest, although I'm not too familiar with Rust's shared memory primitives (maybe need to use a Mutex around a tuple type or something like that).

svmnotn:
`updated` is just to know if the other thread has run, because otherwise it would keep calling itself. Would it work better if i switched the type to be `(bool, Vec<f32>)`?

svmnotn:
instead of using two variables

greshake:
Yep. In this case it is very unlikely that you will see any malfunction with your current implementation, because you have at most two agents, and the second agent can only put out updates every 25 or so seconds. But imagine this:  the update function is called from the main thread, be it through a timer or update signal, and checks `updated` which is true. If at some point the implementation then changed to `values` being written to after setting `updated = true`, the main thread might see an old version of `values` because the thread that set `updated` could get the lock to `values` *after* the main thread (In this case that would still not trigger malfunction because a second update would be triggered right after) . Anyway, it might introduce some bugs that are really hard to find/debug. Using a `Mutex<(bool, Vec<f32>)` (not sure if that exact type def works) solves that issue because it would ensure consistency.

svmnotn:
possible issue should be fixed now

svmnotn:
@greshake conflict issue is solved now, and the possible bug in synchronization should be fixed.

Anything left holding this PR back?

greshake:
No, just me forgetting to revisit the matter. Sorry. Just let me investigate the sudden compiler warning in the CI and I'll merge. Thanks for your work :+1:

greshake:
Somehow a case arm was duplicated in your PR. Fix it and I'll merge :) Compiler warnings for the win!

svmnotn:
@greshake should be fixed now, I had done a rebase and didn't notice that it got duplicated

PR #68, opened 2017-07-15
fix most clippy lints

etrombly:
Went through and fixed most of the clippy lints. Sorry this touches so many files, but most of the changes are pretty minimal. Mostly changing unwrap_or to unwrap_or_else, and overlapping match arms.


greshake:
Thanks :+1:

PR #70, opened 2017-07-20
remove `eprintln!` macro due to it being on stable 1.19

svmnotn:
This means our minimum Rust version is 1.19


greshake:
Thanks!

greshake:
I'll wait for a couple weeks though to decrease the number of people opening issues about compile errors.

PR #76, opened 2017-08-01
Add additional options for CPU block

anthonynguyen:
This makes the info/warning/critical thresholds configurable for the CPU block


svmnotn:
Definitely something we should add to all blocks that do this :+1:

greshake:
Yea most of the thresholds were set by me arbitrarily :D Thanks :+1:

PR #78, opened 2017-08-02
[Added] bad-wolf theme

Drogglbecher:
I added a theme whose color schema matches the bad-wolf airline theme for vim (the "lemon bar" style: http://imgur.com/KOsmyN1). For some people who use this (quiet popular) color schema this one could be interesting ;)


greshake:
Thanks! I like the theme :+1:

PR #90, opened 2017-08-31
[WIP] allow custom block to specify which mouse button to use when clicking

### LINKED ISSUE #89: make custom block command run once
Hello! I am trying to make a block to run only once. I do not know if this is an issue or something i do wrong.

I have this block
[[block]]
block = "custom"
interval = 60
command = "cat /tmp/.email"
on_click = 'python ~/.config/i3/mutt.py'

mutt.py:
import filelock
import os
import sys
import time
parentpid = os.getpid()
lock = filelock.FileLock("/tmp/.mutt_runnning_lock")
try:
    with lock.acquire(timeout = .1):
        pid = os.fork()
        os.kill(parentpid,9)
        os.system('termite -e "mutt";~/bin/checkMail.sh &')
        lock.release()
except:
    exit(0)

what the script does is create a a lockfile. if lockfile can not be acquired it exits.

Problem: when i click multiple times on the block it opens many termite with mutt many times.
When I run the script multiple times from the shell it seems to do what it is supposed to: if mutt is open it does not open multiple instances

svmnotn:
I haven't tested it. @endir could you please see if this could fix #89? This is mostly a proof of concept, so it needs testing but since it compiles, theoretically it _should_ work but who knows.

I'm open to any recommendations on how to do this in a better, more user friendly manner.
However if it's anything related to reading the mouse button name, it would be nice if @pitkley could extend the deserialize impl to handle it ;)

### DIFF
--- a/src/blocks/custom.rs
+++ b/src/blocks/custom.rs
@@ -1,18 +1,18 @@
-use std::time::{Duration, Instant};
-use std::process::Command;
-use std::iter::{Cycle, Peekable};
-use std::vec;
-use std::env;
 use chan::Sender;
+use std::env;
+use std::iter::{Cycle, Peekable};
+use std::process::Command;
+use std::time::{Duration, Instant};
+use std::vec;
 
 use block::{Block, ConfigBlock};
 use config::Config;
 use de::deserialize_duration;
 use errors::*;
-use widgets::button::ButtonWidget;
-use widget::I3BarWidget;
 use input::{I3BarEvent, MouseButton};
 use scheduler::Task;
+use widget::I3BarWidget;
+use widgets::button::ButtonWidget;
 
 use uuid::Uuid;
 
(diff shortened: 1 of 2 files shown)


endir:
Thanks! I will check it in the next few days.

endir:
Hi I am not sure how to enable on_left_click    I tried to add 

on_left_click = "something"

but that does not work

svmnotn:
@endir i made it a bit different, not sure about the syntax in toml but the rust would be `(MouseButton, "something")`

svmnotn:
that way we could do all supported click types, i didnt test how to input it however.... I'll see about doing some research if you cant find it

greshake:
I'll check it out soon, thanks :D

that1communist:
This would also fix #106 and #107

svmnotn:
updated this back to master now that I have more time, I'll look into getting this working properly in the coming days

ammgws:
@svmnotn Sorry for taking so long to get to this!

Are you still using/interested in implementing this feature?

If so, can we do it without breaking existing user configs?
So leave `on_click` as is, but have it only work for the left mouse button.
Then add `on_middle_click` and `on_right_click` as separate parameters.

ammgws:
Closing due to inactivity.

PR #91, opened 2017-09-03
change battery device config to string

etrombly:
I have a device with a battery named max170xx_battery, so the device config being number wasn't working. Changed it to a string to allow for different battery names. Since the documentation is moved to the wiki will need to update that separately.


greshake:
Thanks, I'll update the wiki :)

rubdos:
Perhaps add this to "breaking changes" in the README :-)

greshake:
Good idea :+1:

PR #95, opened 2017-10-19
Adding Config options for 'alert' and 'warning' thresholds.

dijit:
Defaults to previous values.
Should read "warning" and "alert" from config, alert being "Critical".

### DIFF
--- a/example_config.toml
+++ b/example_config.toml
@@ -8,6 +8,8 @@ alias = "/"
 info_type = "available"
 unit = "GB"
 interval = 20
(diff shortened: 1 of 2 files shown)


dijit:
Build is failing due to something in Sound. Not related to my PR.

greshake:
Hi, thanks for contributing! 
I checked and it seems you're right. Only fails because warnings are hard errors for the CI builds... Will have to do some cleanup soon anyways. 

Good work :+1:

Issue #92, opened 2017-09-05
change brightness control from xrandr to sysfs


etrombly:
xrandr changes the brightness of the display using gamma rather than changing the brightness in hardware, so it doesn't really save much power. Using /sys/class/backlight/ directly would require a udev rule for a regular user, but saves a lot more power. Not sure if this would be better as an entirely separate block, or an option.

greshake:
I'm not familiar with the interface and how easy it will be to make this compatible with a lot of screens.. But of course using the interface would be much preferred

etrombly:
It supports multiple displays, the names aren't as nice though. They'll be along the lines of acpi_video0, and intel_backlight. It also has the edid, so you can parse out the actual display name if it reports it correctly, but that's probably too long.

mirh:
Just for the records, xrandr *can* change "hardware brightness". It's just that it's called backlight in there (in the drivers that support this property to begin with)

ammgws:
@mirh Do you mean the `--brightness` option can change hardware brightness depending on the user's hardware? Or is there another option specifically for changing hardware brightness? I couldn't find one.

mirh:
No, brightness is just a hacky software solution, end. 
I meant the `backlight` property. 
https://unix.stackexchange.com/questions/61256/how-to-change-lcd-intensivity-brightness

ammgws:
Hmm those are some weird ranges. We could add an option to the xrandr block to make it use this method, just need to parse out the ranges I guess.

```sh
xrandr --verbose
Screen 0: ...
    BACKLIGHT: 268 
        range: (0, 852)
    Backlight: 268 
        range: (0, 852)
```

mirh:
Or there should also be xbacklight I guess.

PR #98, opened 2017-10-22
Add a simple Backlight (brightness) block

atheriel:
This PR adds a `backlight` block that shows brightness information for a given display device. It does duplicate some aspects of the `xrandr` block, but has two important advantages:

1. It reads the `sysfs` files directly instead of using an external command (like `xrandr`), which means it works on Wayland, and addresses #92; and
2. It uses `inotify` (on a separate thread) to update the block whenever the display brightness is changed, instead of updating at a set interval.

This approach does mean adding the `inotify` crate as a dependency, and it will only work on Linux. I haven't carefully profiled the block for performance issues, but it doesn't look all that suspicious in `htop`.

I've also run `cargo fmt` on the main file, for what it's worth.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ nix = "0.8.1"
 i3ipc = "0.4.2"
 num = "0.1.39"
 chan = "0.1.19"
+inotify = "0.5.0"
 # Used only in debug build mode
 # for profiling blocks
 cpuprofiler = "0.0.3"
--- a/src/blocks/backlight.rs
+++ b/src/blocks/backlight.rs
@@ -18,20 +18,23 @@ use uuid::Uuid;
 
 /// Read a brightness value from the given path.
 fn read_brightness(device_file: &Path) -> Result<u64> {
-    let mut file = try!(OpenOptions::new()
-        .read(true)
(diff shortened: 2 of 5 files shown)


greshake:
Looks real nice, thanks! Will have to incorporate inotify in a few other blocks, too :)

greshake:
Please extend the Wiki with your block!

atheriel:
Thanks! I don't have write access to the wiki, so here's an entry:

## Backlight

Creates a block to display screen brightness. This is a simplified version of the `xrandr` block that reads brightness information directly from the filesystem, so it works under Wayland. The block uses `inotify` to listen for changes in the device's brightness directly, so there is no need to set an update interval.

When there is no `device` specified, this block will display information from the first device found in the `/sys/class/backlight` directory. If you only have one display, this should find it correctly.

### Examples

Show brightness for a specific device:

```toml
[[block]]
block = "backlight"
device = "intel_backlight"
```

Show brightness for a the default device:

```toml
[[block]]
block = "backlight"
```

### Options

-- | -- | -- | --
device | The `/sys/class/backlight` device to read brightness information from. | No | Default device (see wiki for details).

Issue #82, opened 2017-08-11
Improve network block


dakra:
It would be nice to have more display options for the "net" block.
For example I would just like to show the interface IP instead of the current
bandwith.

Also an option to hide (or a different text) the block if the interface is down.

Something like (from my old config):

```
ethernet tun0 {
        format_up = "VPN: %ip"
        format_down = "VPN: down"
}
ethernet net0 {
        format_up = "E: %ip (%speed)"
        format_down = "E: down"
}
ethernet net1 {
        format_up = "E: %ip (%speed)"
        format_down = "E: down"
}
```

Instead of 'E: down' I would actually prefer to only show the
active interface (not possible with i3status).

Nice status bar. Thanks :)

greshake:
@etrombly did the net block, maybe he wants to help you out :)

etrombly:
Shouldn't be too hard to add the ip. I didn't add it originally because there isn't a standard for formatting across blocks, it's handled at the block level, so I was waiting to see if anything was created for that. @greshake if I send an empty vec back for the view function will the block not be drawn at all? If not I can just change the status.

greshake:
Correct, it will not be displayed.
See util.rs:99:
```rust
        if widgets.len() == 0 {
            continue;
        }
```

atheriel:
@greshake @etrombly Any objection to my upstreaming some related changes to the Net block? It would add these features.

etrombly:
I don't have a problem with it. Anything to improve it is welcome.

greshake:
Sure go ahead

PR #102, opened 2017-10-27
Add network status, IP address, SSID, and wireless/wired icons to the net block

atheriel:
This PR adds several new features to the net block:

- An icon to distinguish between wireless and wired connections.
- Automatic hiding of graphs/throughput info for inactive networks (which can be turned off with `hide_inactive = false`).
- The ability to show the IP address (with `ip = true`).
- The ability to show the wireless network SSID (with `ssid = true`). Truncates long SSIDs to `max_ssid_width` characters.

Most of these are discussed as desirable features in #82.

It also fixes a small bug where the initial update of the net block could sometimes show absurdly large upload/download numbers.

This is a pretty big PR that makes a number of substantial changes to the existing implementation, but I've tried to keep the history fairly clean if you want to look at the commits more granularly.

~~There is one undesirable aspect of the implementation for checking the network SSID and IP addresses: both use external shell commands to get the relevant info, so this is a nontrivial amount of useless CPU to burn checking information that shouldn't change that often every second (which is how often `update()` is called to check on the upload/download speed). I'm not sure how big a problem this might be considered or the correct way to solve it, since at the moment there is no way to have different update intervals for different widgets in a block.~~

Network SSID and IP addresses are updated every 30s instead of at the block update interval (which is 1s by default), because these values don't change that often.


greshake:
Nice! Will review soon!

greshake:
Regarding the different update times, you can just store the last time updated in your block state, then compare that time in update() to now and trigger a secondary update if the time elapsed since is greater than X.

greshake:
Also, we should offer users to have the SSID truncated or in a marquee (quite complicated to build and not CPU friendly- I built one marquee example for the music block)

atheriel:
That's a much simpler and more obvious update approach than what I had in mind. Sounds perfect. Also, I think truncated makes more sense, since I doubt that anyone really needs to know every letter of their SSID.

Should have some time to make these changes later today or tomorrow.

greshake:
Nice! 

I ran into a crash while testing the patch on my daily driver, the /net/operstate file didn't exist. It happened after waking from sleep, and because I was travelling I didn't record the exact message.

atheriel:
I didn't know that was possible... I'll update the logic to account for a potentially missing file.

atheriel:
I've made the requested changes, and improved the names of configuration options. Let me know what you think.

src/blocks/net.rs
```diff
@@ -56,37 +210,87 @@ impl NetConfig {
...
+    fn default_ip() -> bool {
+        false
+    }
 }
 
 impl ConfigBlock for Net {
     type Config = NetConfig;
 
     fn new(block_config: Self::Config, config: Config, _tx_update_request: Sender<Task>) -> Result<Self> {
+        let device = try!(NetworkDevice::from_device(block_config.device));
```
greshake: Use the shiny new `?` operator :)

src/blocks/net.rs
```diff
@@ -105,49 +309,99 @@ fn convert_speed(speed: u64) -> (f64, &'static str) {
...
+
+            return Ok(Some(self.update_interval));
+        } else {
+            self.network.set_text("".to_string());
+        }
+
+        // Update SSID and IP address every 30s.
+        let now = Instant::now();
+        if now.duration_since(self.last_update).as_secs() > 30 {
+            if let Some(ref mut widget) = self.ssid {
```
greshake: The renaming from `self.ssid` -> `widget` and `device.ssid` -> `ssid` is kinda confusing to read. Same goes for the part below... Maybe you can find a more readable solution? I won't block the PR for this change though :)

src/blocks/net.rs
```diff
@@ -105,49 +309,99 @@ fn convert_speed(speed: u64) -> (f64, &'static str) {
...
     fn view(&self) -> Vec<&I3BarWidget> {
-        if self.graph {
-            return vec![
-                &self.output_tx,
-                &self.graph_tx,
-                &self.output_rx,
-                &self.graph_rx,
-            ];
+        // Since we can't error here, report errors as non-up.
+        let is_up = match self.device.is_up() {
```
greshake: Careful! The `is_up()` function performs I/O, this should not be done in `view()`. Only use this function to render a presentable version of your widget from current state.  `view()` may be called very often (independently of the update intervals set), so it should only perform simple operations and *no* system calls!

atheriel: Fixed by moving state into the `Net` struct itself.

src/blocks/net.rs
atheriel: Naming should be clearer now.

src/blocks/net.rs
atheriel: Done.

atheriel:
Rebased and comments addressed. Let me know if there's anything else.

greshake:
Thanks for the big contribution!

greshake:
Merged :)

greshake:
I agree, closed

PR #103, opened 2017-10-27
Fix volume increasing by a value less then step

michelboaventura:
It was not possible to increase volume if its value is greater than (100 - volume_step). For example, with volume equals 98% and volume_step equals 5% it was not possible to set volume to 100%. This commit fix the issue by increasing volume by min(volume_step, 100 - volume).


greshake:
Yup, thanks!

PR #105, opened 2017-11-01
Net block: Make tx and rx widgets optional

1wilkens:
Hi,

while setting up the new `net` block, I realized that the throughput widgets are always displayed.
As I don't really want them, I added two config switches similar to `ip` and `ssid` ones defaulting to `true` so existing configurations don't change.

I tried to be as noninvasive as possible in regards to coding style, but if there is anything wrong, let me know.
I also tried to run `cargo fmt` but that generated a lot of changes in code I did not touch, so I didn't commit that.

Cheers,
Florian

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -177,9 +177,6 @@ pub struct NetConfig {
     #[serde(default = "NetConfig::default_device")]
     pub device: String,
 
-    #[serde(default = "NetConfig::default_graph")]
-    pub graph: bool,
-
     /// Whether to show the SSID of active wireless networks.
     #[serde(default = "NetConfig::default_ssid")]
     pub ssid: bool,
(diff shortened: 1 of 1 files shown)


greshake:
You actually should be bolder in your changes here. If the throughput widgets are not enabled, also disable the calculation/ updating of the data in the net block's update function to save CPU time & I/O! 

And of course, thanks for the PR! :+1:

1wilkens:
Ah that makes total sense, I tried a minimal example first. I'll work on it some more. :)

1wilkens:
That should be better.
I also added a small commit to change the intial value of `last_updated` to set ssid and ip address immediately on the first `update`.

src/blocks/net.rs
```diff
@@ -195,6 +195,14 @@ pub struct NetConfig {
     /// Whether to hide networks that are down/inactive completely.
     #[serde(default = "NetConfig::default_hide_inactive")]
     pub hide_inactive: bool,
+
+    /// Whether to show the upload throughput of active networks
+    #[serde(default = "NetConfig::default_tx")]
+    pub tx: bool,
```
atheriel: I don't think it's really necessary to have two configuration options -- I can't imagine anyone wants to see *just* up or *just* down speeds. Why not simplify this to a single `speed` option (which mirrors the language of i3blocks)?

atheriel:
That change to the initial value of `last_updated` fixes a nice bug in my code. 👍

1wilkens:
Yeah I thought about the configuration options and was leaning more in the direction of "the more customizable the better", but I can see the argument for simplicity (never tried i3blocks).
I'll change it to one option, any ideas for the name?
`throughput` seems a bit verbose, `txrx` might be a little cryptic..

1wilkens:
Totally missed the suggestion for `speed`. I decided on that because its ways easier to type then `throughput` constantly ;)

src/blocks/net.rs
greshake: I can imagine. For example, I have a reasonably stable uplink, but my downlink can vary immensely

greshake:
Yo if you're ready to be merged just tell me

1wilkens:
I think this is the version I prefer the most:
- `tx` and `rx` are separate
- `graph` is split into tx and rx components as well to mirror the regular indicators
I also want to include the bitrate (next to the ssid) but I'll do that in another PR.

atheriel:
Maybe `down_speed` and `up_speed`? I definitely think `tx` and `rx` are too cryptic for user-facing configuration options.

1wilkens:
Updated with `speed_up`/`speed_down` mirroring `graph_up`/`graph_down` :)

greshake:

Issue #101, opened 2017-10-26
Scrolling reads all inputs rather than just most recent


that1communist:
If you scroll on a trackpad on the volume icon, you'll notice it won't stop when you stop scrolling, rather it will go through all the inputs you gave it, slowly.

I don't know best how to describe this issue, sorry if the title is poor, I can resubmit.

greshake:
Can't reproduce here, more details please :)

that1communist:
https://photos.app.goo.gl/V0NuUt5giVQ7ncYn1

Here's a video.

even if i move the cursor away after i've done the scrolling, it will continue to change the volume.

greshake:
Okay. This is what we need:
1. OS/ Versions of software involved (i3, i3status-rust, i3bar, do you use a compositor?) 
2. Configuration files for i3, i3status-rust

What you can try to help debug the issue:
- Does the problem occur when using a regular mouse?
- Add another block that reacts to clicks (e.g. music or temp) and see if they react instantly to clicking or are delayed

If no lead turns up, I'll try to sprinkle a few debugging statements and we'll do a logging run :)

that1communist:
1. OS/ Versions of software involved (i3, i3status-rust, i3bar, do you use a compositor?)

The issue happens without a compositor.

```
i3 version 4.14 (2017-09-04)
i3bar version 4.14 (2017-09-04)
i3status-rs 0.9
```


Configuration files for i3, i3status-rust

This is the bar section of my i3 config
```
bar {
  bindsym button1 nop
  status_command /usr/bin/i3status-rs ~/.i3/config.toml
  font pango:Noto Mono, fontawesome 7
  strip_workspace_numbers yes
	colors {
		background		#000000
		statusline		#ffffff
		separator		#383838
		focused_workspace	#ff0000 #440000 #ffffff
		active_workspace	#ff0000 #440000 #383838
		inactive_workspace	#383838 #383838 #ffffff
		urgent_workspace	#d32f2f #d32f2f #000000
    	}
}
```
This is my i3status-rust config
[theme]
name = "plain"
[theme.overrides]
idle_bg = "#000000"
idle_fg = "#FFFFFF"
info_bg = "#000000"
info_fg = "#FFFFFF"
good_bg = "#000000"
good_fg = "#FFFFFF"
warning_bg = "#000000"
warning_fg = "#FFFFFF"
critical_bg = "#000000"
critical_fg = "#FFFFFF"
separator = "20"
[...]

PR #108, opened 2017-11-03
Update the Sound block asynchronously instead of at a set interval

atheriel:
This PR modifies the Sound block to listen for ALSA events as a way to know that the block must be updated, as opposed to checking the volume every two seconds. This should dramatically reduce the number of updates to the block and make it much more responsive to users expecting visual feedback when changing the volume (e.g. as discussed in #101). However, it does mean spinning up another thread, which itself has a cost.

I haven't modified what is accepted in the block configuration -- the `interval` field is now simply ignored -- so existing configurations are not rejected.

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -1,9 +1,11 @@
-use std::time::Duration;
-use std::process::Command;
 use std::cmp::min;
+use std::io::Read;
+use std::process::{Command, Stdio};
+use std::thread;
+use std::time::{Duration, Instant};
 use chan::Sender;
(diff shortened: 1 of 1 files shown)


greshake:
Nice, thanks! I've been working on a single-threaded all-in-one solution based on futures/tokio for this for a while, because writing async code so far has been... unelegant to say the least. Unfortunately, Rust isn't quite ready for that yet...

atheriel:
Yeah, I can see how futures might simplify some of the API. Good luck!

PR #109, opened 2017-11-03
[Net block]: Add bitrate widget (using `iw`)

1wilkens:
This PR adds a bitrate widget to the `net` block via `iwlist` (provided on Arch Linux by `wireless_tools`, a dependency of the original `i3status`).
In my quick search I didn't find a way of getting this information without another dependency but I am sure there is one.
I had `wireless_tools` installed as I still have `i3status` installed but that might of course not be the case for everyone.

Is there a standard way to signal that a certain setting requires a dependency? I thought about returning an error in the `bitrate` function but that would still execute the command every time an update is requested.
If the new dependency is not acceptable, then feel free to close this PR ;)

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -102,7 +102,7 @@ impl NetworkDevice {
                 ],
             )
             .output()
-            .block_error("net", "Failed to exectute SSID query.")?
+            .block_error("net", "Failed to execute SSID query.")?
             .stdout;
 
         if iw_output.len() == 0 {
@@ -131,7 +131,7 @@ impl NetworkDevice {
                 ],
             )
             .output()
-            .block_error("net", "Failed to exectute IP address query.")?
+            .block_error("net", "Failed to execute IP address query.")?
             .stdout;
 
         if ip_output.len() == 0 {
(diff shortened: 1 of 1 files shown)


atheriel:
I think it's a good idea to have a way of signalling dependencies like this, too. For example, the sound block requires several ALSA commands.

One way to do this is to make use of [Cargo's features section](http://doc.crates.io/manifest.html#the-features-section). Package managers could then handle enabling the relevant features when building with cargo.

greshake:
Yea... @atheriel You haven't updated the Wiki for the Net block after your PR :)

greshake:
You can check for the dependency when the block is instantiated and then return a blockerror, which will be displayed to the user. Failing entirely if the configuration isn't right is the way to go imho, instead of just skipping the missing dependency.

1wilkens:
Managed to get the information from `iw` similar to SSID.
Right now the bitrate is refreshed every 10 seconds which seemed like a nice balance between performance and responsiveness. :)

atheriel:
How often does the bitrate actually change? I thought it was a feature of the router itself.

1wilkens:
It depends on the router, your wireless hardware and the quality of the connection. At home the router is in the next room and my bitrate changes between 144 and 360 MB/s.
For me its an indicator of link quality that's why I like to see it in the bar, but 30 seconds are maybe also enough.

greshake:

1wilkens:
Hey thanks for merging, I'll update `blocks.md` in the next days :)

Issue #110, opened 2017-11-06
fix alignment of volume icon.


that1communist:
in i3blocks, if i make the code this (I have only tested noto and noto sans.)

```
AUDIO_HIGH_SYMBOL=' '

AUDIO_MED_THRESH=50
AUDIO_MED_SYMBOL='  '

AUDIO_LOW_THRESH=9
AUDIO_LOW_SYMBOL='     '
```

it stays aligned no matter what the % on the side is, except for 100%.

those could be fixed by just making the percentage always 3, i guess.

Also, it would be better if muting was denoted by the block changing color, rather than having an X, because then you know what the volume is going to be when you unmute, this would also fix alignment (but you'd have to base the number of spaces/half spaces on the %, so the work would be semi-duplicated.)

the spacing would have to be 

```
AUDIO_HIGH_SYMBOL='✖ '

AUDIO_MED_THRESH=50
AUDIO_MED_SYMBOL='✖ '

AUDIO_LOW_THRESH=9
AUDIO_LOW_SYMBOL='✖  '
```

Thank you for your time, sorry I can't push the update directly, i don't understand rust at all.

greshake:
I would encourage you to do a pull request. That requires only basic understanding of git :) Also, I don't see that you made changes to Rust code. On a pull request, we can collaborate easier on improving your solution.

that1communist:
I'm having trouble finding where the LOC for the echo of the icons are in the sound block, could you point me to those lines? After that I can make a pull request.

greshake:
I don't quite understand your question. The icons are defined in icons.rs, and are used by calling the `.with_icon()` function of a widget. The `widgets/text.rs` widget takes care of rendering these icons for all blocks.

greshake:
The ones you are interested in are (those from the Awesome icon set, line 61 onwards):
```Rust
        "volume_full" => " \u{f028} ",
        "volume_half" => " \u{f027} ",
        "volume_empty" => " \u{f026} ",
```

that1communist:
Okay, my bad, I assumed it worked similarly to i3blocks for no good reason. I was looking for the icons in the sound block rather than in widgets/text.

thank you.

that1communist:
```
volume_full = "  "
volume_half = "   "
volume_empty = "    "

```

in the config fixes this entirely, except for while muted, I don't know how to make this default.

that1communist:
[...]

PR #112, opened 2017-11-06
Volume icon alignment

that1communist:
#110

start of resolution, this should fix all cases of icon alignment except while muted.

I don't know how to do that part.

I haven't compiled this yet because it should just work. currently am, but i'm on a really slow laptop.


that1communist:
Nope, doesn't work. i'll try again later.

PR #113, opened 2017-11-07
Weather block, take two

atheriel:
This PR makes a second go at adding a weather block proposed in #25, this time using the OpenWeatherMap API (as suggested by @oschoudhury).

I'm sensitive to giving credit for the work done in #25, so the first commit of this PR is actually the commit from that branch, lightly edited when necessary and rebased to the current master so that it should compile. The second commit introduces a parallel OpenWeatherMap implementation.

My current design points to a way to have multiple API backends for this block by adding a variant to the `WeatherService` enum and making using of `#[derive(Deserialize)]`. So to configure the current block for OpenWeatherMap we have

```toml
[[block]]
block = "weather"
service = { name = "openweathermap", api_key = "XXX", city_id = "6167865", units = "metric" }
```

(You can also use the TOML table syntax; I've used the {} syntax for brevity.)

But if someone adds a `WeatherService::DarkSky` variant it would look something like

```toml
[[block]]
block = "weather"
service = { name = "darksky", token = "XXX", lat = 47.23, long = -41.11, units = "SI" }
```

Another API provider could be added in a similar fashion. Serde will decode all of this transparently; a new author would only need to add the enum variant.

The big outstanding question for me on this PR before merging is: should I be trying to fix up the existing DarkSky code? Or can I safely delete it, leaving anyone interested in adding back the functionality to take a look at this branch for a starting point.

Note that I didn't address all of the comments raised in #25 about that code, either -- I can't actually get all the required crates to build on my system, and any newer version of the `darksky` crate has a different API. I'm not really that interested in trying to port it, since the OpenWeatherMap API is more appealing to me anyway.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,3 @@ inotify = "0.5.0"
 # for profiling blocks
 cpuprofiler = "0.0.3"
 progress = "0.2"
-darksky = { version = "0.5.0", optional = true }
-
-[features]
-default = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,3 +24,7 @@ inotify = "0.5.0"
 # for profiling blocks
 cpuprofiler = "0.0.3"
 progress = "0.2"
+darksky = { version = "0.5.0", optional = true }
+
+[features]
+default = []
--- a/src/blocks/mod.rs
+++ b/src/blocks/mod.rs
@@ -16,6 +16,8 @@ mod focused_window;
 mod xrandr;
 mod net;
 mod backlight;
+#[cfg(feature = "darksky")]
+mod weather;
 
 use config::Config;
 use self::time::*;
@@ -36,6 +38,8 @@ use self::temperature::*;
 use self::xrandr::*;
 use self::net::*;
 use self::backlight::*;
+#[cfg(feature = "darksky")]
+use self::weather::*;
 
 use super::block::{Block, ConfigBlock};
 use errors::*;
@@ -85,6 +89,8 @@ pub fn create_block(name: &str, block_config: Value, config: Config, tx_update_r
             "focused_window" => FocusedWindow,
             "xrandr" => Xrandr,
             "net" => Net,
-            "backlight" => Backlight
+            "backlight" => Backlight,
+            #[cfg(feature = "darksky")]
+            "weather" => Weather
     )
 }
--- a/src/blocks/mod.rs
+++ b/src/blocks/mod.rs
@@ -16,7 +16,6 @@ mod focused_window;
 mod xrandr;
 mod net;
 mod backlight;
-#[cfg(feature = "darksky")]
 mod weather;
 
 use config::Config;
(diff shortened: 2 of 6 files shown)


greshake:
Great to see that work not go to waste! It was a really nice PR, and your work does it justice :) However, the DarkSky Terms of Service still stand and I quote:

> You agree that any application or service which incorporates data obtained from the Service shall prominently display the message “Powered by Dark Sky” in a legible manner near the data or any information derived from any data from the Service. This message must, if possible, open a link to https://darksky.net/poweredby/ when clicked or touched. You may not display or invoke the Service or Dark Sky name or logo in any manner that implies a relationship or affiliation with, sponsorship, promotion, or endorsement by Dark Sky, except as authorized by these Terms of Service.

We can't abide by this requirement, so I can't merge any PR which incorporates the DarkSky API. Other services are welcome if we can manage the terms.

atheriel:
Personally I don't think that the DarkSky terms apply in this case, since we do not "incorporate data obtained from the service" in the bar itself; we merely provide an easy way for users to display information they obtain themselves, using their own API keys. But I'm not a lawyer, and I wouldn't take a wager on winning a legal argument. So I'll kill the DarkSky code.

On a related note, the relevant section of the [OpenWeatherMap terms of service](https://openweathermap.org/terms) indicates that:

>  Free weather API is provided under the terms of the Creative Commons Attribution-ShareAlike 4.0 Generic License. You are free to:
> Share — copy and redistribute the material in any medium or format
> Adapt — remix, transform, and build upon the material for any purpose, even commercially.
> The licensor cannot revoke these freedoms as long as you follow the license terms. Under the following terms:
> Attribution — You must give appropriate credit, provide a link to the OpenWeatherMap.org, and indicate if changes were made. You may do so in any reasonable manner, but not in any way that suggests the licensor endorses you or your use.
> ShareAlike — If you remix, transform, or build upon the material, you must distribute your contributions under the same license as the original.
> No additional restrictions — You may not apply legal terms or technological measures that legally restrict others from doing anything the license permits. 

On my reading, our use in the bar should be OK by these terms. I think that it would be sufficient to indicate in the Weather block documentation that data can be retrieved from their API with a (free) API key -- this seems to me to fulfill the "any reasonable manner" criteria specified by the license.

greshake:
If i3status-rust is not a service that "incorporate data obtained from the service", what is? 

I agree OpenWeatherMap should work just fine :)

atheriel:
Ok, I've removed the DarkSky code and made some minor fixes.

I've also modifed the block to use a format template approach, e.g.

```toml
[[block]]
block = "weather"
# service = ...
format = "{weather} ({location}) {temp}°, {wind} km/s"
```
The four keys shown above are all that are available at the moment, but by default it will only show the weather and temperature.

varjolintu:
Nice one. Is there any particular reason you are using curl shell command instead of e.g. curl or hyper crate?

atheriel:
Purely for simplicity. I’m not opposed to using a curl crate, though. Might
make for better error reporting, too (e.g. when not connected to the
internet).

greshake:

PR #114, opened 2017-11-08
Allow customization of marquee interval and speed

ammgws:
I found myself wanting to adjust the parameters for the marquee so I figured we should allow them to be customized.

The default values have been set to the same as they are now, 10s and 0.5s (500000000ns), so this should not change anything for anyone who hasn't explicitly set the values in their config.

I'm new to Rust so I'm wondering if there is a better way to accomplish lines 144-5 & 147-8?

### DIFF
--- a/src/blocks/music.rs
+++ b/src/blocks/music.rs
@@ -8,6 +8,7 @@ use errors::*;
 use scheduler::Task;
 use input::I3BarEvent;
 use block::{Block, ConfigBlock};
+use de::deserialize_duration;
 use widgets::rotatingtext::RotatingTextWidget;
 use widgets::button::ButtonWidget;
 use widget::{I3BarWidget, State};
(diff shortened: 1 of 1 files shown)


greshake:
Hi, welcome to the project :)

src/blocks/music.rs
```diff
@@ -25,6 +26,8 @@ pub struct Music {
     dbus_conn: Connection,
     player_avail: bool,
     marquee: bool,
+    marquee_interval: Duration,
```
greshake: These fields aren't needed, because you can set the values at block 'construction' time in the RotatingTextWidget, as you did. That's also why the compiler is emitting warnings- these fields are never read.

src/blocks/music.rs
```diff
@@ -122,9 +141,11 @@ impl ConfigBlock for Music {
 
         Ok(Music {
             id: id_copy,
+            marquee_interval: block_config.marquee_interval,
```
greshake: Then you also don't need to specify them here

ammgws:
Cheers, tests should pass now. I hope to contribute more once my Rust gets less "rusty"

ammgws:
#124

PR #115, opened 2017-11-10
Use serde_derive to simplify reading more config enums.

atheriel:
This is basically just a small refactor of something I noticed reading through the code. I'm not sure you're accepting this kind of submission yet.

There are a few enums in the DiskSpace and Memory blocks that are read from the configuration file as strings and then converted into enums through some sort of `from_str()` method. But one can just use `#[derive(Deserialize)]` to read into the enums directly. This will also make it easier to add more possible configuration values in the future -- just add a variant to the enum itself.


greshake:
Yea, thanks!

PR #123, opened 2017-12-07
Added total and used info_types, and option to show percentage to disk_space block

varjolintu:
Added implementation to the two missing info_types, `used` and `total`. Also there's a possibility to show a percentage which value depends on the info_type used. The default value for this is false.

With `free` it would show: `<free> (<free_percent>%)`
With `total`: `<used>/<total> (<used_precent>%)`

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -169,6 +169,7 @@ Key | Values | Required | Default
 `info_type` | Currently supported options are available and free | No | `"available"`
 `unit` | Unit that is used to display disk space. Options are MB, MiB, GB and GiB | No | `"GB"`
 `interval` | Update interval, in seconds. | No | `20`
+`show_percentage` | Show percentage of used/available disk space depending on info_type. | No | `false`
 
 ## Focused Window
 
--- a/src/blocks/disk_space.rs
+++ b/src/blocks/disk_space.rs
@@ -16,12 +16,13 @@ extern crate nix;
 
 use self::nix::sys::statvfs::vfs::Statvfs;
 
-#[derive(Copy, Clone, Debug, Deserialize)]
+#[derive(Copy, Clone, Debug, Deserialize, PartialEq, Eq)]
 pub enum Unit {
     MB,
     GB,
     GiB,
     MiB,
+    Percent
 }
 
 impl Unit {
@@ -31,6 +32,7 @@ impl Unit {
             Unit::GB => bytes as f64 / 1000. / 1000. / 1000.,
             Unit::MiB => bytes as f64 / 1024. / 1024.,
             Unit::GiB => bytes as f64 / 1024. / 1024. / 1024.,
+            Unit::Percent => bytes as f64,
         }
     }
 }
@@ -40,9 +42,8 @@ impl Unit {
 pub enum InfoType {
     Available,
     Free,
-    // TODO: implement
-    //Total,
-    //Used,
+    Total,
+    Used,
 }
 
 pub struct DiskSpace {
@@ -55,6 +56,7 @@ pub struct DiskSpace {
     unit: Unit,
     warning: f64,
     alert: f64,
+    show_percentage: bool,
 }
 
 #[derive(Deserialize, Debug, Clone)]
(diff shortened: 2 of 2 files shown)


blocks.md
```diff
@@ -391,4 +392,4 @@ interval = 1800
 Key | Values | Required | Default
 ----|--------|----------|--------
 bytes | weather to use bytes or bits in the display.<br>(true for bytes, false for bits) | No | false
-interval | Update interval in seconds | No | 1800
\ No newline at end of file
+interval | Update interval in seconds | No | 1800
```
atheriel: I think this line is what is causing the merge conflict. Can you remove this change?

src/blocks/disk_space.rs
```diff
@@ -159,7 +166,10 @@ impl Block for DiskSpace {
         let statvfs = Statvfs::for_path(Path::new(self.path.as_str()))
             .block_error("disk_space", "failed to retrieve statvfs")?;
         let result;
-        let converted;
+        let mut converted = 0.0f64;
+		let mut converted_str = String::new();
```
atheriel: Looks like there are some whitespace issues here. Can you stage `rustfmt`-ified versions of the regions you've modified?

src/blocks/disk_space.rs
```diff
@@ -88,6 +88,9 @@ pub struct DiskSpaceConfig {
     #[serde(default = "DiskSpaceConfig::default_alert")]
     pub alert: f64,
 
+    /// Show percentage
+    #[serde(default = "DiskSpaceConfig::default_show_percentage")]
+    pub show_percentage: bool,
```
atheriel: Instead of adding another configuration option, can you implement percentages as another `Unit` type?

atheriel:
Thanks for the contribution, this is a nice addition to have. There are some small formatting issues that I've annotated, and one comment on the implementation: I think it makes more sense to make `percent` available as a unit, so the configuration option would be `unit = "percent"`.

varjolintu:
Thanks for the feedback! I will do the necessary fixes soon.

src/blocks/disk_space.rs
varjolintu: Making it `Unit` type removes the possibility to show both GB and percentage at the same time. Originally I meant that the percentage is just and extra option, not `Unit`type.

atheriel: I'm not quite sure what the right approach is here. This might be good for now, with the possible addition of a `percent` unit in the future so that users could print *only* the percent used, if so desired. @greshake do you have an opinion on this?

varjolintu: Maybe that could be it. Make a new `Unit` type and still keep the `show_percentage` option which is active if the Unit type is not percentage.

varjolintu:
The new Unit type Percent is now available. I also added support for warning/alert limits with that Unit type. So for example using `total` warning/alert can be 70/80 and when using `free` 20/10 etc.

When using the new `show_percentage` option it doesn't affect to the output formatting when using Percent Unit type. Feel free to give feedback. I'm not quite sure if the `compute_state` function is pretty enough.

varjolintu:
Anything else missing from the PR?

blocks.md
```diff
@@ -378,8 +378,12 @@ step_width = 3
 
 Key | Values | Required | Default
 ----|--------|----------|--------
-`step_width` | The percent volume level is increased/decreased for the selected audio device when scrolling. Capped automatically at 50. | No | `5`
-`on_click` | Shell command to run when the sound block is clicked. | No | None
+path | Path to collect information from | No | /
```
greshake: Uhm, this is does not correspond to the Sound block. Maybe you edited the wrong lines?

src/blocks/disk_space.rs
greshake: Your solution is fine with me. Also shouldn't require any breaking changes for existing configs

src/blocks/disk_space.rs
```diff
@@ -158,8 +182,11 @@ impl Block for DiskSpace {
     fn update(&mut self) -> Result<Option<Duration>> {
         let statvfs = Statvfs::for_path(Path::new(self.path.as_str()))
             .block_error("disk_space", "failed to retrieve statvfs")?;
-        let result;
-        let converted;
+        let mut result;
+        let mut converted = 0.0f64;
```
greshake: ```let mut converted: f64;```

greshake: Dont really care but this works, too

blocks.md
varjolintu: This file just keeps messing itself up for some reason.. trying to solve it again.

atheriel:
This looks good to me. Thanks for your patience.

Issue #107, opened 2017-11-02
Clock should open orage (or some equivalent calendar) on left click


that1communist:
Currently does nothing, so, that'd be nice.

greshake:
Relates to https://github.com/greshake/i3status-rust/issues/106

OneMoreByte:
This should be closed

Issue #106, opened 2017-11-02
Sound block should open pavucontrol on left click.


that1communist:
Or make click actions configurable without recompiling.

greshake:
Good idea. Maybe you can open a PR? I'm short on time for new features

that1communist:
I don't really know anything about how that works. I'm sorry I can't be of more help, could you point me in the right direction?

greshake:
Sure. Basically, you need to do the following to add a click handler/ application launch feature to block X:
1. Add a new configuration option for the application to launch (best provide some useful default)
2. Handle the Left Mouse click event in the click handler of block X and launch the program

greshake:
If you're looking for examples, the custom block allows you to execute a command on left-click

OneMoreByte:
Hey, can I try implementing this? I've been wanting to help the project out..

greshake:
Sure, go ahead!

atheriel:
Closed by #128.

PR #128, opened 2017-12-20
Add onclick actions to time and sound blocks

OneMoreByte:
Adds the features @that1communist wanted in the clock and time blocks. So both #106 and #107 

Let me know if there is anything I can improve! I'm fairly new to rust

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -41,11 +41,10 @@ impl SoundDevice {
             .map(|o| String::from_utf8_lossy(&o.stdout).trim().to_owned())
             .block_error("sound", "could not run amixer to get sound info")?;
 
-        let last_line = &output
-            .lines()
-            .into_iter()
-            .last()
-            .block_error("sound", "could not get sound info")?;
+        let last_line = &output.lines().into_iter().last().block_error(
+            "sound",
+            "could not get sound info",
+        )?;
 
         let last = last_line
             .split_whitespace()
@@ -66,14 +65,16 @@ impl SoundDevice {
 
     fn set_volume(&mut self, step: i32) -> Result<()> {
         Command::new("sh")
-            .args(&[
-                "-c",
-                format!(
-                    "amixer set {} {}%",
-                    self.name,
-                    (self.volume as i32 + step) as u32
-                ).as_str(),
-            ])
+            .args(
+                &[
+                    "-c",
+                    format!(
+                        "amixer set {} {}%",
+                        self.name,
+                        (self.volume as i32 + step) as u32
+                    ).as_str(),
+                ],
+            )
             .output()
             .block_error("sound", "failed to set volume")?;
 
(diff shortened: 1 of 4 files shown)


atheriel:
It looks like there's some noise in these commits from `rustfmt`; the changes themselves are more confined. Here are my unsolicited review comments:

1. I would strongly prefer that the command run by the Sound block be configurable in the same way that it is for the Time block, instead of being hard-coded to run `pavucontrol`.
2. You should use `Option<String>` instead of `String` for the command configurations and handle `None`, instead of defaulting to `sleep 0`, which just seems wasteful.
3. I suggest the name `on_click` for the configuration options instead of the more ambiguous `command`.
4. Document any new configuration options for these blocks in `blocks.md`.

OneMoreByte:
Alright, that shouldn't take to long. I'll get those fixed

atheriel:
Your last commit deletes the `blocks.md` file, which I think is unintentional.

src/blocks/sound.rs
```diff
@@ -228,23 +240,46 @@ impl Block for Sound {
     }
 
     fn click(&mut self, e: &I3BarEvent) -> Result<()> {
+
+
         if let Some(ref name) = e.name {
+            let mut command = "".to_string();
```
atheriel: This logic could probably all be moved into the match arm `MouseButton::Left => { ... }` below.

src/blocks/sound.rs
```diff
@@ -115,6 +118,9 @@ pub struct SoundConfig {
     /// The steps volume is in/decreased for the selected audio device (When greater than 50 it gets limited to 50)
     #[serde(default = "SoundConfig::default_step_width")]
     pub step_width: u32,
+
+    #[serde(default = "SoundConfig::default_on_clicked")]
+    pub on_clicked: Option<String>,
```
atheriel: Can you change this to `on_click`, for consistency with the [Custom block](https://github.com/greshake/i3status-rust/blob/master/blocks.md#custom)?

src/blocks/time.rs
```diff
@@ -30,6 +34,9 @@ pub struct TimeConfig {
     /// Update interval in seconds
     #[serde(default = "TimeConfig::default_interval", deserialize_with = "deserialize_duration")]
     pub interval: Duration,
+
+    #[serde(default = "TimeConfig::default_on_clicked")]
+    pub on_clicked: Option<String>,
```
atheriel: Can you change this to `on_click`, for consistency with the [Custom block](https://github.com/greshake/i3status-rust/blob/master/blocks.md#custom)?

OneMoreByte:
Oh geez, I didn't realize you'd get spammed with my git stuff. I'm kinda struggling with git right now. I'll fix that after I get everything on track again

atheriel:
No worries, git lives up to its namesake. If you need suggestions I can try and be helpful.

OneMoreByte:
Alright, I think I got it. Let me know If I broke anything. I'm going to get those things corrected quick

atheriel:
I've merged this manually into master, but unfortunately GitHub doesn't realize it. Thanks!

PR #133, opened 2018-01-11
battery: display remaining time until battery is depleted/charged.

matthiaskrgr:
Screenshots: 
time_charging
discharging

Note: only tested on my device.

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -127,8 +127,6 @@ impl Block for Battery {
             0
         };
 
-
-
         let energy_full = if file_exists(&format!("{}energy_full", self.device_path)) {
              read_file(&format!("{}energy_full", self.device_path))?
                 .parse::<u64>()
@@ -136,7 +134,6 @@ impl Block for Battery {
         } else {
             0
         };
-
 
         let power_now = if file_exists(&format!("{}power_now", self.device_path)) {
             read_file(&format!("{}power_now", self.device_path))?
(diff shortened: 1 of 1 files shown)


atheriel:
Thanks, this is a nice idea for an addition. Do you happen to know how accurate this linear estimate is? Or if other programs estimate charging time differently?

src/blocks/battery.rs
```diff
@@ -120,9 +119,55 @@ impl Block for Battery {
 
         let state = read_file(&format!("{}status", self.device_path))?;
 
+
+        #[allow(unused_assignments)]
+        let mut energy_now = 0;
```
atheriel: I'm not sure this annotation is necessary. This code could probably be refactored to be more idiomatic and remove the need for this annotation, e.g.
```rust
let energy_now = if file_exists ... { ... } else { 0 };
```
The same is true for the other let bindings here, unless I'm very much mistaken. You can use tuple destructuring (e.g. `let (hours, minutes) = if ... { } else { }`) as well to fix the one below.

matthiaskrgr:
>Thanks, this is a nice idea for an addition. Do you happen to know how accurate this linear estimate is?
>Or if other programs estimate charging time differently?

It's in line with what powertop says. I can't remember any other battery widgets right now that seemed to try to calculate some kind of average battery duration.

src/blocks/battery.rs
matthiaskrgr: The ````(hours,minutes)```` thing is a bit more tricky because I reuse ````hours```` when calculating minutes.

src/blocks/battery.rs
```diff
@@ -120,9 +119,52 @@ impl Block for Battery {
...
+                (0,0)
+            }
+        } else {
+            (0,0)
+        };
+
         // Don't need to display a percentage when the battery is full
         if current_percentage != 100 && state != "Full" {
-            self.output.set_text(format!("{}%", current_percentage));
+            self.output.set_text(format!("{}% {}:{:02}", current_percentage, hours, minutes));
```
matthiaskrgr: I removed the comma separator that can be seen in the screenshots between the percentage and the duration.

atheriel:
The changes look good.

The last thing here is the fact that this changes the default display of the battery block, which may be undesirable for some users, and so I think showing the time remaining ought to have its own configuration option. But that's not a reason to hold back this current PR.

PR #137, opened 2018-01-12
readme: fix link to powerline font package in arch repos.

matthiaskrgr:
32bit support was removed from which probably broke the i686 url.


atheriel:
Thanks!

PR #139, opened 2018-01-12
Battery: implement "show" config option to display percentage, time until (dis)charged or both.

### LINKED ISSUE #138: Add a configuration option to customize time/percent display in the Battery block
As mentioned in #133. The easy option here is to have support for e.g. `percent = true` and `time = false` configuration options. The alternative is to implement a `format` configuration option for the block.

matthiaskrgr:
Example:
 ````       show = "percentage"````

Possible values: "both", "time", "percentage".
Default: "both".
    
Closes #138

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -19,6 +19,7 @@ pub struct Battery {
     max_charge: u64,
     update_interval: Duration,
     device_path: String,
+    show: String,
 }
 
 #[derive(Deserialize, Debug, Default, Clone)]
@@ -31,6 +32,10 @@ pub struct BatteryConfig {
     /// Which BAT device in /sys/class/power_supply/ to read from.
     #[serde(default = "BatteryConfig::default_device")]
     pub device: String,
+
+    /// Show only percentage, time until (dis)charged or both
+    #[serde(default = "BatteryConfig::default_show")]
+    pub show: String,
 }
 
 impl BatteryConfig {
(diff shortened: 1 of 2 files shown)


src/blocks/battery.rs
```diff
@@ -161,7 +171,18 @@ impl Block for Battery {
 
         // Don't need to display a percentage when the battery is full
         if current_percentage != 100 && state != "Full" {
-            self.output.set_text(format!("{}% {}:{:02}", current_percentage, hours, minutes));
+            if self.show  == "both" {
```
matthiaskrgr: Hm, I should use ````match```` here

src/blocks/battery.rs
```diff
@@ -31,6 +32,10 @@ pub struct BatteryConfig {
     /// Which BAT device in /sys/class/power_supply/ to read from.
     #[serde(default = "BatteryConfig::default_device")]
     pub device: String,
+
+    /// Show only percentage, time until (dis)charged or both
+    #[serde(default = "BatteryConfig::default_show")]
+    pub show: String,
```
atheriel: Can you implement this as an enum, so it can be handled directly by `serde`? There are some examples of this elsewhere in the code, including [this one](https://github.com/greshake/i3status-rust/blob/master/src/blocks/disk_space.rs#L38-L46). Using an enum would also remove the need to handle the `_` case in the match block below, since it would error out (with a helpful message) when reading the config file, instead.

matthiaskrgr: Good point, thanks!

atheriel:
I'm happy with this implementation. Can you fix the merge conflicts generated by the documentation, or should I?

matthiaskrgr:
rebased

PR #140, opened 2018-01-13
[on hold]  temperature: add option "show =" which accepts "max", "average" and "both" keys.

matthiaskrgr:
I found it annoying that the block would display two temps so I added an option to switch between both, only max and only average.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -394,6 +394,7 @@ Creates a block which displays the system temperature, based on lm_sensors' `sen
 block = "temperature"
 collapsed = false
 interval = 10
+show = "max"
 ```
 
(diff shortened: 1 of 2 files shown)


atheriel:
I've been silent on this because I have an implementation of `format` for this block in another branch, which provides the same functionality. Will get a PR up soon so we can discuss pros/cons of each approach.

shazow:
I submitted a PR which adds a format field for temperature, somewhat related: https://github.com/greshake/i3status-rust/pull/162

atheriel:
Closing this in favour of @shazow's #162.

PR #141, opened 2018-01-13
uptime: add uptime block

matthiaskrgr:
The block will display system uptime. To save space it will only display the 2 biggest units, so weeks and days or days and hours, or hours and minutes or minutes and seconds.

### DIFF
--- a/src/blocks/uptime.rs
+++ b/src/blocks/uptime.rs
@@ -1,18 +1,14 @@
 use std::time::Duration;
 use chan::Sender;
-
 use block::{Block, ConfigBlock};
 use config::Config;
 use de::deserialize_duration;
 use errors::*;
 use widgets::text::TextWidget;
 use widget::I3BarWidget;
-use input::I3BarEvent;
 use scheduler::Task;
-use std::fs::OpenOptions;
-use std::io::prelude::*;
-
 use uuid::Uuid;
+use blocks::lib::*;
 
 pub struct Uptime {
     text: TextWidget,
@@ -52,22 +48,9 @@ impl ConfigBlock for Uptime {
     }
 }
 
-fn read_file(path: &str) -> Result<String> {
-    let mut f = OpenOptions::new()
-        .read(true)
-        .open(path)
-        .block_error("uptime", &format!("failed to open file {}", path))?;
-    let mut content = String::new();
-    f.read_to_string(&mut content)
-        .block_error("uptime", &format!("failed to read {}", path))?;
-    // Removes trailing newline
-    content.pop();
-    Ok(content)
-}
-
 impl Block for Uptime {
     fn update(&mut self) -> Result<Option<Duration>> {
-        let uptime_raw = match read_file("/proc/uptime") {
+        let uptime_raw = match read_file("uptime", "/proc/uptime") {
             Ok(file) => file,
             Err(e) => {
                 return Err(BlockError(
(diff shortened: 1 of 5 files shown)


atheriel:
I'm hesitant to accept this, since it's so easy to accomplish with the Custom block. I have

```toml
[[block]]
block = "custom"
command = "uptime --pretty | sed -E 's/ week(s)?,/w/g; s/ day(s)?,/d/; s/ hour(s)?,/h/; s/ minute(s)?/m/'"
interval = 600
on_click = "true"
```

in my own configuration file, which does almost exactly the same thing.

matthiaskrgr:
A few years back when I was using awesomeWM with a lot of (shell-scripty) custom widgets (same as blocks in i3wm), however in total I was probably calling sed/grep maybe 15 times for a complete panel update which was causing measurable overhead. :/
Implementations in rust should be a lot faster which is why I am trying to port some of the widgets to rust.

greshake:
I agree with this PR, since I was trying to eliminate a lot of the avoidable system calls and overhead by creating i3status-rust and reading the relevant information from the kernel directly, instead of opening a shell process every time. Especially since Meltdown, saving recurring system calls and resources in general is always good. Thanks!

src/blocks/uptime.rs
```diff
@@ -0,0 +1,137 @@
...
+            id: Uuid::new_v4().simple().to_string(),
+            update_interval: block_config.interval,
+            text: TextWidget::new(config.clone()),
+            tx_update_request: tx_update_request,
+            config: config,
+        })
+    }
+}
+
+fn read_file(path: &str) -> Result<String> {
```
greshake: I'm quite sure this is a duplicate. Would you mind finding the original and either putting it into a util file or accessing it instead?

src/blocks/uptime.rs
```diff
@@ -0,0 +1,137 @@
...
+        //debug:  let text = format!("uptime: {}w, {}d, {}h, {}m, {}s", weeks, days, hours, minutes, seconds);
+        self.text.set_text(text);
+        Ok(Some(self.update_interval))
+    }
+
+    fn view(&self) -> Vec<&I3BarWidget> {
+        vec![&self.text]
+    }
+
+    fn click(&mut self, _: &I3BarEvent) -> Result<()> {
```
greshake: Remove the handler entirely. It has a default implementation,

matthiaskrgr: Ok, I copied most of this from template.rs

src/blocks/uptime.rs
matthiaskrgr: Can do, this is from the battery block.

src/blocks/uptime.rs
```diff
@@ -0,0 +1,116 @@
...
+                    "Uptime failed to read uptime string.".to_owned(),
+                ));
+            }
+        };
+
+        let total_seconds = match uptime.parse::<f64>() {
+            Ok(uptime) => uptime as u32,
+            Err(e) => {
+                return Err(BlockError(
+                    "Uptime".to_owned(),
```
atheriel: For consistency, can you downcase this and the error above to match other blocks? And there's no need to prefix the error message itself with "Uptime..." either, since it's displayed to the user already. "Failed to convert /proc/uptime to an integer" would fit the style of other blocks better.

atheriel:
It's fair to make a performance argument here. I had thought it was just shelling out to `uptime`, but it seems that's not the case.

atheriel:
Is there any reason this can't be merged, @greshake?

greshake:
Mh I was going to add the suggestion you mentioned in your review, but I guess its not that important.

PR #142, opened 2018-01-14
cpu: if cpu utilization is below 100% aka in the double digits, try...

matthiaskrgr:
..to pad string with spaces to ensure consistent text length to prevent other widgets from being shoved back and forth all the time.

It seems the default font is not monospace so I could't get it to fit perfectly, however this is already a lot less annoying.

### DIFF
--- a/src/blocks/cpu.rs
+++ b/src/blocks/cpu.rs
@@ -134,8 +134,11 @@ impl Block for Cpu {
             _ => State::Idle,
         });
 
-        self.utilization.set_text(format!("{:02}%", utilization));
-
+        if utilization < 100  {// 2 digits, pad with extra space to ensure consistent length
+            self.utilization.set_text(format!("  {:02}%", utilization));
+        } else {
+            self.utilization.set_text(format!("{:02}%", utilization));
+        }
         Ok(Some(self.update_interval))
     }
 


src/blocks/cpu.rs
```diff
@@ -134,8 +134,11 @@ impl Block for Cpu {
             _ => State::Idle,
         });
 
-        self.utilization.set_text(format!("{:02}%", utilization));
-
+        if utilization < 100  {// 2 digits, pad with extra space to ensure consistent length
+            self.utilization.set_text(format!("  {:02}%", utilization));
+        } else {
+            self.utilization.set_text(format!("{:02}%", utilization));
```
atheriel: Why not `"{:03}%"` instead? Wouldn't that do the same thing?

matthiaskrgr: Well, it will pad with a zero instead of spaces, but ````099%````, ````055%```` or ````011%```` looks a bit stupid imo.

greshake: I'd prefer to have less duplication here, we can use an expression for the optional space instead of an surrounding if:
```Rust
format!(" {}{:02}%", if utilization < 100 {" "} else {""}, utilization)
```

greshake: Nevermind. `format!("{:3}%", utilization)` is what we're looking for

atheriel: Oops, my bad. That's the one I meant.

matthiaskrgr: ````{:3}```` makes jumping even worse here... :/ 
I guess a space does not have the same width as number in the default font?

greshake: Then just use a Monospace font

matthiaskrgr: Well I do use ````      font pango:DejaVu Sans Mono, FontAwesome 12```` as suggested in the readme.md...
I wonder what is happening there.

PR #146, opened 2018-01-17
memory: add format variants of precentages as integers (without decimal points).

matthiaskrgr:
IMO knowing that "33%" of memory is in use is enough and knowing that "33.42%" memory is used does not add that much value so I added integer-variants of percentage formats. ````{MAp}```` gets an additional ````{MApi}```` format, etc...


greshake:
Thanks, sorry for the delay :(

PR #147, opened 2018-01-17
cpu block: add optional cpu frequency display

matthiaskrgr:
Adds option to display average cpu core frequency in cpu block


greshake:
Thanks :+1:

Issue #96, opened 2017-10-20
Battery module does not find /sys/class/power_supply/BAT0/charge_full


xdnicoco:
I get an `Error in block 'battery': failed to open file /sys/class/power_supply/BAT0/charge_full` when trying to add a battery block.
Using arch linux 4.13.7-1-ARCH on lenovo yoga 700. 
Result of `ls -l /sys/class/power_supply/BAT0`:
```-rw-r--r-- 1 root root 4096 Oct 20 10:39 alarm
-r--r--r-- 1 root root 4096 Oct 20 10:39 capacity
-r--r--r-- 1 root root 4096 Oct 20 10:39 capacity_level
-r--r--r-- 1 root root 4096 Oct 20 10:39 cycle_count
lrwxrwxrwx 1 root root    0 Oct 20 10:39 device -> ../../../PNP0C0A:00
-r--r--r-- 1 root root 4096 Oct 20 10:39 energy_full
-r--r--r-- 1 root root 4096 Oct 20 10:39 energy_full_design
-r--r--r-- 1 root root 4096 Oct 20 10:39 energy_now
-r--r--r-- 1 root root 4096 Oct 20 10:39 manufacturer
-r--r--r-- 1 root root 4096 Oct 20 10:39 model_name
drwxr-xr-x 2 root root    0 Oct 20 10:39 power
-r--r--r-- 1 root root 4096 Oct 20 10:39 power_now
-r--r--r-- 1 root root 4096 Oct 20 10:39 present
-r--r--r-- 1 root root 4096 Oct 20 10:39 serial_number
-r--r--r-- 1 root root 4096 Oct 20 10:39 status
lrwxrwxrwx 1 root root    0 Oct 20 09:56 subsystem -> ../../../../../../../../../class/power_supply
-r--r--r-- 1 root root 4096 Oct 20 10:39 technology
-r--r--r-- 1 root root 4096 Oct 20 10:39 type
-rw-r--r-- 1 root root 4096 Oct 20 10:39 uevent
-r--r--r-- 1 root root 4096 Oct 20 10:39 voltage_min_design
-r--r--r-- 1 root root 4096 Oct 20 10:39 voltage_now```

greshake:
https://github.com/greshake/i3status-rust/issues/16

greshake:
It should default to capacity, which you have...

atheriel:
Having the same issue here, too. I have a `capacity` file but no `charge_full` file.

[...]

Issue #16, opened 2017-05-20
Bug: BAT0/charge_full can't be read


keeslinp:
My batter doesn't support doing charge_full but BAT0/capacity does work (exactly the same as far as I can tell). How would you feel about me adding an option so that someone can specify the proper endpoint. Alternatively if there are only a few proper endpoints across the majority of devices it can just detect which one works.

greshake:
Phew... Capacity actually reports finished battery percentage for me. But, if the battery is on ac and full reports over 100 (since my battery doesn't charge up to full capacity anymore). Can you look for docs or something showing that capacity is actually better to read from? Does your capacity report mAh or a percentage?

keeslinp:
I verified that the /capacity does change as the battery goes down so that's a good sign. I couldn't find anything specific about whether one was better than the other though.

greshake:
I went to the official docs: https://www.kernel.org/doc/Documentation/power/power_supply_class.txt
```
Attributes/properties detailed
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

~ ~ ~ ~ ~ ~ ~  Charge/Energy/Capacity - how to not confuse  ~ ~ ~ ~ ~ ~ ~
~                                                                       ~
~ Because both "charge" (ÂµAh) and "energy" (ÂµWh) represents "capacity"  ~
~ of battery, this class distinguish these terms. Don't mix them!       ~
~                                                                       ~
~ CHARGE_* attributes represents capacity in ÂµAh only.                  ~
~ ENERGY_* attributes represents capacity in ÂµWh only.                  ~
~ CAPACITY attribute represents capacity in *percents*, from 0 to 100.  ~
~                                                                       ~
~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~
[...]
CAPACITY - capacity in percents.
CAPACITY_ALERT_MIN - minimum capacity alert value in percents.
CAPACITY_ALERT_MAX - maximum capacity alert value in percents.
CAPACITY_LEVEL - capacity level. This corresponds to
POWER_SUPPLY_CAPACITY_LEVEL_*.
[...]
Q: Suppose, my battery monitoring chip/firmware does not provides capacity
   in percents, but provides charge_{now,full,empty}. Should I calculate
[...]

PR #149, opened 2018-01-29
Added support for energy

Miniwoffer:
- added another else if to support energy_full/_current

A possible fix to #96, but given i know little about linuxs power supply class there might be a better fix.


greshake:
See https://github.com/greshake/i3status-rust/issues/16.
Problem is that I wasn't sure we can infer correct or accurate values from the energy/charge attributes, as the documentation warns. But since we did implement this simple calculation for charge, I'll let this pass. However, it'd be great if someone with a battery that provides energy attributes checks if they are kinda accurate.

Cheers

Issue #158, opened 2018-03-07
Use $SHELL for on_click


lausek:
At the moment, `sh` is hardcoded for execution of custom commands

https://github.com/greshake/i3status-rust/blob/881979bd7f356465d5e84fe898db08705309bab5/src/blocks/custom.rs#L118

I think that the `$SHELL` variable should be used here. That would also allow calling user defined commands like `fish` functions.

ammgws:
Second this. Currently forced to explicitly call `fish` like so:

```
[[block]]
block = "custom"
command = "fish --command=ime_status"
on_click = "fish --command=ibus-toggle"
```

atheriel:
On the face of things this makes sense to me, but I'd like to verify that `$SHELL` will always be what users expect it to be, and handle cases where the variable is empty.

lausek:
@atheriel How should those cases be handled?

bemeurer:
@atheriel `$SHELL` is set, at least on Linux, with `chsh`, and is visible to all user processes, as far as I know. The `env::var()` call used in @lausek's implementation should work flawlessly, at least on Linux. I'm going to test it on FreeBSD and see what happens.

lausek:
@atheriel As @bemeurer said, you can use `chsh` to set the `$SHELL` variable.

Regarding your second comment, are you thinking of something specific that `sh` can do but e.g. `bash` can't?

atheriel:
I'm happy enough with the `chsh` answer (TIL!), but to clarify on the second, yes, I am wondering if assuming `bash`/`fish` will act like `sh` is a mistake. On doing some research I think this is probably OK:

According to [this Stackoverflow answer](https://stackoverflow.com/questions/5725296/difference-between-sh-and-bash#answer-5725402), `bash` is not a POSIX-compliant implementation of `sh`, but the behaviour of `bash` is so similar as to be symlinked in most distributions. I assume the same is true of `fish`. So it's not unreasonable to expect that users who set `$SHELL` can be expected to ensure that their shell choice is compatible with `sh`, and we can safely merge #159!

bemeurer:
@atheriel Watch out, also `bash` is quasi-POSIX compliant, the same is **not** true for `fish`. Fish is *extremely* non-POSIX, it's perhaps the most non-POSIX shell that is actually used by people, and most, if not all, of your csh/bash/POSIX-shell scripts will *not* run on Fish without considerable patching.

[...]

PR #159, opened 2018-03-07
Send custom commands to $SHELL

lausek:
This implements the feature requested in #158. If `$SHELL` was not set, we'll fall back to `sh`.

### DIFF
--- a/src/blocks/custom.rs
+++ b/src/blocks/custom.rs
@@ -116,10 +116,8 @@ impl Block for Custom {
         let mut update = false;
 
         if let Some(ref on_click) = self.on_click {
-            match env::var("SHELL") {
-                Ok(name) => Command::new(name), 
-                _ => Command::new("sh")
-            }.args(&["-c", on_click]).output().ok();
+            Command::new(env::var("SHELL").unwrap_or_else(|_| "sh".to_string()))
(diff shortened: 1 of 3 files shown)


src/blocks/custom.rs
```diff
@@ -115,7 +116,8 @@ impl Block for Custom {
         let mut update = false;
 
         if let Some(ref on_click) = self.on_click {
-            Command::new("sh").args(&["-c", on_click]).output().ok();
+            Command::new(env::var("SHELL").unwrap_or_else(|_| "sh".to_owned()))
```
RealOrangeOne: As it's intended to use a literal here, simply using `unwrap_or` here might be better?

lausek: Oh, yeah sure. Pretty new to this so I always appreciate comments ^^

atheriel:
This implementation looks good to me, but I'm waiting on a response to my comments in the issue to before merging.

Also, if you have the time/energy, can you add a note to the `blocks.md` entry for this block about the new behaviour around `$SHELL`? (I won't block the merge on that request, however.)

atheriel:
Looks good to me. Thanks! And my thanks to @RealOrangeOne for reviewing the initial code, too.

PR #162, opened 2018-03-22
blocks/temperature: Add format field

shazow:
This adds an optional `format` field to the `temperature` block.

The default format remains the same as before, so it should be fully backwards compatible. The format code was borrowed from the `load` block implementation.

The goal was to allow more compact temperature strings like:

`format = "{min}~{max}°"`

(particularly since I don't care much about the average.)

I'm a beginner rustacean, so please excuse any code faux pas. Feedback is welcome. I am particularly unsure if there is a better way to get the min/max/avg in one pass (instead of 3).

### DIFF
--- a/src/blocks/temperature.rs
+++ b/src/blocks/temperature.rs
@@ -1,5 +1,6 @@
 use std::time::Duration;
 use std::process::Command;
+use util::FormatTemplate;
 use chan::Sender;
 use scheduler::Task;
 
@@ -19,7 +20,28 @@ pub struct Temperature {
     collapsed: bool,
     id: String,
     update_interval: Duration,
+    format: FormatTemplate,
 }
+
+#[derive(Deserialize, Debug, Default, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct LoadConfig {
+    #[serde(default = "LoadConfig::default_format")]
+    pub format: String,
+    #[serde(default = "LoadConfig::default_interval", deserialize_with = "deserialize_duration")]
+    pub interval: Duration,
+}
+
+impl LoadConfig {
+    fn default_format() -> String {
+        "{avg}° avg, {max}° max".to_owned()
+    }
+
+    fn default_interval() -> Duration {
+        Duration::from_secs(5)
+    }
+}
+
 
 #[derive(Deserialize, Debug, Default, Clone)]
 #[serde(deny_unknown_fields)]
@@ -31,6 +53,10 @@ pub struct TemperatureConfig {
     /// Collapsed by default?
     #[serde(default = "TemperatureConfig::default_collapsed")]
     pub collapsed: bool,
+
+    /// Format override
+    #[serde(default = "LoadConfig::default_format")]
+    pub format: String,
 }
 
 impl TemperatureConfig {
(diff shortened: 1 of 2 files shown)


src/blocks/temperature.rs
```diff
@@ -54,6 +80,8 @@ impl ConfigBlock for Temperature {
             output: String::new(),
             collapsed: block_config.collapsed,
             id,
+            format: FormatTemplate::from_string(block_config.format)
+                .block_error("load", "Invalid format specified for temperature")?,
```
atheriel: "load" should be "temperature" here.

src/blocks/temperature.rs
```diff
@@ -101,9 +129,17 @@ impl Block for Temperature {
...
                 .max()
                 .block_error("temperature", "failed to get max temperature")?;
+            let min: i64 = *temperatures
+                .iter()
+                .min()
+                .block_error("temperature", "failed to get min temperature")?;
             let avg: i64 = (temperatures.iter().sum::<i64>() as f64 / temperatures.len() as f64).round() as i64;
 
-            self.output = format!("{}° avg, {}° max", avg, max);
+            let values = map!("{avg}" => avg,
```
atheriel: Can you use `{average}` instead of `{avg}`? I'd like to keep away from abbreviations in most cases.

atheriel:
Hi, sorry for the delay. This patch looks great, I've just left some minor comments. To answer your query: while I think it would be possible to write a custom map to compute the min, max, and sum in one pass, but I'm not sure this optimization would yield much gain.

If you are comfortable doing so, can you add documentation for the new `format` parameter to the `blocks.md` file?

src/blocks/temperature.rs
shazow: Do we want `{minimum}` and `{maximum}` also?

src/blocks/temperature.rs
shazow: Good catch, will update shortly.

shazow:
> If you are comfortable doing so, can you add documentation for the new format parameter to the blocks.md file?

Will do, I'll ping when the changes are pushed.

shazow:
@atheriel Pushed the changes you requested, please take another look.

Personally I feel `{average}` looks weird next to `{min}` and `{max}` (I'd lean to keeping it as `{avg}`), but your call.

src/blocks/temperature.rs
```diff
@@ -19,8 +20,29 @@ pub struct Temperature {
...
+    #[serde(default = "LoadConfig::default_interval", deserialize_with = "deserialize_duration")]
+    pub interval: Duration,
+}
+
+impl LoadConfig {
+    fn default_format() -> String {
+        "{average}° avg, {max}° max".to_owned()
+    }
+
+    fn default_interval() -> Duration {
```
shazow: Actually I think this is redundant?

shazow: Ah woops that's a leftover from Load, fixing.

shazow: :+1:

atheriel:
:+1:

I think that the "min" and "max" abbreviations are commonplace, while "avg" is not, which is why I prefer `{average}`. Perhaps the correct thing to do is use `{mean}` here -- which is both shorter and less colloquial. I'm happy to leave that up to you. Otherwise this is ready to merge IMO.

shazow:
@atheriel I'm happy to keep it as-is. I don't have write access (I guess technically the review wasn't set to approved yet), so please merge at will! :)

Issue #165, opened 2018-04-13
Create protocol for Custom block to set state


BrendanBall:
from https://github.com/greshake/i3status-rust/issues/4
I mentioned it would be great if the Custom block, ie. Scriptable Block to execute shell commands can do more than just set the text of the output.

Basically I want the command to determine what state it's in, and therefore need to create some kind of protocol. I've so far just gone and implemented the most basic version possible using json. see https://github.com/BrendanBall/i3status-rust/commit/fc6817aef16b7221566ffde35ceb67f542d0b4b2 . It allows you to print out json for your custom command to change the state. see below:
```
[[block]]
block = "custom"
command = "echo '{\"state\":\"Critical\", \"text\": \"beep\"}'"
interval = 100
```

This is all I need for now so I'll just use my fork for now, but if you guys want to put more thought into this that'd be great. If we actually want to add this feature, I would put this protocol behind a feature flag so that not every custom command you want to run has to output json or whatever protocol we choose. So give the option to either just consume text or consume the protocol per custom block.

ps. I also haven't really written much rust yet, so might need some guidance if we're going to put more effort into this.

atheriel:
Alteratively, we could add a new custom block that accepts this JSON protocol. That way we don't have to check if input is valid JSON before just treating it as text every time its updated.

BrendanBall:
That sounds like a really good idea. Any idea what to call it? JSONPlugin? I don't know if we'd ever want to support multiple protocols, but if you do then every protocol can have its own block? Or just call it Plugin?

ohk2kt3t4:
@BrendanBall Are you still using the plugin block you came up with in #166?

I was wondering if we can implement this as part of the `custom` block instead of creating a new one.
For example if you add a new option to the `custom` block called `output_format` with choices `standard` (default) and `json`. When json is selected then apply the logic from #166, otherwise block behaviour stays the same as now.

This allows solutions for issues like #414 and #454 as those users could now use the JSON option to further customise their custom block.

@jcgruenhage You mentioned wanting to "a command that returns a JSON blob that contains the text, icon and state for the text widget". The solution above covers at least the text and state, can you see if support for setting the icon can be trivially implemented as well?

jcgruenhage:
[...]

PR #166, opened 2018-04-15
add plugin block

BrendanBall:
plugin block is the same as custom block except that it implements a
basic json protocol to allow the custom command to set the state.
see https://github.com/greshake/i3status-rust/issues/165


ammgws:
@BrendanBall 
This feature is now implemented via the `json` option of the Custom block.

PR #168, opened 2018-04-18
battery block: remove assignment and comment regarding rust 1.20, rust 1.25 is stabel now

matthiaskrgr:
Didn't see any compiler warning with stable.


atheriel:
Yes, this is probably safe to remove now. If anyone has an objection (ie. they are stuck on nightly circa last August), now is the time to chime in.

PR #170, opened 2018-04-20
battery block: add support for having multiple batteries in one block.

matthiaskrgr:
While implementing this, I assumed that we could not have one battery charging while another one discharges.

We can now specify
device = "BAT0, BAT1"
in status.toml.
This will make the battery block display percentage seperately but calculate expected runtime taking both batteries into account.

Previously we could use one battery block for each battery but then it would only display the expected runtime for the battery that is currently discharging and not "total" runtime.
battery_block

If only one battery is present, nothing should change from previous behaviour.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -79,6 +79,15 @@ interval = 10
 show = "both"
 ```
 
+If you have several batteries in your laptop, instead of using one block per battery,
+you can specify them like so to display both battery percentages but only one total expected runtime:
+```toml
+[[block]]
+block = "battery"
+show = "both"
+device = "BAT0, BAT1"
+```
+
 ### Options
 
 Key | Values | Required | Default
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -1,22 +1,29 @@
-use std::time::Duration;
-use chan::Sender;
-use scheduler::Task;
 use block::{Block, ConfigBlock};
+use blocks::lib::*;
+use chan::Sender;
 use config::Config;
 use de::deserialize_duration;
 use errors::*;
-use widgets::text::TextWidget;
-use widget::{I3BarWidget, State};
+use scheduler::Task;
+use std::time::Duration;
 use uuid::Uuid;
-use blocks::lib::*;
+use widget::{I3BarWidget, State};
+use widgets::text::TextWidget;
 
 pub struct Battery {
     output: TextWidget,
     id: String,
     max_charge: u64,
     update_interval: Duration,
-    device_path: String,
+    devices: String,
     show: ShowType,
+}
+
+#[derive(PartialEq, Eq)]
+enum DeviceState {
+    Charging,
+    Neutral,
+    Discharging,
 }
 
 #[derive(Deserialize, Copy, Clone, Debug)]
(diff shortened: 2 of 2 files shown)


src/blocks/battery.rs
```diff
@@ -66,107 +73,198 @@ impl ConfigBlock for Battery {
...
+        // these calculations might get wrong if one battery charges while one discharges
+        // however I have not seen this yet, so let's just assume this does not happen and both
+        // batteries are are always in similar state (one charging or discharging and others idle/unknown)
 
-        let (hours, minutes) = if power_now > 0 && energy_now > 0 {
-            if state == "Discharging" {
-                let h = (energy_now as f64 / power_now as f64) as u64;
-                let m = (((energy_now as f64 / power_now as f64) - h as f64) * 60.0) as u64;
+        let (hours, minutes) = if total_power_now > 0 && total_energy_now > 0 {
+            if state < 0 {
```
matthiaskrgr: I should use the enum here as well...

atheriel:
What's the advantage to this over having two battery blocks? Is it purely the better charge/discharge time estimate?

matthiaskrgr:
The estimate was my main motivation, but it's also slightly more space (i3 panel space) efficient.
both

blocks.md
```diff
@@ -79,6 +79,15 @@ interval = 10
 show = "both"
 ```
 
+If you have several batteries in your laptop, instead of using one block per battery,
+you can specify them like so to display both battery percentages but only one total expected runtime:
+```toml
+[[block]]
+block = "battery"
+show = "both"
+device = "BAT0, BAT1"
```
atheriel: Is it possible for this to become a TOML list like `["BAT0", "BAT1"]` instead? Or would that break existing configurations?

src/blocks/battery.rs
```diff
@@ -66,107 +73,198 @@ impl ConfigBlock for Battery {
...
+                state -= 1;
+            } else if battery.state == "Unknown" {
+                // a battery may be Unknown when it's just sitting there not doing anything while the
+                // other one is being (dis/)charged
+                state += 0;
+            }
+        }
+
+        let device_state: DeviceState;
+        if state == 0 {
```
atheriel: Can we use `match` here instead? It would obviate the need for `unreachable!()`.

matthiaskrgr: done

droslean:
That is a really good idea, but I don't like the approach. Showing both of the batteries is not good in the calculation. I guess the we will have to calculate a general percentage for the whole batteries capacity and output a single battery in the status bar. We can check how gnome are already doing this.

matthiaskrgr:
It have hot-swappable batteries which means as soon as the external one is depleted, I could remove it (the internal one will take over for that moment) and insert another backup battery, thus it makes sense to see both percentages side by side.

droslean:
In that case, your solution makes sense. But what if there are 2 internal batteries ?

droslean:
By the way. In this PR the state of battery does not change.
Can you check fix this ?

matthiaskrgr:
Uhm, what do you mean with the state of the battery does not change?
I have been using this in my own fork since a few months and didn't notice any problems.

matthiaskrgr:
Ok, there's another nasty can of merge conflicts, I don't care anymore.

droslean:
For example, when your batteries are almost empty, the state doesn't change and you still see the green color.
Also, your PR needs rebase. Please take a look. it will be nice to have this feature.

PR #171, opened 2018-04-21
Block for nvidia gpu.

lgAshanl:
For proprietary nvidia drivers. Has no dependencies on xorg.


greshake:
Very nice! Thanks :+1:

atheriel:
This looks great!

blocks.md
```diff
@@ -338,6 +338,40 @@ Key | Values | Required | Default
...
+```
+
+### Options
+
+Key | Values | Required | Default
+----|--------|----------|--------
+`gpu_id` | GPU id in system | No | 0
+`label` | Display custom gpu label | No | ""
+`interval` | Update interval, in seconds. | No | `1`
+`utilization` | Display gpu utilization. In percents. | No | `true`
```
atheriel: Should these be `show_utilization`, `show_memory`, `show_temperature`, `show_fan_speed`, and `show_clocks` instead? That would fit the style of other blocks more closely.

lgAshanl: I have not seen this in other blocks, but this is logical.

atheriel: I should clarify: the style we *aspire* to with other blocks. Thanks for the changes!

PR #174, opened 2018-04-23
change formatting line width from 100 to 200.

matthiaskrgr:
Not sure why 200 was chosen, but it looks *really* bad imo, and since my editor auto-runs rustfmt it would constantly reflow code of a file I am editing.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,2 +1,2 @@
-max_width = 200
+max_width = 100
 write_mode = "Overwrite"


atheriel:
Please note that "It looks really bad imo" is not a particularly helpful way to frame a wide-ranging change of this kind. For the original rationale for the existing style, we can ask @svmnotn (who wrote the original configuration).

Also, since 100 [appears to be the default](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#max_width), we could simply remove the entry instead.

Finally, "max width" seems to me to be for forcing the wrapping of very long lines (typically because they contain very long type signatures, or in our case, equations). It is not a license to *write* code so that it will produce long lines, which I think is what you find so visually unappealing.

matthiaskrgr:
Well, 36 of 38 rust files of this project seem to not honor the current coding style as suggested by the file.
Formatting to the current rustfmt manifest or to the rustfmt default is going to be wide-ranging either way which is why I tried to update the formatting of block as I do changes to it.

What do you suggest? Should we just remove the rustfmt config file and use whatever is rustfmts default?

svmnotn:
I think the 200 was because some comments were being wrapped weirdly, try removing it, running rustfmt, and compare the changes, perhaps make a pr and everyone can check if the code looks better with the defaults.

atheriel:
Ok, I think if comments are the motivating issue we should be looking to control that [separately instead](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#comment_width). Otherwise I support moving to the default 100 for line width.

As for actually adhering to rustfmt: so far in practice we've only really been updating the format of code as it has actively been worked on, which is not a bad policy. I know this is results in ugly inconsistency, but it does have the advantage of preserving git history on line changes more clearly, and it is not unusual in open-source projects to favour this over style adherance.

But I wouldn't mind a single, one-time "rustfmt entire repository" change, either. @greshake, do you have an opinion on this?

greshake:
I'm ok with some code style changes.

matthiaskrgr:
Ok, I made a showcase branch with 2 commits: https://github.com/matthiaskrgr/i3status-rust/commits/cargo_fmt_default

This is the diff from current formatting to current rustfmt.toml 
https://github.com/matthiaskrgr/i3status-rust/commit/45954d03c5604f64fb4731d161152bbeb5b763d8

This is the diff from current rustfmt.toml to rustfmt default: 
https://github.com/matthiaskrgr/i3status-rust/commit/50e2d4e0dafd00457c7d5954a0a59bb0a6584b47

This is the diff from current formatting to rustfmt default:
https://github.com/greshake/i3status-rust/compare/master...matthiaskrgr:cargo_fmt_default

svmnotn:
rustfmt default doesn't look bad, and I think it will make vscode's format on save not make tons of changes...

Issue #172, opened 2018-04-22
don't use sed/grep/awk in blocks


matthiaskrgr:
It seems there are some blocks using external commands for string manipulation where we could use native rust instead:

sed is used in net block
grep is used in net and xrandr block
awk is used in net block

atheriel:
Is there an advantage to replacing this code? Do you think it will e.g. be more performant?

matthiaskrgr:
1) users might no have some of the external programs installed
2) most likely using rust is going to be more performant since when we use an external program, starting up the program, have it parse its args and take in stdin and write to stdout etc all takes up resources which we can avoid by using rust
3) using native rust makes the execuded code honor local optimization levels (opt-level=3/z, target-cpu=native and gives us more control
4) if one of the programs happens to change its output, it might break the block without us noticing

atheriel:
I don't see why (3) would matter or using Rust would help us with (4). And on (1), note that all three of `grep`, `sed`, and `awk` are part of POSIX, so they will always be available on the platforms we target.

But (2) makes sense to me.

I think what we would really be trading off here is debugability for performance. At the moment you can paste complete commands into a shell and see what the output looks like. All of the external command parsing is highly localized. This has been immensely helpful for debugging user issues and finding corner cases during development. If we go down this route and do all `grep`/`sed`/`awk`-like transformations in Rust, in order to diagnose problems in the future we will likely be recreating the Rust-based transformations with these programs *anyway* so that, again, we can paste them into a shell (or ask users to do so).

In light of that, I don't think this is a good idea.

greshake:
As I've already said here multiple times, I'm super in favor of native Rust implementation of everything. In these cases however it is a minor thing right now. We have to execute a subshell anyway, and all of these commands are extremely performant and optimized, Rust probably only has a minor edge over them. In the long run, this is surely something to improve on. @matthiaskrgr If you feel so strongly about this, you should make a PR. Its not a priority for us to implement this ourselves right now

PR #175, opened 2018-04-23
[wip] network block: use rust code instead of grep and sed, fix clippy lints 

matthiaskrgr:
Part of resolving #172.
This branch also contains #174 and a rustfmt run on the block because I wanted to have leave the formatting changes out of the refactoring commits.

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -91,24 +91,28 @@ impl NetworkDevice {
             ));
         }
         let mut iw_output = Command::new("sh")
-            .args(&[
-                "-c",
-                &format!(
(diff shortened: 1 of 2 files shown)


src/blocks/net.rs
```diff
@@ -118,15 +122,7 @@ impl NetworkDevice {
...
-            .args(
-                &[
-                    "-c",
-                    &format!(
-                        "ip -oneline -family inet address show {} | sed -rn \"s/.*inet ([\\.0-9/]+).*/\\1/p\"",
-                        self.device
-                    ),
-                ],
-            )
+            .args(&["-c", &format!("ip -oneline -family inet address show {} | sed -rn \"s/.*inet ([\\.0-9/]+).*/\\1/p\"", self.device)])
```
matthiaskrgr: ah, looks like I missed this...

PR #177, opened 2018-04-29
blocks/toggle: Add icon overrides, make text optional

shazow:
Allow the user to customize the toggle button icon so that the icon can
represent what the various toggle blocks are doing instead of having to
rely on labels.

By default, the behaviour remains the same and backwards-compatible with older configs.

I currently have multiple toggles in my status bar and they're taking up a lot of space for both the toggle icon and the label, so this allows for many compact toggles without labels.

Here's an example of one I'm using for toggling redshift:

```toml
[[block]]
block = "toggle"
command_state = "systemctl --user -q is-active redshift && echo on"
command_on = "systemctl --user start redshift"
command_off = "systemctl --user stop redshift"
icon_on = "backlight_empty"
icon_off = "backlight_full"
interval = 60
```

### DIFF
--- a/src/blocks/toggle.rs
+++ b/src/blocks/toggle.rs
@@ -18,6 +18,8 @@ pub struct Toggle {
     command_on: String,
     command_off: String,
     command_state: String,
+    icon_on: String,
+    icon_off: String,
     update_interval: Option<Duration>,
     toggled: bool,
     id: String,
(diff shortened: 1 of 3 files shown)


src/blocks/toggle.rs
```diff
@@ -39,10 +41,33 @@ pub struct ToggleConfig {
...
+    #[serde(default = "ToggleConfig::default_icon_on")]
+    pub icon_on: String,
+
+    /// Icon ID when toggled off (default is "toggle_off")
+    #[serde(default = "ToggleConfig::default_icon_off")]
+    pub icon_off: String,
+
     /// Text to display in i3bar for this block
+    #[serde(default = "ToggleConfig::default_text")]
     pub text: String,
```
atheriel: Would it make sense to move to using an `Option<String>` instead? Serde should handle things correctly at the TOML level if so.

atheriel:
I like this change. Also I'm totally stealing your example.

src/blocks/toggle.rs
shazow: I was looking for a way to do that! Thanks for the pointer. I'll play around with that.

shazow:
@atheriel Nice! Just need to find a better font-awesome icon override so it doesn't collide with my brightness icon.

Btw just noticed you're a fellow Torontonianite. We need an i3status-rust meetup/hackathon. :D

atheriel:
@shazow Yes, that sounds like fun. It seems that the Toronto Rust community has been pretty quiet of late, even though Mozilla has an office here and there are several prominent Rust community members in the area.

src/blocks/toggle.rs
shazow: @atheriel PTAL at https://github.com/greshake/i3status-rust/pull/177/commits/49768cc2f001123017376c2049829499e727e97d, feels a bit better but not sure about the `with_content` builder.

shazow:
@atheriel Please take another look when you have a chance. :)

shazow:
@atheriel Poke poke.

shazow:
Ping ping. :)

atheriel:
Whoops, this totally fell off my radar. Thanks for your patience.

shazow:
Np, thanks for getting to it! :)

Issue #167, opened 2018-04-17
Check if execution was successful in toggle block


lausek:
The toggle block doesn't check if the toggling action itself was successful. For example, I added a block which lets me switch to my headset instead of my speakers. If the headset is unplugged, the switch can't happen and the buttons state gets inconsistent.

Therefore, I propose to add a check on the output of `command_on` and `command_off`. Possible fixes:
- Just like `command_state`, initial output means `false` - filled means `true`
- `command_state` must output something like `=true|false` as last line. If this line doesn't exist, we fallback on empty check, so this shouldn't break existing configs
- Check on exit code (*Why doesn't the current implementation use that as indicator?*)

Maybe we could also add a slow cool down in which the background color turns to error state, so the user gets feedback immediately. 

**Note:** This could break existing modifications, but updating has to be done manually atm

lausek:
I modified my fork for this [issue](https://github.com/lausek/i3status-rust/commit/f8b9ca9556e71665e0634e2a37d24ff8113b8182).

atheriel:
Of the proposed options, checking the exit code makes the most sense to me, both from a Unix-y perspective and an implementation one. I worry about the potential cost of supporting ad-hoc input protocols like the ones you propose, even if they are sensible in this particular use case.

lausek:
@atheriel In my understanding error code `1` could also be seen as an internal error. What if the script fails because of some logic inside?

lausek:
@atheriel anything wrong with the request?

ammgws:
Implemented via #648

PR #180, opened 2018-05-03
implement enhancement #167

lausek:
Check if the command was executed correctly before changing the visual state of the toggle button


ammgws:
Implemented via #648

PR #181, opened 2018-05-03
blocks: add pacman2 block.

### LINKED ISSUE #179: pacman block v2 idea dump
instead of calling pacman in fakeroot env, I believe we can simply
1) use "checkupdates" to check for official package updates (does not require fakeroot)
this is nice because it seems to be shipped with pacman anyway
2) use "cower -du" to check for aur update.
would require cower to be installed
3) count total number of output lines for number of updates

alternatively we could display one number for aur updates and one for official updates.

matthiaskrgr:
This block has a simpler design than the original pacman block and does not require fakeroot.
It uses checkupdate to get official updates and requires cower binary to get aur updates which are displayed side by side.

Closes #179


matthiaskrgr:
screenshots:
pacman2_2
pacman2_1

atheriel:
I have a strong preference to see the discussion in the related issue #179 proceed before accepting this.

PR #185, opened 2018-05-10
Make temperature block only switch on left clicks

svmnotn:
This is a personal annoyance, so feel free to close if the 'works with any click' functionality is desired. I could also see it being a left or right click thing, I could make that change... I just don't want it going off on scrolling.


atheriel:
Sorry for the delay. This makes perfect sense to me, and is likely unintentional. Thanks!

PR #186, opened 2018-05-10
dependency updates

svmnotn:
Updated all the dependencies, dbus and nix were the only ones that changed anything that alters our code


atheriel:
This is great, thank you! Let me confirm on my end before merging.

Issue #27, opened 2017-05-26
Random crash after running for a while


Svenito:
I run i3status fine but after some time it crashes. I've run it with a debug build of the latest master (1517618d) to provide a detailed stacktrace

```
thread 'main' panicked at 'other was less than the current instant', /checkout/src/libstd/sys/unix/time.rs:276
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:355
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:371
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:549
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:511
   6: <std::time::Instant as core::ops::Sub>::sub
             at /checkout/src/libstd/sys/unix/time.rs:276
             at /checkout/src/libcore/result.rs:706
             at /checkout/src/libstd/sys/unix/time.rs:275
             at /checkout/src/libstd/time/mod.rs:184
             at /checkout/src/libstd/time/mod.rs:249
   7: i3status_rs::scheduler::UpdateScheduler::do_scheduled_updates
             at src/scheduler.rs:84
   8: i3status_rs::main
             at src/main.rs:181
   9: __rust_maybe_catch_panic
             at /checkout/src/libpanic_unwind/lib.rs:98
  10: std::rt::lang_start
             at /checkout/src/libstd/panicking.rs:433
             at /checkout/src/libstd/panic.rs:361
             at /checkout/src/libstd/rt.rs:56
  11: main
  12: __libc_start_main
  13: _start
```
System
```
LSB Version:    :core-4.1-amd64:core-4.1-noarch:cxx-4.1-amd64:cxx-4.1-noarch:desktop-4.1-amd64:desktop-4.1-noarch:languages-4.1-amd64:languages-4.1-noarch:printing-4.1-amd64:printing-4.1-noarch
Distributor ID: CentOS
Description:    CentOS Linux release 7.3.1611 (Core) 
Release:        7.3.1611
Codename:       Core
[...]

PR #188, opened 2018-05-15
Fix panick when subtracting "now" Instant

avdv:
Similar to the problem reported in #27, which was fixed in f91a3d4, I saw this
error:

```
thread 'main' panicked at 'other was less than the current instant', libstd/sys/unix/time.rs:292:17
```
(sorry, no backtrace)

Since I am using the music block with text rotation, I am best-guessing that the
error is caused by the Instant subtraction in rotatingtext widget.

### DIFF
--- a/src/widgets/rotatingtext.rs
+++ b/src/widgets/rotatingtext.rs
@@ -137,18 +137,19 @@ impl RotatingTextWidget {
 
     pub fn next(&mut self) -> Result<(bool, Option<Duration>)> {
         if let Some(next_rotation) = self.next_rotation {
-            if next_rotation > Instant::now() {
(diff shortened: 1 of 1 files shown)


avdv:
> I'm happy to merge this, but I am totally confused as to how this error could ever arise. Is there no guarantee that successive calls to Instant::now() are monotonically increasing...?

Indeed it is, but that is exactly the problem:

You first call `Instant.now` and check "is it less than X?", then you call `Instant.now` again and subtract the new value from X. If, in the meantime, `Instant.now` increased to be > X, you get the panic.

atheriel:
Ah right, of course. Thanks!

PR #192, opened 2018-05-31
Fixes #190 (underflow causing 2**64-1 update count)

### LINKED ISSUE #190: pacman block displaying 18446744073709551615 updates ...
Still learning rust, but cannot see an obvious explanation for this.

After updating pacman itself to v5.1.0, the pacman block now displays 2^64-1 required updates when `pacman -Su -p` shows no updates required, and 0 updates required when I've downgraded a package and `pacman -Su -p` suggests an upgrade.

I've tried wiping out the tmp db, and running the command manually, to no avail.

yawpitch:
Tested this on either side of the pacman 5.1 update, the `pacman -Qu` command prints the same number of lines in both earlier and later versions.

### DIFF
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -122,14 +122,14 @@ fn get_update_count() -> Result<usize> {
             Command::new("sh")
                 .args(&[
                     "-c",
-                    &format!("fakeroot pacman -Su -p --dbpath \"{}\"", updates_db),
+                    &format!("fakeroot pacman -Qu --dbpath \"{}\"", updates_db),
                 ])
                 .output()
                 .block_error("pacman", "There was a problem running the pacman commands")?
                 .stdout,
         ).block_error("pacman", "there was a problem parsing the output")?
             .lines()
-            .count() - 1,
+            .count(),
     )
 }
 
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -120,6 +120,7 @@ fn get_update_count() -> Result<usize> {
     Ok(
         String::from_utf8(
             Command::new("sh")
+                .env("LC_ALL", "C")
                 .args(&[
                     "-c",
                     &format!("fakeroot pacman -Qu --dbpath \"{}\"", updates_db),
@@ -129,6 +130,7 @@ fn get_update_count() -> Result<usize> {
                 .stdout,
         ).block_error("pacman", "there was a problem parsing the output")?
             .lines()
+            .filter(|line| !line.contains("[ignored]"))
             .count(),
     )
 }


src/blocks/pacman.rs
```diff
@@ -122,14 +122,14 @@ fn get_update_count() -> Result<usize> {
             Command::new("sh")
                 .args(&[
                     "-c",
-                    &format!("fakeroot pacman -Su -p --dbpath \"{}\"", updates_db),
+                    &format!("fakeroot pacman -Qu --dbpath \"{}\"", updates_db),
```
pitkley: Unfortunately this is not identical. `pacman -Su -p` skips ignored packages, while `pacman -Qu` will show them.

Sample output:

```console
$ pacman -Su -p
$ pacman -Qu
linux 4.16.11-1 -> 4.16.12-1 [Ignoriert]
linux-headers 4.16.11-1 -> 4.16.12-1 [Ignoriert]
```

This would mean that the available update count is different, and why `checkupdates` [performs an additional grep](https://git.archlinux.org/pacman-contrib.git/tree/src/checkupdates.sh.in#n60) excluding any lines that contain a pair of square brackets:

```sh
pacman -Qu --dbpath "$CHECKUPDATES_DB" 2> /dev/null | grep -v '\[.*\]'
```

Simply adding the `grep -v` to the command executed works:

```diff
diff --git a/src/blocks/pacman.rs b/src/blocks/pacman.rs
index 701b434..57b6da2 100644
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -122,7 +122,7 @@ fn get_update_count() -> Result<usize> {
             Command::new("sh")
                 .args(&[
                     "-c",
-                    &format!("fakeroot pacman -Qu --dbpath \"{}\"", updates_db),
+                    &format!("fakeroot pacman -Qu --dbpath \"{}\" | grep -v '\\[.*\\]'", updates_db),
                 ])
                 .output()
                 .block_error("pacman", "There was a problem running the pacman commands")?
```

An alternative solution could be to force the localization pacman uses and specifically filter out lines that have `[ignored]` in them:

```console
$ LC_ALL=C pacman -Qu
linux 4.16.11-1 -> 4.16.12-1 [ignored]
linux-headers 4.16.11-1 -> 4.16.12-1 [ignored]
```

```diff
diff --git a/src/blocks/pacman.rs b/src/blocks/pacman.rs
index 701b434..9c0acda 100644
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -120,15 +120,17 @@ fn get_update_count() -> Result<usize> {
     Ok(
         String::from_utf8(
             Command::new("sh")
+                .env("LC_ALL", "C")
                 .args(&[
                     "-c",
                     &format!("fakeroot pacman -Qu --dbpath \"{}\"", updates_db),
                 ])
                 .output()
                 .block_error("pacman", "There was a problem running the pacman commands")?
                 .stdout,
         ).block_error("pacman", "there was a problem parsing the output")?
             .lines()
+            .filter(|line| !line.contains("[ignored]"))
             .count(),
     )
 }
```

While I think the second solution is cleaner, I'm not sure if there are any other edge-cases the `\[.*\]` trick would capture, that this doesn't.

atheriel: Could we not just use `.filter(|line| !line.contains("["))` instead of `grep`?

yawpitch: Good catch @pitkley. Personally I think the cleaner option is better as well, and I'll update the PR to reflect that.

A quick check over the pacman source code seems to indicate that this the only such special case, and honestly I'd say it's better to address other outliers as they're found.

yawpitch:
Updated PR for fix by @pitkley.

src/blocks/pacman.rs
pitkley: > Could we not just use `.filter(|line| !line.contains("["))` instead of grep?

@atheriel I am not sure if any of the output produced by `pacman -Qu` could contain a `[` that is not for ignored packages, leading to false positives. I agree with @yawpitch that we can handling the `[ignored]` case specifically is probably fine, and extend this if the need comes up.

PR #195, opened 2018-06-02
Add todo block.

Terae:
Block to count and display TODOs contained on a file.


atheriel:
There seems to be an unrelated commit here -- can you explain the purpose of that one or open a separate PR?

As for the todo list functionality: is this a common format for such things? Is there a standard program that handles/creates these files? I'm trying to understand how this would have general appeal to other users.

PR #196, opened 2018-06-03
update deps

svmnotn:
Updated chrono-tz, lazy_static, chan, uuid, and inotify, to their latest versions.
The update to uuid was to the 0.7-dev version, but so far there are already some changes.
We probably should wait until the full 0.7 version is ready.

@greshake @atheriel any preference on how often we should update dependencies? I'm thinking about checking (aka running `cargo update` and seeing if anything happens) around once a month. But maybe it would be better to only update when there would be a code change, or some kind of issue is found on a dependency. 

If we go with the latter feel free to close this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,9 +1,9 @@
 [[package]]
 name = "aho-corasick"
-version = "0.6.4"
+version = "0.6.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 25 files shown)


atheriel:
Sorry that I was so late to reply to this. I think we need a more formal approach to updating -- see my linked issue. Unfortunately, these particular changes are pretty stale, so I'm going to close this particular issue.

PR #197, opened 2018-06-03
allow trailing commas in macros

svmnotn:
this should reduce the lines changed when a new block is added, by one... not much of a change, but it also means no compile issues if the comma is or isn't there.


svmnotn:
I just realized `icons` has the same "issue," I'll be adding a commit here to address that.

PR #199, opened 2018-06-03
Fix the sound toggle function with amixer.

Terae:
`amixer set Master toggle` mutes the sound, `amixer -D pulse set Master toggle` toggles it.

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -85,7 +85,7 @@ impl SoundDevice {
 
     fn toggle(&mut self) -> Result<()> {
         Command::new("sh")
-            .args(&["-c", format!("amixer set {} toggle", self.name).as_str()])
+            .args(&["-c", format!("amixer -D pulse set {} 1+ toggle", self.name).as_str()])
             .output()
             .block_error("sound", "failed to toggle mute")?;
 


atheriel:
I'm trying to verify this at the moment. There seems to be no mention of this in the man pages. Why is it that the `-D` flag does not default correctly to the `pulse` device for this functionality? Or do you have multiple devices on your system?

Terae:
I only have a second monitor but the initial command works to mute the sound and nothing appears by clicking again on the block; I do not master `amixer` so I don't know if the fix `-D pulse` is portable. Btw, I searched why it works only in one direction on my computer but found nothing.

atheriel:
What is the output of `amixer info` before and after running `amixer set Master toggle`? Maybe it switches the default device at some point?

Terae:
No, `amixer info` returns me that in both cases:
```txt
$ amixer info
Card default 'PCH'/'HDA Intel PCH at 0xdf420000 irq 132'
  Mixer name	: 'Realtek ALC898'
  Components	: 'HDA:10ec0899,146211d6,00100003'
  Controls      : 30
  Simple ctrls  : 12
```

and the `toggle` command:
```txt
$ amixer set Master toggle
Simple mixer control 'Master',0
  Capabilities: pvolume pvolume-joined pswitch pswitch-joined
  Playback channels: Mono
  Limits: Playback 0 - 87
  Mono: Playback 72 [83%] [-11.25dB] [on]
```
The only difference for the toggle is `[on]` and `[off]` but the sound is however muted.

atheriel:
I'm currently wondering if the correct way to address these issues is allow the user to configure the device name. Let me investigate further.

atheriel:
We haven't had any bug reports for this, so I believe it is due to an idiosyncratic issue.

Issue #131, opened 2018-01-11
On_click option should be available for all blocks


Kommynct:
This is just kinda obviously an issue, but I was here to report another bug so i figured i might as well add it to the issues checklist.

atheriel:
I don't think this is high on the list of priorities for maintainers, but I agree this would be a nice feature for blocks that do not otherwise have a natural on-click behaviour already. You are free to take a crack at it yourself, or propose which blocks need it most.

lausek:
@atheriel Is there a reason why the default implementation doesn't automatically pass the command to the shell? I mean, if the block needs another behaviour in `click` it could just override the trait method, right?

atheriel:
This is complicated by the fact that clicks are at the "widget" level, not the "block" level. This is why there's no default block clicking behaviour in the current implementation, although I'm not opposed to it.

atheriel:
@boogy Note that several blocks have on click behaviour already. We couldn't introduce this for all blocks without breaking some existing features.

mankinskin:
Why has this issue got closed?

karthink:
Adding my vote to this request. on_click would be very useful on most modules that don't have them yet.

PR #200, opened 2018-06-03
Graphical interfaces

Terae:
By right-clicking on blocks, the appropriate GUI interface opens.

I updated these blocks converting `TextWidget` by `ButtonWidget`:
* **Battery**: launch `gnome-power-statistics` by default
* **CPU**: launch `gnome-system-monitor` by default
* **Net**: launch `nmtui` by default. Actually, it works only by clicking on text but not on a graphic.

The `on_click` argument is an `Option<String>` so it is still possible to open nothing like this:
```toml
[[block]]
name = "my_block"
on_click = ""
```

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -1,3 +1,5 @@
+use std::env;
+use std::process::Command;
 use std::time::Duration;
 use chan::Sender;
 use scheduler::Task;
@@ -5,18 +7,20 @@ use block::{Block, ConfigBlock};
 use config::Config;
 use de::deserialize_duration;
 use errors::*;
-use widgets::text::TextWidget;
+use input::{I3BarEvent, MouseButton};
+use widgets::button::ButtonWidget;
 use widget::{I3BarWidget, State};
 use uuid::Uuid;
 use blocks::lib::*;
 
 pub struct Battery {
-    output: TextWidget,
+    output: ButtonWidget,
     id: String,
     max_charge: u64,
     update_interval: Duration,
     device_path: String,
     show: ShowType,
+    on_click: Option<String>,
 }
 
 #[derive(Deserialize, Copy, Clone, Debug)]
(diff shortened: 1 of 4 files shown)


atheriel:
FWIW, I'm strongly opposed to setting defaults that expect the user to have GNOME applications installed. This is very likely to annoy non-GNOME users -- which, for an i3-related application, is most of them.

I also think we should link this to a related issue: #131. It's important to note that if we decide to go this route -- left click to open a related application -- it prevents us from adding alternative left-click behaviours for these blocks in the future. For this reason I'm not terribly supportive of the change. What do you think, @greshake?

(There are also a few blocks that have on-click behaviour already, for example the temperature block.)

Terae:
I agree in the fast we shouldn't setting defaults to GNOME applications, but does they are default GUI application installed on almost all systems or not? The easiest should be to set by default `on_click = ""` and let the user to choose its own GUI app.

Some blocks are toggled by the left-click event that is why I used the right-click behaviour to open graphic instances, except the temperature block (which doesn't differenciate teh mouse event yet) I fount no block which use this event so it seems to be a good and non destructive improvement no?

BrendanBall:
I would certainly be one of those non-GNOME users. And no @Terae gnome apps are not installed on most systems. I would think a minority of i3 users have gnome installed. I would definitely leave it blank by default.

src/blocks/battery.rs
```diff
@@ -215,6 +229,30 @@ impl Block for Battery {
         vec![&self.output]
     }
 
+    fn click(&mut self, event: &I3BarEvent) -> Result<()> {
```
l4l: Seems like all of these functions are the same, so it mgiht be a better idea to move it out from impl (maybe to ButtonWidget?) with a signature like `fn handle_click(id: &String, on_click: &String, event: &I3BarEvent) -> Result<()>` and for the impl left just a call for `handle_click(&self.id, &self.on_click, event)`

l4l:
Is there any activity there? I think that feature is kinda useful

atheriel:
@l4l I don't think there is particularly resounding support for this feature. Partly this is due to the non-trivial implementation, since we don't have any abstraction levels for blocks in the configuration file (and so every change needs to be made for blocks individually). And partly, I think, this is because it doesn't seem very "i3-like" to click the bar instead of using an application launcher like `drun` or `rofi`. Of course, that's more a cultural argument than a technical one.

PR #210, opened 2018-06-27
Add mail block

tablet-mode:
Adds a block which displays the number of unread mails in configured inboxes. Currently only supports maildir format. It has pretty minimal functionality but might be useful to some users.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -19,6 +19,7 @@
 - [Toggle](#toggle)
 - [Weather](#weather)
 - [Xrandr](#xrandr)
+- [Mail](#mail)
 
 ## Backlight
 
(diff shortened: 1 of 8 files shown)


atheriel:
This is a great idea!

My one comment would be that `mail` might be too generic, given that this only supports Maildir, and I think in practice it would be very difficult to support other kinds of mail notification with the same constraints. How about just calling it the `maildir` block?

Alternatively, you could consider something like the design of the `weather` block, which attempts to abstract out the configuration of potential backends from the main block configuration.

blocks.md
```diff
@@ -600,3 +601,27 @@ Key | Values | Required | Default
...
+
+### Examples
+
+```toml
+[[block]]
+block = "mail"
+interval = 60
+inboxes = ["/home/user/mail/local", "/home/user/mail/gmail/Inbox"]
+threshold_warning = 1
+threshold_warning = 10
```
atheriel: This is probably supposed to be `threshold_critical`.

src/blocks/mail.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use uuid::Uuid;
+
+pub struct Mail {
+    text: TextWidget,
+    id: String,
+    update_interval: Duration,
+
+    //useful, but optional
+    #[allow(dead_code)]
```
atheriel: There's no need to include these yet -- you might as well drop them and avoid the confusion of dead code.

tablet-mode:
I'm glad you like the idea :-)

> My one comment would be that mail might be too generic, given that this only supports Maildir, and I think in practice it would be very difficult to support other kinds of mail notification with the same constraints. How about just calling it the maildir block?

I was thinking about that as well and sort of hoped that it might become more generic and support more formats. However I don't think I'll actually get to it anytime soon so it would probably be best to just rename it to `maildir`.

I'll have a look at the `weather` block though, thanks for the pointer. Also thanks for reviewing!

tablet-mode:
OK, I think I fixed everything now and renamed the block to `maildir`. 

I still like the idea of a more powerful `mail` block but I looked at some of the available mail libs and I don't think I'll be able to quickly make that happen. So in the meantime it would be awesome to merge this as `maildir` block if you're happy with the code.

joefromct:
Hi, is there anyway to get t his to work with `Maildir++` style directories? 

For instance, my dovecot has sub-directories prefixed with a `.`, so i think it's not reporting any of the mail.

PR #214, opened 2018-07-05
Battery block: ability to calculate remaining time from charge/current

GladOSkar:
Remaining time didn't work for me, because my machine doesn't have the `energy_now` and `power_now` files but only `charge_now` and `current_now`.

This makes it so that `energy/power` is used if available, then if not, `charge/current` is used, and if neither is present, an error is thrown.

I am relatively new to rust, let me know if i've done something suboptimally :)

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -112,42 +112,56 @@ impl PowerSupplyDevice {
     /// Query the estimated time remaining, in minutes, before (dis)charging is
     /// complete.
     pub fn time_remaining(&self) -> Result<u64> {
-        let energy_full = match self.energy_full {
-            Some(val) => val,
-            None => {
-                return Err(BlockError(
-                    "battery".to_string(),
-                    "Device does not support reading energy".to_string(),
-                ))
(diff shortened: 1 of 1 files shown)


atheriel:
This looks good to me. Do you find that it gives a reasonable estimate of the time remaining?

GladOSkar:
@atheriel It gives a reasonably good, usable estimate. I never really measured how much it deviates from actual discharge times. Maybe i'll do that some time.

GladOSkar:
OK I did some science to it:

ert-measurements

The current implementation varies a lot depending on the current power usage and basically isn't averaged at all. But even if averaged over 10 minutes (see green line), it does not give a consistent trajectory.

The charge line on the other hand is surprisingly consistent, so i would recommend calculating the remaining time by averaging over charge amounts.

The averaging interval should probably be between 2 and 10 minutes, maybe customizable with 5 minutes as a default?

atheriel:
That's excellent investigative work, thank you. I'll file a follow-up issue.

PR #215, opened 2018-07-11
WIP: Add block for notmuch email users

bobthemighty:
This PR adds a new notmuch block that queries a notmuch database and displays the count of messages. It directly uses the notmuch C lib rather than shelling out.

The simplest configuration will return the total count of messages in the notmuch database stored at $HOME/.mail

```
[[block]]
block = "notmuch"
```

Optionally, we can apply labels, thresholds, or suppress the mail icon.

```
[[block]]
block = "notmuch"
query = "tag:inbox"
threshold_good = 1
name = "I"

[[block]]
block = "notmuch"
query = "tag:alert and not tag:trash"
threshold_warning = 1
threshold_critical = 10
name = "A"
no_icon = true

[[block]]
block = "notmuch"
query = "tag:todo"
threshold_info = 1
name = "T"
no_icon = true
```
mail

### DIFF
--- a/src/blocks/notmuch.rs
+++ b/src/blocks/notmuch.rs
@@ -17,6 +17,7 @@ use self::libc::c_char;
 use std::env;
 use std::ffi::CString;
 use std::ptr;
+use std::result;
 use widget::State;
 
 use uuid::Uuid;
@@ -86,21 +87,11 @@ pub enum notmuch_database_mode_t {
 
 #[link(name = "notmuch")]
 extern "C" {
-    pub fn notmuch_query_count_messages(
-        query: *mut notmuch_query_t,
-        count: *mut u16,
-    ) -> notmuch_status_t;
+    pub fn notmuch_query_count_messages(query: *mut notmuch_query_t, count: *mut u16) -> notmuch_status_t;
 
-    pub fn notmuch_query_create(
-        database: *mut notmuch_database_t,
-        query_string: *const c_char,
-    ) -> *mut notmuch_query_t;
+    pub fn notmuch_query_create(database: *mut notmuch_database_t, query_string: *const c_char) -> *mut notmuch_query_t;
 
-    pub fn notmuch_database_open(
-        path: *const c_char,
-        mode: notmuch_database_mode_t,
-        database: *mut *mut notmuch_database_t,
-    ) -> notmuch_status_t;
+    pub fn notmuch_database_open(path: *const c_char, mode: notmuch_database_mode_t, database: *mut *mut notmuch_database_t) -> notmuch_status_t;
 
     pub fn notmuch_database_destroy(database: *mut notmuch_database_t) -> notmuch_status_t;
 
@@ -110,8 +101,8 @@ pub struct Notmuch {
     text: TextWidget,
     id: String,
     update_interval: Duration,
-    query: *mut notmuch_query_t,
-    db: *mut notmuch_database_t,
+    query: CString,
+    db: CString,
     threshold_info: u16,
     threshold_good: u16,
     threshold_warning: u16,
@@ -130,9 +121,7 @@ pub struct Notmuch {
 #[serde(deny_unknown_fields)]
 pub struct NotmuchConfig {
     /// Update interval in seconds
-    #[serde(
-        default = "NotmuchConfig::default_interval", deserialize_with = "deserialize_duration"
-    )]
+    #[serde(default = "NotmuchConfig::default_interval", deserialize_with = "deserialize_duration")]
     pub interval: Duration,
     #[serde(default = "NotmuchConfig::default_maildir")]
     pub maildir: String,
@@ -149,7 +138,7 @@ pub struct NotmuchConfig {
     #[serde(default = "NotmuchConfig::default_name")]
     pub name: Option<String>,
     #[serde(default = "NotmuchConfig::default_no_icon")]
-    pub no_icon: bool
+    pub no_icon: bool,
 }
 
 impl NotmuchConfig {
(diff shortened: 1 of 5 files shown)


src/blocks/notmuch.rs
```diff
@@ -0,0 +1,282 @@
...
+    update_interval: Duration,
+    query: CString,
+    db: CString,
+    threshold_info: u16,
+    threshold_good: u16,
+    threshold_warning: u16,
+    threshold_critical: u16,
+    name: Option<String>,
+
+    //useful, but optional
```
atheriel: I'd advise deleting these if you aren't using them.

src/blocks/notmuch.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+#[repr(C)]
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub enum notmuch_database_mode_t {
+    NOTMUCH_DATABASE_MODE_READ_ONLY = 0,
+    NOTMUCH_DATABASE_MODE_READ_WRITE,
+}
+
+#[link(name = "notmuch")]
```
atheriel: What will happen if this is not linked correctly?

atheriel:
This is a nice piece of work. I'm a little unsure of having all the extern C code inside of this project (as opposed to encapsulated in some `notmuch-sys` crate), but it's a reasonable place to start from.

Why do you consider this WIP? What would you like to add or improve?

src/blocks/notmuch.rs
bobthemighty: If the package is unavailable, then the project fails to build, that's what led me to ask about a plugin system.

bobthemighty:
There're a couple of notmuch crates already that do exactly this, I'd be happy to use one of those instead (there's a particularly nice one that would improve the code significantly). I was unsure of the ettiquette of adding new dependencies to the project. This is the first Rust code I've written so I'm fumbling a little.

I still need to add some error handling for things like bad configuration or errors from the underlying lib, but other than that I'm happy. The problem is that as it stands the code depends on the notmuch c-lib being available to link.

atheriel:
My thinking is that we should use [cargo's support for feature flags](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section) here, so that users can opt-in to notmuch support. We haven't used this approach so far, but I do think it's the right choice here.

The approach will make one of the notmuch crates an optional compile-time dependency, and only include this block's module if that feature flag has been set. Since you are new to Rust, I am happy to help you with that part. Go ahead and use the crate you have in mind, and then we'll move it behind a feature flag.

bobthemighty:
The feature flags seem like a good compromise. I'm happy to have a go at that and ping you when it's working. The only question I have is what that means for CI. Would you want to perform a build with that flag enabled or not?

atheriel:
I'd like to have it enabled on Travis, yes. Shouldn't be too hard to ensure that the required notmuch library is installed via the repo's config file.

bobthemighty:
Good morrow! 

Sorry this has been a little delayed. I've updated the code to use the notmuch crate, which means I can get rid of all the C extern cruft. I've played a little with the feature flag functionality, but I'm unclear on how to use it to remove the block.

It's simple enough to conditionally `use self::notmuch` in mod.rs, but I'm not sure how to conditionally include the block in the call to `blocks!`.

atheriel:
Hmm, I hadn't thought about that either. Let me take a look this weekend.

uinput:
@bobthemighty This looks awesome!

What about updating the `block!` macro like so:
```rust
macro_rules! blocks {
    ( $name:ident, $block_config:ident, $config:ident, $tx_update_request:ident ; $( $block_name:expr, $feature:expr => $block_type:ident ),+ ) => {
        match $name {
            $(
                $block_name if $feature == "" || cfg!(feature = $feature) => block!($block_type, $block_config, $config, $tx_update_request),
             )*
            _ => Err(BlockError($name.to_string(), "Unknown block!".to_string())),
        }
    }
}
```
This adds a guard for conditional compilation. You can then rewrite the `create_block()` method. I'm pretty sure there is a more elegant way to do this. However right now, this is the best I could come up with. You can specify one feature per block, and if it should be included by default, an empty string is required like so:
```rust
[...]
"net", "" => Net,
"notmuch", "notmuch" => Notmuch,
[...]
```

rschulman:
I would love to have this block! Is there a reason that this PR stalled? Any way I can help out?

atheriel:
@rschulman It's likely that the original author just has other things on his plate. I would also be happy to see someone take this feature up. My suggestion would be to try this branch and see what works and what needs to be done. Then, if you would like to continue working on it:

1. Squash the commits and rebase against master to fix the merge conflicts (which look very simple, but I can advise you if needed). Attribute this commit to the original author.
2. Add your own commit(s) as needed.
3. Open a new PR that we can mark as superseding this one.

If you need a model for what this process looks like, take a look at #113.

bobthemighty:
Hey. I've been happily using my own build since opening this PR. From my
perspective it works well but I don't have the rust experience to fix the
issue of optional dependencies.

If someone else wants to push this forward, they can do so with my
blessings

atheriel:
@bobthemighty If the only outstanding issue here is just the optional dependency bit, we can absolutely fix that in a jiffy and get this merged. Is the block otherwise ready, in your view?

atheriel:
I squashed and merged this manually into master, fixing the merge conflicts. Thanks for your contribution! At the moment, you must opt-in to this block with `cargo build --feature notmuch`; feel free to try it out and report any issues you encounter.

PR #227, opened 2018-08-08
Add alpha channel to color utils.

uinput:
This commit introduces an alpha channel to the current color arithmetic. That allows for a completely transparent background.


atheriel:
This looks like a nice addition, thank you.

PR #228, opened 2018-08-08
Add coretemp block.

uinput:
This pull request adds a new block to measure the current CPU temperature efficiently.

It does not depend on `sensors` nor does it run any external program to retrieve its measurements. By default, the block relies on the sysfs filesystem being present. You can specify any glob pattern to look for files.
The file list will only be retrieved once at startup, since I don't think hot-swapping of CPUs has to be supported.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,6 +235,11 @@ version = "0.3.54"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "glob"
+version = "0.2.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "i3ipc"
 version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -254,6 +259,7 @@ dependencies = [
  "clap 2.31.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "cpuprofiler 0.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "dbus 0.6.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "glob 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "i3ipc 0.8.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "inotify 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


atheriel:
I'm not quite sure I understand how this works, though the code looks good. What does the file pattern look for? Is this a common `sysfs` interface? Can you provide a sample configuration?

uinput:
Hi, thank you for the feedback. I'm sorry I didn't go into detail here.
The `sysfs` interface is actually the same that `i3status` is using, as explained [here](https://i3wm.org/i3status/manpage.html#_cpu_temperature). The example path on that page isn't that recent, the current implementation should at least support Arch Linux and Fedora out of the box (It's looking for files matching the default glob pattern `/sys/devices/platform/coretemp.0/hwmon/*/temp*_input`).

As for the configuration, it's quite similar to the `temperature` block. I'm using 
```toml

[[block]]
block = "coretemp"
format = "{max}°"
interval = 3
```
You could also set your own `pattern`, which supports glob patterns very similar to `bash`.

atheriel:
Ah, so this block is mostly about bypassing `lm_sensors` and reading the `sysfs` interface directly, right?

atheriel:
It might be worth considering refactoring the existing temperature block to use this approach instead, so that we could remove the dependency on `lm_sensors`. I had a branch somewhere that implemented something like the glob pattern in that block, but this seems like a better alternative.

(Relatedly, we might also consider making a separate "fan speed" block, since that's likely the only other thing people want out of `lm_sensors`.)

uinput:
Do you want me to completely remove the current temperature implementation in favor of directly reading sysfs? This would raise compatibility problems, since we cannot provide the same functionality as before.
Is there a standard way of getting fan speeds using sysfs? I could only find my GPU's fan speeds.

atheriel:
Oh, I'm sorry, it's been too long since I took a look at the sensors/sysfs stuff. IIRC the hwmon interface has pretty mixed support depending on your hardware. I'm looking at the `/sys/class/hwmon` directory, which contains both platform and virtual devices on my system that implement the `hwmon` temperature interface. You can list them with

```
$ ls /sys/class/hwmon/*/temp*_input
```

(They're all symlinks.) I'm fairly sure that `lm_sensors` is just reading out of these files, which was the reason I thought we might just bypass it entirely. You can see all fan sensors with `*/fan*_input`, too.

I'm not sure that I understand the difference between these and the coretemp reading that you're picking up here, but is it reasonable to think we should work towards combining this approach with the existing code?

uinput:
The glob pattern you suggested deals with symlinks, which are not currently well-supported by the `glob` crate. That's why I was using `/sys/devices/platform/coretemp.0/hwmon/*/temp*_input` in the first place. Do you know of any other crate supporting glob patterns? I do not want to implement this myself, as that might be very inefficient. Also, how can we know what fan speeds users are interested in?

Also, `coretemp` and `temperature` would not be a name where users expect to display fan speeds. I think we should try to merge those features into one block, but split them into `coretemp` and `fanspeed` or something like that.

atheriel:
Just a head's up: the master branch now has support for filtering the `temperature` block to only show a given chip name. You can use this to restrict readings to coretemp with `chip = "coretemp-*"`.

ammgws:
@uinput Does the filtering feature introduced to the temperature block solve this for you?

ammgws:
Closing due to inactivity.

PR #229, opened 2018-08-08
Add padding to percent values of memory block.

uinput:
This commit adds zero-padding to the percent values of the `memory` block. On my system for example, memory usage is always around `10%`. This leads to the memory block changing from one digit to two digits and vice-versa quite frequently.

I'd like to have less movement in my status bar, while keeping the values recent. The idea was to pad single-digit values with `0` in front as done in the `cpu` block. We could also pad using space.

### DIFF
--- a/src/blocks/memory.rs
+++ b/src/blocks/memory.rs
@@ -351,7 +351,7 @@ impl Memory {
         );
         self.values.insert(
             "{MFpi}".to_string(),
-            format!("{}", mem_free.percent(mem_total) as i32),
+            format!("{:02}", mem_free.percent(mem_total) as i32),
         );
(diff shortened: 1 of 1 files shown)


atheriel:
I think this makes sense, but I'm not sure whether zero-padding or space-padding is better. Does anyone have an opinion?

uinput:
Should we support an option to make padding customizable? In case we use space, do we want the one-wide value to be left aligned so there isn't more space between it and the symbol in contrast to the two-wide case?

atheriel:
Zero-padding is consistent with the CPU block, so let's go with that as a rule of thumb. Thanks!

PR #231, opened 2018-08-10
Add upower block.

uinput:
This pull request adds a new block to read the battery level efficiently.

It triggers updates via UPower's dbus interface. The default device is derived from UPower as well, by iterating over all possible devices and taking the first valid one. This means UPower is required to use this block, but I think this is fair since there will only be updates triggered when necessary.

Sample configuration:
```
[[block]]
block = "upower"
device = "battery_BAT0"
```

### DIFF
--- a/src/blocks/mod.rs
+++ b/src/blocks/mod.rs
@@ -1,49 +1,52 @@
-mod time;
-mod template;
-mod load;
-mod memory;
+mod backlight;
+mod battery;
 mod cpu;
-mod music;
-pub mod battery;
 mod custom;
 mod disk_space;
+mod focused_window;
+mod load;
+mod maildir;
+mod memory;
+mod music;
+mod net;
+mod nvidia_gpu;
 mod pacman;
-mod temperature;
-mod toggle;
 mod sound;
 mod speedtest;
-mod focused_window;
-mod xrandr;
-mod net;
-pub mod backlight;
-mod weather;
+mod temperature;
+mod template;
+mod time;
+mod toggle;
+mod upower;
 mod uptime;
-pub mod nvidia_gpu;
-pub mod maildir;
+mod weather;
+mod xrandr;
 
 use config::Config;
-use self::time::*;
-use self::template::*;
-use self::music::*;
-use self::cpu::*;
-use self::load::*;
-use self::memory::*;
+
+use self::backlight::*;
 use self::battery::*;
+use self::cpu::*;
 use self::custom::*;
 use self::disk_space::*;
+use self::focused_window::*;
+use self::load::*;
+use self::maildir::*;
+use self::memory::*;
+use self::music::*;
+use self::net::*;
+use self::nvidia_gpu::*;
 use self::pacman::*;
 use self::sound::*;
 use self::speedtest::*;
-use self::toggle::*;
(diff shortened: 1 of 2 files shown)


atheriel:
Is there any reason that the dbus approach to polling for updates can't be applied to the existing battery block?

uinput:
No, I just didn't want to introduce any compatibility issues. If you request to replace the existing battery block, I'll do that. Be aware that this block requires UPower to be installed, while the current battery block doesn't.

src/blocks/upower.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+}
+
+impl From<u32> for DeviceType {
+    fn from(id: u32) -> Self {
+        match id {
+            // https://upower.freedesktop.org/docs/Device.html#Device:Type
+            // TODO: derive this automatically.
+            1 => DeviceType::LinePower,
+            2 => DeviceType::Battery,
```
atheriel: If the only device type you care about is this one, why bother will all of this other boilerplate? Just have `Battery` and `Other`.

src/blocks/mod.rs
```diff
@@ -1,49 +1,52 @@
-mod time;
```
atheriel: I think that the changes to this file are unintentional format changes. Can you remove changes that are not relevant to the new block?

atheriel:
That's a fair point.

src/blocks/upower.rs
```diff
@@ -0,0 +1,306 @@
...
+                 interface='org.freedesktop.DBus.Properties',\
+                 member='PropertiesChanged'",
+                device_path);
+
+            c.add_match(&rule).unwrap();
+
+            loop {
+                let timeout = 100000;
+
+                for _event in c.iter(timeout) {
```
atheriel: Do you know how often dbus sends battery events? It's not a bad idea to have a minimum wait period to prevent event spam from triggering too many bar updates.

src/blocks/mod.rs
uinput: Yes, I sorted them in order to get cleaner looking code. Should I just remove it, or is a pull request welcome?

src/blocks/upower.rs
atheriel: Oh, that's a pretty substantial difference. I suppose that it's safe enough to dispense with a fallback, then.

atheriel:
Unfortunately, I can't get this to work. `upower -i /org/freedesktop/UPower/devices/battery_BAT0` lists the correct device information on my machine, but if I configure this block with

```
[[block]]
name = "upower"
device = "battery_BAT0"
```

I get "Device is invalid". Further, if I leave it blank, I get no percentage reading (just the "-"), and no error.

Any ideas?

uinput:
Does `dbus-monitor --system "type='signal',path='/org/freedesktop/UPower/devices/battery_BAT0',interface='org.freedesktop.DBus.Properties',member='PropertiesChanged'"` work for you just fine or is there any error message?

atheriel:
It turns out that this was likely an idiosyncrasy with my dbus/upower setup.

atheriel:
Closed as superseded by #252.

PR #232, opened 2018-08-12
Add networkmanager block.

uinput:
This pull request adds a block to receive dbus signals from NetworkManager as suggested in #4.

It's a very basic implementation right now. The block can determine the network status and the device you're connected with your network (primary device).

Suggestions are welcome.

### DIFF
--- a/src/blocks/mod.rs
+++ b/src/blocks/mod.rs
@@ -20,6 +20,7 @@ mod weather;
 mod uptime;
 pub mod nvidia_gpu;
 pub mod maildir;
+mod networkmanager;
 
 use config::Config;
 use self::time::*;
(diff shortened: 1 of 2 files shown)


atheriel:
What's the advantage of this over the current network block?

uinput:
The current block refreshes every `n` seconds, depending on your configuration. In #4, @greshake himself listed a NetworkManager block using dbus as the most important new block. This pull request tries to build a base for this idea, updates are only issued if there is actually something to update.

As I understand, this project tries to be resource friendly, so I think this would be a good addition. There are two advantages over the current block:
- we don't update unnecessarily
- changes are visible immediately.

Issue #145, opened 2018-01-16
Errors in blocks shouldn't make the bar unusable


lovesegfault:
I have a laptop with a pretty nasty battery, sometimes it shows up, sometimes it doesn't, and it's a mess. In i3status, or i3pystatus, whenever my battery went rogue the 'block' on the statusbar would show a helpful "no battery". On i3status-rust, however, I get a huge red message covering the entire bar telling me my battery isn't there.

Moreover, if for example my weather API key isn't working (activation delay), I get another huge red message telling me it's invalid that covers the entire bar.

I think warnings and errors should either be pop-ups (i3-nag style), or be contained within their respective block, instead of covering the entire bar, which makes it unusable.

atheriel:
I generally agree, although I think it's correct to error out in some cases. I'm working on a few patches to fix battery and network-related errors at the moment, which would seem to cover most of the annoying crashes due to temporarily failing `sysfs` reads or shell commands.

bemeurer:
Well, sure, I don't think erroring is a bad thing. I agree that it should error out in some cases (e.g. invalid API key for weather block), what I'm saying is *how* it should error. Instead of the whole bar becoming an error message, maybe just the particular block (and if the message is too big you could use the scrolling thing implemented in the music/media player block).
This way if for some reason the weather API is down, or my WiFi card is going nuts, or some other error, I can see that it is happening, but it also doesn't impair the rest of the functionality of the status bar.

bemeurer:
Just to provide an example of this happening. Every time I boot my computer it takes a few seconds for the Wi-Fi to connect proper, and because of that upon boot the statusbar is like this:
![screenshot from 2018-01-24 13-38-39](https://user-images.githubusercontent.com/7243783/35358608-0f92df00-010c-11e8-9ac5-3e8984b1a5d4.png)
At which point I have to reload i3 for it to go back to normal. All I was trying to argue for in my previous comments was that errors like this should be shown in the block where they come from, like in i3status :)

atheriel:
The key problem with this for the current implementation is that errors are propagated up to the bar level. The infrastructure for differentiating between fatal and non-fatal errors simply does not exist, and until it does I don't think there is a good way of addressing these problems. As a stopgap measure, I do think that we can (and will, when some time for this project materializes) silently fail on some of the obvious network/battery issues, but that is not a long-term solution.

greshake:
[...]

Issue #85, opened 2017-08-14
Account for not having a battery


RealOrangeOne:
A strange sounding issue I know...

I sync config between my desktop and laptop, 1 of which doesnt have a battery. Because of the missing battery, the entire bar fails to load. 

IMO the nicest solution would be for the block not to render. However some people may want it to display `NO BAT` or similar, which could probably be user configured?

burdickjp:
Here's a solution using a shell script to generate the config conditionally from templates:
https://faq.i3wm.org/question/1367/anyway-to-include-in-config-file/%3C/p%3E.html

svmnotn:
It would be nice to provide a way to do this without an external shell script. I believe @greshake had an idea about this with "profiles" or something.

greshake:
Yea... At first I assumed this is a rather rare issue but now I'll consider moving to a solution for this more quickly. I still have not decided which solution makes sense; There are elegant solutions (e.g. profiles) that fix this but require significant implementation effort and I fear overcrowding the project with features that solve rare use cases and thereby diminishing maintainability in the long run.

uinput:
@greshake Has there been any progress on this? What about adding an option for the blocks to silently fail? Users could specify the battery block to not render if the battery was not found.

uinput:
I think the same issue applies to the `backlight` block. I thought about implementing a `fail_silently = True` option for those. Do you think that would suffice for our use case?

burdickjp:
I think failing silently would fix a lot of things across the board.

atheriel:
I think we could implement something similar to the [Weather block fallback](https://github.com/greshake/i3status-rust/blob/master/src/blocks/weather.rs#L200-L203). That is, we show an `×` when the battery appears to be missing. I'd prefer this to hiding the block, because otherwise users with errors/typos in their configurations will be confused when the block is simply missing. An `×` provides some visual feedback to those users.

atheriel:
In addition, this would be a pretty easy change to make, because the battery device already returns `Result<T>` for its `sysfs` queries.

uinput:
That would permanently take unused space in case there is no battery/backlight. An explicit hint for users about what the option `fail_silently` does shouldn't lead to such problems, I think.

uinput:
As for the implementation, what about returning a `SilentError` when the block is initialized? That way we could completely remove the block, saving computation time.

uinput:
[...]

Issue #50, opened 2017-06-23
Add flag to suppress errors on startup


michelboaventura:
I use i3status-rs on both my pc and notebook, sharing the same configuration. But since only the notebook has a battery, the configuration doesn't work on my PC and I have to always comment the battery block.

It would be great if we have a flag to just ignore any block with errors, but show the other ones.

greshake:
I don't really like the proposed solution, seems hacky and might lead to confusion down the road (issues being openend because error messages are hidden because someone only copy-pasted the config/cmd line args). Also it would just make sense to have separate configs in your use case. But I'll await further judgement from others, maybe there is more demand for this feature than I estimate.

greshake:
@jheyens @pitkley @keeslinp @svmnotn Any comment?

michelboaventura:
I understand your point, and maybe I'm the only one sharing the same i3status-rs/i3/entire world of configs between my pcs :)

svmnotn:
Maybe have all blocks have an `optional = false` tag that can be changed to _true_.
That way if the block is optional, any errors will simply cause the block to not be displayed?
That seems like a more complete solution, specially if all blocks default to _false_ so that users would have to opt-in to the feature.

keeslinp:
I think one helpful thing would a message saying which module failed. Are we currently displaying any information with the error message? I'm running a pretty old version so I'm not sure. When I first tried out this app I had to just comment out stuff until it worked and go from there.

keeslinp:
I agree that silently failing is probably a bad idea. I haven't updated in a while so I'm not sure about how it's all handled at the moment. I'll try to test some things out this weekend if I get the time (It's been pretty crazy lately) and I'll try to have a more informed opinion.

Two very early thoughts are this:
1. I like the idea of an optional flag that defaults to fail explicitly because it allows people to have more seamless usage across devices without messing with the average user.
2. If we can contain failure to an individual component it could add buttons to a dead component to either restart or hide it. That also helps with one-off errors so people don't have to restart all of i3 to get it back up.

svmnotn:
@keeslinp I like the idea that
> If we can contain failure to an individual component it could add buttons to a dead component to either restart or hide it. That also helps with one-off errors so people don't have to restart all of i3 to get it back up.

Now that we have most things return a `Result<_>` it should be possible to do, but that should probably be its own issue.

jheyens:
I definitely would not like to have a global mute switch for all blocks simultaneously. 
[...]

PR #233, opened 2018-08-13
Add removable option for battery

lschuermann:
I recently bought a laptop with a dual hot-swappable battery configuration.
To avoid `i3status-rust` throwing an error every time it is swapped, this adds the option to mark a battery as removable.

Currently, I'm just ignoring all errors from `.status()`, in case the battery is removable. Is this fine or do I should I rather check for the specific errors? Additionally, I'm not quite sure whether my code would be seen as idiomatic Rust. Can this be improved?

Thanks!

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -178,8 +178,10 @@ pub struct Battery {
     output: TextWidget,
     id: String,
     update_interval: Duration,
-    device: PowerSupplyDevice,
+    device_name: String,
+    device: Option<PowerSupplyDevice>,
     show: ShowType,
+    removable: bool,
 }
 
 /// Options for displaying battery information.
@@ -207,6 +209,11 @@ pub struct BatteryConfig {
     /// Options for displaying battery information.
     #[serde(default = "BatteryConfig::default_show")]
     pub show: ShowType,
+
+    /// Whether the battery is removable. If it can't be found, no error will
+    /// be thrown.
+    #[serde(default = "BatteryConfig::default_removable")]
+    pub removable: bool,
 }
 
 impl BatteryConfig {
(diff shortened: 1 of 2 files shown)


src/blocks/battery.rs
```diff
@@ -221,34 +228,94 @@ impl BatteryConfig {
...
-            device: try!(PowerSupplyDevice::from_device(block_config.device)),
+            device_name: block_config.device,
+            device,
             show: block_config.show,
+            removable: block_config.removable,
         })
     }
 }
 
+// TODO: Is there a builtin for this?
```
lschuermann: I'm almost certain there is something in the standard library for this. Can't find it however.

lschuermann:
I don't have my laptop with the removable battery anymore, so this PR isn't really relevant to me personally. However, I think that this change would still be useful, especially in the age of convertibles and these other fancy gadgets.

A friend of mine recently bought a T480 with a removable battery option. Now that the commit has merge conflicts, I'd be willing to rebase this PR on the current master. It would be nice to know, whether this has any chance of being merged before I do so.

atheriel:
Hey, thanks for your patience. There are a few issues with this problem historically, but the biggest one is that none of the core contributors have a removable battery to test against. However, I do wish to see better support for those that do. My suggestions if you decide to return to this PR would be:

1. Experiment with the interaction between Upower's D-Bus signaling and removable batteries. I'm simply not sure what happens when these two are combined.
2. Don't bother asking users to mark their batteries as `removable`; instead, just treat all batteries that cannot respond to status requests as "removed".

We can also update the docs to indicate that removed batteries are still polled for updates every `interval` (for non-Upower blocks).

Does that make sense?

JohnAZoidberg:
I don't think it's a good idea to drop the `removable` tag. If the battery isn't removable it should fail and tell the user about the failure.

GladOSkar:
@JohnAZoidberg i disagree. If it is not removable and i see it is "removed", i will know something is wrong. It does not need to crash the status bar.

atheriel:
> I think this is a general question we have to face for all blocks.

It might appear that way at first, but in practice the only complaints we get for annoying crashes under normal operation are for the battery and network-related blocks. Usually the current failure mode is just fine.

> we should absolutely display a "battery not found even though it should be there symbol"

Yes, I assumed we would do that. The correct symbol is `×`, which we also use for the network and sound blocks. I think it is reasonable to do this in all cases where the battery is not responsive, either because it has been removed or (much more rarely) because it is misconfigured.

JohnAZoidberg:
After looking at the issues it seems that everyone who's complaining (#50, #85, #145) is sharing their config between laptop and desktop; and not actually the block failing because there is a problem.

At least one of them doesn't like the idea of having the `x` there constantly on their desktop.
I think the last solution proposed there (`if_command`) is a good way to solve this because it can hide the block on one computer and show it (with possible desired errors) on the other.

Alternatively it would be solved by marking the battery `removable` - it's essentially "removed" by switching from one device to the other.
Then we can have the `x` only for the failing blocks that are not removable.

atheriel:
I believe this functionality is now unnecessary, since we can now set `device = "DisplayDevice"` with UPower and avoid these problems. Is this incorrect?

JohnAZoidberg:
Yes, that works and shows both batteries combined as one. Also it doesn't break when the other battery is unplugged.
But I'd still quite like to see both individually so that I know when to swap out the external battery for a full one.

ammgws:
@lschuermann @JohnAZoidberg Not sure whether you're still using i3status-rs, but I believe this is covered by #835 which was merged recently.

JohnAZoidberg:
Awesome! Thanks for letting us know.

PR #242, opened 2018-08-22
Fix alternating colors in slick theme

GladOSkar:
When i recently recompiled i3 to its newest version i noticed the alternating colors of the slick theme not working correctly.

So i looked at the code and noticed that this was implemented very weirdly. The slick theme did not use the alternating tint, which is meant to be used for this, but rather used a (probably) non-intentional behavior of the tinting, in that it removes the fourth color channel (alpha?). This way, the alternating blocks did not have a fourth channel while every other block had one. Now my new version of i3 apparently does not display this fourth channel anymore, which made all the blocks have the same background color.

This change makes the slick theme use the tint. This also inverts the light and dark blocks compared to the previous version, which is not avoidable since we can only tint positively. Also, all the grey backgrounds are a little brighter. I think it looks alright.

Let me know if i got something wrong :)

### DIFF
--- a/src/themes.rs
+++ b/src/themes.rs
@@ -2,21 +2,21 @@ use std::str::FromStr;
 
 lazy_static! {
     pub static ref SLICK: Theme = Theme {
-        idle_bg: "#424242de".to_owned(),
+        idle_bg: "#424242".to_owned(),
         idle_fg: "#ffffff".to_owned(),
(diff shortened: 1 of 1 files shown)


atheriel:
I actually think this might be related to the recent support for alpha channels with #225. In any case this is a better approach, anyway. Thanks!

PR #243, opened 2018-08-29
Add option to hide cpu usage.

derekdreery:
Because I have very limited screen real estate on my portrait window, it's useful to hide the cpu usage, and just show status changes as it reaches the thresholds. This PR adds an option to do just that.

Also uses cpu-monitor to abstract getting usage.

# Stuff to do
 - [x] update docs

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,9 +1,9 @@
 [[package]]
 name = "aho-corasick"
-version = "0.6.4"
+version = "0.6.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


atheriel:
There are a couple of reasons I'm hesitant about this:

- I am willing to accept that three characters makes a big difference to you, but keep in mind that we need to support this change, and I'm a little doubtful that it will get much use by anyone else. It's generally a good idea to open an issue for discussion before submitting a PR like this.

- This behaviour is not consistent with other blocks. Why is CPU usage special in that it ought to be hidden when below a threshold? Should this be true about the memory or temperature blocks?

- You've introduced a fairly large new dependency (take a look a the lock file changes) to replace code that was, albeit not terribly simple, certainly working. This new dependency is marked as version 0.1 and has all of 21 downloads, which doesn't give me feelings of confidence. If you want to add a new crate, I think you'd need some independent justification. Do we get any other improvements from using this package?

atheriel:
Yes, please try to make `skeptic` an optional or dev dependency. If you can do that I'll accept the new dependency.

> Without the patch, I can set the text for memory and load, but not cpu. I set temp to click to show.

This makes sense to me. But it does make me think that you should be using a `format` config option instead (the same suggestion I made in in #244). That would be more consistent with the existing blocks. If you need help implementing that I can assist you.

> This PR is meant to be the start of a discussion - I find it's easier to discuss something when you have a concrete implementation.

This is mostly so that you don't write code/blocks that we don't end up accepting, which is demoralizing for you, as an author.

derekdreery:
> Yes, please try to make skeptic an optional or dev dependency. If you can do that I'll accept the new dependency.

I don't mind either way. I made the crate because it felt like a gap in the ecosystem, but I can remove it if that's desired. You can pin to the specific version if you want to guarantee it continues to work.

> This makes sense to me. But it does make me think that you should be using a format config option instead (the same suggestion I made in in #244). That would be more consistent with the existing blocks. If you need help implementing that I can assist you.

No worries I'll have a think about what a format option would look like for cpu, to match other format options.

> This is mostly so that you don't write code/blocks that we don't end up accepting, which is demoralizing for you, as an author.

Honestly I don't mind :).

derekdreery:
I think this has kinda been handled now.

PR #244, opened 2018-08-31
Add power consumption (watt)

SWW13:
I added the battery power consumption, but there are some things which needs discussion:
- Adding new values to the `ShowType`-enum for all variants doesn't feel right (see https://github.com/SWW13/i3status-rust/tree/battery_consumption2). I switched to an `Vec` for now.
- I'd love to have an format string for more customization, but that involves adding some sort of template engine and would require some more refactoring. Any plans for generic format solution yet?

battery_consumption

```toml
[[block]]
block = "battery"
interval = 10
show = ["time", "percentage", "watt"]
```

### TODO:
- [x] Finalize config
- [x]  Update Documentation
- [ ] Add/Update Tests

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -88,7 +88,15 @@ Key | Values | Required | Default
 ----|--------|----------|--------
 `device` | The device in `/sys/class/power_supply/` to read from. | No | `"BAT0"`
 `interval` | Update interval, in seconds. | No | `10`
-`show` | Show remaining 'time', 'percentage' or 'both' | No | `percentage`
+`show` | Format string | No | `{level}%`
+
+### Format string
+
+Placeholder | Description
+------------|-------------
+`{percentage}` | Battery level (in percent) excluding %
+`{time}` | Remaining time (in hh:mm)
+`{power}` | Power consumption (in watt) from battery or from power supply for charging
 
 ## CPU Utilization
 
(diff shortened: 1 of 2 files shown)


SWW13:
I switched to `FormatTemplate` and added backwards compatibility.

blocks.md
```diff
@@ -88,7 +88,15 @@ Key | Values | Required | Default
...
 `device` | The device in `/sys/class/power_supply/` to read from. | No | `"BAT0"`
 `interval` | Update interval, in seconds. | No | `10`
-`show` | Show remaining 'time', 'percentage' or 'both' | No | `percentage`
+`show` | Format string | No | `{level}%`
+
+### Format string
+
+Placeholder | Description
+------------|-------------
+`{level}` | Battery level (in percent) excluding %
```
atheriel: These should probably be "percentage", "time", and "power".

atheriel:
The format changes look good. I may not have been very clear about the backwards compatibility issue, though. I think we should add a *new* option, `format`, but keep the old option, `show`, as well. Using `show` in a configuration file has the effect of changing the default format, but is otherwise ignored in favour of the new `format` option. Is that more clear?

atheriel:
Changes look good. Are you able to resolve the merge conflicts? And is there anything else outstanding on this PR?

atheriel:
Thanks!

Issue #20, opened 2017-05-23
[Sound Block] Supporting different audio setups


NullVoxPopuli:
```
 $ RUST_BACKTRACE=1 ./i3status-rs -d ../../../i3status-rust.json
{"version": 1, "click_events": true}
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /checkout/src/libcore/option.rs:323
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:355
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:371
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:549
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:511
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:495
   7: rust_begin_unwind
             at /checkout/src/libstd/panicking.rs:471
   8: core::panicking::panic_fmt
             at /checkout/src/libcore/panicking.rs:69
   9: core::panicking::panic
             at /checkout/src/libcore/panicking.rs:49
  10: <i3status_rs::blocks::sound::Sound as i3status_rs::block::Block>::update
  11: i3status_rs::main
  12: std::panicking::try::do_call
             at /checkout/src/libstd/panicking.rs:454
  13: __rust_maybe_catch_panic
             at /checkout/src/libpanic_unwind/lib.rs:98
  14: std::rt::lang_start
             at /checkout/src/libstd/panicking.rs:433
             at /checkout/src/libstd/panic.rs:361
             at /checkout/src/libstd/rt.rs:57
  15: __libc_start_main
  16: _start

```

I just tried using this for the first time, cause this project looks awesome. :-)

However, I got the above the error.

I'm using this version: 86ff6bc053274ed3f56ffc7b8eb68545e258c118

i3status-rust.json
```
 $ cat i3status-rust.json 
[
  {"block": "temperature", "interval": 10, "collapsed": false},
[...]

PR #245, opened 2018-08-31
PulseAudio Support

SWW13:
I started working on PulseAudio support (#20).

Currently the connection to the PulseAudio server works. Retrieving the current state will require callbacks as the PulseAudio library doesn't offer synchronous calls. Besides getting the callbacks to work the whole sound block should be refactored to work with different audio systems.

Current State:
- [x] retrieve volume / mute state
- [x] listen to state changes
- [x] increase / decrease volume
- [x] mute / unmute
- [x] pulse audio / alsa fallback
- [x] clean up
- [x] documentation

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -536,6 +536,10 @@ pub struct Sound {
 #[serde(deny_unknown_fields)]
 pub struct SoundConfig {
     /// ALSA / PulseAudio sound device name
+    #[serde(default = "SoundDriver::default")]
+    pub driver: SoundDriver,
+
+    /// ALSA / PulseAudio sound device name
     #[serde(default = "SoundConfig::default_name")]
     pub name: Option<String>,
 
@@ -545,6 +549,21 @@ pub struct SoundConfig {
 
     #[serde(default = "SoundConfig::default_on_click")]
     pub on_click: Option<String>,
+}
+
+#[derive(Deserialize, Copy, Clone, Debug)]
+#[serde(rename_all = "lowercase")]
+pub enum SoundDriver {
+    Auto,
+    Alsa,
+    #[cfg(feature = "pulseaudio")]
+    PulseAudio,
+}
+
+impl Default for SoundDriver {
+    fn default() -> Self {
+        SoundDriver::Auto
+    }
 }
 
 impl SoundConfig {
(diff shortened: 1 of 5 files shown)


atheriel:
The general approach looks fine to me, though we might want to make PulseAudio support opt-in. What's the advantage of routing through PulseAudio instead of using ALSA directly?

src/blocks/sound.rs
```diff
@@ -177,48 +389,28 @@ impl ConfigBlock for Sound {
...
         }
 
-        let sound = Sound {
+        let mut sound = Sound {
             text: ButtonWidget::new(config.clone(), &id).with_icon("volume_empty"),
             id: id.clone(),
-            devices: vec![SoundDevice::new("Master")?],
+            devices: vec![
+                // TODO: find better solution for mixed types
+                match PulseAudioSoundDevice::new(0) {
```
atheriel: It seems unlikely that users will have mixed types if they are opting in to PulseAudio for one. However, we clearly need a way of configuring the device in use here. I don't think we should be controlling the volume of multiple devices from the same block.

SWW13:
> The general approach looks fine to me, though we might want to make PulseAudio support opt-in.

Then I'll add a feature-gate for PulseAudio.

> What's the advantage of routing through PulseAudio instead of using ALSA directly?

Many advanced features such as per application audio settings (see https://en.wikipedia.org/wiki/PulseAudio#Features).

SWW13:
I'm about to finish the PulseAudio support, could you please review the current state?
Some notes about the changes: 
- I also cleaned up the ALSA code and added device name support for both device types.
- I couldn't test the ALSA events as they are not working on my system because of the PulseAudio usage.
- The PulseAudio Device uses a global state (with lazy_static) because `libpulse_binding` needs static-lifetime callbacks and does not offer a synchronous API.
- There are two threads running with a PulseAudio server connection, one for sending requests and one for running the event loop to receive callbacks for subscribed events. There are two needed to block for new requests and for new events at the same time. That's the cleanest solution I came up with so far without the need to pull for events.

I have been using the PulseAudio implementation now for some days without any problems.

1wilkens:
FWIW, I also use this branch for a few days and it greatly improves the usefulness of the bar in pulseaudio setups, so I would like to see this merged.
Is anything else missing besides documentation?

atheriel:
I can confirm that this seems to be working, and that the compile-time feature flag seems to work as well. I'm a little uncomfortable with the complexity of the code, and I don't know if I have the bandwidth to really understand it. However, I don't think that's necessarily a barrier to merging it.

The one change that I would like to see is a move away from the PulseAudio -> ALSA fallback system. I'd prefer that users be able to chose their "driver" via the configuration. Could we have an explicit `driver = "pulseaudio" | "alsa" | ""` configuration option?

SWW13:
> I'm a little uncomfortable with the complexity of the code, and I don't know if I have the bandwidth to really understand it. However, I don't think that's necessarily a barrier to merging it.

I'm also not happy with the complexity, but I don't see how it could be reduced much further. 

> Could we have an explicit `driver = "pulseaudio" | "alsa" | ""` configuration option?

That's something I also thought about. Adding an option to explicitly select an driver seems like a good decision.
But I'd like to keep the fallback as default `"auto"` option for easier usage. If pulseaudio is installed it should definitely be used, as the default device is maybe only available in pulseaudio (e.g. bluetooth headphones). But having pulseaudio as default would break alsa only systems when no driver is set.

atheriel:
Yep, I'm happy with the fallback option being the default. That was what I'd expect if the user does not explicitly specify a `driver`.

atheriel:
@SWW13 Happy to merge this now. Did you want me to keep the history intact or can I squash the commits?

SWW13:
I'd prefer squashing.

atheriel:
Thanks for all your work on this. :+1:

Issue #223, opened 2018-07-29
Music block doesn't seem to honour max_width


SergeantBiggs:
I've just started using i3status-rust today and I noticed that the music player block always seems to take the same amount of space, even with "max_width" turned to a lower number. If I do that, it uses even less space from the block (scrolls more), but stays the same size overall. It this wanted?
![bug_i3status_rust](https://user-images.githubusercontent.com/19983492/43369701-b7b60754-9372-11e8-9d89-c60e1edd14da.png)

ammgws:
Related to #155?

SergeantBiggs:
No, doesn't seem like it since I use it, not sway

ammgws:
Pretty sure it is due to [this line](https://github.com/greshake/i3status-rust/blob/bb49eb575fede955a2facc98b050fe9105b85784/src/widgets/rotatingtext.rs#L129) in `RotatingTextWidget` which tells i3 to set the minimum width of the block to 240 pixels. Where as the `max_width` setting for the music block is actually only used for setting the number of characters after which it will start to scroll, and not for setting the width fo the block ([see here](https://github.com/greshake/i3status-rust/blob/bb49eb575fede955a2facc98b050fe9105b85784/src/widgets/rotatingtext.rs#L63-L67).
(Even though the music block docs say "Max width of the block in characters, not including the buttons")

@greshake Is there a reason why the widget is hardcoded to 240 pixels?

SergeantBiggs:
Yes it works now. Thank you!

PR #250, opened 2018-09-07
Fix rotating text widget width

ammgws:
Currently the minimum size of the block is hardcoded to 240 pixels for reasons unknown.
This leads to issues such as #223 with the music block.

Proposed changes:
- set `min_width` based on the user provided `width` value
- `min_width` should be set by characters (by passing a string) instead of by pixels

EDIT:
**1) Fixed the max_width setting for the widget**
Before (latest i3status-git, max_width set to 10 yet block size is large)
ezgif-3-bd71b56c94

After (this PR):
ezgif-3-367da31595

**2) Fixed the 'jitter' during scrolling:**
Before (latest i3status-git):
ezgif-3-78bdb17443

After (this PR):
ezgif-3-1cc0242860 1

### DIFF
--- a/src/widgets/rotatingtext.rs
+++ b/src/widgets/rotatingtext.rs
@@ -126,7 +126,7 @@ impl RotatingTextWidget {
                                 self.get_rotated_content()),
             "separator": false,
             "separator_block_width": 0,
-            "min_width": if self.content == "" {"".to_string()} else {"0".repeat(self.width)},
+            "min_width": if self.content == "" {"".to_string()} else {"0".repeat(self.width+5)},
             "align": "left",
             "background": key_bg,
             "color": key_fg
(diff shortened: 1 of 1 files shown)


ammgws:
Fixing the 'jitter' during scrolling:

Before (latest i3status-git):
ezgif-3-78bdb17443

After (this PR):
ezgif-3-1cc0242860 1

ammgws:
Updated first post with visual of fixes.

atheriel:
This looks good to me, but I'm not a consumer of the music block and I know it's @greshake's baby. We'll wait a bit to see if he has anything to say.

greshake:
All good. To quote the i3bar protocol docs: 

> The value can also be a string. In this case, the width of the text given by min_width determines the minimum width of the block. This is useful when you want to set a sensible minimum width regardless of which font you are using, and at what particular size. 

Nice, I didn't know!

PR #251, opened 2018-09-09
Allow use of pango markup in separator

ammgws:
By using [pango markup](https://i3wm.org/docs/i3bar-protocol.html) we can set the font type, size, etc individually just for the separator block, which can be useful in debugging/fixing font issues or just in terms of customisation.

eg in config.toml:
```
[theme.overrides]
separator = "<span font='FontJustForSeparator 12'></span>"
```

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -136,7 +136,7 @@ pub fn print_blocks(order: &Vec<String>, block_map: &HashMap<String, &mut Block>
                     "separator_block_width": 0,
                     "background": if sep_bg.is_some() { Value::String(sep_bg.unwrap()) } else { Value::Null },
                     "color": sep_fg,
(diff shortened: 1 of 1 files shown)


atheriel:
I think this is missing the definition of the `pango` symbol. The build failure is a real one.

ammgws:
Sorry that's my bad. It should be a string.

atheriel:
As long as there's no big performance cost to Pango rendering, why wouldn't we just set `markup: "pango"` as the default?

ammgws:
Isn't that what I've done? Or do you mean for normal blocks as well?

atheriel:
Ah, I misinterpreted -- I thought you wanted it to be a variable configured somewhere.

PR #252, opened 2018-09-10
Add dbus/upower support to the battery block

atheriel:
This feature is built on the block submitted by @uinput in #231, and includes that branch in the first commit.

I thought that the upower code was sufficiently useful for the main battery block that we should just move it there. As a result, battery updates are both (1) more responsive to change in stage, e.g. when unplugging, and (2) considerably more infrequent. Since not all users will have upower installed, you must opt in to this feature by setting `upower = true` in your battery block. Otherwise we fall back on sysfs and update intervals.

Since I believe that this superceeds all features of the contributed upower block, I've removed it. Does this resolution make sense to you, @uinput?


uinput:
Yes, I think this is a good solution for all users. Thank you!

JohnAZoidberg:
The documentation in /blocks.md is missing for this.

atheriel:
@JohnAZoidberg Added in 7d1703d.

src/blocks/battery.rs
```diff
@@ -191,13 +206,154 @@ impl PowerSupplyDevice {
...
+                .block_error("battery", "Failed to read UPower Percentage property.")?;
+
+        if capacity.0 > 100.0 {
+            Ok(100)
+        } else {
+            Ok(capacity.0 as u64)
+        }
+    }
+
+    fn time_remaining(&self) -> Result<u64> {
```
GladOSkar: Hi @atheriel i just noticed that "Time to full" does not work with this PR even though upower does support it.

This looks like the "Remaining time" (the one to be shown) is only set to the "TimeToEmpty" here.

It should be set to "TimeToFull" while charging.

atheriel: Ah, you are correct.

Issue #249, opened 2018-09-05
When a block is clicked, why check block id in each block?


cljoly:
I’m wondering why it is up to the `click` function from each block to check [(as advised here)](https://github.com/greshake/i3status-rust/blob/bb49eb575fede955a2facc98b050fe9105b85784/src/block.rs#L18).

Why isn’t this checked [at this point](https://github.com/greshake/i3status-rust/blob/bb49eb575fede955a2facc98b050fe9105b85784/src/main.rs#L241)? This way, each block wouldn’t have to figure out whether it was clicked.

atheriel:
I always thought it was an artifact of the way the i3bar protocol sends events. But it's quite clunky. If you can come up with an alternative I'd happily refactor it.

greshake:
It's because every block can have multiple sub-widgets (buttons, mostly) that can be clicked independently. One example is the music block which has three different buttons:
```rust
fn click(&mut self, event: &I3BarEvent) -> Result<()> {
        if let Some(ref name) = event.name {
            let action = match name as &str {
                "play" => "PlayPause",
                "next" => "Next",
                "prev" => "Previous",
                _ => "",
            };
            if action != "" {
                let m = Message::new_method_call(
                    format!("org.mpris.MediaPlayer2.{}", self.player),
                    "/org/mpris/MediaPlayer2",
                    "org.mpris.MediaPlayer2.Player",
                    action,
                ).block_error("music", "failed to create D-Bus method call")?;
                self.dbus_conn
                    .send(m)
                    .block_error("music", "failed to call method via D-Bus")
                    .map(|_| ())
            } else {
                Ok(())
            }
        } else {
            Ok(())
        }
}
```

greshake:
Also we have the flexibility of allowing blocks to react to arbitrary clicks- I believe this was something I thought was valuable for some advanced concepts I never got to implement (in particular removable storage- imagine a new block popping up for any removable storage with mount/unmount buttons)

greshake:
In a future rewrite, I'd obviously prefer a method where every widget has an implementable click function, but right now that would throw us into some turbulent waters in terms of lifetimes and references.

leowzukw:
Thanks for your answers.

At least, the code to check for the simple case may be factored. If you are ok with that, I would do a PR.

greshake:
We could rename `click` -> `click_any` and add a method `click` that is only called when a widget from the current block is called with an empty default implementation

leowzukw:
[...]

PR #253, opened 2018-09-10
[WIP] Factor code to check block id when clicked

leowzukw:
Implement [this comment](https://github.com/greshake/i3status-rust/issues/249#issuecomment-419442576) from #249.

We want to factor the code called for each block, that checks whether the block was clicked or not.

- [x] Refactor `click` to `click_any`.
- [ ] Add the new `click` method, relaying click inside the block only.
- [ ] Use this new method in blocks where it makes sense.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -15,6 +15,14 @@ pub trait Block {
     fn view(&self) -> Vec<&I3BarWidget>;
 
     #[allow(unused_variables)]
+    /// This function is only called for click on the block. You may need
+    /// function click_any for advanced use. Please note that both this function
+    /// and click_any, will be executed.
+    fn click(&mut self, event: &I3BarEvent) -> Result<()> {
+        Ok(())
+    }
+
+    #[allow(unused_variables)]
     /// This function is called on every block for every click.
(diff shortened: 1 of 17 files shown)


greshake:
This PR should be joined with the implementation of the new click method that only relays clicks on the same block

leowzukw:
If the `click` function from 69941f0 is what you want, I could start to use it instead of `click_any` in some block.

greshake:
Yea that's it, my bad :D

greshake:
Just noticed: The problem is that the event name isn't the same as the block id. Since any block can have multiple widgets (like the play/pause buttons on the music block) the event name matches the widget id instead, so we'd need to know which widgets belong to which block

leowzukw:
I will have a look at this.

leowzukw:
Finally found time to work on this.

So, I’m not sure to follow, you want me to add a method to block to get a list of each widget id in that block? Or should we use the click method for any block with widget?

leowzukw:
ping @greshake ?

leowzukw:
Poke poke 😄

PR #262, opened 2018-09-17
Wind direction

incub77:
Added the option to show wind direction via the weather block.
I absolutely don't know, whether this implementation is a rusty way to do thing, as I don't know that language :) Comments are welcome!

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -577,7 +577,7 @@ Show detailed weather in San Francisco through the OpenWeatherMap service:
 ```toml
 [[block]]
 block = "weather"
-format = "{weather} ({location}) {temp}°, {wind} km/s"
+format = "{weather} ({location}) {temp}°, {wind} m/s"
 service = { name = "openweathermap", api_key = "XXX", city_id = "5398563", units = "metric" }
 ```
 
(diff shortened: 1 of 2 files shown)


src/blocks/weather.rs
```diff
@@ -138,6 +147,29 @@ impl Weather {
                     }
                 };
 
+                // Convert wind direction in azimuth degrees to abbreviation names
```
atheriel: The more idiomatic pattern here is to `match` on the range of values, e.g.
```rust
match direction {
  23..67 => "NE".to_string(),
  ...
  293..337 => "NW".to_string(),
  _ => "N".to_string(),
}
```

incub77: Changed it, tanks for the hint!

blocks.md
```diff
@@ -577,7 +577,7 @@ Show detailed weather in San Francisco through the OpenWeatherMap service:
 ```toml
 [[block]]
 block = "weather"
-format = "{weather} ({location}) {temp}°, {wind} km/s"
+format = "{weather} ({location}) {temp}°, {wind} km/s {direction}"
```
atheriel: I believe this should be `m/s`, as you pointed out in #261.

atheriel:
I've confirmed this works and will merge it shortly -- can you fix the typo or should I?

blocks.md
incub77: Done.

PR #263, opened 2018-09-25
Add IBus block

ammgws:
`IBus` (Intelligent Input Bus) is an input method user interface which is used to type in languages other than English, such as Japanese.

This block displays the current input method ("global engine") in the status bar so that you can check before starting to type something. It updates instantly by subscribing to D-Bus signal event emitted by IBus.

Have been using this for a couple of weeks with no issues so far, however I kindly request a code review due to being a Rust noob.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -205,9 +205,9 @@ Key | Values | Required | Default
 ----|--------|----------|--------
 `max_width` | Truncates titles to this length. | No | `21`
 
-## ibus
+## IBus
 
-Creates a block which displays the current global engine set in ibus. Updates are instant as D-Bus signalling is used.
+Creates a block which displays the current global engine set in [IBus](https://wiki.archlinux.org/index.php/IBus). Updates are instant as D-Bus signalling is used.
 
 ### Examples
 
--- a/blocks.md
+++ b/blocks.md
@@ -6,6 +6,7 @@
 - [Custom](#custom)
 - [Disk Space](#disk-space)
 - [Focused Window](#focused-window)
+- [ibus](#ibus)
 - [Load](#load)
 - [Maildir](#maildir)
 - [Memory](#memory)
@@ -203,6 +204,17 @@ max_width = 21
 Key | Values | Required | Default
 ----|--------|----------|--------
 `max_width` | Truncates titles to this length. | No | `21`
+
+## ibus
+
+Creates a block which displays the current global engine set in ibus. Updates are instant as D-Bus signalling is used.
+
+### Examples
+
+```toml
+[[block]]
+block = "ibus"
+```
 
 ## Load
 
--- a/src/blocks/ibus.rs
+++ b/src/blocks/ibus.rs
@@ -154,8 +154,10 @@ fn parse_msg(ci: &ConnectionItem) -> Option<&str> {
 fn get_ibus_address() -> Result<String> {
     // TODO: Check IBUS_ADDRESS variable, as it seems it can be manually set too.
 
+    // TODO: Don't fail if $XDG_CONFIG_HOME is not set. 
+    // Next try $HOME/.config, then only error if that $HOME is not set.
     let config_dir = env::var("XDG_CONFIG_HOME")
-        .block_error("ibus", "Problem getting $XDG_CONFIG_HOME")?;
+        .block_error("ibus", "$XDG_CONFIG_HOME not set")?;
 
     // TODO: Check /var/lib/dbus/machine-id if /etc/machine-id fails
     let mut f = File::open("/etc/machine-id")
--- a/src/blocks/ibus.rs
+++ b/src/blocks/ibus.rs
@@ -185,7 +185,7 @@ fn get_ibus_address() -> Result<String> {
         .block_error("ibus", &format!("Could not open {}", ibus_socket_path))?;
     let mut ibus_address = String::new();
     f.read_to_string(&mut ibus_address)
-        .block_error("ibus", &format!("Error reading contents of {}", ibus_address))?;
+        .block_error("ibus", &format!("Error reading contents of {}", ibus_socket_path))?;
     let re = Regex::new(r"IBUS_ADDRESS=(.*),guid").unwrap(); // valid regex expression will not cause panic
     let cap = re.captures(&ibus_address)
         .block_error("ibus", &format!("Failed to extract address out of {}", ibus_address))?;
--- a/src/blocks/ibus.rs
+++ b/src/blocks/ibus.rs
@@ -1,0 +1,170 @@
+use std::env;
+use std::fs::File;
+use std::io::prelude::*;
+use std::sync::{Arc, Mutex};
+use std::thread;
+use std::time::{Duration, Instant};
+use chan::Sender;
+
+use block::{Block, ConfigBlock};
+use config::Config;
+use errors::*;
+use widgets::text::TextWidget;
+use widget::I3BarWidget;
+use input::I3BarEvent;
+use scheduler::Task;
+
+use uuid::Uuid;
+
+extern crate dbus;
+extern crate regex;
+use self::dbus::{Connection, ConnectionItem, stdintf, arg};
+use self::regex::Regex;
+
+pub struct IBus {
+    id: String,
+    text: TextWidget,
+    engine: Arc<Mutex<String>>,
+}
+
+#[derive(Deserialize, Debug, Default, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct IBusConfig {
+    /// Set to display engine name as icon.
+    #[serde(default = "IBusConfig::default_as_icon")]
+    pub as_icon: bool,
+}
+
+impl IBusConfig {
+    fn default_as_icon() -> bool {
+        true
+    }
+}
+
+impl ConfigBlock for IBus {
+    type Config = IBusConfig;
+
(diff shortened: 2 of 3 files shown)


src/blocks/ibus.rs
```diff
@@ -0,0 +1,185 @@
...
+// Example file contents:
+// ```
+// # This file is created by ibus-daemon, please do not modify it
+// IBUS_ADDRESS=unix:abstract=/tmp/dbus-8EeieDfT,guid=7542d73dce451c2461a044e24bc131f4
+// IBUS_DAEMON_PID=11140
+// ```
+fn get_ibus_address() -> Result<String> {
+    // TODO: Check IBUS_ADDRESS variable, as it seems it can be manually set too.
+
+    let config_dir = env::var("XDG_CONFIG_HOME")
```
atheriel: Do I understand correctly that this will error out if this variable is unset? Or is it just for errors in actually retrieving the environment variable?

src/blocks/ibus.rs
```diff
@@ -0,0 +1,185 @@
...
+    // Hence on sway you will need to reload the bar once after login to get the block to work.
+    let display = env::var("DISPLAY")
+        .block_error("ibus", "$DISPLAY not set. Try restarting block if on sway")?;
+    let re = Regex::new(r"^:(\d{1})$").unwrap();
+    let cap = re.captures(&display).unwrap();
+    let display = &cap[1].to_string();
+
+    let hostname = String::from("unix");
+
+    let ibus_socket_path = format!("{}/ibus/bus/{}-{}-{}", config_dir, machine_id, hostname, display);
```
atheriel: Is there any reason to expect that there will be more than one file in `ibus/bus`? If not, you could avoid a great deal of effort here and just open the only file in that directory, if it exists.

src/blocks/ibus.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+impl Block for IBus {
+    fn id(&self) -> &str {
+        &self.id
+    }
+
+    // Updates the internal state of the block.
+    fn update(&mut self) -> Result<Option<Duration>> {
+        let string = (*self.engine
```
atheriel: Maybe call this `engine`, not `string`.

src/blocks/ibus.rs
```diff
@@ -0,0 +1,185 @@
...
+        let info: arg::Variant<Box<arg::RefArg>> = p.get("org.freedesktop.IBus", "GlobalEngine")
+            .block_error("ibus", "Couldn't get ibus address")?;
+
+        // `info` should contain something containing an array with the contents as such:
+        // [name, longname, description, language, license, author, icon, layout, layout_variant, layout_option, rank, hotkeys, symbol, setup, version, textdomain, icon_prop_key]
+        // Refer to: https://github.com/ibus/ibus/blob/7cef5bf572596361bc502e8fa917569676a80372/src/ibusenginedesc.c
+        // e.g.                   name           longname        description     language
+        // ["IBusEngineDesc", {}, "xkb:us::eng", "English (US)", "English (US)", "en", "GPL", "Peng Huang <shawn.p.huang@gmail.com>", "ibus-keyboard", "us", 99, "", "", "", "", "", "", "", ""]
+        //                         ↑ We will use this element (name) as it is what GlobalEngineChanged signal returns.
+        let current = info.0.as_iter().unwrap().nth(2).unwrap().as_str().unwrap_or("??");
```
atheriel: Can you avoid using `unwrap()` here (and really anywhere) in favour of `block_error()` and friends?

blocks.md
```diff
@@ -204,6 +205,17 @@ Key | Values | Required | Default
 ----|--------|----------|--------
 `max_width` | Truncates titles to this length. | No | `21`
 
+## ibus
+
+Creates a block which displays the current global engine set in ibus. Updates are instant as D-Bus signalling is used.
```
atheriel: You might want to include a link explaining what "ibus" is here.

atheriel:
The biggest thing I notice in the code is the use of `unwrap()` and `expect()` -- please try to get rid of those unless absolutely necessary.

Also, what will happen if IBus is not installed/available? Is there a reason to think this might happen?

src/blocks/ibus.rs
ammgws: @atheriel
Actually I have more than one file in there since $DISPLAY isn't constant and ibus doesn't clear the folder at any stage.

src/blocks/ibus.rs
ammgws: @atheriel agreed

ammgws:
> The biggest thing I notice in the code is the use of `unwrap()` and `expect()` -- please try to get rid of those unless absolutely necessary.

I am working on this at the moment.

> Also, what will happen if IBus is not installed/available? Is there a reason to think this might happen?

I think it would be strange to enable the ibus block if you don't have it installed... regardless, this is what (should) happen in that case:

| Situation | Expected Result |
| --- | --- |
| IBus not running/installed, stale file in ~/.config/ibus/bus/ | block_error: failed to connect to DBus |
| IBus not running/installed, no file in ~/.config/ibus/bus/ | block_error: could not get IBus address |
| IBus restarted while block is running | undefined atm. doesn't crash the block though |

For the last one I will try implement IBusBus::disconnected signal to catch when it is restarted and retry the connection gracefully.

src/blocks/ibus.rs
ammgws: > The returned result is Ok(s) if the environment variable is present and is valid unicode. If the environment variable is not present, or it is not valid unicode, then Err will be returned.

Errors if not set. I guess I should update the block_error message to make it clearer.

src/blocks/ibus.rs
ammgws: Have done this now. Only these ones are left:

```rust
^^/i3status-rust >>> rg "unwrap()" src/blocks/ibus.rs
// Do these need to be replaced? All the other blocks using D-Bus were like this:
77:            let c = Connection::open_private(&ibus_address).unwrap();
80:            ).unwrap();
// this should be OK
// "If another user of this mutex panicked while holding the mutex, then this call will return an error once the mutex is acquired."
84:                            let mut engine = engine_original.lock().unwrap();
// These return Options on the received dbus message item. Don't see why these would error..
134:    if &*m.interface().unwrap() != "org.freedesktop.IBus" { return None };
135:    if &*m.member().unwrap() != "GlobalEngineChanged" { return None };

// These will only fail if the given regex expression is invalid.
176:    let re = Regex::new(r"^:(\d{1})$").unwrap();
189:    let re = Regex::new(r"IBUS_ADDRESS=(.*),guid").unwrap();

File: src/blocks/ibus.rs (hunk 1 above)
Comment (atheriel) [reply src/blocks/ibus.rs]: You should also fall back on the default specified by the XDG standard, which is `~/.config` I believe.

**Comment by atheriel** [comment]
@ammgws OK, those failure modes make sense to me.

File: src/blocks/ibus.rs (hunk 2 above)
Comment (atheriel) [reply src/blocks/ibus.rs]: If it's just `$DISPLAY` we expect to change we can just look for the file matching that, right? No need to muck around by guessing the hostname and machine ID. Or do users end up with more than one machine ID?

We could also let (or expect that) users set a `socket_path` manually in their configuration.

File: src/blocks/ibus.rs (hunk 1 above)
Comment (ammgws) [reply src/blocks/ibus.rs]: Agreed: 
> If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

File: src/blocks/ibus.rs (hunk 2 above)
Comment (ammgws) [reply src/blocks/ibus.rs]: Actually that is a good point. So can simplify the code by looking at the directory conents and pattern match 32charstring-string-$DISPLAY. Or perhaps first check the directory and if there is only one file just go with that, otherwise if more than one file then pattern match to find the right one.

> The machine ID is usually generated from a random source during system installation or first boot and stays constant for all subsequent boots. Optionally, for stateless systems, it is generated during runtime during early boot if necessary. The machine ID is a single newline-terminated, hexadecimal, 32-character, lowercase ID.

File: src/blocks/ibus.rs (hunk 4 above)
Comment (ammgws) [reply src/blocks/ibus.rs]: Done. 

By the way, off-topic to this PR but there's about 50 or so `unwrap()`s used in the other blocks in the codebase. Ideally should these be looked at and replaced one day? If some still remain after refactoring then do we assume nothing is obvious and add a comment to each one explaining why they will not fail?

File: src/blocks/ibus.rs [hunk 6]
```diff
@@ -0,0 +1,197 @@
...
+//       ccc = display number from $DISPLAY
+// Refer to: https://github.com/ibus/ibus/blob/7cef5bf572596361bc502e8fa917569676a80372/src/ibusshare.c
+//
+// Example file contents:
+// ```
+// # This file is created by ibus-daemon, please do not modify it
+// IBUS_ADDRESS=unix:abstract=/tmp/dbus-8EeieDfT,guid=7542d73dce451c2461a044e24bc131f4
+// IBUS_DAEMON_PID=11140
+// ```
+fn get_ibus_address() -> Result<String> {
```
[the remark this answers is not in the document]
atheriel: Just use `get_ibus_address()?` on L51.

atheriel:
I've rebased and merged this in 5104f57. Thanks for your work! The master branch also contains your suggested changes to `get_ibus_address()` -- try it out and see if it works for you.

ammgws:
Cheers for finishing this off! It's working well so far.

PR #268, opened 2018-09-30
fallback on nmcli to get ssid if iw is not present

mathroc:


wmww:
Looks good to me; Does what it should do on my machine.

PR #269, opened 2018-09-30
Update nix crate to 0.11

dywedir:
Should fix build on aarch64 (see https://github.com/NixOS/nixpkgs/pull/47566#issuecomment-425752132)


atheriel:
Looking at the `.lock` file, we're actually dropping a few dependencies, which is nice.

PR #271, opened 2018-10-01
Move themes docs from wiki to repo

### LINKED ISSUE #226: Update wiki to reflect new overrides.
I stumbled across the challenge to change the theme to a fully transparent background. What I didn't understand at first was why I'd get a background for some block while others successfully turned transparent. The issue was the alternating of the colors. After reading the source, I tried to override the alternating color, which worked partially (but I wasn't there yet)
```
alternating_tint_bg = "#00000000"
```
For this to take effect one also needs alpha channel support, which I added in #225.
I'd be really glad if that theme override could be added to the wiki page `Theming & Icons`, since it would have saved me some time.

Also, the backlight icon overrides are missing there, namely `backlight_full`, `backlight_partial1`, `backlight_partial2`, `backlight_partial3`, and `backlight_empty`. Maybe you could add them as well.

ammgws:
Fixes #226 and also allows future changes to be done via PR.

### DIFF
--- a/themes.md
+++ b/themes.md
@@ -1,4 +1,5 @@
-The bar can be themed either by specifying a pre-complied theme or overwriting defaults in the configuration. We differentiate between themes and icon sets.
+The bar can be themed either by specifying a pre-complied theme or overwriting defaults in the configuration.  
+We differentiate between themes and icon sets.
 
 In order to change either, add them to your configuration:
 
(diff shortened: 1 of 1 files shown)


themes.md
```diff
@@ -0,0 +1,81 @@
...
+bat = " | | "
+bat_full = " |X| "
+bat_charging = " |^| "
+bat_discharging = " |v| "
+```
+
+Example configurations can be found as `example_theme.toml` and `example_icon.toml`.
+Here's a full list of available overrides:
+
+| Theme        		    | Icons
```
atheriel: Can you just turn this into two bullet lists? Maintaining a table is going to turn into a whitespace nightmare.

atheriel:
Thank you!

themes.md
ammgws: What do you think about a preparing a small script to generate the table from the respective source files? Would eliminate mistakes plus we could keep the table formatting.

ammgws: For now I will change it to two bullet lists.

ammgws:
OK should be good to go now.

ammgws:
@atherial Anything else need to be done?

PR #273, opened 2018-10-03
Fix parsing of input JSON from new sway

wmww:
Previously, if the separating comma was at the end of an input buffer, `serde_json::from_str(...)` would return `Err` and the program would panic. The same would happen if the input read hit `EOF`. The former issue was surfaced by running in the latest swaybar (compiled from git), and the latter while debugging the former.

With my changes, only what's between the first open and last close curly brace is sent to Serde, so both problems are fixed. We also don't need to special-case the initial `[` anymore.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -44,12 +44,12 @@ pub fn process_events(sender: Sender<I3BarEvent>) {
         let mut input = String::new();
         io::stdin().read_line(&mut input).unwrap();
 
-        if !input.starts_with('[') {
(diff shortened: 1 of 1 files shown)


atheriel:
Thanks, this is a nice bugfix. Can we confirm that this does not break under older versions of `swaybar` or `i3bar`?

wmww:
Tested just now in i3 with i3bar. Seems to work fine to me.

Issue #275, opened 2018-10-07
PulseAudio support causes compilation error


crablab:
@SWW13 I think your most recent change may have broken something as I can no longer compile Master. 

```
[...]

PR #276, opened 2018-10-08
Update Cargo.toml to fix pulseaudio library issues

crablab:
Changes `libpulse-binding` requirements to fix issue #275 as suggested by @atheriel


atheriel:
Ah, thanks!

PR #282, opened 2018-10-08
fix getting wifi ssid on ubuntu

### LINKED ISSUE #280: SSID lookup in net block fails because of erronous parameters
The recently introduced fallback in the "net" block for the SSID lookup introduced a new option to the `/bin/sh` `Command` call, `-o pipefail`, which is not supported by most `sh` implementation, notably `dash`, which is [shipped with Ubuntu](https://packages.ubuntu.com/bionic/dash).

I would propose to just drop the `-o pipefail`.

Reference: mathroc@93845e414dd5b5b5840a5e6614d46be878c94fe3

mathroc:
on ubuntu and probably other systems, /bin/sh does not support `-o pipefail`

closes #280


mathroc:
@moritzheiber, @wmww can you test that I did not mess something else up ?

moritzheiber:
How about splitting the command call and evaluating the result in Rust instead of the shell? That would also make the function as a whole a lot more testable and less reliant on outside factors (shell, tool output changed etc)

mathroc:
@moritzheiber true, maybe I'll take a look at how it could be done on the Rust side

atheriel:
Can we confirm that this works correctly on Ubuntu?

moritzheiber:
I can confirm this is working on Ubuntu 18.04.

faulesocke:
Can we get this merged?

PR #294, opened 2018-10-21
Add solarized-light theme

JohnAZoidberg:
I'm a fan of light themes :) If there's `solarized-dark` there should be `solarized-light`.

Also adds comments to the existing solarized-dark theme to identify
which solarized colors the hex codes represent.


atheriel:
Looks good to me. Thanks!

PR #295, opened 2018-10-22
sound: Add config show_volume_when_muted

JohnAZoidberg:
The sound block doesn't show the volume when it's muted. This commit
adds a config option to enable just that.

It is useful to see how high the volume is turned up before unmuting.
This also allows you to adjust the volume while muted and then unmute at
the right volume.


JohnAZoidberg:
I don't actually think we need the config option- just enable it by default. Would anybody not want this?
Fewer config options -> less room for bugs.

JohnAZoidberg:
What do you think about this?
I've been applying this patch for the last 7 months.

atheriel:
Just hadn't properly reviewed it. Looks good, thanks for your patience! I think keeping the existing behaviour by default is the right option.

Issue #220, opened 2018-07-17
Calculate the capacity from multiple batteries.


droslean:
With more batteries all the calculations are wrong. 
We should summarise the capacity from all the batteries and output the percentage to the bar.

In the configuration, we would be able to specify more batteries there, instead of getting the default BAT0.
Which in most cases the default battery is not BAT0

GladOSkar:
I think it should be an option, so you can choose if you want the batteries displayed separately or as one.

droslean:
In which cases you need to check the batteries seperatly?

GladOSkar:
For example if one of the batteries is your UPS or if you have a hot-swappable battery you want to put into an external charger when it's empty

atheriel:
There was an attempt to support multiple batteries in #170, but it fell apart. At the moment we make very simple calculations for capacity and (dis)charge time, but I recently refactored the internal implementation to separate battery devices from the block itself, so it's quite possible to write code that accounts for multiple batteries when estimating these values. I would also like to see a more sophisticated estimate of (dis)charge time in the future -- surely there are models out there.

Unfortunately, we're somewhat stymied by the fact that the battery block only expects a single device, and I'm skeptical there are enough users with multiple batteries to justify breaking compatibility with everyone else's config. At the moment, I think that one possible way forward is to write a new block that uses the same underlying device type but allows for more idiosyncratic setups. Unfortunately, I don't personally have a second battery, so I'm not the best person to bang out an implementation.

GladOSkar:
@atheriel I assume that refactor is why you haven't commented on my PR yet. I can look into time estimation and multiple batteries some more in mid-august.

atheriel:
Nope, that just fell off my radar. Sorry!

atheriel:
This should now be possible (after #356) by setting `device = "DisplayDevice"`. I don't have a second battery; can someone confirm this?

droslean:
@atheriel Nope. #356 , Is not related.

atheriel:
@droslean Can you explain why? The Upower docs point to it doing exactly what you've asked for here -- or is that not what happens in practice?

droslean:
@atheriel Sry I was wrong. The #356 it actually solves this problem!!!

/close

atheriel:
Ah, I'm glad to hear that. Credit to @frececroka for finding a solution.

PR #297, opened 2018-10-27
Add support for multiple batteries

mattclement:
I use a Thinkpad with a removable battery, and made this to combine the stats into one. I've been running this successfully for a week now. I'm very new to rust, so if I did things incorrectly please let me know since I'd like to learn.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -83,16 +83,25 @@ interval = 10
 format = "{percentage}% {time}"
 ```
 
+Show the combined stats of multiple batteries:
+```toml
+[[block]]
+block = "battery"
+devices = ["BAT0", "BAT1"]
+```
+
 ### Options
 
 Key | Values | Required | Default
 ----|--------|----------|--------
 `device` | The device in `/sys/class/power_supply/` to read from. | No | `"BAT0"`
+`devices` | List of devices in `/sys/class/power_supply` to read from. | No | `[]`
 `interval` | Update interval, in seconds. | No | `10`
 `format` | A format string. See below for available placeholders. | No | `"{percentage}%"`
 `show` | Deprecated in favour of `format`. Show remaining `"time"`, `"percentage"` or `"both"` | No | `"percentage"`
 
 The `show` option is deprecated, and will be removed in future versions. In the meantime, it will override the `format` option when present.
+`devices` will take priority over `device`.
 
 ### Format string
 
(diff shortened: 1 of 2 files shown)


GladOSkar:
Related: #220

GladOSkar:
The config syntax makes sense to me, but i can't test it. @droslean could you test it?

Also a `merge = [true/false]` option would be great if it fits into this PR

GladOSkar:
That is a good question actually, i didn't think about that.

If we had 2 Battery blocks that would not be very nice since i think the estimated remaining time should always be merged. Having 2 remaining times does not sound useful at all.

But having Labels may be useful since every usecase i have for wanting the blocks to be separate is to know the state of a specific device (BT Mouse, UPS, etc.)

I think having configurable labels (while it may be a little feature-creepy) is the way to go. Maybe we could use `/sys/class/power_supply/BAT0/model_name` as a default?

droslean:
@GladOSkar  I've tested it. It seems that its working fine.

atheriel:
Thanks for your patience. I have a few comments here, although I am supportive of this feature.

First, what's the interaction between multiple devices and Upower? Or do we only support polling `sysfs` when you have multiple devices? I'd very much like to get Upower support, since it gives such a huge performance gain for this particular block.

Secondly, I'm somewhat unhappy with the `devices` configuration field, because I think it's a workaround we might not really need. I have an inkling that we can write some custom `serde` code to consume strings as a one-element vector, which would mean that we could just use the existing `device` and handle it differently for single vs. multiple batteries. However, writing `serde` code is extremely beginner-unfriendly in my experience, so I can't ask that you do this yourself. I will have to look into it when I have time. (For anyone who's particularly adventurous, you can see our existing `serde` deserialization hacks [here](https://github.com/greshake/i3status-rust/blob/master/src/de.rs)).

Finally, some comments with respect to @GladOSkar's comments on formatting. I'm not sure I really agree that you'd always want the time remaining to be merged -- it's certainly relevant information that one of your batteries is close to empty. Moreover, I'm quite sure that time remaining estimates we make via `sysfs` for a battery that is not discharging will be highly inaccurate, so combining multiple time estimates would actually give you a pretty meaningless result. It's possible that combining estimates from Upower would be more accurate, but I'd like to see some investigation along those lines as well. What do those with multiple batteries see in terms of time estimates when they have two blocks at the moment? How does it look with `sysfs` vs. Upower?

atheriel:
@mattclement Sorry to necro this thread, but is it possible that the support for Upower's `DisplayDevice` in #356 can supersede this PR? Or is there additional functionality here that I'm missing?

mattclement:
@atheriel I think that support does the same thing, yeah.

PR #299, opened 2018-10-30
Use SHELL env variable for custom and toggle block

JohnAZoidberg:
With the custom block a user can specify a custom command for a block.
By default it is evaluated by `sh`. The `on_click` command is however
evaluated by `$SHELL` if it's set. That behaviour should also be used
for `command` because `sh` is very limited and the user expects the
behaviour to be the same as their default shell.

This commit checks if `$SHELL` is set and uses this for all user
specified commands in the `custom` and `toggle` blocks.


atheriel:
Thanks!

PR #301, opened 2018-10-31
Configurable threshold in temperature block

leowzukw:
It’s pretty handy to adapt the threshold from one state to another to
your own hardware.

This is also a feature mentioned as missing in `blocks.md`.

### DIFF
--- a/src/blocks/temperature.rs
+++ b/src/blocks/temperature.rs
@@ -20,6 +20,10 @@ pub struct Temperature {
     collapsed: bool,
     id: String,
     update_interval: Duration,
+    maximum_good: i64,
+    maximum_idle: i64,
+    maximum_info: i64,
+    maximum_warning: i64,
     format: FormatTemplate,
 }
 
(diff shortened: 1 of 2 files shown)


src/blocks/temperature.rs
```diff
@@ -128,13 +169,15 @@ impl Block for Temperature {
...
             }
 
-            self.text.set_state(match max {
-                0...20 => State::Good,
-                21...45 => State::Idle,
-                46...60 => State::Info,
-                61...80 => State::Warning,
-                _ => State::Critical,
-            });
+            let state =
```
atheriel: I'd suggest moving back to the `match` syntax here. It's a little more verbose in this case but easier to reason about.

atheriel:
Thanks, this is a nice addition. Can you take a look at the merge conflicts?

src/blocks/temperature.rs
leowzukw: Done in dd0cd52

leowzukw:
Dealt with merge conflict.

Issue #310, opened 2018-12-01
Internal: requests for update should not be repeated again and again for ever.


rtxm:
In https://github.com/greshake/i3status-rust/blob/287282b06316a3b4525578386ebc4dd21ead6622/src/main.rs#L227 request for updates are added to the scheduler that keeps repeating forever as soon as the block `.update` method always returns a time interval.

This affects the following blocks:

- music (each time a bus event is received);
- custom (each time you click it);
- memory (each time you click it).

This explains #290 and is not limited to the music block. It happens that the music blocks receives many more external events that are converted to requests for update internally than any other block and thus exhibits much sooner higher CPU usage.

atheriel:
Yes, this is clearly an issue, and probably the heart of #290. If anyone's keen to take on the scheduling code we're very happy to take PRs.

rtxm:
I propose not to schedule *requests for update* and process them immediately as illustrated in PR #311.

atheriel:
Has this been resolved by #311? If so, we should close the issue.

xxxserxxx:
@atheriel, I don't think so.  See #569, which I traced down to the music block -- unless what I found is unrelated to the root cause of this.

ammgws:
@rtxm @xxxserxxx With #906 the music block should not spam as many updates anymore. I haven't been able to reproduce this issue for a while (my main player has been `mpd` this whole time) so I cannot confirm whether things like #569 have been solved or not.

Issue #290, opened 2018-10-16
Music block ramps up updates/sec over time, causing increasing CPU usage


sebastianst:
I use the most recent git version of i3status-rust. On my 8 threaded desktop, the `i3bar` process shows an ever increasing CPU usage.
```bash
# ps -ewwo %cpu,args --sort=-%cpu | head      
%CPU COMMAND
 8.0 i3bar --bar_id=bar-0 --socket=/run/user/$UID/i3/ipc-socket.14743
 4.6 /usr/lib/firefox/firefox
 3.1 /usr/lib/firefox/firefox ...
 2.5 /usr/lib/Xorg -nolisten tcp :0 ...
 2.4 /usr/share/spotify/spotify
 1.5 /usr/lib/thunderbird/thunderbird
 1.4 i3status-rs /home/$USER/.config/i3/status.toml
```
So since start the average was 8%, but at the time of writing, in `htop` it hovers around 17%.
Restarting i3 in-place, which restarts `i3bar`, brings its CPU usage back down to below 1%.
Everything else works just fine btw.

My config
```toml
theme = "solarized-dark"
[icons]
name = "awesome"
[icons.overrides]
cpu = "⚡"
music = ""

[[block]]
block = "music"
player = "spotify"
max_width = 36
buttons = ["play", "next"]

[[block]]
block = "disk_space"
path = "/"
alias = "/"
info_type = "available"
unit = "GB"
interval = 20

[[block]]
name = "networkmanager"

[[block]]
block = "memory"
format_mem = "{Mug}GB({Mup}%)"
format_swap = "{SUg}GB({SUp}%)"
display_type = "memory"
icons = true
clickable = true
interval = 5
warning_mem = 80
warning_swap = 80
critical_mem = 95
critical_swap = 95

[[block]]
block = "cpu"
info = 15
warning = 50
interval = 1
frequency = true

[[block]]
block = "load"
format = "{1m} {5m}"
interval = 3

[[block]]
block = "weather"
format = "{weather} {temp}° {wind} m/s {direction}"
service = { name = "openweathermap", api_key = "<secret>", city_id = "<my_city_id>", units = "metric" }

[[block]]
block = "time"
format = "%a %d/%m %T"
timezone = "Europe/Berlin"
interval = 60

[[block]]
block = "sound"
step_width = 3
on_click = "pavucontrol"

[[block]]
block = "custom"
command = "/home/$USER/bin/xkb-layout-icon.sh"
interval = 1
[...]

PR #311, opened 2018-12-01
Process requests for update only  once

rtxm:
A proposed fix for #310 and very probably #290.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -224,7 +224,12 @@ fn run(matches: &ArgMatches) -> Result<()> {
             },
             // Receive async update requests
             rx_update_requests.recv() -> res => if let Some(request) = res {
-                    scheduler.schedule(request);
+                // Process immediately and forget
+                block_map
+                    .get_mut(&request.id)
+                    .internal_error("scheduler", "could not get required block")?
+                    .update()?;
+                util::print_blocks(&order, &block_map, &config)?;
             },
             // Receive update timer events
             ttnu.recv() => {
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -57,10 +57,6 @@ impl UpdateScheduler {
         UpdateScheduler { schedule }
     }
 
-    pub fn schedule(&mut self, task: Task) {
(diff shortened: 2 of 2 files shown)


atheriel:
@greshake I'd like you to review this, since you are the authority on the scheduler.

greshake:
Could you please elaborate on how this could be the cause of said issued?

Comments:
1. This commit ignores that asynchronously scheduled tasks can request a delayed update by passing an update time in the future
2. The async updates are only processed once already as far as I can tell; They don't get 'special treatment' so to say, instead they land in the same queue as the rest of the planned updates to allow for 'bundeling' of updates which occur at the same time

greshake:
I agree that this should fix the issues we've had, but it is a 'dirty' solution in a way, for several reasons:

1. It ignores the ability of the update requests to contain a delay (arguably a feature we don't really need and don't use, so that is okay for now)
2. It leads to duplication. We now have two places where block updates are executed

Because I don't see any simpler solution right now, we should accept this. Going forward, these issues will be entirely resolved in the new architecture I'm planning...

@rtxm Thanks for investigating, the issue is now clear to me. Your solution is straight forward and should work :) Just to have it on record, here is my take on the bug:
Affected: Every block which schedules updates via the return parameter of the update function every time and also schedules async updates.
Example: Take a block which has a counter: It ticks every second and increments it by 1. The update function returns an interval of one second. When clicked, it schedules an update via the async pipeline to increment the value by an additional 1. 
When started, the block works as expected, incrementing the value every second. If clicked at, say 1.5s after start, the value is again incremented, but now the updater schedules and additional update for 2.5s (because the update function noted it wants to be called 1s later). The block now increments every half second.

rtxm:
I'm aware the solution is not "clean" as I'd love it to be ;-) ; the intent was to spotlight the problem and start a discussion as didn't want to touch the architecture in anyway. 

The bug is exactly as you describe it, you explain it better than me.

I enjoy very much i3status-rust and I'm willing to help. Do not hesitate to open a discussion about the new architecture.

PR #318, opened 2018-12-30
Update Cargo.lock

Moredread:
With the stable version of cargo I get 

```
error: failed to parse lock file at: /home/user/i3status-rust/Cargo.lock

To learn more, run the command again with --verbose.
```

I updated the Cargo.lock file with `cargo check` and the latest cargo nightly version. With that also Cargo stable works.

I ran into this problem when updating the package on NixOS, as even downloading the dep crates failed. Using rust nightly didn't help there, I'm not sure why atm.


atheriel:
I'd like to figure out why Cargo is indicating a parse failure and how that could possibly be related to bumping the version of `libpulse-binding` before merging this. Can anyone confirm the same issue?

ln-nl:
> I'd like to figure out why Cargo is indicating a parse failure

Probably because there is no `libc 0.2.40` in the Cargo.lock

atheriel:
That seems like a plausible explanation. We could probably make that change explicitly but this PR is a good enough solution as is. 👍

PR #329, opened 2019-02-15
blocks/music.rs: Support on_collapsed_click configuration

### LINKED ISSUE #328: Open music player on click of collapsed music block
### Problem Statement

When i3 starts, I don't have my music player running; I would like to be able to just click on the collapsed music icon in my status bar and have my music player start.

### Proposal

I propose adding an `on_collapsed_click` config option to the music block, which I could then use to configure my desired behaviour.

#### Rejected Alternatives

I considered using the `player` config option as the source for this, but I don't believe there's any guarantee that the value of `player` and the binary used to launch that player will be the same, so this would be brittle. (I would also like to retain the current behaviour of detecting any player, whilst still being able to launch my preferred player if no player is detected.)

OddBloke:
This implementation was what I reached when playing around with the codebase to work out how I could do this; I'm more than happy to throw this away if we reach a different conclusion on what we want the interface to look like.

Fixes #328

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -371,6 +371,14 @@ block = "music"
 buttons = ["play", "next"]
 ```
 
+Start Spotify if the block is clicked whilst it's collapsed:
+
+```toml
+[[block]]
+block = "music"
+on_collapsed_click = "spotify"
+```
+
 
 ### Options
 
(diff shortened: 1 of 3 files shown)


greshake:
Thaks for the PR, I like it. Unfortunately there is some buggy behaviour still when it comes to closing the music player. The bar sometimes doesn't notice and collapse, so you can't restart the player- at least some times. But that is separate from the stuff implemented here.

PR #332, opened 2019-02-18
Use cargo fmt

JP-Ellis:
You have `rustfmt.toml`; however, it seems like it hasn't been run/followed in a long time.

This pull request simply ran `cargo fmt`.

If you want, I can further modify this pull request to add `cargo fmt` as a stage to the continuous integration, thus ensuring that all pull requests are correctly formatted before being merged.

PR #333, opened 2019-02-18
Allow for Missing Network

### LINKED ISSUE #121: Treat missing network device as down
Currently, the network block does not play very nicely with removable network devices.

For example, I have a USB Ethernet adapter and the associated device (in `/sys/class/net`) is only present if the adapter is plugged in.  When the adapter is not there, I get an error instead.

Do you think `hide_inactive` should treat absent devices as being inactive too?  Or perhaps a new `ignore_missing` flag should be added for this?

JP-Ellis:
This pull requests adds a new option for the network block: `hide_missing`.  If set to `true` (`false` by default), a missing network device will not raise an error.

This is useful if the device is not always present (such as a USB ethernet adapter for example).

This pull request closes #121.

This pull request should be merged after my other pull request (#332).

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -1,9 +1,9 @@
+use chan::Sender;
 use config::Config;
 use errors::*;
-use scheduler::Task;
-use chan::Sender;
-use std::time::Duration;
 use input::I3BarEvent;
+use scheduler::Task;
+use std::time::Duration;
 use widget::I3BarWidget;
 
 pub trait Block {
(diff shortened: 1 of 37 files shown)


atheriel:
1. What will the block show if it is missing? and
2. Why not make this the default behaviour? I don't think we really need a configuration option -- this is the right thing to do in almost all cases (as I mentioned in #121).

JP-Ellis:
Closed in favour of #344

Issue #257, opened 2018-09-13
IWD block


cnt0:
Hi. I created a block for [IWD](https://wiki.archlinux.org/index.php/Iwd). There's an [example](https://github.com/cnt0/i3status-rust/blob/master/blocks.md#iwd). It can address issues like #247. Would you like to merge it?

(decided to make an issue instead of PR because readme.md says so)

atheriel:
Ah, so the idea is that this can provide some information for those without `wpa_supplicant` or `iw`?

cnt0:
@atheriel it doesn't require `wpa_supplicant` or `iw`. The only requirement is running `iwd` daemon, which is essentially a replacement of `wpa_supplicant`. Read the archwiki article from the top post to learn more

atheriel:
Hey, thanks for your patience on this. I'd like to have support for IWD users. My main concern would be incorporating this support into the existing network block. Given your work implementing the block, how hard do you think this would be?

ammgws:
Closing due to inactivity. Feel free to re-open if there are any updates.

oati:
I'd really like iwd support. I recently switched to using systemd-networkd + iwd, and as far as I could tell, I don't have any other way to get a network block. What's preventing this from being added?

ammgws:
Someone who will do the work.

PR #335, opened 2019-02-25
blocks/networkmanager: Upgrades

kennylevinsen:
This PR gives the NetworkManager block the full treatment. It looks quite nice now, if I may say so myself.

Instead of simply showing primary device type and availability, we can now see all active connections, the types of their devices (and if applicable, SSID of connected AP), as well as their IPs. Configuration options exist to filter noise out, such as only showing the primary connection, showing SSIDs and showing IPs.

All information is pulled over D-Bus whenever we receive a PropertiesChanged event.

A click handler was also added, allowing one to easily open nmtui in a popup or similar.

The only things that the net block can do that we can't would be bitrate and current speed. The bitrate is accessible, but it might also get a bit noisy within the same block.

### DIFF
--- a/src/blocks/networkmanager.rs
+++ b/src/blocks/networkmanager.rs
@@ -1,18 +1,22 @@
+use std::ffi::OsStr;
 use std::fmt;
-use std::time::{Duration, Instant};
+use std::net::Ipv4Addr;
+use std::process::Command;
 use std::thread;
+use std::time::{Duration, Instant};
 
 use chan::Sender;
 use uuid::Uuid;
 
+use block::{Block, ConfigBlock};
+use blocks::dbus::arg::{Array, Iter, Variant};
+use blocks::dbus::{BusType, Connection, ConnectionItem, Message, MessageItem, Path};
 use config::Config;
 use errors::*;
+use input::{I3BarEvent, MouseButton};
 use scheduler::Task;
-use block::{Block, ConfigBlock};
 use widget::{I3BarWidget, State};
-use widgets::text::TextWidget;
-use blocks::dbus::{BusType, Connection, Message, MessageItem};
-use blocks::dbus::arg::Variant;
+use widgets::button::ButtonWidget;
 
 enum NetworkState {
     Unknown = 0,
@@ -28,7 +32,7 @@ enum NetworkState {
 impl From<u32> for NetworkState {
     fn from(id: u32) -> Self {
         match id {
-            // https://developer.gnome.org/NetworkManager/unstable/nm-dbus-types.html#NMState
+            // https://developer.gnome.org/NetworkManager/stable/nm-dbus-types.html#NMState
             10 => NetworkState::Asleep,
             20 => NetworkState::Disconnected,
             30 => NetworkState::Disconnecting,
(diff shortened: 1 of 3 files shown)


kennylevinsen:
(just fixed an error in the commit message—it bugged me)

kennylevinsen:
The failure is deprecation warnings on unrelated code, probably due to a bump in rustc on travis. I'll rebase it later today, assuming they have been fixed in master.

kennylevinsen:
I have opened #342 to fix the deprecation errors.

src/blocks/networkmanager.rs
```diff
@@ -28,125 +32,371 @@ enum NetworkState {
...
         }
     }
 }
 
-enum ConnectionType {
-    Ethernet,
-    Wireless,
-    Other,
+#[derive(Debug)]
+enum DeviceType {
```
atheriel: So I understand why you've included all possible devices, but many of them will not be encountered by usters. Can you simplify the code by amalgamating the more unusual devices into an `Other` discriminant?

src/blocks/networkmanager.rs
```diff
@@ -28,125 +32,371 @@ enum NetworkState {
...
-            ConnectionType::Ethernet => write!(f, "net_wired"),
-            ConnectionType::Wireless => write!(f, "net_wireless"),
-            ConnectionType::Other => write!(f, "net_other"),
+            DeviceType::Ethernet => Some("net_wired".to_string()),
+            DeviceType::Wifi => Some("net_wireless".to_string()),
+            DeviceType::Modem => Some("net_modem".to_string()),
+            DeviceType::Bridge => Some("net_bridge".to_string()),
+            DeviceType::TUN => Some("net_bridge".to_string()),
+            DeviceType::Wireguard => Some("net_vpn".to_string()),
+            _ => None,
```
atheriel: This list of device types looks more reasonable to me. Most of the others should be removed.

src/blocks/networkmanager.rs
```diff
@@ -28,125 +32,371 @@ enum NetworkState {
...
-            .block_error("networkmanager", "Failed to read property")?;
+        let active_connections: Variant<Array<Path, Iter>> = m.get1().block_error("networkmanager", "Failed to read active connections")?;
 
-        Ok(ConnectionType::from(connection_type.0))
+        Ok(active_connections.0.into_iter().map(|x| NMConnection { path: x }).collect())
+    }
+}
+
+#[derive(Clone)]
+struct NMConnection<'a> {
```
atheriel: I think this should be `NmConnection`, in keeping with Rust's naming convention for types containing acronyms.

src/blocks/networkmanager.rs
```diff
@@ -28,125 +32,371 @@ enum NetworkState {
...
+    fn devices(&self, c: &Connection) -> Result<Vec<NMDevice>> {
+        let m = ConnectionManager::get(c, self.path.clone(), "org.freedesktop.NetworkManager.Connection.Active", "Devices").block_error("networkmanager", "Failed to retrieve connection device")?;
+
+        let devices: Variant<Array<Path, Iter>> = m.get1().block_error("networkmanager", "Failed to read devices")?;
+        Ok(devices.0.into_iter().map(|x| NMDevice { path: x }).collect())
+    }
+}
+
+#[derive(Clone)]
+struct NMDevice<'a> {
```
atheriel: Same naming issue as above.

src/blocks/networkmanager.rs
```diff
@@ -28,125 +32,371 @@ enum NetworkState {
...
 pub struct NetworkManagerConfig {
-    /// Whether to show the connection type or not.
-    #[serde(default = "NetworkManagerConfig::default_show_type")]
-    pub show_type: bool,
+    #[serde(default = "NetworkManagerConfig::default_on_click")]
+    pub on_click: Option<String>,
+
+    /// Whether to only show the primary connection, or all active connections.
+    #[serde(default = "NetworkManagerConfig::default_only_primary_connection")]
+    pub only_primary_connection: bool,
```
atheriel: I would suggest this be simplified to `primary_only`.

src/blocks/networkmanager.rs
```diff
@@ -28,125 +32,371 @@ enum NetworkState {
...
+
+    /// Whether to show an unknown device icon instead of name for unknown devices.
+    #[serde(default = "NetworkManagerConfig::default_unknown_device_icon")]
+    pub unknown_device_icon: bool,
```
atheriel: Is there a compelling reason why we'd imagine a user would configure the icon in this way?

atheriel:
Thanks, this is some nice work. One thing I'm curious about: you've emulated the Net block configuration here by including `ip` and `ssid`, but an alternative is to implement a `format` string. Does that make sense to you?

src/blocks/networkmanager.rs
kennylevinsen: Well, for an unknown device type, you can either get a question mark (uninformative, but pretty), or a device type text (informative, but very ugly and not "in theme").

I felt like people might have differing opinions, looked at how many options things like the "net" block had, and added both. I'm all ears for external ideas here.

src/blocks/networkmanager.rs
kennylevinsen: Noted, I'll update it.

src/blocks/networkmanager.rs
kennylevinsen: Noted. Was stuck in Go mode, where all acryonyms are upper/lower-cased as a unit.

src/blocks/networkmanager.rs
kennylevinsen: Noted as above.

src/blocks/networkmanager.rs
kennylevinsen: Of course. The documentation link should be sufficient to extend as necessary.

I believe I primarily added them all for exploratory reasons.

src/blocks/networkmanager.rs
kennylevinsen: Yup

kennylevinsen:
A format string is not a terrible idea. So, a per-connection formatter, with `{icon}`, `{type}`, `{ssid}`, `{ip}`, `{ips}` (for all)? Maybe the entire icon/no icon for unknown types could also be dealt with here...

kennylevinsen:
I believe all your feedback has been implemented, but give it a look and tell me what you think.

The biggest change is that there are are now two formatters (device and connection), which allows for greater flexibility and simultaneously replaces a few of the previous config options.

The unknown icon option no longer exists, as the list of devices with assigned icons is equal to the list of implemented device types. However, it is now possible to use the connection type name through the device formatter if one prefers to do so.

As for the removed device types, I guess we'll just add them as someone bumps into them on a normal system.

kennylevinsen:
Also, we should definitely make a version of FormatTemplate that can also accept string generating functions instead of plain strings. That way, we won't have to waste time hitting D-Bus with lookups if the user doesn't want to see the SSID or IPs.

atheriel:
Did some local testing on this. Looks great. One final question: can you explain why IPs belong to the connection, not to a device?

kennylevinsen:
Yeah, it's a bit odd. For some reason, NetworkManager exposes IPs per connection.

However, I do realize now that they are *also* exposed per device. I think I'd prefer moving them there, as it makes more sense.

kennylevinsen:
I have moved the IP4Config readout to the devices, which makes more sense. I have also taken the liberty of doing the following:

1. Add connection ID to the connection format string
2. Create a separate AP format string
3. Add AP signal strength and frequency to the AP format string

Very minor and easy to review changes. Tested on my laptop and desktop.

greshake:
This is a great PR. @atheriel you worked on this already; What is the status?

atheriel:
I do actually have a view on outstanding changes that need to be made here, but haven't had the time to detail them. The summary would be: I think the current PR absorbs way too much of the complexity of the D-Bus API for Network Manager instead of using the existing design of the `net` block.

I would very much like to see a world in which we have Network Manager, the existing sysfs/iw/ip, and `iwd` backends for the net block, based on this and the work in #257. I plan to manually pull these changes into a new branch, rebase as necessary, and integrate them to the net block code at some point in the future. Until then, I'm pretty hesitant to merge this and commit to maintaining backwards compatibility to the complex interface presented here. Make sense?

kennylevinsen:
While the idea of a common network frontend seems sensible, it would probably end up complicating matters more than one would gain from it.

Off the top of my head, I see that going either one of two ways:
1. A simple abstraction closer to the existing net block, which handicaps NetworkManager considerably, and requires some broken heuristics to navigate connection objects to simulate raw devices.
2. A NetworkManager-like abstraction, which while functioning might complicate simpler backends considerably, needing to simulate a deeper hierarchy where there is none.

Unless #2 is the way you want to go, then I suspect it would be better to keep high-level connection managers like NetworkManager separate from simple network tools used by the net block.

However, as long as functionality is not noticably reduced, I have no objections to such effort. I'm cozy over on my fork until something ends up making it upstream.

greshake:
@kennylevinsen makes a valid point. Unifying such interfaces (be it network, music or something else) will inevitably lead to high complexity or lost functionality. I would be okay with separating 
a NetworkManager block from simpler interfaces. 

In the long run however, we should have a way of 'grouping' these blocks that makes sense to the user- but it will become ever more fragmented, as the linux ecosystem is as well.

atheriel:
That's fair, it's possible that there's no good unified interface. I haven't given up on trying, though, or perhaps meeting at some reasonable point in the middle. Several other blocks have `driver` options that have worked well enough.

And in case it's not clear from my earlier comment: @kennylevinsen has done a ton of useful work here that I don't intend to give up on.

ammgws:
I didn't realise this block already existed since it's not documented in `blocks.md`

I don't see the harm in merging this since it's an improvement on an existing block

PR #337, opened 2019-02-25
blocks/net: Add on_click support

kennylevinsen:
Add on_click to net block. Convenient for opening network configuration panels.

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -3,16 +3,19 @@ use std::io::prelude::*;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::time::{Duration, Instant};
+use std::ffi::OsStr;
 use chan::Sender;
 
 use block::{Block, ConfigBlock};
 use config::Config;
 use de::deserialize_duration;
 use errors::*;
-use widgets::text::TextWidget;
+use widgets::button::ButtonWidget;
 use widgets::graph::GraphWidget;
 use widget::I3BarWidget;
 use scheduler::Task;
+use input::{I3BarEvent, MouseButton};
+
 
 use uuid::Uuid;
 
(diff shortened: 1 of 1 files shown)


HD1ex:
It would also be nice to have the on_click option on every block.
Maybe this would generalize this PR a bit more, giving the user to run custom (shell) commands on click.
For example on the pacman block I'd like to run the following command:
`on_click = "notify-send "$(printf "The following updates are available:\n%s\n" "$(fakeroot pacman -Qu --dbpath /tmp/checkup-db-$USER/)")""`

But maybe such functionality should be implemented in Rust, if I understand this project correctly :)

atheriel:
The `on_click` behaviour of blocks has been discussed several times in the past, and I'm not sure that there is strong enough support for me to merge this PR. The problem with implementing a generic "run this shell command on click" is that it prevents us from introducing other on-click behaviours for blocks -- of which there are several already.

@HD1ex Notification support is probably beyond the scope of the Pacman block. But for your example I'd actually suggest using `cron` instead of relying on your status line to check for updates.

kennylevinsen:
I understand this approach. However, at the same time, given no other functionality, it is awfully convenient to use to open network configuration.

Maybe there should just be a way to assign on_click optionally to any block, so that a block can implement on_click itself *or* a user could override any click functionality? I do not have that strong opinions about this, but I do see the convenience of being able to open configuration panels where no other functionality is available or sensible

kennylevinsen:
I'm entirely OK with just having this closed and replaced by a general discussion with regards to global on_click.

greshake:
The block clicking system is already pretty convoluted and intransparent, I hesitate to add complexity. To achieve your goal, you could also just add a custom block right next to it which serves as your shortcut. We'll keep an eye out for this feature when (or if) we rewrite most of the project eventually with futures. The event system should also be overhauled then, probably becoming more composable.

Thanks for the PR!

greshake:
#335 also adds this feature to the NetworkManager block, so I revisited my opinion.. Since it is such a requested feature, I'll let individual handlers go through but I'm not going to make this a general feature. I hope that is alright for now.

PR #339, opened 2019-02-25
blocks/net: Command cleanup

kennylevinsen:
Removes unnecessary shell instantiations, as well as silly "xyz | grep | awk" constructs.


atheriel:
LGTM. :+1:

PR #342, opened 2019-03-06
Replace trim_left/right with trim_start/end

kennylevinsen:
This should fix the deprecated trim_left/trim_right warnings.


atheriel:
Thanks!

PR #344, opened 2019-03-10
Add `hide_missing` option for Network devices

### LINKED ISSUE #121: Treat missing network device as down

### LINKED ISSUE #131: On_click option should be available for all blocks

JP-Ellis:
If `hide_missing` is `false` (the default), the network symbol is shown with an
'×'.

If `hide_missing` is `true`, the network block is completely omitted.

Closes #121

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -25,26 +25,22 @@ pub struct NetworkDevice {
 impl NetworkDevice {
     /// Use the network device `device`. Raises an error if a directory for that
     /// device is not found.
-    pub fn from_device(device: String) -> Result<Self> {
+    pub fn from_device(device: String) -> Self {
         let device_path = Path::new("/sys/class/net").join(device.clone());
-        if !device_path.exists() {
-            return Err(BlockError(
-                "net".to_string(),
-                format!(
-                    "Network device '{}' does not exist",
-                    device_path.to_string_lossy()
-                ),
-            ));
-        }
 
         // I don't believe that this should ever change, so set it now:
         let wireless = device_path.join("wireless").exists();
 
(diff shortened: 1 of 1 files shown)


atheriel:
The Travis build appears to be failing for a legitimate reason. Are you able to replace your usage of the deprecated interface?

JP-Ellis:
@atheriel This has been fixed in #342 (which has been merged after I created this pull request).  I can rebase if you wish, or leave it as is.

atheriel:
Yes, my mistake. I edited my comment to address the error almost immediately, but not fast enough I see.

JP-Ellis:
I've rebased the commit on top of the `master` so that the tests can pass.

Is there anything more you would like from this pull request before it is merged?

atheriel:
Nope, just did some local testing. Looks good. Thanks for following through with this. :+1:

GladOSkar:
@atheriel @JP-Ellis One thing i noticed: This messes up alternating color schemes because the block is just invisible, not removed.

image

JP-Ellis:
I don't use an alternating color scheme, so didn't notice that.  @GladOSkar, perhaps you should open a new issue so this can be tracked?

GladOSkar:
@JP-Ellis Done #349

PR #346, opened 2019-03-19
Allow custom block to use pango formatting

ammgws:
This enables pango markup for the button widget so that users can add colour to their `custom block` output using their shell command output (example below).

20190319_22h57m01s

More examples (background colour, different font all in the same output):
20190319_23h23m36s

### DIFF
--- a/src/widgets/button.rs
+++ b/src/widgets/button.rs
@@ -26,7 +26,8 @@ impl ButtonWidget {
                 "separator": false,
                 "separator_block_width": 0,
                 "background": "#000000",
-                "color": "#000000"
+                "color": "#000000",
+                "markup": "pango"
             }),
             config,
             cached_output: None,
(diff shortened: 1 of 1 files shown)


src/widgets/button.rs
```diff
@@ -26,7 +26,8 @@ impl ButtonWidget {
                 "separator": false,
                 "separator_block_width": 0,
                 "background": "#000000",
-                "color": "#000000"
+                "color": "#000000",
+                "markup":" pango"
```
atheriel: Seems to be a `"` in the wrong place, still.

ammgws:
Sorry about the noise, was trying to squash the commits and failed.
Anyway, it should be good to go now.

atheriel:
Squashed it via the merge. Thanks!

PR #350, opened 2019-03-25
blocks/net: Support tun device

d3m3vilurr:
- Initial support tunnel device
- Bind specific [icon](https://fontawesome.com/icons/lock?style=solid) for VPN network device.


atheriel:
Can you explain what's going on here?

d3m3vilurr:
sorry i forgot adding description. i'll also update commit message

if we use tun(tunnel) device via `tun` kernel module, `/sys/class/net/{device}/operstate` is always return `unknown`.
this problem occurs `is_up` method return the `false`.
so we should use another information to detect `tap` device or not, and I used the file `/sys/class/net/{device}/tun_flags` for this like detecting `wireless` device.

atheriel:
Ah, that makes sense. However, if `tun` devices will always report as up, then what happens when they are actually down (e.g. if the vpnc script removes the interface after you log of of the VPN)?

d3m3vilurr:
Hm... I didn't think about that before you asked me.
because most vpn services make the device when connect vpn and remove after disconnect.
but yes we can call `down` via `ip` command.
I also tested this, that device's `operstate` move to `down`

IMO, we have multiple options;
1. `Ok(operstate == "up")` change to `Ok(operstate != "down")`; but some device driver would get another state, compare with `up` is best option in this case
2. or just split cases like 
```
} else {
  let operstate = read_file(&operstate_file)?;
  if self.tun {
    Ok(operstate != "down")
  } else {
    Ok(operstate != "up")
  }
}
```

what do you prefer? I'll change it after receiving your opinion.

atheriel:
Hey, sorry for the late reply here. I did some local testing, and two things came to mind:

(1) I definitely led you astray with regards to the disconnection issue. The original behaviour -- all tun networks are "up" -- is probably the best one.
(2) You only change the icon to the VPN one when the block is created. However, if the device is missing when that happens (e.g. the VPN is not yet set up), it will show as an Ethernet device instead, and never switch to the VPN icon. This is probably relevant, as many users will start the bar on login, when the tun device has not yet been set up.

Any thoughts?

d3m3vilurr:
1. you mean, do I revert e9e97ce?
2. it just my opinions
    - I thought `have tun device` means `connected/connecting vpn`.
      but technically, vpn icon didn't mean `can access vpn network`.
      some vpn service like `tinc` doesn't remove the device whenever connect or not until stop the service,
      but some others (like openvpn, but I didn't test it yet) would remove the device when vpn disconnected.
    - personally I set with `hide_inactive = true` and `hide_missing = true` for this device.
      so when I boot up, bar didn't see the this device, then show icon after vpn network was connected.
    - this block is `net` device and already we used two icons for `wifi` and `ethernet`.
      still can't add the icon?
    - and anyway I have no idea until today. can we use tun device for non-vpn work? just internal network?
      well.. if that maybe just [network](https://fontawesome.com/icons/network-wired?style=solid) icon is better.

I'll follow your opinion :)

atheriel:
Yes, I did mean to revert e9e97ce. If you're not comfortable doing that then I can. Otherwise I think that (2) is just a corner case that should not prevent this from being merged. We can fix it later.

Thanks for your patience on this.

d3m3vilurr:
Reverted that ;) thank you for hard review

PR #356, opened 2019-04-02
Add support for UPower's DisplayDevice.

frececroka:
I have a laptop with two batteries. I am not interested in the charge and time remaining of the individual batteries, but would rather see aggregate values. In particular, how long do I have until both are empty?

UPower supports this out of the box, via the concept of a DisplayDevice. From what I understand, this is basically a logical power supply representing all currently relevant power supplies. From the UPower documentation:

> [...] The "display device" [is] a composite device that represents the status icon to show in desktop environments.

For more information see https://upower.freedesktop.org/docs/UPower.html#UPower.GetDisplayDevice.

The current battery block does not allow using the DisplayDevice, because of how the D-Bus path for the battery is constructed. For the device `BAT0`, the D-Bus path `/org/freedesktop/UPower/devices/battery_BAT0` is used, but the path for the DisplayDevice is `/org/freedesktop/UPower/devices/DisplayDevice` (and not `/org/freedesktop/UPower/devices/battery_DisplayDevice`).

This PR changes how the configuration option `device` of the battery block is handled. All values of `device` that are not `"DisplayDevice"` are handled exactly as before. But if `device` has the value `"DisplayDevice"` and UPower is enabled (i.e., `upower = true` in the block configuration), the `UpowerDevice` is instantiated with the correct path to the DisplayDevice, i.e. with `/org/freedesktop/UPower/devices/DisplayDevice` instead of `/org/freedesktop/UPower/devices/battery_DisplayDevice`.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -72,7 +72,7 @@ Creates a block which displays the current battery state (Full, Charging or Disc
 
 The battery block collapses when the battery is fully charged -- or, in the case of some Thinkpad batteries, when it reports "Not charging".
 
-The battery block supports reading charging and status information from `sysfs`, or optionally through the [Upower](https://upower.freedesktop.org/) D-Bus interface on systems where that is available.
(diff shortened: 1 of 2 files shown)


atheriel:
Thanks, this is a nice addition.

PR #357, opened 2019-04-04
Add a block for Bluetooth devices

### LINKED ISSUE #178: Bluetooth device status
Would be nice to glance at

- connected bluetooth devices
- their battery level (if available)

Or is it outside the scope of this project?

atheriel:
This PR adds a new block and the relevant documentation. I'm using the D-Bus API, so it should be extremely light on resources. No new dependencies, either.

(I am indebted to [this blog post](http://smartspacestuff.blogspot.com/2016/02/i-got-figurin-out-dbus-bluez.html) for figuring out some aspects of the D-Bus API.)

At the moment, the block recognizes headphones, keyboards, mice, and joysticks and chooses an appropriate icon. If the device supports the `org.bluez.Battery1` interface, the block will also report the battery level.

Note that since I don't actually own a device that supports the battery interface, I'm not actually sure that it works -- any testers would be much appreciated.

Otherwise, this block doesn't do much at the moment. It is possible that we could add e.g. (dis)connect on right click, for example, or auto hide/show behaviour to match some other blocks.

Closes #178. cc @srid

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -2,6 +2,7 @@
 
 - [Backlight](#backlight)
 - [Battery](#battery)
+- [Bluetooth](#bluetooth)
 - [CPU Utilization](#cpu-utilization)
 - [Custom](#custom)
 - [Disk Space](#disk-space)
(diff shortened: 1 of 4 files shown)


frececroka:
Thanks, this is useful! I have a Bluetooth headset, but while Android is sporadically able to read the battery level, Bluez can't pick up on it. This seems to be an issue with Bluez (see [here](https://stackoverflow.com/questions/49078659/check-battery-level-of-connected-bluetooth-device-on-linux/)).

I like right-click to connect/disconnect. Maybe opening Bluetooth Manager on left click could also be useful?

atheriel:
Yes, I've found the battery support in Bluez very unreliable. Oh well, maybe after the next kernel update...

I implemented the right-click functionality. I'll punt on the left-click open-an-application feature for now, that's a matter of contention in a few other PRs/issues.

GladOSkar:
@atheriel Can confirm battery status works for my MX Master 2s:

image

marzzzello:
Thanks for this. I just have the problem that when the bluetooth is off that I get `Error in block 'bluetooth': Failed find a device with matching MAC address.` It would be good to ignore that case and not to let i3status-rust fail

atheriel:
@marzzzello Why would you have this block in your bar if your bluetooth is off?

marzzzello:
@atheriel because I normally have bluetooth always off. I just turn it on when I use it.

marzzzello:
The best would be to let the icon disappear when bluetooth is off

atheriel:
@marzzzello Discuss in #374.

PR #358, opened 2019-04-04
blocks/battery: When charging, show remaining time with UPower.

frececroka:
This is another small enhancement for the battery block with `upower = true`.

For a `PowerSupplyDevice`, the method `time_remaining` returns the time until the battery is empty if it is currently being drained, and it returns the time until the battery is full if it is currently being charged.

For a `UpowerDevice`, the method `time_remaining` always returns the time until the battery is empty, even when it is charging.

This PR changes the behavior of the `UpowerDevice` to match the `PowerSupplyDevice` w.r.t. the `time_remaining` method.

On a related note, how about changing the method `fn BatteryDevice::status(&self) -> Result<String>;` to return an enum `BatteryStatus` or so?

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -332,10 +332,11 @@ impl BatteryDevice for UpowerDevice {
     }
 
     fn time_remaining(&self) -> Result<u64> {
+        let property = if self.status()? == "Charging" { "TimeToFull" } else { "TimeToEmpty" };
         let time_to_empty: dbus::arg::Variant<i64> =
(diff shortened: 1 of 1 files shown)


atheriel:
LGTM. :+1: I'm not sure why I never filed this as a bug; it certainly was one.

As for why we don't have an enum: I think it's mostly because the UPower implementation came later, so we were basically just following what could be read directly from the file under `sysfs`. In this case, I don't think moving to an enum would simplify the code very much; in fact, you'd have to do more conversions from the stringy representation. It might help clarify the API, but then, we don't really have anything consuming our API to begin with.

PR #359, opened 2019-04-08
Migrate code to edition 2018

rumpelsepp:
This enables to remove now useless `extern crate` statements.
I ran cargo fix --edition.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ name = "i3status-rs"
 version = "0.9.0"
 authors = ["Kai Greshake <development@kai-greshake.de>",
            "Contributors on GitHub (https://github.com/greshake/i3status-rust/graphs/contributors)"]
+edition = "2018"
 
 [features]
 default = ["pulseaudio"]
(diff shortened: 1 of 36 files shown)


atheriel:
Two thoughts here:

1. In your view, is there any other advantage to using the 2018 edition right now? Or just some simpler syntax here?
2. What compiler version does this limit us to? I can't seem to find it [in the 2018 edition docs](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html). If we already require that version, then this is a much easier change to make.

rumpelsepp:
It is 1.31: https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html

atheriel:
I think we currently compile on 1.30 (October), so 1.31 (December) isn't too big a jump. I don't think we currently have any NLL-related hacks, but it is appealing to have them nonetheless.

Are you able to rebase this and fix the conflicts?

rumpelsepp:
> Are you able to rebase this and fix the conflicts?

done

Issue #272, opened 2018-10-02
Establish a dependency inclusion/update policy


atheriel:
As @svmnotn casually mentions in #196, we don't really have a policy for when we should update dependencies, and as I've said to some recent contributors, we don't really have criteria for inclusion of new crate dependencies, either. I think this has become increasingly important. On the current master branch (11c2a216) we have a total of 91 crate dependencies.

This is what the dependency graph looks like at the moment, using the excellent `cargo graph` visualization tool:

![Dependency graph as of 2018-10-02](https://user-images.githubusercontent.com/1448326/46376110-2d5e0b00-c663-11e8-8105-d040bb496f16.png)

There is clearly some bloat that can be removed here, but less than you might think. We can also see the consequence of a recent inclusion: the `maildir` crate added in #210 pulls in 10 new dependencies.

I think we need to establish a policy for when crates should be updated (in the `.lock` file or in the `.toml` file), and at least some guidelines for including new crates. These are my suggestions.

Please feel free to comment or provide additional rationale. It's my intention to eventually make something like this part of a CONTRIBUTING.md document.

## Updating Dependencies

Updating a dependency can be justified for the following reasons, both of which should be obvious:

1. We need/want functionality introduced in a newer version. For example, bumping `i3ipc` to 0.8.4 might make our focused window block work with `sway`; or
2. A newer version fixes a bug we have encountered. E.g. #269.

We may also opt to update a dependency for these additional reasons, although they are less compelling:

3. A newer version has substantial performance improvements relevant to this project.
4. A newer version removes a transient dependency. For example, updating `thread_local` to 3.6 would remove the `unreachable` crate.

## Adding New Dependencies

As new blocks are introduced, they may require new dependencies. Two things I'd like to see in this case:

1. New dependencies are relatively "proven" crates; and 
2. Heavier dependencies are introduced to provide *core functionality* that will benefit a large number of users.

For cases where this is not true, I think we should make more use of Cargo's feature flags. For instance, we might want to move the `maildir` dependency behind such a feature flag, and I have also suggested that the work-in-progress PulseAudio (#245) support and Notmuch block (#215) take this approach.

atheriel:
Also relevant here is Cargo's recent move to supporting building crates with a `--minimum-versions` flag, which has discovered all kinds of problems with `.toml` files in the Rust community.

greshake:
[...]

PR #360, opened 2019-04-08
Update dependencies

rumpelsepp:
Let's follow the upstream versions a bit more. At the next step I'm looking at replacing `chan` with `chrossbeam` since, `chan` is very deprecated...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,9 +1,9 @@
 [[package]]
 name = "aho-corasick"
-version = "0.6.4"
+version = "0.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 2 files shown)


atheriel:
Hey, did you notice any bug fixes or features we actually need these updates for? We've tried to move away from arbitrarily updating the dependencies. (See #272.)

rumpelsepp:
Ah. I can understand this. My intention was to create a baseline for #361 and IMO tracking upstream more closely safes work for porting later. Maybe some tests can be added to the repo which can detect breakage in case of dependency updates.

greshake:
Let's come back to this when we have the policy in #272 figured out. By that time this PR will be outdated too

Issue #264, opened 2018-09-25
"chan" crate is deprecated


ammgws:
Just saw that the [chan crate](https://crates.io/crates/chan) has been deprecated, and is now superseded by [crossbeam-channel](https://github.com/crossbeam-rs/crossbeam-channel).

On that note, what is the reason for using `chan` instead of [std::sync::mpsc](https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html)?

If there is no real reason not to use the standard library, we could port the code to use that and remove the dependency on `chan`. Otherwise, I guess we should move over to the new crate at some point?

atheriel:
I've been curious about the use of `chan` for a while, but hadn't noticed it was deprecated. I believe only @greshake knows the reasoning behind the choice.

ammgws:
I see everything has been updated to crossbeam so can close this now.

(Though the question still remains whether we can remove the dependency and use the std lib)

PR #361, opened 2019-04-08
Remove chan and add crossbeam_channel

rumpelsepp:
As promised in #360, here is the port to crossbeam_channel. The pull request needs a bit more cleanup.


atheriel:
Awesome! Thanks for taking this on!

cc @greshake: the concurrency code is mostly your domain, in case you have any comments.

Closes #264.

rumpelsepp:
IMO the code needs a bit more polish. Since I haven't been fully soaked into the codebase I can't assess if it's ready yet. But it builds and seems to work. :)

ammgws:
> This library is an alternative to std::sync::mpsc with more features and better performance.

Perhaps could revisit whether we can use the standard library instead?

rumpelsepp:
> Perhaps could revisit whether we can use the standard library instead?

Valid point. I didn't investigate if the multiple producer, multiple consumer pattern is absolutely required. If not, I might just update to `std` and multiple producer, single consumer instead.

atheriel:
I believe stabilization of `std::Future` is on the horizon, too, so we could possibly adopt that instead.

rumpelsepp:
I tried the `Futures` stuff a year ago and it was so confusing... Maybe once `async`/`await` has settled it might be worth porting. For now I suggest sticking to the synchronous code.

greshake:
Thanks for the PR! 
I want to redo the entire system with async await and/or futures at some point, but it has proven to be a really frustrating experience. My goal is to improve the ergonomics of writing async blocks, but it seems to just grow in complexity. While Futures were stabilized in 1.36 we are still a long way from actually meeting the ergonomic requirements and so we should certainly switch to crossbeam in the meantime. I'll check this out and merge if it works, it should be a seamless transition.

greshake:
I merged the PR manually to add soem corrections for the current version. Most of the crossbeam dependency could also be omitted since we only need crossbeam_channel. 
Cheers :)

PR #362, opened 2019-04-09
block/time: Clean up zombie processes.

### LINKED ISSUE #355: Clicking time block makes zombie processes with `on_click` option
### Configuration:
```
[[block]]
block = "time"
interval = 5
format = "%Y %-jth, %a, %-m/%d %R"
on_click = "gsimplecal"
```
[`gsimplecal`](https://github.com/dmedvinsky/gsimplecal) is a pop-up calendar program. When I click time block in `i3status-rs` then `gsimplecal` shows well, but after closing it (by Ctrl-w or Ctrl-q or whatever) it seems to make zombie processes, one zombie per each try. I don't think it's `gsimplecal`'s issue; manually running it in command line doesn't creates a zombie process.

### Maybe related code:
https://github.com/greshake/i3status-rust/blob/b3146bcbdbcca79fd1daba675205324284f90cbd/src/blocks/time.rs#L100-L108

frececroka:
Spawning a command without waiting for it leaves behind a zombie process. This pull request waits for the spawned commands.

The time and sound blocks use the same code to spawn a child process, so I moved this code to a new module `subprocess.rs`.

If spawning the child failed, an error is returned.

Fixes #355.

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -5,7 +5,6 @@ use std::io::Read;
 use std::process::{Command, Stdio};
 use std::thread;
 use std::time::{Duration, Instant};
-use std::ffi::OsStr;
 #[cfg(feature = "pulseaudio")]
 use std::rc::Rc;
 #[cfg(feature = "pulseaudio")]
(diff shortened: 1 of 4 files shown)


atheriel:
Looks good to me. :+1:

The new module might come in handy if we ever figure out whether to implement `on_click` for all blocks, too.

Issue #3, opened 2017-05-19
[Low Prio] New Themes


greshake:
We should have more themes than solarized-dark, especially a simple text/i3status style one. To add them, just edit the src/themes.rs file. It should be obvious how to add a new theme. Every theme is Json, containing foreground and background colors for each block state (Warning, Info, Critical etc). Also here you can define the separator symbol.

hansondr:
Is there any plan to support custom themes directly in the TOML without completely overriding an internal theme?

currently:
```toml
[theme]
name = "solarized-dark"
[theme.overrides]
idle_bg = "#123456"
idle_fg = "#abcdef"
# ...
```

perhaps something like the following:
```toml
[[theme]]
idle_bg = "#123456"
idle_fg = "#afcdef"
# ...
```

greshake:
Not really. You can select 'plain' as your base theme, but I mean it doesn't matter if you specify/override all the attributes, you got yourself a new theme. I want to encourage people to publish their themes in a PR, not sure if this does, but at worst it's a semantic oddity, without functional impairment.

MarcelPa:
Just to pitch an idea: the terminal colors from Xresources could be used as a theme. This would enable selecting themes from sites like xcolors.net or similar.
I would be willing to implement this, but I am still in a process of learning Rust. Basically, this should "only" involve parsing the Xresources file and getting all RGBA values needed; not the hardest problem I guess.
Any thoughts on this?

greshake:
Parsing the Xresources won't be a problem... Converting it into a i3status-rust theme won't be either. Ideally, we want to allow for specifiying a path (to an Xresources file) as the theme parameter. Easiest way to imlement is in the theme.rs file, there is a function mapping a theme name to a theme instance. We simply need to distinguish paths and known theme names and try to generate a theme instance for paths on the fly. Should be quite easy. Just see if there is a theme with the string as name available and, if not, try to treat it as an Xresources file.

lausek:
I really dislike the idea of having another configuration file just for colors. @hansondr idea sounds great for me. We could utilize the `[theme]` section inside the status config for setting colors and everything else.

So, I think we should remove all hardcoded themes except `PLAIN`. In my opinion, the following would be a good order for applying styles:

1. Read `Xresources` of user - typically at `~/.Xresources`
2. Read the attributes specified under `[theme]` inside the configuration file and override attributes placed by the previous step
3. Everything else should be taken from the `PLAIN` template

[...]

PR #366, opened 2019-04-16
Allow for reading theme information from standalone files

atheriel:
This PR introduces a new `theme.file` key to the TOML configuration, which makes it possible to use complete themes not provided by `i3status-rs` itself. We look in the following locations for this file:

- The actual file location, if it is a real path.
- `$XDG_CONFIG_HOME/i3status-rs/themes`; and
- `/usr/share/i3status-rs/themes`

The theme file must be "complete", e.g. contain all of the keys, for it to be serialized correctly. 

In the process, I did a lot of cleanup for deserialization of the `[theme]` section itself to make it clearer and more transparent; it should be much easier to add functionality there in the future.

As an aside: this could make it possible for us to extract the built-in themes and install them in, e.g. `/usr/share/i3status-rs/themes`.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,18 +1,18 @@
 use crate::de::*;
 use crate::icons;
-use serde::de::{self, Deserialize, Deserializer};
 use toml::value;
+use serde::de::{Deserialize, Deserializer, Error};
 use std::collections::HashMap as Map;
 use std::marker::PhantomData;
 use std::ops::Deref;
 use std::str::FromStr;
-use crate::themes::{self, Theme};
+use crate::themes::{Theme, ThemeConfig};
 
 #[derive(Deserialize, Debug, Clone)]
 pub struct Config {
     #[serde(default = "icons::default", deserialize_with = "deserialize_icons")]
     pub icons: Map<String, String>,
-    #[serde(default = "themes::default", deserialize_with = "deserialize_themes")]
+    #[serde(deserialize_with = "deserialize_themes")]
     pub theme: Theme,
     #[serde(rename = "block", deserialize_with = "deserialize_blocks")]
     pub blocks: Vec<(String, value::Value)>,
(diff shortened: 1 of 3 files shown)


atheriel:
[Related comment](https://github.com/greshake/i3status-rust/issues/3#issuecomment-377714954) in #3.

atheriel:
I've just realized that this breaks `theme = name`-style configurations, and can't be merged as-is.

greshake:
What is your status on this?

atheriel:
Still like the idea, have to fix the broken config issue. The custom deserialization code here is pretty hairy and  I wasn't able to find a solution at the time. I'll get back to it.

PicoJr:
> I've just realized that this breaks `theme = name`-style configurations, and can't be merged as-is.

Hello, after cherry-picking this PR, I tried to fix the issue above.

I have duplicated `Config` struct into `LegacyConfig`.

```toml
# LegacyConfig
theme = "solarized-dark"
```

vs

```toml
# Config
[theme]
name = "solarized-dark"
```

can be deserialized into `LegacyConfig` struct using serde (as before).

I have implemented `From<LegacyConfig> for Config` that way `LegacyConfig` can be transformed into `Config` right away.

When loading config, we try deserialize it to `Config` first and if it fails we try to deserialize it to `LegacyConfig`:

```rust
// main.rs
    let config: Result<Config> = deserialize_file(config_path.to_str().unwrap());
    let config = config.or_else(|_| {
        let legacy_config: Result<LegacyConfig> = deserialize_file(config_path.to_str().unwrap());
        legacy_config.map(|legacy| legacy.into())
    })?;
```

see https://github.com/PicoJr/i3status-rust/commit/40fe25febaa192fb23748d8210b7726e706b05b3

What do you think?

atheriel:
Oh, that's a clever solution. Yes, I think we could move forward with this, then.

ammgws:
Closing in favour of #611

PR #368, opened 2019-04-16
Adds a new block for displaying the keyboard layout

### LINKED ISSUE #319: Keyboard layout indicator
Hello. I want show in status bar current keyboard layout.
How i can do this?
I try this lib [https://github.com/nonpop/xkblayout-state](https://github.com/nonpop/xkblayout-state) and custom-block, but i can't do it.
Thx

atheriel:
On distributions with systemd, the `org.freedesktop.locale1` bus exposes a property for reading the current keyboard layout (in the X.org sense). This block either reads that property, or queries the layout using the venerable `setxkbmap`.

Outstanding issues:

- ~~I've named the block "Locale" after the bus and the associated `localectl` program. The only other property we could be interested in is the locale itself; but I'm not sure that anyone really wants that displayed in their bar. So maybe we should just narrow this to being a "keyboard" block instead.~~

- Updates do not seem to work as expected under Sway. Using `localectl` to change the layout emits the correct signal for the bus but does not seem to affect Sway, while using `setxkbmap -layout` *does* seem to affect Sway, but does not emit the correct signal via D-Bus. I'd like to get to the bottom of this, and if anyone is willing to test it let me know.

~~I'll write some docs when these issues are resolved.~~

Edit: I've written a `setxkbmap` driver for the block as well, so we should be covered.

Closes #319.


JohnAZoidberg:
It seems like in Wayland, because it doesn't have a server, every window manager needs to implement their own keyboard layout setting. In Sway until recently it was only possible to change it by setting environment variables before starting it. Now it's possible using `setxkbmap`.

I think `setxkbmap` is probably the most compatible command because it works on X and on Wayland compositors that support this (probably also if they're using xwayland, not sure).

It is really cool to use dbus but if it doesn't work properly, that sadly doesn't help. I'd suggest querying `setxkbmap -print`. It's not the prettiest but likely to work in most cases.

I encourage to look further into the possible methods of achieving this. I just wanted to suggest my opinion for something that works.

atheriel:
@JohnAZoidberg Thank you for the suggestions. I've now added a `setxkbmap` driver and set it as the default. Perhaps the D-Bus approach will work correctly in the future.

Issue #365, opened 2019-04-15
net block stays unresponsive if device path does not exist at startup


p-hamann:
if the tun0 interface is missing before starting i3status-rust, then the net block for tun0 device will not respond after enabling tun0 interface. I can only fix this by restarting i3status-rust after enabling the tun0 interface.

atheriel:
I thought something like this might crop up, although I'm not sure how to fix it. We avoid crashing on missing interfaces at startup to avoid annoying users.

p-hamann:
It is noticeable that if exists(&self) returns true on startup, the directory (/sys/class/tun0) can be removed again by stopping the VPN connection, but the block still remains responsive. Only if the directory does not exist when starting i3status-rust, the block remains unrespronsive.
Otherwise, do you know a workaround, so that the tun0 directory exists, without VPN gets started before?

PR #369, opened 2019-04-17
blocks/net: Prevent crash of tun block

p-hamann:
fixes issue #365


atheriel:
Not about this solution. It seems to be possible for a device to have `tun_flags` but not be named `tun*` (e.g. TAP devices).

p-hamann:
Oh, indeed. Fixed it now.

atheriel:
Seems like a reasonable solution. Thanks!

Issue #373, opened 2019-04-26
The SSID in `net` block is not loaded during initialization and "break" displaying


Delgan:
Hi. :)

While configuring your awesome `i3status-rust` tool, I noticed that the `SSID` of the `net` block is not displayed until the first `interval` delay is elapsed. As a result, the block is not properly displayed (see the discontinuity after the `net` icon).

![56815326-56cdc680-6841-11e9-98ea-babd6468706b](https://user-images.githubusercontent.com/4193924/56816588-191e6d00-6844-11e9-817e-c797eafe395a.png)

This can be triggered by reloading i3 with `$mod+r`, and this is particularly annoying while starting the computer, as the `net` doesn't need to be often updated, it may remain several seconds in this "broken" state.

I'm using `i3status-rs` v0.9, and my config looks like this:

```
[[block]]
block = "net"
interval = 30
device = "wlp0s12f0"
ssid = true
ip = false
speed_up = false
speed_down = false
bitrate = false
```

Do you think this would be possible to fetch the SSID during the initialization? Or delaying it until 1 second for example? Displaying an space `" "` meanwhile would probably also avoid breaking the block formatting.

atheriel:
Yep, we should provide a fallback to avoid this issue.

atheriel:
Fixed by #376.

Delgan:
Great! Thanks @crablab. :wink:

PR #376, opened 2019-05-07
Fix issue with SSID default causing rendering issues

crablab:
Fix issue reported in #373 where SSID is not initialised to any default value causing rendering issue. 

This inits it to a single space, solving the rendering issue. You could load the current SSID by default there but that would require some structural changes and I'm not sure if that is the desired outcome.


atheriel:
Thanks!

PR #382, opened 2019-05-14
Maildir enhancements

mwillsey:
This PR adds two enhancement to the maildir block: 

1. It lets you choose whether to count the `new`, `cur`, or both folders of the maildir(s)
2. It lets you opt out of the icon.

I'm shooting for a pair of blocks that look like this: " :email: 1 | 10 ", where the first number is the unread mails (using the `new` option), and the other is all the mails in that folder (using the `all` option). Helps me go for inbox zero. Note how there's no mail icon on the second block.

I chose defaults that preserve the existing behavior of the maildir block, so these features are totally opt-in.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -289,6 +289,7 @@ interval = 60
 inboxes = ["/home/user/mail/local", "/home/user/mail/gmail/Inbox"]
 threshold_warning = 1
 threshold_critical = 10
+display_type = "new"
 ```
 
 ### Options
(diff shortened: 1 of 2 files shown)


atheriel:
This is a nice feature to have, thanks. I do have a suggestion that would hit closer to your desired outcome: why not introduce a `format` field to the configuration? This approach is used by several other blocks that can display a variety of metrics. You could then set `format = "{new} | {all}"`.

mwillsey:
@atheriel I had that thought, but it makes thresholding more complicated. If my format is "{new} - {cur} - {all}", when do I highlight it is warning or critical? You'd almost need a warning/critical threshold for each number (so 6 thresholds). And what if you're not displaying all the information? Then I guess the user would have to set all the thresholds to unlimited.

greshake:
I'm okay with this in its current state. As @mwillsey mentioned this allows us to have much more information available with individual thresholds.

I like the approach in general to have multiple instances of a block to display different things. While slightly inconsistent between blocks, we use this approach a lot already. For example the storage block, we add one block per device.

Issue #379, opened 2019-05-10
CPU usage bars


fbreitwieser:
Thanks a lot for i3status-rust, really like it!

In my previous bar configuration with another tool, I configured to have a small barplot for CPU utilization with a separate UTF block character for each core's CPU utlization.

E.g., on my 4-core laptop, it would show something like '▇▅▁▁' if the first two cores are in use and the last two cores aren't. [Additionally, it would color the blocks individually; e.g. the first block would be red in this example and the second one orange.]

i3status-rust uses UTF ramp characters so far for different levels of the blacklight and volume. What do you think about adding that functionality to the cpu module, especially with separate characters for each core? I could make a pull request. What do you think would be the best way to support it? E.g. with a format key?

fbreitwieser:
I made an implementation that adds this kind of barchart. It adds a format option, and with `{barchart} {utilization}% {frequency}GHz` it shows the following:

![image](https://user-images.githubusercontent.com/516060/57661096-74c45680-75b7-11e9-8f34-d6a104193fa1.png)
![image](https://user-images.githubusercontent.com/516060/57661196-c53bb400-75b7-11e9-9aeb-ecbe75a73819.png)

I can submit a pull request after polishing up the code. To do that - Is there a good way to ask a FormatTemplate whether it has a key?

atheriel:
This looks like a neat feature, we'd be happy to have your PR. `FormatTemplate` is a bit limited at the moment, unfortunately.

fbreitwieser:
Cool, will submit it. Right now I have the following block in `update` to set `has_frequency` and `has_barchart`. I could put the functionality into a `FormatTemplate` method `pub fn has_vars(Vec<&'static str>) -> Vec<bool>`, or something like that:
 
```rust
impl Block for Cpu {
    fn update(&mut self) -> Result<Option<Duration>> {
        if self.first_update {
            let mut fmt = &self.format;
            loop {
                match fmt {
                    FormatTemplate::Var(ref s, ref next) => {
                        match &s as &str {
                            "{frequency}" => { self.has_frequency = true; },
                            "{barchart}" =>  { self.has_barchart = true; },
                            _ => {},
                        }
                        if let Some(fmt1) = next {
                            fmt = fmt1;
                        } else {
                            break;
                        }
                    }
                    FormatTemplate::Str(ref s, ref next) => {
                        if let Some(fmt1) = next {
                            fmt = fmt1;
                        } else {
                            break;
                        }
                    }
[...]

PR #383, opened 2019-05-15
Add CPU barchart (ref #379)

fbreitwieser:
As discussed in #379, this pull request adds a `format` key to the cpu block, and removes the `frequency` key. The format key can have the placeholders barchart, utilization and frequency.

The default of the format key is currently `{barchart} {utilization}% {frequency}GHz`.

The barchart displays UTF8 block characters for up to 31 distinct CPU cores, each of which has 8 different states ('▁', '▂', '▃', '▄', '▅', '▆', '▇', '█').

Result:
image

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -153,7 +153,7 @@ Update CPU usage every second:
 [[block]]
 block = "cpu"
 interval = 1
-frequency = true
+format = "{barchart} {utilization}% {frequency}GHz"
 ```
 
 ### Options
@@ -164,7 +164,7 @@ Key | Values | Required | Default
 `warning` | Minimum usage, where state is set to warning. | No | `60`
 `critical` | Minimum usage, where state is set to critical. | No | `90`
 `interval` | Update interval, in seconds. | No | `1`
-`frequency` | Shows avg cpu frequency in GHz | No | `false`
+`format` | A format string. Possible placeholders: `{barchart}` (barchart of each CPU's core utilization), `{utilization}` (average CPU utilization in percent) and `{frequency}` (CPU frequency). | No | `"{barchart} {utilization}% {frequency}GHz"`
 
 ## Custom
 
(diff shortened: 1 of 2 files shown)


src/blocks/cpu.rs
```diff
@@ -46,12 +49,16 @@ pub struct CpuConfig {
     #[serde(default = "CpuConfig::default_critical")]
     pub critical: u64,
 
-    /// Display frequency
-    #[serde(default = "CpuConfig::default_frequency")]
-    pub frequency: bool,
```
atheriel: OK, so we don't actually want to remove this, as it will break existing user configurations. Instead, we should permit this key and set `format` to `{utilization} {frequency}GHz` if it is present but `format` is absent. This is a pattern we've used elsewhere when introducing a `format` key.

blocks.md
```diff
@@ -164,7 +164,7 @@ Key | Values | Required | Default
 `warning` | Minimum usage, where state is set to warning. | No | `60`
 `critical` | Minimum usage, where state is set to critical. | No | `90`
 `interval` | Update interval, in seconds. | No | `1`
-`frequency` | Shows avg cpu frequency in GHz | No | `false`
```
atheriel: Don't remove this; instead, add that it is "deprecated in favour of the `format` key".

src/blocks/cpu.rs
```diff
@@ -46,12 +49,16 @@ pub struct CpuConfig {
...
+    /// Format override
+    #[serde(default = "CpuConfig::default_format")]
+    pub format: String,
 }
 
 impl CpuConfig {
+    fn default_format() -> String {
+        "{barchart} {utilization}% {frequency}GHz".to_owned()
```
atheriel: We should probably match the existing block output as the default, which I believe is equivalent to `{utilization}%`.

src/blocks/cpu.rs
fbreitwieser: added frequency back and its documentation back

greshake:
Seems good to me, I think all the requested changes were made. Thank you!

PR #388, opened 2019-05-26
blocks/net: parse SSID containing whitespace correctly

alarsyo:
While being connected to an SSID containing whitespace, I noticed that the current method only shows the first word of the SSID. Here's a more universal solution (discussion welcome on a better command accomplishing the same goal)


atheriel:
Looks good to me. Thanks!

PR #389, opened 2019-05-26
blocks/net: join all IP addresses of an interface in one line

### LINKED ISSUE #386: Network block IPs displayed in a wierd way when IP aliasing configured
When IP aliasing is configured on the network interface, network block correctly obtains all addresses, but each address is in a separate line - and it is also drawn that way in status bar, which makes it unreadable - as seen here:

It is this way because IP is obtained by cutting it out with `sed` from `ip` output that has each address on separate line:
https://github.com/greshake/i3status-rust/blob/master/src/blocks/net.rs#L140

I think the output of sed should be merged into one line, I've even tried to do that by modifying sed command, but while it works fine when I run the command manually, it stops printing any address altogether in `i3status-rs` output, here is what I tried to do:
```
diff --git src/blocks/net.rs src/blocks/net.rs
index 63dffb5..ec548f2 100644
--- src/blocks/net.rs
+++ src/blocks/net.rs
@@ -137,7 +137,7 @@ impl NetworkDevice {
                 &[
                     "-c",
                     &format!(
-                        "ip -oneline -family inet address show {} | sed -rn \"s/.*inet ([\\.0-9/]+).*/\\1/p\"",
+                        "ip -oneline -family inet address show {} | sed -rn -e \"s/.*inet ([\\.0-9/]+).*/\\1/; G; s/\n/ /;h\" -e \"$ P;\"",
                         self.device
                     ),
                 ],
```

I do not know why this approach does not work, although my rust knowledge is limited so maybe I am missing something here.

maciejjo:
When IP aliasing is configured on the selected NIC, 'ip' will print
each address on separate line. If the lines are not joined, it results
in being unreadable in the status line.

Closes #386

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -137,7 +137,7 @@ impl NetworkDevice {
                 &[
                     "-c",
                     &format!(
-                        "ip -oneline -family inet address show {} | sed -rn \"s/.*inet ([\\.0-9/]+).*/\\1/p\"",
+                        "ip -oneline -family inet address show {} | sed -rn -e \"s/.*inet ([\\.0-9/]+).*/\\1/; G; s/\\n/ /;h\" -e \"$ P;\"",
                         self.device
                     ),
                 ],


atheriel:
Thanks!

Issue #380, opened 2019-05-10
Man page


crozbo:
It will be nice to have terminal/offline help, man pages or just pure terminal commands for information's like in wiki pages. Listing available blocks, themes, and icons, blocks, blocks options, theme colors...

atheriel:
Yes, I think having a `man` page would be a nice addition. It should be possible to produce one from the existing documentation via Pandoc, too.

crablab:
I'm happy to look into this, but could you explain a little more about a) where the existing docs are b) how you intend to use Pandoc? (I use it solely for markdown/LaTeX conversion) @atheriel

atheriel:
Pandoc can output the man page format, so we could render the existing `blocks.md` documentation as, say, the **CONFIGURATION** section of a man page pretty easily. There is some additional documentation on the wiki as well, although it is not maintained as well as the block documentation.

crablab:
I have a draft PR #391 to do this. 

The man page is generate by the CI. How do we want to 'deploy' this or otherwise commit back to the repo?

PR #391, opened 2019-05-30
Auto-generate man pages from markdown

crablab:
As in #380, generate proper man pages from the markdown. 

Nb. this generates the man page at CI. Still clarifying how we want to 'deloy' it...


crablab:
So it [looks like](https://travis-ci.community/t/draft-pull-requests-not-being-built/2434/11) Travis doesn't run CI on draft PRs :cry:

greshake:
Should we maybe add a simple pandoc man page generator script for local builds? Then we can push a static version with every tagged release. I don't want to push the pandoc dependency on to the people building the packages

atheriel:
I'd prefer that approach. I don't think this PR goes far enough, and it made me think that we need to actually write a man page, or at least sections of it.

greshake:
Unfortunately our cmdline interface library doesn't support generating man pages yet :( https://github.com/clap-rs/clap/issues/552

crablab:
This PR was never meant to be particularly great :man_shrugging:  

I'm not really sure what direction the maintainers would like to take here. I feel like clap supporting man page generation would be the best solution, but that doesn't seem like it's coming anytime soon.

atheriel:
@crablab To be clear, I did not intend to attack you or the quality of your submission. I just think the scope of this issue is larger than what can be solved using the existing documentation, mostly because a proper man page would need to include sections that simply have not been written yet.

crablab:
@atheriel Oh yeah for sure :blush:  Offence was never taken! I generally agree - my comments are simply that I don't really know what direction people want to take here. 
The problem with manually writing manpages is they quickly tend to get out of sync with developments in the project.

atheriel:
#556 should fulfill the ideas in this PR. Thanks for your work!

PR #392, opened 2019-05-31
 add an option for TB and TiB for disk_space 

marzzzello:


atheriel:
Yep, makes sense. Thanks!

PR #393, opened 2019-06-03
Use `bat_quarter` and `bat_three_quarters` icons from font awesome for battery ranges

Ma27:
This is another indicator how charged the battery currently is.


GladOSkar:
I think you mean "quarter"...

Ma27:
ouch... yes Indeed :sweat_smile:

src/blocks/battery.rs
```diff
@@ -508,7 +508,12 @@ impl Block for Battery {
             self.output.set_icon(match status.as_str() {
                 "Discharging" => "bat_discharging",
                 "Charging" => "bat_charging",
-                _ => "bat",
+                _ => match capacity {
+                    Ok(0...15) => "bat_quarter",
```
atheriel: May I ask why you chose these particular ranges?

Ma27: Originally, I wanted to match the ranges of the coloring and after that I realized that there are only icons for a quarter :sweat_smile: 

But I guess that it's more intuitive to use quarters for the range :+1:

Ma27: fixed :)

src/blocks/battery.rs
```diff
@@ -508,7 +508,12 @@ impl Block for Battery {
             self.output.set_icon(match status.as_str() {
                 "Discharging" => "bat_discharging",
```
atheriel: I'm not quite sure I understand what will happen here. Won't these icons only appear if the battery fails to report correctly as charging/discharging? How do you trigger them on your system?

Also, note that `bat_discharging` is the 50% battery icon; perhaps that could be used as well.

Ma27: Oof, so it seems as my battery doesn't manage to properly report the status (I often have "UNK" with the default i3bar for instance) which is why I probably misread the code. I fixed it now, in case of an unknown status it still tries to determine the status now :)

Ma27:
@atheriel anything else for me to fix? Or did I miss something? :)

atheriel:
CI failures are due to the use of deprecated syntax, but this can be fixed in master.

Issue #364, opened 2019-04-10
[sway] Internal error in context 'util'


Sangeppato:
On sway, I get the following error when I try to use i3status-rust, with both the default configuration and a simpler one, I get the following [error](https://imgur.com/a/8nw3F2Z). The package was installed on Arch Linux with i3status-rust-git from the AUR, thanks

atheriel:
Can you run the program directly in a terminal? It will print out a better error in that case. You can also use `RUST_BACKTRACE=1` to get more information.

Sangeppato:
Actually I don't get any error running it from the shell... This actually makes me think that the problem is on the sway side

atheriel:
Can you post the bar part of your Sway config?

atheriel:
Closing this due to perceived inactivity. Feel free to reopen if you think there's something more to address.

r3drock:
I get the exact same error, also with sway on arch linux. I compiled from source.
When ran in terminal:
```sh
~ ./i3status-rs $HOME/.config/sway/config.toml
{"version": 1, "click_events": true}
[[{"background":"#000000","color":"#dc322f","full_text":" Internal error in context 'util': failed to open file.\nCause: No such file or directory (os error 2) ","separator":false,"separator_block_width":0}]


Internal error in context 'util': failed to open file.
Cause: No such file or directory (os error 2)
```
But the program does not close after the error, it just halts.
So I don't know how to make a backtrace.

r3drock:
I see my mistake now: I named my config config.toml and referenced it as status.toml in my sway config.
After fixing this it all works nicely.

r3drock:
This error had been way easier to see for me, when the errormessage told me which file was missing.
I changed the error message to tell me which file was missing.

atheriel:
I suppose it is obvious to us but not to a user that "failed to open file" can only mean "failed to open the config file", since the config file is the only file we actually open. We should indeed try to fix this.

PR #394, opened 2019-06-03
Make the "missing config" error more verbose

r3drock:
This tells the user which file is missing e.g. in case the config file
is missing. See #364.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -28,7 +28,7 @@ where
 {
     let mut contents = String::new();
     let mut file = BufReader::new(File::open(file)
-        .internal_error("util", "failed to open file")?);
+        .internal_error("util", &("failed to open file".to_owned() + file))?);
     file.read_to_string(&mut contents)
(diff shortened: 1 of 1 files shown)


src/util.rs
```diff
@@ -28,7 +28,7 @@ where
 {
     let mut contents = String::new();
     let mut file = BufReader::new(File::open(file)
-        .internal_error("util", "failed to open file")?);
+        .internal_error("util", &("failed to open file".to_owned() + file))?);
```
atheriel: Is it possible for you to use the same `format!` approach used elsewhere in the project? Using `+` for string concatenation is not very idiomatic.

r3drock: Sure, I did this in https://github.com/greshake/i3status-rust/pull/394/commits/3cbfae5bb5a9981fe43d436c57a28c0e93ed5fcb.

atheriel:
Thanks!

PR #406, opened 2019-07-17
src/blocks/disk_space.rs: fix build on i686

### LINKED ISSUE #403: Broken on 32bit targets
https://github.com/greshake/i3status-rust/blob/master/src/blocks/disk_space.rs#L185-L262 is broken on 32bit targets, as you can see on https://travis-ci.org/void-linux/void-packages/builds/558597989. It seems that nix returns u32 in some cases on those architectures instead of u64.

Gottox:
fixes #403.


atheriel:
Thanks!

PR #407, opened 2019-07-20
Weather Block: Provide default value for missing wind fields, fixes #395

### LINKED ISSUE #395: weather block incorrectly treats missing field as malformed JSON
Sometimes JSON from openweathermap is missing a field (in case of my location it is "deg" parameter of wind is not present when wind speed is low), which causes whole status to disappear, displaying "Error in block 'weather': Malformed JSON." According to thread on OWM support [1], response with some fields missing is valid and means that given characteristic was not measured.
I think that it would be better if missing field was just omitted in weather block format.

Also, maybe error in block could be displayed in the block itself rather than replacing the whole status? Now when the weather block throws BlockError, the whole status is gone and only the error is shown.

[1] https://openweathermap.desk.com/customer/portal/questions/16989445-how-to-treat-missing-json-fields-

bramvdbogaerde:
This pull requests introduces changes to the weather block.

It takes the fact into account that some fields might not be present in the response of the OpenWeatherMap API (see issue #395). Specifically, it provides default values for the fields:

* speed (by default 0.0)
* deg (converted to an Option, displayed as the wind direction if Some(f64) or "-" if None)

Additionally, the pull request suggests some code changes in order to to reduce the amount of code duplication by abstracting the `malformed json` error into its own function, and by using a combination of  built-in operators on `Option`s and `Result`s and the `?` (try) operator.

### DIFF
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -48,6 +48,13 @@ pub struct Weather {
     update_interval: Duration,
 }
 
+fn malformed_json_error() -> Error {
+    BlockError(
+            "weather".to_string(),
+            "Malformed JSON.".to_string(),
+    )
+}
+
 impl Weather {
     fn update_weather(&mut self) -> Result<()> {
         match self.service {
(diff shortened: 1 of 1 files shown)


atheriel:
Nice work, this looks good to me.

Issue #320, opened 2019-01-19
Get API key from env. variable


yangmillstheory:
Thanks for the great project :). I version my dotfiles and I'm planning on using the Weather block, but you need to specify an API key in the .toml file, which risks checking the key into source control. 

Can we implement some way of pulling this information from the process environment?

atheriel:
Sure, that would make sense. If anyone's willing to take it on I will happily review the PR, otherwise it might take a while to get to.

fennifith:
@yangmillstheory I have no Rust experience either, but if you want a "different" solution (that works for other config files too) you could use something like [this script](https://gist.github.com/fennifith/4221a79ec5f39c21085fec4c41d7d160#file-gen-bintray-descriptor-sh) on your config file to format environment variables into it before i3 is started (this is what I'm doing - it allows me to get the network device for the "net" block from `iw dev` as well).

Of course, you still have to remember not to add the formatted file to VCS after the script has been run (maybe you could have it store the formatted version in a different location to avoid this?), but it makes it much easier to manage the configuration across multiple devices.

atheriel:
@fennifith That's a clever example, thanks. We have a few issues filed around sharing configurations across machines, and that serves as a good workaround for many of them.

@yangmillstheory It shouldn't be too hard to get started with Rust, and you can build this project with Cargo. But it's not the most welcoming of languages, either.

Implementation-wise, you should be able to use [the standard library's support for environment variables](https://doc.rust-lang.org/std/env/fn.var.html). Here is how I would do it:

* Change the `api_key` from a `String` to an `Option<String>` in the `OpenWeatherMap` structure (in `src/blocks/weather.rs`). This will allow it to be missing/empty when parsing the configuration file.

* In the logic that reads the configuration (also in that file), check if the `api_key` is `None` and then look in the `OPENWEATHERMAP_API_KEY` environment variable. If both are missing, error out.

The actual tricky bit here will be handling `None` and `Err` correctly. It will end up being a 3-4 line change in the end, but that understates the work needed to get there.

ammgws:
For privacy reasons I'd also want to apply this to `city_id`.

atheriel:
Closed by #410.

PR #410, opened 2019-07-24
Make Weather API Key and City Id Configurable with Environment Variables

nicholasfagan:
Adds functionality requested from issue #320.

To use, leave the weather block's `service.api_key` and/or `service.city_id`
blank, and set them instead with the `OPENWEATHERMAP_API_KEY` and
`OPENWEATHERMAP_CITY_ID` environment variables.

Implemented by making `api_key` and `city_id` `Option<_>`'s, and
providing `serde` with a default function, which tries to get the members
instead from `std::env`. Then just adding another match branch in
`Weather::update_weather` to handle the case when the members are `None`
and return a useful error.

### DIFF
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::env;
 use std::process::Command;
 use std::time::Duration;
 use crossbeam_channel::Sender;
@@ -15,6 +16,11 @@ use crate::util::FormatTemplate;
 use crate::widgets::button::ButtonWidget;
 use crate::widget::I3BarWidget;
 
+lazy_static! {
+    static ref OPENWEATHERMAP_API_KEY_ENV: &'static str = "OPENWEATHERMAP_API_KEY";
+    static ref OPENWEATHERMAP_CITY_ID_ENV: &'static str = "OPENWEATHERMAP_CITY_ID";
+}
+
 #[derive(Clone, Debug, Deserialize)]
 #[serde(tag = "name", rename_all = "lowercase")]
 pub enum WeatherService {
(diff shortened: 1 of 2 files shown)


src/blocks/weather.rs
```diff
@@ -15,6 +16,11 @@ use crate::util::FormatTemplate;
 use crate::widgets::button::ButtonWidget;
 use crate::widget::I3BarWidget;
 
+lazy_static! {
```
greshake: Using lazy_static is not necessary, you can also define it as a compile time constant:
```rust
const TEXT: &str = "Test";
```
See https://doc.rust-lang.org/rust-by-example/custom_types/constants.html for more details :)

src/blocks/weather.rs
```diff
@@ -26,12 +32,23 @@ pub enum WeatherService {
...
+        api_key: Option<String>,
+        #[serde(default = "WeatherService::getenv_openweathermap_city_id")]
+        city_id: Option<String>,
         units: OpenWeatherMapUnits,
     },
 }
 
+impl WeatherService {
+    fn getenv_openweathermap_api_key() -> Option<String> {
+        env::var(OPENWEATHERMAP_API_KEY_ENV.to_string()).ok()
```
greshake: The `to_string()` can be omitted here

src/blocks/weather.rs
```diff
@@ -177,6 +194,27 @@ impl Weather {
                                   "{direction}" => convert_wind_direction(raw_wind_direction),
                                   "{location}" => raw_location);
                 Ok(())
+            },
+            WeatherService::OpenWeatherMap { ref api_key, ref city_id, .. } => {
+                if api_key == &None {
```
greshake: Use `api_key.is_none()` instead or use a pattern matching expression like `if let None = api_key`. See https://doc.rust-lang.org/std/option/enum.Option.html

nicholasfagan:
Thanks for the feedback! I'll get to it in a few hours.
I'll also update the blocks reference to mention the new environment variables.

nicholasfagan:
I think this now closes #320 :)

atheriel:
Ideally these errors would be issued when loading the configuration, but otherwise this looks great, thank you. I'll close #320.

PR #413, opened 2019-07-29
Add a docker block

jlevesy:
## What does this PR do ? 

This PR adds a `docker` block which displays current counts of containers on the local docker host.

It works by calling the [system information](https://docs.docker.com/engine/api/v1.40/#operation/SystemInfo) through the local docker socket using `curl`.

Unfortunately I did not find a way to easily make this call using an in process HTTP client because of the unix socket. If there is any better solution, feel free to suggest it ^^.

## How to test this PR ? 

Add the following snippet to your status bar, and restart your bar.

```
[[block]]
block = "docker"
format = "{running}"
```

And here's a screenshot of  the block running on my setup :tada:.

2019-07-29-192100_1297x103_scrot

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -794,3 +794,33 @@ Key | Values | Required | Default
 `resolution` | Shows the screens resolution | No | `false`
 `step_width` | The steps brightness is in/decreased for the selected screen (When greater than 50 it gets limited to 50) | No | `5`
 `interval` | Update interval, in seconds. | No | `5`
+
+## Docker
+
+Creates a block which shows the local docker daemon status (containers running, paused, stopped, total and image count).
+
+### Examples
+
+```toml
+[[block]]
+block = "docker"
+interval = 2
+format = "{running}/{total}"
+```
+
+### Options
+
+Key | Values | Required | Default
+----|--------|----------|--------
+`interval` | Update interval, in seconds. | No | `5`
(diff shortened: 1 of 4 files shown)


SWW13:
Code LGTM, but I'm not a maintainer.

---

You could use the `unix_socket` crate and just skip the headers. Maybe `HTTP/1.0` allows some edge cases I'm not aware of.

```rust
use std::io::prelude::*;
use unix_socket::UnixStream;
fn main() {
    let mut stream = UnixStream::connect("/var/run/docker.sock").unwrap();
    stream.write_all(b"GET /info HTTP/1.0\r\nHost: api\r\n\r\n").unwrap();
    let mut response = String::new();
    stream.read_to_string(&mut response).unwrap();
    println!("{}", response);
}
```

atheriel:
This is a nice contribution, thank you. I'm going to merge as-is because the unresolved questions do not conflict with it.

I think @SWW13 has provided a nice way to avoid the `curl` dependency, but it's arguable that we want to support non-Unix sockets as well, and `curl` can handle those nicely. Docker permits [using TCP](https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-socket-option) for remote access to the daemon, but I'm not sure how common that actually is on personal machines. Docker also allows you to set the socket path -- which is hard-coded in this PR -- but again, not sure if anyone actually uses that in our domain.

Issue #331, opened 2019-02-15
Allow custom blocks to (configurably) indicate state via exit code


OddBloke:
I have a systemd job that from time-to-time fails; I've written a custom block which calls `systemctl` to find out what the status of the unit is. `systemctl` will return 0 if the unit is running as expected, or 1-4 if it isn't (with each return code meaning a different failure state).

I would like to extend the custom plugin to have configuration along the lines of:

```toml
[[block]]
block = "custom"
command = "systemctl ..."
good_exitcode = 0
warn_exitcode = 3
critical_exitcode = 4
```

When my command exits with those specific exitcodes, I would then like the custom plugin to set the state appropriately.

(This proposal is related to #165, but I believe it is complimentary. Integrating at the exit code level works with a broad range of existing programs without JSON-producing wrappers, but isn't as rich an interface for programs that are looking to integrate specifically with i3status-rust.)

atheriel:
The question I have about this is that I'm not sure your example is very compelling: it's pretty arguable that any non-zero exit code from `systemctl status` would be worthy of a warning or error status. Anyone wanting to use this feature would likely have to write custom scripts to output very carefully-chosen exit codes, since most existing programs only issue non-zero exit codes for errors.

Isn't a straightforward alternative to add an option like `warn_on_nonzero_exit = true` instead?

nicholasfagan:
With this being a custom script, any no zero exit can be changed to something else, i.e. `systemctl status || exit 1`.

OddBloke:
I agree in retrospect that the `systemctl `example isn't very compelling. A more interesting use case, perhaps, _is_ a specifically written custom script that can use exit codes to signal block state. This would make the custom block interface a little richer for people who are happy to write slightly more involved commands, without affecting the majority of use cases.

In a scenario where people are expected to be writing adapters anyway, perhaps my original proposal could be changed to a single configuration boolean? Something like `warn_on_exit_1 = true` would mean that exit code 1 would be a warning (instead of an error as under usual operation)?

What do you think?

ammgws:
@OddBloke The JSON option that was just merged in #515 may be of use to you. Could use it like this:

i3status-rs config
```
[[block]]
block = "custom"
command = "<run your script>"
json = true
```
script (sorry this is in fish script, but you get the idea)
```fish
systemctl status random-unit
switch $status
    case 0
        set state Good
    case 3
        set state Warn
    case 4
        set state Critical
    case '*'
        set state Warn
end
[...]

PR #414, opened 2019-07-30
Add support to match exit codes to states for custom block

nicholasfagan:
Adds feature requested in #331, also related to #165. 
The default state is idle. Any exit codes that should be an exception to that can be listed.

Is this easy enough to set multiple exit codes, or should it be improved by allowing selecting ranges of exit codes? This would significantly increase complexity but would make it easier for certain tasks, like mark all exit codes > 0 as warning. Let me know what you think.

### DIFF
--- a/src/blocks/custom.rs
+++ b/src/blocks/custom.rs
@@ -10,7 +10,7 @@ use crate::config::Config;
 use crate::de::deserialize_duration;
 use crate::errors::*;
 use crate::widgets::button::ButtonWidget;
-use crate::widget::I3BarWidget;
+use crate::widget::{I3BarWidget, State};
 use crate::input::I3BarEvent;
 use crate::scheduler::Task;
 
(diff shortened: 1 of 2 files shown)


src/blocks/custom.rs
```diff
@@ -89,13 +109,26 @@ impl Block for Custom {
...
+                } else if self.warning_exit_codes.contains(&statuscode) {
+                    State::Warning
+                } else if self.good_exit_codes.contains(&statuscode) {
+                    State::Good
+                } else if self.info_exit_codes.contains(&statuscode) {
+                    State::Info
+                } else {
+                    State::Idle
+                }
+        );
```
nicholasfagan: To improve performance in extreme cases (for example if the user configured a list of 1 million numbers), this could be changed to have a `HashMap<i32, State>` that could be populated in `Custom::new`.

atheriel: That seems like a silly usecase for us to support, so the existing approach makes more sense to me.

ohk2kt3t4:
Are there any blockers to merging this?

ammgws:
I think if #166 is reimplemented as part of the current custom block then you can have the functionality mentioned in OP by writing wrappers for whatever command you're running.

ammgws:
This use-case can be covered by the new `json` option for the Custom block. See example given here: https://github.com/greshake/i3status-rust/issues/331#issuecomment-617185419

PR #415, opened 2019-07-31
Space villain theme

JeremyBellows:
Inspired by cartoon villain color schemes

Colors can be seen here
https://coolors.co/06060f-394049-00223f-c1c1c1-2d1637

space-villain-theme


ammgws:
@atheriel What do you reckon?

atheriel:
@ammgws I like the idea of including more theme options but I don't like the fact that we embed them in the source code, and that users cannot supply their own. I'd like to get #366 off the ground to make adding themes relatively painless.

PR #418, opened 2019-08-08
Fixes #387: Display WiFi signal strength in the net block

### LINKED ISSUE #387: Wireless signal quality
Is there a way to display the wireless signal quality similarly to what was done in i3status?

bnjbvr:
Hi! I wanted to see the Wifi signal strength in the status bar, and saw there was an open issue with some hints to do it. Turns out that `iw` provides this information too, so we can parse its output to figure it out. It's a bit inefficient because we may call iw several times to grep for some of its information, but I think it's fine since it doesn't happen a lot.

I've reused the formulas hinted at in the issue, and I can confirm they coincide with the results provided by the original i3status bar on my machine.

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -130,6 +130,62 @@ impl NetworkDevice {
         }
     }
 
+    fn absolute_signal_strength(&self) -> Result<Option<i32>> {
+        let up = self.is_up()?;
+        if !self.wireless || !up {
+            return Err(BlockError(
+                "net".to_string(),
+                "Signal strength is only available for connected wireless devices."
+                    .to_string(),
+            ));
+        }
+        let mut iw_output = Command::new("sh")
+            .args(&[
+                "-c",
+                &format!(
+                    "iw dev {} link | sed -n 's/^\\s\\+signal: \\(.*\\) dBm/\\1/p'",
+                    self.device
+                ),
+            ])
+            .output()
+            .block_error("net", "Failed to execute signal strength query.")?
+            .stdout;
+        if iw_output.is_empty() {
+            Ok(None)
+        } else {
+            iw_output.pop(); // Remove trailing newline.
+            String::from_utf8(iw_output)
+                .block_error("net", "Non-UTF8 signal strength.")
+                .and_then(|as_str| as_str.parse::<i32>()
+                    .block_error("net", "Non numerical signal strength."))
+                .map(Some)
+        }
+    }
+
+    fn relative_signal_strength(&self) -> Result<Option<u32>> {
+        let xbm = if let Some(xbm) = self.absolute_signal_strength()? {
+            xbm as f64
+        } else {
+            return Ok(None);
+        };
+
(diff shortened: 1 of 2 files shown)


src/blocks/net.rs
```diff
@@ -130,6 +130,62 @@ impl NetworkDevice {
         }
     }
 
+    fn absolute_signal_strength(&self) -> Result<Option<i32>> {
+        let up = self.is_up()?;
+        if !self.wireless || !up {
+            return Err(BlockError(
+                "net".to_string(),
+                "SSIDs are only available for connected wireless devices."
```
SWW13: The error message should be changed.

bnjbvr: Good catch! Fixed.

atheriel:
Thanks!

rsteube:
Oh nice, it's already in.
@bnjbvr i'm getting `command failed: Operation not permitted (-1)` for the `iw dev .. link` command - am i missing a group or is a sudo entry necessary?

(fyi: network-manager gives out bars like `▂▄▆_` with: `nmcli device wifi list | grep '^*' | awk '{print $8}' )

bnjbvr:
@rsteube I have honestly no clues regarding the issue; a quick search seems to tell it could be a driver issue.

Regarding the nmcli line: it also contains the relative signal strength (under the SIGNAL column), so we could directly copy it out from there. Nice thing is that it displays the same value as my code :) (which is obvious, since the code I added is copied from network-manager's...)

rsteube:
Yes would be nice to have nmcli as option since it seems iw won't work without sudo on some systems.
This looks great for parsing: `nmcli -t -f in-use,active,ssid,signal dev wifi`:
```
nmcli -t -f in-use,active,ssid,signal dev wifi
 :no:guestnet:80
 :no:skynet:79
*:yes:skynet:66
```
(only works if network-manager is used though)

rsteube:
This works for me: https://github.com/rsteube/i3status-rust/commit/fcf4571c9a2ff45c29fa4209c55150d0893bb49b

BenFradet:
weird thing, building from master I'm getting:

```
[[{"background":"#000000","color":"#dc322f","full_text":" Configuration error: failed to deserialize block config.\nCause: unknown field `signal_strength`, expected one of `interval`, `device`, `ssid`, `max_ssid_width`, `bitrate`, `ip`, `hide_inactive`, `hide_missing`, `speed_up`, `speed_down`, `graph_up`, `graph_down`, `on_click` ","separator":false,"separator_block_width":0}]


Configuration error: failed to deserialize block config.
Cause: unknown field `signal_strength`, expected one of `interval`, `device`, `ssid`, `max_ssid_width`, `bitrate`, `ip`, `hide_inactive`, `hide_missing`, `speed_up`, `speed_down`, `graph_up`, `graph_down`, `on_click`
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: "SendError(..)"', src/libcore/result.rs:999:5
```

when building from 374e70978e2e3619aef82d56c974ed32ee1cfb7c

BenFradet:
nvm, the wrong artifact was picked up appreantly

PR #419, opened 2019-08-11
Add basic support for wireguard devices

vvrein:
This is only basic support, because wireguard device can be named at will:
```
[root@tokk wireguard]# ip -4 a s
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    inet 127.0.0.1/8 scope host lo
2: enp6s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    inet 192.168.2.3/24 brd 192.168.2.255 scope global noprefixroute enp6s0
33: vpnwireguarg0: <POINTOPOINT,NOARP,UP,LOWER_UP> mtu 1420 qdisc noqueue state UNKNOWN group default qlen 1000
    inet 10.0.0.3/24 scope global vpnwireguarg0
[root@tokk wireguard]# grep wireguard /sys/class/net/vpnwireguarg0/uevent
DEVTYPE=wireguard
```
And the only way to (as far as I understand) distinguish wg device from any other - is to read file /sys/class/net/$DEVICE/uevent, and find there `wireguard` string.

I tried to do it this way:
```rust        
        let wg_uevent_path = device_path.join("uevent");
        // pseudocode
        if wg_uevent_path.exists() then { 
            let uevent_data = read_to_string(&wg_uevent_path);
            let wg = uevent_data.contains("wireguard");
       } else { 
            let wg = false;
       };
```
but I never had any experience with Rust, so nothing happened...

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -1,4 +1,5 @@
 use std::fs::OpenOptions;
+use std::fs::read_to_string;
 use std::io::prelude::*;
 use std::path::{Path, PathBuf};
 use std::process::Command;
@@ -24,6 +25,7 @@ pub struct NetworkDevice {
     device_path: PathBuf,
     wireless: bool,
     tun: bool,
+    wg: bool,
 }
 
 impl NetworkDevice {
@@ -36,11 +38,24 @@ impl NetworkDevice {
         let wireless = device_path.join("wireless").exists();
         let tun = device_path.join("tun_flags").exists() || device.starts_with("tun") || device.starts_with("tap");
 
+        fn is_wg_device(wg_uevent_path: &PathBuf) -> bool {
+            if wg_uevent_path.exists() {
+                let uevent_data = read_to_string(&wg_uevent_path).expect("Unable to read file");
+                return uevent_data.contains("wireguard");
+            } else {
+                return false;
+            };
+        };
+
+        let wg_uevent_path = device_path.join("uevent");
+        let wg = is_wg_device(&wg_uevent_path);
+
         NetworkDevice {
             device,
             device_path,
             wireless,
             tun,
+            wg,
         }
     }
 
@@ -58,6 +73,8 @@ impl NetworkDevice {
             // opposed to erroring out the entire block.
             Ok(false)
         } else if self.tun {
+            Ok(true)
+        } else if self.wg {
             Ok(true)
         } else {
             let operstate = read_file(&operstate_file)?;
@@ -86,7 +103,7 @@ impl NetworkDevice {
 
     /// Checks whether this device is vpn network.
     pub fn is_vpn(&self) -> bool {
-        self.tun
+        self.tun || self.wg
     }
 
     /// Queries the wireless SSID of this device (using `iw`), if it is
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -38,17 +38,11 @@ impl NetworkDevice {
         let wireless = device_path.join("wireless").exists();
         let tun = device_path.join("tun_flags").exists() || device.starts_with("tun") || device.starts_with("tap");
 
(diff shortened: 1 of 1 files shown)


atheriel:
I think the idea of detecting Wireguard devices is a good one, but I'm not sure why they would need to be distinguished from other tunnel/VPN devices. Also, does this approach rely on yet-unmerged kernel support?

vvrein:
>Also, does this approach rely on yet-unmerged kernel support?

I did not quite understand, but it seems yes, wireguard support is not yet included in the kernel, only a separate kernel module.
By the way, I studied rust a little, maybe after a while I can add the correct wg detection with `uevent_data.contains("wireguard");`

Question by code - is it better to add wg support as
```let tun = device_path.join ("tun_flags"). exists () || device.starts_with ("tun") || device.starts_with ("tap") || device.starts_with ("wg");``` ?

**Comment by vvrein** [comment]
In general, I've added wg as separate from tun/tap as wireguard works on very separate protocol.
But I'm not pretty sure that it is right decision in scope of i3status-rust :)
Any suggestions?

**Comment by atheriel** [comment]
At the moment I think the only special aspect of tun/tap devices is that they get a special VPN icon, which is what you've done for Wireguard as well. I think that's the right choice.

>  I did not quite understand, but it seems yes, wireguard support is not yet included in the kernel, only a separate kernel module.

I was wondering if you were running a custom kernel, and whether that would affect your detection method (i.e. reading the `uevent` file). But you are not, so this should work for other users.

Is there a reason you don't want to include the `uevent` file reading in this PR? I think it's more robust than looking for a `wg` prefix.

**Comment by vvrein** [comment]
>Is there a reason you don't want to include the uevent file reading in this PR?  

Because I did not have any experience in programming on the rast until this time, I know a little bash and a little Python, and first changes was easy for me.

Now I know Rust a little bit more so i maked detection by uevent file :)

I'm wondering is my approach suitable for you (function in top of netdev definition, if path exists) ?

**Comment by vvrein** [comment]
Hmm, strange things, i still can build my & upstream version of i3status-rust with rust 1.37.0

**Comment by vvrein** [comment]
Aw, but i can't with RUSTFLAGS="-D warnings"

File: src/blocks/net.rs [hunk 1]
```diff
@@ -36,11 +38,24 @@ impl NetworkDevice {
         let wireless = device_path.join("wireless").exists();
         let tun = device_path.join("tun_flags").exists() || device.starts_with("tun") || device.starts_with("tap");
 
+        fn is_wg_device(wg_uevent_path: &PathBuf) -> bool {
+            if wg_uevent_path.exists() {
+                let uevent_data = read_to_string(&wg_uevent_path).expect("Unable to read file");
```
atheriel: In general we want to avoid `panic`-ing, is it possible to use `?` here instead of `expect()`?

atheriel:
It looks like the newest compiler has deprecated some common patterns in the codebase, but there are no errors related to your changes, so don't worry about it.

src/blocks/net.rs
atheriel: Actually, since we don't return a `Result` here, maybe just treat failures to read this file as `false` for the Wireguard check.

vvrein: Read some examples with error handling - this way should be more concise, and without panicking since any error is interpreted as `false`
Does it satisfy our requirements?
```rust
        let wg_uevent_path = device_path.join("uevent");
        let wg = match read_to_string(&wg_uevent_path) {
                Ok(s) => s.contains("wireguard"),
                Err(_e) => false,
        };
```

atheriel: Looks good to me.

vvrein:
Pushed new changes

atheriel:
Thanks for your patience on this. Looks good to me.

PR #423, opened 2019-08-15
Blocks/battery: handle devices with no "battery" prefix in UPower

### LINKED ISSUE #421: Battery block can't show UPower batteries without battery_ prefix
Since #356 while using ```upower``` as ```driver``` anything but ```DisplayDevice``` in ```device``` is treated as a ```battery_``` suffix.
With the following list of batteries, I am not able to read the battery status of my mouse while using UPower
```
 ~ % upower --enumerate
/org/freedesktop/UPower/devices/line_power_AC
/org/freedesktop/UPower/devices/battery_BAT0
/org/freedesktop/UPower/devices/mouse_hidpp_battery_0
```

freswa:
fixes #421

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -15,6 +15,7 @@ use uuid::Uuid;
 use crate::block::{Block, ConfigBlock};
 use crate::blocks::dbus;
 use crate::blocks::dbus::stdintf::org_freedesktop_dbus::Properties;
+use crate::blocks::dbus::arg::Array;
 use crate::config::Config;
 use crate::de::deserialize_duration;
 use crate::errors::*;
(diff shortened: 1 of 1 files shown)


src/blocks/battery.rs
```diff
@@ -232,7 +237,10 @@ impl UpowerDevice {
             .block_error("battery", "Failed to read UPower Type property.")?;
 
         // https://upower.freedesktop.org/docs/Device.html#Device:Type
-        if upower_type != 2 {
+        // consider any peripheral, UPS and internal battery
+        let list: [u32; 5] = [2,3,4,5,6];
+
+        if !list.contains(&upower_type) {
```
atheriel: It may be better to simply disallow type `1` ("Line Power"). I'd also appreciate if the link to the docs remained in a comment so we have it for future reference (as happened here).

src/blocks/battery.rs
```diff
@@ -221,7 +221,12 @@ impl UpowerDevice {
     /// if D-Bus cannot connect to this device, or if the device is not a
     /// battery.
     pub fn from_device(device: &str) -> Result<Self> {
-        let device_name = if device == "DisplayDevice" { device.to_string() } else { format!("battery_{}", device) };
+        let device_name =
+            if device == "DisplayDevice" || !device.starts_with("BAT") {
```
atheriel: This is a good intermediate solution and I won't hold off merging with it in place, but it's possible that this will surface issues if batteries are not given `BAT*`-style names on all systems.

src/blocks/battery.rs
freswa: Done

src/blocks/battery.rs
freswa: I've changed the logic to either use ```DisplayDevice``` or otherwise search for ```device``` as suffix in ```upower --enumerate```

PR #425, opened 2019-08-18
Add GitHub block

jlevesy:
## What does this PR do ? 

This PR adds a `github` block which displays current count of unread notifications for an user. 
It enables to display a total count of unread notifications and also a count per "notification reasons" [see here](https://developer.github.com/v3/activity/notifications/#notification-reasons).

To support this, I had to make a change on the token regexp of the `FormatTemplate`, to make it also support tokens with a `_` or a `-` character in it (see bf71cd0).

It relies on the [Notifications API](https://developer.github.com/v3/activity/notifications/) of the Github API v3 and iterates through all the notifications pages.

## How to test this PR ? 

Add the following snippet to your status bar, and restart your bar.

```
[[block]]
block = "github"
format = "{total}|{author}|{comment}|{mention}|{review_requested}"
```

And here's a screenshot of  the block running on my setup.

2019-08-18-135311_338x97_scrot

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -361,7 +361,7 @@ Key | Values | Required | Default
 `interval` | Update interval, in seconds. | No | `30`
 `format` | A format string. See below for available placeholders. | No | `"{total}"`
 
-It requires a github token, which must be passed using the `GITHUB_TOKEN` environment variable.
+It requires a Github [personal access token](https://github.com/settings/tokens/new) with the "notifications" scope. It must be passed using the `GITHUB_TOKEN` environment variable.
 
 ### Available Format Keys
 
--- a/blocks.md
+++ b/blocks.md
@@ -360,13 +360,28 @@ Key | Values | Required | Default
 ----|--------|----------|--------
 `interval` | Update interval, in seconds. | No | `30`
 `format` | A format string. See below for available placeholders. | No | `"{total}"`
+`api_server`| API Server URL to use to fetch notifications. | No | `https://api.github.com`
 
-It requires a Github [personal access token](https://github.com/settings/tokens/new) with the "notifications" scope. It must be passed using the `GITHUB_TOKEN` environment variable.
+It requires a Github [personal access token](https://github.com/settings/tokens/new) with the "notifications" scope. It must be passed using the `I3RS_GITHUB_TOKEN` environment variable.
 
 ### Available Format Keys
 
-The block maintains a count of all notifications under the `total` format key.
-It also maintains count per "notification reason". All specified [reasons](https://developer.github.com/v3/activity/notifications/#notification-reasons) are available as format keys.
+ Key | Value
+-----|-------
+`{total}` | Total of all notifications
+`{assign}` | Total of notifications related to issues you're assigned on.
+`{author}` | Total of notifications related to threads you are the author of.
+`{comment}` | Total of notifications related to threads you commented on.
+`{invitation}` | Total of notifications related to invitations.
+`{manual}` | Total of notifications related to threads you manually subscribed on.
+`{mention}` | Total of notifications related to content you were specifically mentioned on.
+`{review_requested}` | Total of notifications related to PR you were requested to review.
+`{security_alert}` | Total of notifications related to security vulnerabilities found on your repositories.
+`{state_change}` | Total of notifications related to thread state change.
+`{subscribed}` | Total of notifications related to repositories you're watching.
+`{team_mention}` | Total of notification related to thread where your team was mentioned.
+
+For more information about reasons, please see the [API documentation](https://developer.github.com/v3/activity/notifications/#notification-reasons).
 
 ## IBus
 
--- a/blocks.md
+++ b/blocks.md
@@ -9,6 +9,7 @@
 - [Disk Space](#disk-space)
 - [Docker](#docker)
 - [Focused Window](#focused-window)
+- [Github](#github)
 - [IBus](#ibus)
 - [Keyboard Layout](#keyboard-layout)
 - [Load](#load)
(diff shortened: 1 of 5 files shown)


atheriel:
I like the idea of this block, but it's hard to swallow an estimated 100 new dependencies for it (see #272). Is it possible to use `curl` instead, as we do in other blocks?

jlevesy:
> I like the idea of this block, but it's hard to swallow an estimated 100 new dependencies for it (see #272). Is it possible to use curl instead, as we do in other blocks?

Yeah, you're right.

Using `curl` will add a bit of complexity because I need to collect some header to iterate through all notifications, so I'll have to do this by hand.

Or maybe... Isn't there a more "self contained" http client in rust that would fit our use case ? Having the bar able to make a request without requiring curl might be interesting too.

Also we should probably mention in documentation that `curl` is required to use some of the blocks :sweat_smile: . 

WDYT ?

ammgws:
@jlevesy Are you still interested in working on this? It's fine to use `curl` as we already require it for the weather block.

jlevesy:
@ammgws I'll do something this week, thanks for the remidner !

ammgws:
Looking forward to it!

jlevesy:
I rebased and did the asked changes. I'm waiting a bit before marking this as ready for review because I want to properly test it in conditions 😄.

ammgws:
Good stuff!
Seems to be working for me, or at least the `total` count anyway. Will need to try it out some more to check the others.
image

Did you forget to add an icon for the block? Your initial post seems to have one.

blocks.md
```diff
@@ -341,6 +342,32 @@ Key | Values | Required | Default
...
+```
+
+### Options
+
+Key | Values | Required | Default
+----|--------|----------|--------
+`interval` | Update interval, in seconds. | No | `30`
+`format` | A format string. See below for available placeholders. | No | `"{total}"`
+
+It requires a github token, which must be passed using the `GITHUB_TOKEN` environment variable.
```
ammgws, line 364: Looks like GitHub calls them "personal access tokens" so perhaps we should align our wording with that for maximum clarity. Also might be good to mention that only the `notifications` scope is required when generating a new token for the github block.

jlevesy:
> Did you forget to add an icon for the block? Your initial post seems to have one.

Nice catch, thanks, will fix.

blocks.md
jlevesy: Alright, will do. Thanks !

ammgws:
Could you @ me so I can see if my bar changes?

jlevesy:
@ammgws Adressed your comments. feel free to review :)

ammgws:
image
It might be nice to match the weather block's "x" instead of "N/A". I think some other blocks use the "x" as well.

src/blocks/github.rs
```diff
@@ -0,0 +1,277 @@
+extern crate lazy_static;
```
ammgws, line 1: I don't believe this is needed, so we can remove this line.

src/blocks/github.rs
```diff
@@ -0,0 +1,277 @@
...
+
+use crate::blocks::Update;
+use crate::blocks::{Block, ConfigBlock};
```
ammgws, line 4: Can you group these into one line?

src/blocks/github.rs
```diff
@@ -0,0 +1,277 @@
...
+pub struct GithubConfig {
+    /// Update interval in seconds
+    #[serde(
+        default = "GithubConfig::default_interval",
+        deserialize_with = "deserialize_duration"
+    )]
+    pub interval: Duration,
+
+    #[serde(default = "GithubConfig::default_api_server")]
+    pub api_server: String,
```
ammgws, line 44: Did you intend to document this option, or is it hidden on purpose?

src/blocks/github.rs
```diff
@@ -0,0 +1,277 @@
...
+    fn default_format() -> String {
+        "{total}".to_owned()
+    }
+}
+
+impl ConfigBlock for Github {
+    type Config = GithubConfig;
+
+    fn new(block_config: Self::Config, config: Config, _: Sender<Task>) -> Result<Self> {
+        let token = match std::env::var(GITHUB_TOKEN_ENV).ok() {
```
ammgws, line 69: It seems like GITHUB_TOKEN might be used by some official GitHub tools: https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#about-the-github_token-secret

Do you think it's better to rename the env to something like I3RS_GITHUB_TOKEN?

src/blocks/github.rs
```diff
@@ -0,0 +1,277 @@
...
+            "{author}" => format!("{}", aggregations.get("author").unwrap_or(&default)),
+            "{comment}" => format!("{}", aggregations.get("comment").unwrap_or(&default)),
+            "{invitation}" => format!("{}", aggregations.get("invitation").unwrap_or(&default)),
+            "{manual}" => format!("{}", aggregations.get("manual").unwrap_or(&default)),
+            "{mention}" => format!("{}", aggregations.get("mention").unwrap_or(&default)),
+            "{review_requested}" => format!("{}", aggregations.get("review_requested").unwrap_or(&default)),
+            "{security_alert}" => format!("{}", aggregations.get("security_alert").unwrap_or(&default)),
+            "{state_change}" => format!("{}", aggregations.get("state_change").unwrap_or(&default)),
+            "{subscribed}" => format!("{}", aggregations.get("subscribed").unwrap_or(&default)),
+            "{team_mention}" => format!("{}", aggregations.get("team_mention").unwrap_or(&default))
```
ammgws, line 127: Since these are hardcoded here we might as well mention them in `blocks.md` to save the user having to go check the github api page.

src/blocks/github.rs
```diff
@@ -0,0 +1,277 @@
...
+
+        if self.next_page_url == "" {
+            return Ok(None);
+        }
+
+        let result = Command::new("sh")
+            .args(&[
+                "-c",
+                &format!(
+                    "curl -s -D - -H \"Authorization: Bearer {token}\" -m 3 \"{next_page_url}\"",
```
ammgws, line 195: Could you change this to long options instead?

src/blocks/github.rs
jlevesy: Yeah it makes sense. If one wants to use one different token per app, that's a valid use case.  Will change to `I3RS_GITHUB_TOKEN`.

jlevesy: Yeah it makes sense. If one wants to use one different token per app, that's a valid use case.  Will change to `I3RS_GITHUB_TOKEN`.

src/blocks/github.rs
jlevesy: Oops, I forgot to document this option. Nice catch, thanks !

ammgws:
Thanks!

PR #428, opened 2019-08-19
Fixes #427: Update bare trait objects and use inclusive range syntax.

### LINKED ISSUE #427: Travis builds are failing due to use of deprecated features
From my reading, we need to fix bare trait objects (via the `dyn` keyword) and the inclusive range syntax.

dauc:
I cheated and just ran `cargo fix`; turns out the only changes were `dyn` and the range syntax.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -12,7 +12,7 @@ pub trait Block {
         Ok(None)
     }
     /// Returns the view of the block, comprised of widgets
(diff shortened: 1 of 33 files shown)


dauc:
Second commit removing duplicate semicolons should fix nightly build as well.

atheriel:
Oh, excellent! I was not aware of `cargo fix`. Thanks!

Issue #432, opened 2019-08-28
No Humidity or "Feels like"


ryanswilson59:
The Weather block doesn't have an option to display humidity or the "feels like" temperature. 
I wrote a small patch that adds this functionality that I pushed to my fork. I have not programmed rust before so I am sure there are some code style/formatting issues.

ammgws:
Added via #640

PR #433, opened 2019-08-28
Humidity and Apparent Temperature

ryanswilson59:
This pull request is related to the issue I opened, https://github.com/greshake/i3status-rust/issues/432 about adding the ability to display humidity and "feels like" temperature.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -756,6 +756,8 @@ Key | Value
 ----|-------
 `{location}` | Location name (exact format depends on the service).
 `{temp}` | Temperature.
+`{apparent}` | Apparent Temperature.
+`{humidity}` | Humidity.
 `{weather}` | Textual description of the weather, e.g. "Raining".
 `{wind}` | Wind speed.
 `{direction}` | Wind direction, e.g. "NE".
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -126,6 +126,10 @@ impl Weather {
 
                 let raw_temp = json.pointer("/main/temp").and_then(|v| v.as_f64()).ok_or_else(malformed_json_error)?;
 
+                let raw_humidity = json.pointer("/main/humidity")
+                    .and_then(|v| v.as_f64())
+                    .ok_or_else(malformed_json_error)?;
+
                 let raw_wind_speed: f64= json.pointer("/wind/speed")
                     .map_or(Some(0.0), |v| v.as_f64()) // provide default value 0.0
                     .ok_or_else(malformed_json_error)?; // error when conversion to f64 fails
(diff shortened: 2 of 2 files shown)


src/blocks/weather.rs
```diff
@@ -126,6 +126,10 @@ impl Weather {
 
                 let raw_temp = json.pointer("/main/temp").and_then(|v| v.as_f64()).ok_or_else(malformed_json_error)?;
 
+                let raw_humidity = json.pointer("/main/humidity")
+                    .and_then(|v| v.as_f64())
```
atheriel: It's likely that this will be missing in some locations. Best to use `.map_or(Some(0.0), |v| v.as_f64())` instead, as for wind speed below.

src/blocks/weather.rs
```diff
@@ -141,6 +145,35 @@ impl Weather {
                     .ok_or_else(malformed_json_error)?;
 
                 // Convert wind direction in azimuth degrees to abbreviation names
+                // Compute the felt temperature according to Australian
+                // apparent temperature as found on wikipedia
+                // unsure about how to properly make if statement
+                let check = match *units {
```
atheriel: A more idiomatic approach in Rust is to use `match` statements whenever possible, instead of `if`. I would suggest moving everything into one `match` statement here:

```rust
let apparent_temp = match *units {
  OpenWeatherMapUnits::Metric => {
    // All calculations for metric units here.
  },
  OpenWeatherMapUnits::Imperial => {
    // All calculations for imperial units here.
  },
};
```

This will mean some duplication of code, but it will make it much easier to understand what is happening.

atheriel:
Thanks for your patience. I left some comments. If possible, can you rebase against master to fix the merge conflicts?

blocks.md
```diff
@@ -756,6 +756,8 @@ Key | Value
 ----|-------
 `{location}` | Location name (exact format depends on the service).
 `{temp}` | Temperature.
+`{apparent}` | Apparent Temperature.
```
atheriel: What about `{feelslike}` instead? Also, it might be worth mentioning that we use the "Australian" method with a link to Wikipedia here, rather than in your code.

src/blocks/weather.rs
ryanswilson59: I structured the code the way I did because the formula is only published for metric units. So I figured it would make sense to do all the calculations in metric and then just do unit conversions for "imperial" units.

blocks.md
ryanswilson59: I was also thinking of maybe providing multiple different heat indexes? Different countries use different methods of calculation and there seems to be little logic in why to use one over another. These are some of the different ones I have found.
https://en.wikipedia.org/wiki/Humidex
https://en.wikipedia.org/wiki/Heat_index
https://en.wikipedia.org/wiki/Wind_chill (this one also has the formula for the Australian method)

ammgws:
@ryanswilson59 Are you still interested in working on this?

src/blocks/weather.rs
ammgws: Yeah I guess it's OK for now since I didn't find an imperial formula from a quick Google.

blocks.md
ammgws: I think apparent is fine as the other ones could be added later. Clarified in the docs.

ammgws:
Merged via #640 

Thanks!

PR #434, opened 2019-09-04
Reduce dependencies

ohk2kt3t4:
`regex` 1.3.0 allows for shrinking the dependency tree by disabling crate features. As far as I can tell we are only using the regex crate for some simple ASCII strings so we do not need the Unicode tables, thus we can safely disable those crate features to reduce our dependencies:

Before
image

After
image

(graphs generated with `cargo deps | dot -Tpng > graph.png`)

Due to `parse-zoneinfo` the dependencies have not completely gone away but if `parse-zoneinfo` updates too then the regex 0.2 dependencies should go away too.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,9 +1,5 @@
 [package]
 name = "i3status-rs"
-description = "A feature-rich and resource-friendly replacement for i3status, written in Rust."
-repository = "https://github.com/greshake/i3status-rust/"
-readme = "README.md"
-license = "GPLv3"
 version = "0.10.0"
 authors = ["Kai Greshake <development@kai-greshake.de>",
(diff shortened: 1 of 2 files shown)


atheriel:
Thanks, this looks good. I don't see any changes to the `Cargo.lock` file, though -- am I missing something?

Cargo.toml
```diff
@@ -1,9 +1,5 @@
 [package]
 name = "i3status-rs"
-description = "A feature-rich and resource-friendly replacement for i3status, written in Rust."
```
atheriel: These lines have probably been deleted by accident. Please revert.

ammgws:
Should be good now.

ammgws:
FYI once chronotope/parse-zoneinfo#6 is merged the other dependencies mentioned above will go away :)

PR #436, opened 2019-09-07
Use more accurate calc for strings with multibyte Unicode chars

### LINKED ISSUE #435: Music block behaves strangely when multibyte Unicode text in song title
When using Spotify and playing https://open.spotify.com/track/68zMaaxfx3mN8lFNtp0vUU?si=zMmIVqEMRYWTzhJbO90LtQ, I noticed some unusual behaviour from the music block.

Normally, the text scrolls right-to-left. With the RTL song title, the text instead gets added one character at a time left-to-right. When the characters reach the right side of the block, the block expands a certain amount (until, I guess, it hits a hard cap on its width). Then, nothing happens for a while (presumably while the remaining song title characters are appended out of view) and then the text starts scrolling right-to-left (as usual) and the block shrinks back down to its regular size.

This repeats over again, as you would expect.

ammgws:
To fix #435 .

Ideally I would like to go through all the logic for the text rotation to make sure this is the best fix, but it appears to be working so will submit for review now.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -744,6 +744,10 @@ Key | Values | Required | Default
 `city_id` | OpenWeatherMap's ID for the city. | Yes | None
 `units` | One of `metric` or `imperial`. | Yes | None
 
+The options `api_key` and/or `city_id` can be omitted from configuration,
+in which case they must be provided in the environment variables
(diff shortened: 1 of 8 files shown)


ammgws:
Oops has some unrelated commits. Will remove them later.

ammgws:
Screwed up the rebase so I'll just submit again

PR #437, opened 2019-09-07
Change rot. text length calcfor better fit with multi-byte Unicode chars

### LINKED ISSUE #435: Music block behaves strangely when multibyte Unicode text in song title

ammgws:
To fix #435 .

Ideally I would like to go through all the logic for the text rotation to make sure this is the best fix, but it appears to be working so will submit for review now.


atheriel:
I don't have a way to test this myself -- do you consider it ready to merge as-is? Additionally, you might need to rebase against master to get the new formatting changes.

ammgws:
I've been using it these past couple months so I'd consider it good to go. I'll rebase soon.

ammgws:
OK I'm at a loss - how do I get rid of commits d0c5d30 ~ d5605d5 from this branch?

ammgws:
@atheriel Rebased, good to go.

PR #443, opened 2019-09-17
Fixes a panic in the iBus block due to the use of Perl regex features

ammgws:
After #434 was merged, [this](https://github.com/greshake/i3status-rust/blob/f90bc9874d060a24276c2808275f4068d30ba47b/src/blocks/ibus.rs#L185) line started to cause a panic when using the `ibus` block. 
Adding the `unicode-perl` feature to the crate resolves this issue.

```
> i3status-rs  ~/.config/i3status-rust/config.toml
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Syntax(
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regex parse error:
    ^:(\d{1})$
       ^^
error: Unicode-aware Perl class not found (make sure the unicode-perl feature is enabled)
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
)', src/libcore/result.rs:999:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
[⏎     
```

### DIFF
--- a/src/blocks/ibus.rs
+++ b/src/blocks/ibus.rs
@@ -182,7 +182,7 @@ fn get_ibus_address() -> Result<String> {
         // Hence on sway you will need to reload the bar once after login to get the block to work.
         let display_var = env::var("DISPLAY")
             .block_error("ibus", "$DISPLAY not set. Try restarting bar if on sway")?;
-        let re = Regex::new(r"^:(\d{1})$").unwrap(); // Valid regex is safe to unwrap.
+        let re = Regex::new(r"^:([0-9]{1})$").unwrap(); // Valid regex is safe to unwrap.
         let cap = re
             .captures(&display_var)
             .block_error("ibus", "Failed to extract display number from $DISPLAY")?;


atheriel:
Nice catch. Does this result in any changes to `Cargo.lock`?

ammgws:
Argh sorry I forgot to check up on this and I just saw a new release was cut.

atheriel:
There's no reason we couldn't release a quick patch version with this change, if that's what you're worried about.

atheriel:
Is it possible to remove the need for the Perl-compatible regex? I think `\d` could be replaced with `[0-9]`, right?

ammgws:
Should be good to go now.

PR #448, opened 2019-10-10
Add support for kbdd (keyboard library for per-window keyboard layout)

sashomasho:
A new "driver" option is added: kbddbus, if kbdd (https://github.com/qnikst/kbdd, also in AUR and Debian) is running, the new keyboard layout monitor will observe dbus for keyboard layout changes, which can be triggered by the user or by changing the current window focus (then kbdd will restore the previous layout for this window).

kbdd provides only layout names, like "English (US)", not layout symbols like "us", which consume a lot of space, so the value provided by it is used only if neither setxkbmap/xset combo is found on the system, nor xkblayout-state (which is not typically installed by default, but probably is the fastest way to
obtain the current layout).

I'm fairly new to rust, but I like it, and i3status-rs seems to be my favorite, so I decided to test myself with this change, because I badly needed it :) Any criticism or feedback will be highly appreciated.

### DIFF
--- a/src/blocks/keyboard_layout.rs
+++ b/src/blocks/keyboard_layout.rs
@@ -1,10 +1,13 @@
-use std::process::Command;
+use std::process::{Command, Stdio};
+use std::sync::{Arc, Mutex};
 use std::thread;
 use std::time::{Duration, Instant};
 
 use crossbeam_channel::Sender;
 use dbus;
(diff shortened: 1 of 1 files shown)


sashomasho:
I've found some issues on a different distribution than mine, so I need to take a second look at it before submitting it again.

PR #449, opened 2019-10-11
fix builds on armv6

jcgruenhage:
You can see in this CI job how it failed before: https://travis-ci.org/void-linux/void-packages/jobs/595561765

### DIFF
--- a/src/blocks/disk_space.rs
+++ b/src/blocks/disk_space.rs
@@ -189,16 +189,16 @@ impl Block for DiskSpace {
         let mut result;
         let mut converted = 0.0f64;
         let mut converted_str = String::new();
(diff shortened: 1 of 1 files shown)


atheriel:
Surprising that this wasn't caught/fixed by #406.

jcgruenhage:
> Surprising that this wasn't caught/fixed by #406.

Was surprised too. This fix already existed as a patch against the version from before #406 in the void repo.. Seems like @Gottox just accidentally didn't submit the whole patch to upstream

jcgruenhage:
well... shit. This broke more stuff than it fixed, because `statvfs.blocks()` also returns different data types depending on the architecture. armv7 and i686 are broken now, will send a PR for fixing that soon. Details over at https://travis-ci.org/void-linux/void-packages/builds/596540496

PR #450, opened 2019-10-12
fix builds on armv7 and i686

jcgruenhage:
These were broken by #449, which naively fixed armv6 builds.
This commit should now work on all platforms.
It wraps all return values of statvfs from the nix crate in
`as u64`, which means that the platform specific data type
(which might be u32 instead) is always converted to u64 when
necessary. Before this, only values which were known to cause
an issue were converted.

### DIFF
--- a/src/blocks/disk_space.rs
+++ b/src/blocks/disk_space.rs
@@ -189,16 +189,16 @@ impl Block for DiskSpace {
         let mut result;
         let mut converted = 0.0f64;
         let mut converted_str = String::new();
-        let total = (statvfs.blocks() * (statvfs.fragment_size() as u64)) as u64;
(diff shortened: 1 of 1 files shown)


jcgruenhage:
CI with that a patched disk_space.rs is running at https://travis-ci.org/void-linux/void-packages/builds/596963981

atheriel:
Thanks for all your work on this. Do you think it would benefit Void to release a patch version with these fixes? I am happy to do so.

jcgruenhage:
Not really. The version shipped in void s already patched like this anyway

jcgruenhage:
ftr: https://github.com/jcgruenhage/void-packages/blob/03b322f00fefb11b5045efb2587a7a634c18b1cd/srcpkgs/i3status-rust/patches/fix-32bit.patch is the patch

PR #451, opened 2019-10-12
Keyboard layout kbdd - per window keyboard layout

sashomasho:
A new "driver" option is added: kbddbus, if kbdd (https://github.com/qnikst/kbdd, also in AUR and Debian) is running, the new keyboard layout driver will observe dbus for keyboard layout changes (which can be triggered by the user or by changing the current window focus upon which kbdd will restore the previous layout for the new window).

kbdd provides layout as integer through dbus which correspond to one in the output from setxkmap -query, so after the index is obiained the current layout is extracted from setxkbmap.

Test/verify:
in .i3/config:
exec_always setxkbmap -layout us,de -option "grp:alt_shift_toggle"
exec_always pkill kbdd; kbdd

modify keyboard_layout block configuration with:
driver = "kbddbus"

I'm fairly new to rust, but I like it, and i3status-rs seems to be my favorite, so I decided to test myself with this feature, mainly because I badly needed it :) Any criticism or feedback will be highly appreciated.

### DIFF
--- a/src/blocks/keyboard_layout.rs
+++ b/src/blocks/keyboard_layout.rs
@@ -251,7 +251,7 @@ impl KeyboardLayoutMonitor for KbdDaemonBus {
     }
 }
 
-struct KbddMessageHandler(Arc<Mutex<Option<String>>>);
+struct KbddMessageHandler(Arc<Mutex<u32>>);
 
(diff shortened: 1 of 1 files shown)


atheriel:
This looks good to me. Thanks!

PR #453, opened 2019-10-20
Add Pomodoro timer block

ghedamat:
Hey there,

I decided to learn me some rust and added a "pomodoro timer" block

I'm pretty sure my rust code is probably not good enough to be merged in the mainline branch but I thought I'd open a PR  and if you are interested in giving me some feedback I can work on this a bit more so that others can use it

If this is of no interest feel free to close the PR immediately, just wanted to ask :)

Thanks

## What it does

Adds a pomodoro block with 5 config options
```
[[block]]
block = "pomodoro"
length = 25
break_length = 5
message = "pomodoro done"
break_message = "back to work!"
```

On left click the timer goes from stop to started and if clicked again it pauses

if it completes it starts the break phase, here left click starts the new "pomodoro" phase

on right click it always resets to stopped and resets the total pomodoro count

## Things to fix

* commit history
* in general no idea if this is decent rust code


ghedamat:
@greshake not sure if you can see drafts pr (never used one :D)

atheriel:
I think this could benefit from taking a simpler approach. Instead of using this complex (and hard to understand) state machine and the `machine` package, I would suggest a simpler implementation:

1. Use an enum to represent the state.
2. Store the current state and the timestamp of the last state change on the `Pomodoro` struct.
3. Each time the block is updated, check how much time has elapsed since the last state change and potentially transition to a new state.

Does that make sense to you?

I have left some additional comments in the code itself. Also:

- Don't worry about the Git stuff. We can squash it when it's ready.
- I would suggest calling the enum variant `OnBreak`.

ghedamat:
@atheriel it was great to learn about state machines in rust but this is much simpler indeed :) thanks for taking me off the ledge!

Let me know if there are more things I should improve and thanks again for the help!

shell.nix
```diff
@@ -0,0 +1,31 @@
...
+      curl
+      dbus
+      libpulseaudio
+    ];
+
+    shellHook = ''
+        export OPENSSL_DIR="${openssl.dev}"
+        export OPENSSL_LIB_DIR="${openssl.out}/lib"
+    '';
+  }
```
ghedamat: If possible I would like this to be added,
makes getting up and running with development so much easier (happy to also add a README.md section about how to use this if you like)

src/blocks/pomodoro.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    fn update(&mut self) -> Result<Option<Duration>> {
+        self.tick();
+        self.set_text();
+
+        match &self.state {
+            State::Started => {
+                if self.elapsed >= self.length {
+                    let message = self.message.to_owned();
```
ghedamat: there must be a better way to do this that avoids using `move` and instead uses lifetimes, but this seemed ok?

shell.nix
atheriel: There might be a case for adding it -- although I suspect very few contributors are running NixOs -- but it is absolutely out of scope for this PR.

src/blocks/pomodoro.rs
atheriel: Once you remove the thread spawn this should resolve itself; you can just borrow `self.message`.

shell.nix
ghedamat: you don't need NixOS at all to use this :) 
works on any linux and OSX with a user-space install

but noted. will remove

src/blocks/pomodoro.rs
```diff
@@ -0,0 +1,338 @@
...
+        let id: String = Uuid::new_v4().simple().to_string();
+        let id_copy = id.clone();
+
+        Ok(Pomodoro {
+            id: id_copy,
+            time: TextWidget::new(config).with_icon("pomodoro"),
+            state: State::stopped("\u{25a0} 0:00".to_string()),
+            length: block_config.length * 60,             // convert to minutes
+            break_length: block_config.break_length * 60, // convert to minutes
+            update_interval: Duration::from_millis(1000),
```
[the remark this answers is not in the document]
ghedamat: a timer that lasts 25 minutes that doesn't update every second didn't seem a requirement. but if you think it is I can do the approach you suggested and compare times instead. it will just be a bunch more code I think

src/blocks/pomodoro.rs
ghedamat: but I will need `thread.spawn` to be able to use an external command and don't wait for it to return whether it's nagbar or something else 

am I missing something?

src/blocks/pomodoro.rs
```diff
@@ -0,0 +1,338 @@
...
+                            }
+                        });
+
+                        self.state = self.state.on_break(Break);
+                        self.count = self.count + 1;
+                    }
+                }
+                State::Breaking(_state) => {
+                    if seconds >= &self.break_length {
+                        std::thread::spawn(|| -> Result<()> {
```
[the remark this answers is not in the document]
ghedamat: understood (this comment was not showing in the "files" view, thanks!

ghedamat:
## TODO

- [ ] remove `shell.nix`
- [ ] remove use of `thread.spawn` 
- [ ] make usage of `i3-nagbar` optional
- [ ] allow custom update intervals (rework all code to not use `tick()`

/cc @atheriel will ping once I have the above working (or at least some of it) thanks again!

src/blocks/pomodoro.rs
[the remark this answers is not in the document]
ghedamat: ok, sorry for hammering on this point but

I don't see anything here
https://doc.rust-lang.org/std/process/struct.Command.html

that allows me to NOT deal with the return value and have a "fire" and forget version of this.

`i3-nagbar` "returns" only when the user clicks on X on the bar, so the command keeps running until then

when the command returns it gives an output similar to

```
libi3] libi3/font.c Using Pango font monospace, size 8
button pressed on x = 1911, y = 17
button released on x = 1911, y = 17
```
I'm ignoring the output in my code 
but for some reason that crashes the main thread and triggers and invalid JSON format error, 

I saw the `spawn` pattern used a lot in the rest of i3status-rust when it comes to deal with external commands so I assumed that was the way to go.

do you have tips on where to look for debugging information if you want me to make this change?

Thanks again!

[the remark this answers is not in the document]
ghedamat: the error specifically is from i3

`Error:: Could not parse JSON ...`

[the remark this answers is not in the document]
ghedamat: nevermind, the issues was due to the fact that `spawn` inherits the `stdin` and `stdout` from the parent process so when `i3-nagbar` returns it breaks the output

current fix is in this commit

https://github.com/greshake/i3status-rust/pull/453/commits/a8238ff7e100221d4281dd13fe93958ce7027224

I'll mark this as resolved but expect you to comment further once the pr is in review again

thanks for helping me rubber duck the issue :)

src/blocks/pomodoro.rs
```diff
@@ -126,15 +126,7 @@ impl Block for Pomodoro {
...
-                    std::thread::spawn(move || -> Result<()> {
-                        match Command::new("i3-nagbar").args(&["-m", &message]).output() {
-                            Ok(_raw_output) => Ok(()),
-                            Err(_) => {
-                                // We don't want the bar to crash if i3-nagbar fails
-                                Ok(())
-                            }
-                        }
-                    });
+                    Command::new("i3-nagbar").stdout(Stdio::null()).args(&["-m", &message]).spawn().expect("Failed to start i3-nagbar");
```
ghedamat: syntax taken from
https://doc.rust-lang.org/std/process/struct.Command.html#method.stdout

also note that this and the block below align differently because of the `rustfmt` rules of the project

src/blocks/pomodoro.rs
```diff
@@ -126,15 +126,11 @@ impl Block for Pomodoro {
...
-                        match Command::new("i3-nagbar").args(&["-m", &message]).output() {
-                            Ok(_raw_output) => Ok(()),
-                            Err(_) => {
-                                // We don't want the bar to crash if i3-nagbar fails
-                                Ok(())
-                            }
-                        }
-                    });
+                    Command::new("i3-nagbar")
+                        .stdout(Stdio::null())
```
ghedamat: syntax taken from
https://doc.rust-lang.org/std/process/struct.Command.html#method.stdout

ghedamat:
@atheriel made all the changes you requested but not the one about custom interval.

Doing the math around updates is a bit of a pain because I really wanna support pausing of the timer so storing the start time is not sufficient.

It's doable of course and maybe just I'm being lazy but I don't really see the value of adding that complexity. is it that resource intensive to update every second?

Let me know what you think and feel free to also let me know if the other fixes are good for you


thanks again!

ghedamat:
also @atheriel question, `cargofmt` is failing on CI but I don't understand

locally for me all the files in the project BUT my file are not respecting the formatting. am I missing something?  

thanks!

ghedamat:
Nevermind, found that you ended up doing a big commit upstream. rebased :)

all good just need answer to 

https://github.com/greshake/i3status-rust/pull/453#issuecomment-546623466

to move forward

thanks!

ghedamat:
@atheriel not sure if you saw my last comment, and sorry for nagging

do you think this is good enough for merge or you really want the ability to update with a different interval then 1 second? (see https://github.com/greshake/i3status-rust/pull/453#issuecomment-546623466 for context)

thanks!

atheriel:
No, I don't think that is required for the initial implementation. Thanks for your patience.

PR #454, opened 2019-10-20
separate command to get state for custom block

jcgruenhage:
This is an alternative to \#414. I implemented it without knowing \#414
existed. But since it already existed and it's not merged yet,
I'm pushing this as a simpler alternative.

### DIFF
--- a/src/blocks/custom.rs
+++ b/src/blocks/custom.rs
@@ -113,8 +113,7 @@ impl Block for Custom {
                 .args(&["-c", &state_command.clone()])
                 .output()
                 .map(|o| String::from_utf8_lossy(&o.stdout).trim().to_owned())
-                .unwrap_or_else(|e| e.description().to_owned())
-                .try_into()
+                .map(|o| o.try_into().unwrap_or(State::Critical))
                 .unwrap_or(State::Critical);
 
             self.output.set_state(state);
--- a/src/blocks/custom.rs
+++ b/src/blocks/custom.rs
@@ -3,6 +3,7 @@ use std::process::Command;
 use std::iter::{Cycle, Peekable};
 use std::vec;
 use std::env;
+use std::convert::TryInto;
 use crossbeam_channel::Sender;
 
 use crate::blocks::{Block, ConfigBlock};
@@ -11,6 +12,7 @@ use crate::de::deserialize_duration;
 use crate::errors::*;
 use crate::widgets::button::ButtonWidget;
 use crate::widget::I3BarWidget;
+use crate::widget::State;
 use crate::input::I3BarEvent;
 use crate::scheduler::Task;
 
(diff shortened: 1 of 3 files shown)


src/blocks/custom.rs
```diff
@@ -97,6 +108,18 @@ impl Block for Custom {
 
         self.output.set_text(output);
 
+        if let Some(state_command) = self.state_command.clone() {
+            let state = Command::new(env::var("SHELL").unwrap_or("sh".to_owned()))
+                .args(&["-c", &state_command.clone()])
+                .output()
+                .map(|o| String::from_utf8_lossy(&o.stdout).trim().to_owned())
+                .unwrap_or_else(|e| e.description().to_owned())
```
atheriel: I'm not sure I understand what is happening here. My reading is: a UTF-8 parsing error will yield a string describing the error, which then generates an error in `try_into()` that suppresses this description, which then itself is suppressed by the `unwrap_or()`. If that is the case, what is the point of all the error handling?

src/blocks/custom.rs
```diff
@@ -97,6 +108,18 @@ impl Block for Custom {
 
         self.output.set_text(output);
 
+        if let Some(state_command) = self.state_command.clone() {
+            let state = Command::new(env::var("SHELL").unwrap_or("sh".to_owned()))
+                .args(&["-c", &state_command.clone()])
+                .output()
```
atheriel: What happens if the command fails?

src/blocks/custom.rs
jcgruenhage: `from_utf8_lossy` will not return an error. Bytes that are not valid utf8 will just be replaced with �. The error conversion in unwrap_or_else is just so that we get back a string in any case, whether the command execution worked or not. We then take that string (which is either what the command returns, or an error message why that command failed) and try to create a state from that. If that fails, we assume "critical". We could save a bunch of code here with `map_or_else`, meaning we take the output and convert it to a state (and if that fails assume "critical") and in the same step convert errors from the command execution and map them to "critical".

jcgruenhage: tbh, I'm a bit confused right now, [the docs](https://doc.rust-lang.org/std/result/enum.Result.html#method.map_or_else) describes map_or_else as a nightly only API with 2 parameters, and we use it with one parameter on stable...

src/blocks/custom.rs
jcgruenhage: Described it in https://github.com/greshake/i3status-rust/pull/454#discussion_r336893795

jcgruenhage:
I'm thinking about implementing this usecase again in a separate PR, with a separate block, that has a command that returns a JSON blob that contains the text, icon and state for the text widget. Out of those 3 designs for the usecase, which one do you (@atheriel) think is best?

jcgruenhage:
any feedback on this? @atheriel maybe?

jcgruenhage:
Superseded by #515

PR #458, opened 2019-10-29
strip keyboard variant from setxkbmap output to preserve space

sashomasho:
I'm mostly using i3status-rust on my laptop and I've noticed that after undocking the laptop sometimes (not always) the output of setxkmap is changed from:
>...
> layout:     us,bg
> variant:    ,bas_phonetic
> 

to
>...
> layout:     us,bg:bas_phonetic

and the width of the indicator becomes way too big when switching between the layouts because it includes and the variant, so I've stripped that from the output, since originally it was not supposed to be there.


atheriel:
Thanks!

Henriquelay:
Any way to get this to show again? I switch variants but not the layout. In my case br-qwerty and br-dvorak

PR #464, opened 2019-11-10
Add Fedora installation info

tim77:
Please add Fedora installation info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,9 @@ Optional:
 
 We strongly recommend using the development version. Older versions are not supported at this time.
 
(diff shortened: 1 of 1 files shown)


atheriel:
Awesome, thanks! Quick question: is this a "blessed" package that is part of Fedora itself, or is it hosted by a third party (similar to a PPA for Ubuntu, if you are familiar with that).

tim77:
This is similar to PPA and somewhat AUR in Arch and generally not recommended using random COPR's. But you can check is owner of COPR maintainer in Fedora or not so this at least add some scores to trust (i hope). But good news: i am started packaging `i3status-rust` for official repos as well so i hope we will see it soon in official repos.

atheriel:
Oh I see, this is your work. Thanks!

I'm curious as to how you went about building the `.rpm`, if you'd care to share.

Issue #468, opened 2019-11-20
Icon spacing too big / inconsitent


Rikorose:
The icon spacings in different widgets are very inconsistent, as it can be observed in the attached image. I.e. the spacings before and between the music control buttons are very large, also the space between the wifi symbol and the ssid ('eduroam' in my case). 

![snippet](https://user-images.githubusercontent.com/16517898/69222866-387b3f80-0b72-11ea-801f-163c9288e267.png)

Also, all icons should consume the same space. This prevents 'jumping' if an icon changes (e.g. play icon changes to pause icon).

Edit: Added test configuration and output of i3status-rs.
As you can see in the output, the icons also have inconsistent padding with spaces on the left and right side.
Edit2: Updated output, since previously the wifi ssid was not loaded yet

`test.toml`:
```toml
[[block]]
block = "music"
buttons = ["prev", "play", "next"]

[[block]]
block = "net"
device = "wlp61s0"
hide_inactive = true
ssid = true
signal_strength = true
ip = false
speed_up = false
speed_down = false
graph_up = false
interval = 5
```

`$ i3status-rs test.toml` output:
```json
[...]

PR #470, opened 2019-11-21
Fix spaces around play/pause icons

ammgws:
As per discussion in #468.


atheriel:
As per that discussion, I think this is only the first step. But it's clearly in the right direction.

PR #473, opened 2019-11-24
Allow custom formatting of the pacman block

ifreund:
This PR expands the customization potential for the pacman block by allowing specification of custom format strings.  See the updated documentation in `blocks.md` for details

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -554,6 +554,7 @@ Update the list of pending updates every ten seconds:
 block = "pacman"
 interval = 10
 format = "{count} updates available"
+format_singular = "{count} update available"
 format_up_to_date = "system up to date"
 ```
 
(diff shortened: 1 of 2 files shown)


atheriel:
What do you imagine this might be used for?

ifreund:
I'm using the config I used in the example for the docs right now.  Maybe I'm weird, but I prefer a nice message assuring me that things are up to date rather than a 0 or no message at all.

My current layout for reference:
image
It's a bit more bare bones than the default but really appeals to me for whatever reason.  There's a bit of funkiness going on with the spacing around some of the blocks though which I mean to track down soon.

I suppose I could have just started a fork and hardcoded in my desired changes, but there's relatively little overhead required to support these options, and I suspect that others may also make use of them.  Basically I just wanted to add support for further customization of this great program in the spirit of giving users complete and easily configurable control over all aspects of the bar.

ammgws:
I would use this if it got merged.

@atheriel LGTM?

PR #476, opened 2019-12-02
Add missing documentation for hide_missing and hide_inactive to net

bascht:
Just noticed that #344 adds an option I was searching for and couldn't find in the docs. :smiley_cat: 

Copied over from src/blocks/net.rs


atheriel:
Thanks!

PR #479, opened 2019-12-07
battery: Add configurable thresholds

Thra11:
Add critical and warning thresholds to the block configuration. Use
State::Idle when above the warning threshold.

The battery block is unlike most other blocks, in that it doesn't use the Idle state when it has nothing interesting to say, meaning it stands out when it shouldn't. The way I see it, the current battery capacity is uninteresting for most of the discharge cycle, and the block should only draw attention when it is starting to get low (warning) or close to empty (critical). I have chosen 30% and 10% as the default thresholds, but I am open to suggestions. The option names, `warning` and `critical` are inspired by the CPU and similar blocks.

I'm somewhat new to rust, so let me know if my proposed changes are unidiomatic or simply incorrect.

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -376,6 +376,8 @@ pub struct Battery {
     device: Box<dyn BatteryDevice>,
     format: FormatTemplate,
     driver: BatteryDriver,
+    critical: u64,
+    warning: u64,
 }
 
 #[derive(Deserialize, Debug, Clone)]
@@ -422,6 +424,14 @@ pub struct BatteryConfig {
 
     /// The "driver" to use for powering the block. One of "sysfs" or "upower".
     pub driver: Option<BatteryDriver>,
+
+    /// The threshold below which the remaining capacity is shown as critical
+    #[serde(default = "BatteryConfig::default_critical")]
+    pub critical: u64,
+
+    /// The threshold below which the remaining capacity is shown as warning
+    #[serde(default = "BatteryConfig::default_warning")]
+    pub warning: u64,
 }
 
 impl BatteryConfig {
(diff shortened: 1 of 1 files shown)


Thra11:
I suppose this could be somewhat controversial as it changes the behaviour for existing users and doesn't offer any way to to get back the previous behaviour. I'd be interested to hear if anyone actually likes the previous/current behaviour, as it doesn't make sense to me.

GladOSkar:
Shouldn't it be possible to have the default configuration if none of the thresholds are set? I do think we shouldn't break existing installations.

Regarding the current behaviour i never had any complaints

Thra11:
> Shouldn't it be possible to have the default configuration if none of the thresholds are set? I do think we shouldn't break existing installations.

If none of the thresholds are set, it will simply use the default threshold values. The issue is that I've changed the states which it goes through as it discharges from this:
```Good -> Info -> Warning -> Critical```
to this:
```Idle -> Warning -> Critical```
I did this because while the words "Good" and "Info" sound ok, the way most of the themes display them, "Good" is very distracting, in the style of "Look! Look! Something good has happened!" rather than "Everything's good, nothing to report".

In theory, we could try making it even more configurable, so that the config provides a list of pairs containing a threshold and state (Think in TOML it would be a table?). This would allow both the current behaviour and my preferred behaviour, but at the cost of complexity.

**Comment by GladOSkar** [comment]
Why not have `good` and `info` thresholds as well and then the default would be:

```
good = 100
info = 60
warning = 30
critical = 15
```

and your configuration would be

```
good = 30
info = 30
warning = 30
critical = 15
```

This would require the thresholds to be checked in the right order in order for the most critical state to be choosen for the value but that should be simple to implement and IMO fairly elegant.

Thra11:
I've implemented something like your suggestion as a separate pull request (#480) The only real difference to your suggestion is that in the default config, the idle state is "hidden" between good and info, not above good. I think it makes more sense that way.

ammgws:
Superseded by #480

PR #480, opened 2019-12-10
battery: Add configurable thresholds

Thra11:
Alternative implementation following discussion on https://github.com/greshake/i3status-rust/pull/479:

Add critical, warning, info and good thresholds to the block configuration. The order of states as the battery discharges is as follows:
Good -> Idle -> Info -> Warning -> Critical
This slightly overcomplicated choice of states allows the default configuration to remain as it was before (with the Idle state non-existant between 60 & 60), while also allowing a more conventional battery status indicator that doesn't draw attention except at low and critically low levels (Setting info == warning and good == 100 effectively disables these states).

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -376,6 +376,10 @@ pub struct Battery {
     device: Box<dyn BatteryDevice>,
     format: FormatTemplate,
     driver: BatteryDriver,
+    good: u64,
+    info: u64,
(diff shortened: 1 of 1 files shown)


karolchmist:
I'd be very interested in this feature. I use TLP for fixing the battery level at 60%, which means that currently it's always highlighted.

ammgws:
Seems OK to me

PR #481, opened 2019-12-10
Update blocks.md

hellow554:
change udev rule, some tt changes


atheriel:
The formatting changes look good for the most part, but are you sure the `udev` rule is wholly equivalent?

hellow554:
I did it this way on my arch and yes it's working. I wanted to update the arch wiki as well, but I'm too lazy to open up an account for that. 
It's not 100% equivalent, because if you have a rule that gives also every user permission to write to that file (`o+w`), then this udev rule would overwrite it, but **if** every user would have access, why bothering with this udev rule, right? 
Default for the `brightness` device is `644`, adding `g+w` makes it `664`.

PR #483, opened 2019-12-12
chore: bump deps

tim77:


ammgws:
What fixes/etc do we get from these specific ones?

tim77:
I am trying to package i3status-rust for official Fedora repos and this one of the requirements
https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/#_versions

It compiles fine when i tested this patch on local machine but not investigated yet why CI build failed.

ammgws:
Might be failing because you didn't commit the changes to `Cargo.lock`?

atheriel:
Ah, that's an unfortunate policy conflict with Fedora. I would suggest maintaining a patch as part of the package in the short term, since we are reticent to update dependencies unless it provides an actual benefit.

tim77:
Yep, i already patched and made one compat package already so `i3status-rust` will be very soon in official repos. So i can close this PR, or update `Cargo.lock` and this should pass CI. As you wish.

ignatenkobrain:
> Ah, that's an unfortunate policy conflict with Fedora. I would suggest maintaining a patch as part of the package in the short term, since we are reticent to update dependencies unless it provides an actual benefit.

So you don't consider being in sync with distirbution packages as a benefit? You prefer not collaborating with those maintainers? What will happen when you will receive some bugreport from Fedora users for example?

ignatenkobrain:
Also you never know what bugs people have fixed in their versions. And you should simply follow dependencies release. IF they add unwanted dependency, you can always raise the question. For example, I noticed that @BurntSushi is doing that.

ammgws:
Personally I think that if it doesn't add any new dependencies and the changes are tested by the submitting the PR then we should be open to those.

ammgws:
@atheriel This can be closed now :)

PR #488, opened 2019-12-16
Update Fedora installation instructions

tim77:
Packaged for [official repos](https://src.fedoraproject.org/rpms/rust-i3status-rs). Now available in Rawhide and soon will be for F31 release.


atheriel:
Wow, that's excellent news. Thanks!

PR #489, opened 2019-12-17
chore: Update multiple dependencies

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,10 +2,10 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.6.4"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


ignatenkobrain:
This is alternative to #483

ignatenkobrain:
Even though there are some new deps, overall this reduces dependency list by 2 crates :)

ammgws:
Dependency graphs

Before
before

After
after

ignatenkobrain:
@ammgws is that good or bad? Sorry, not very clear to me :)

tim77:
Just want to add my 2 cents: binary size with new deps decreased - 6,7M vs 6,3M. But probable this is not clear comparison  because this is native Fedora binary vs upstream cargo.

ignatenkobrain:
Strictly speaking, depending on compilation options it is different:

```
⋊> ~/t/i3status-rust on master ↑ du -h target{-before,}/release/i3status-rs{,.stripped}                                                                                     18:38:18
45M	target-before/release/i3status-rs
52M	target/release/i3status-rs
5.5M	target-before/release/i3status-rs.stripped
6.2M	target/release/i3status-rs.stripped
```

I have compiled it before and after with `-Ccodegen-units=1 -Copt-level=3 -Cdebuginfo=2 -Clink-arg=-Wl,-z,relro,-z,now`

ammgws:
@ignatenkobrain just to try visualise the changes and see where the reductions occurred

atheriel:
I believe @ammgws has done quite a lot of related work on updating dependencies to reduce transitive deps, so I'm not sure this is relevant anymore (and would need to be rebased in any case).

PR #491, opened 2019-12-18
Remove 6 optional dependencies used by `clap`

ammgws:
Disable optional dependencies allows us to get rid of 6 dependencies from our project:
- ansi_term
- atty
- strsim
- vec_map
- redox_termios
- termion

Reason we can disable these optional dependencies:
- "suggestions": Turns on the Did you mean '--myoption'? feature for when users make typos.
 **Reason**: Not needed for our case as `i3status-rs` isn't a CLI app that we are manually invoking all the time. It's set once in our i3/sway config and left alone. Also, our only CLI flags are `exit-on-error`, `help` and `version`.

- "color": Turns on colored error messages. This feature only works on non-Windows OSs. 
  **Reason**: As per above.

- "vec_map": Use VecMap internally instead of a BTreeMap. This feature provides a slight performance improvement.
  **Reason**: Performance seems about the same (which makes sense given what our app does...)

Dependency graphs:
Before
before

After
aft

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,24 +9,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "winapi 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
-name = "atty"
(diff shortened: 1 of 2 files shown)


AzeemJiva:
Can we get this merged?

ignatenkobrain:
LGTM

atheriel:
Nice work!

PR #492, opened 2019-12-18
Remove 5 dependencies by replacing `num` with `num-traits`

ammgws:
Removes the following:
- num-bigint
- num-complex
- num-iter
- num-rational
- rustc-serialize

The only place we use the `num` crate is in the graph widget, however it only uses `clamp` and `ToPrimitive` which can be provided by just the `num-traits` subcrate.

Dependency graphs:
Before:
before

After:
aftx11

Future work: Remove this dependency all together as writing a clamp function is only a few lines, and can probably work around the need for ToPrimitive as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -420,32 +420,9 @@ name = "num"
 version = "0.1.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "num-bigint 0.1.43 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


AzeemJiva:
Can we get this merged?

ignatenkobrain:
@ammgws why don't you directly depend on num-* subcrates which implement what you need?

ignatenkobrain:
I have switched to num_traits in https://github.com/greshake/i3status-rust/pull/489, that has removed all aforementioned deps.

ammgws:
@ignatenkobrain an oversight on my behalf

PR #493, opened 2019-12-23
Remove 10 dependencies by bumping `chrono-tz` to 0.5

ammgws:
We use `chrono-tz`, and it has a dependency `parse-zoneinfo` which just had a PR merged that disables unused regex features (like what we did for our regex dep a while back). This results in a reduction of 10 dependencies for us :)

Before
before

After
after


atheriel:
Needs a rebase but otherwise LGTM.

ammgws:
@atheriel Done

PR #497, opened 2019-12-30
Mention that the focused window block works with sway

NilsIrl:


atheriel:
I appreciate trying to improve the docs, but isn't this due to sway's 100% compatibility with i3's IPC? I don't think we do anything special on our end to support it.

NilsIrl:
Yes, it's just that someone who reads the docs might not be aware of that. I was actually going to post an issue asking for support of this and then thought better of trying first and I was right because it did work so I think it is best to make it explicit.

I added a commit to mention that as well.

PR #499, opened 2020-01-02
set backlight brightness via DBus when sysfs is inaccessible

majewsky:
After not getting sysfs access to work, even with the udev rule mentioned in the documentation, I investigated how other applications/desktops are controlling the backlight brightness.

As it turns out, on systems governed by systemd-logind (or elogind), the correct way [according to the systemd developers](https://github.com/systemd/systemd/issues/14463#issuecomment-570190340) is the DBus interface of logind.

This commit uses this method as a fallback for when we cannot open the sysfs file. This works great on my system, but I'm not particularly attached to this method if you find it too surprising. If you'd rather like a configuration option to switch between two methods, I can implement that as well.

### DIFF
--- a/src/blocks/backlight.rs
+++ b/src/blocks/backlight.rs
@@ -110,6 +110,12 @@ impl BacklitDevice {
 
     /// Set the brightness value for this backlit device, as a percent.
     pub fn set_brightness(&self, value: u64) -> Result<()> {
+        let safe_value = match value {
+            0..=100 => value,
+            _ => 100,
+        };
+        let raw = (((safe_value as f64) / 100.0) * (self.max_brightness as f64)).round() as u64;
+
         let file = OpenOptions::new()
             .write(true)
             .open(self.device_path.join("brightness"));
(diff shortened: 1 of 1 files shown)


atheriel:
This is a neat idea.

atheriel:
I'm a bit confused about what's going on here, though, and I have almost no understanding of what is being discussed in your linked systemd issue, which makes it hard to assess whether this will be useful to others.

Can you explain why systemd prevents you from modifying a sysfs file? I believe that udev rule was cribbed from [the Arch wiki](https://wiki.archlinux.org/index.php/Backlight) at some point, although I can't find it now.

majewsky:
I applied the udev rule on my system (Arch Linux with systemd, SDDM and Sway) and it did nothing. The relevant files in sysfs are still `root:root`. But anyway, setting the backlight brightness should not require messing with udev rules. Everyone else (Plasma, Gnome, etc.) can do it without messing with udev rules. And that's because they go through the systemd-logind DBus interface.

atheriel:
> Everyone else (Plasma, Gnome, etc.) can do it without messing with udev rules. And that's because they go through the systemd-logind DBus interface.

That makes sense to me. Incidentally, I realize that #481 recently changed the `udev` rule in the documentation -- does using the old rule fix your issue?

majewsky:
> Incidentally, I realize that #481 recently changed the `udev` rule in the documentation -- does using the old rule fix your issue?

I just tried, and it doesn't work for me either.

atheriel:
> I just tried, and it doesn't work for me either.

It was a long shot. I'm happy to merge this provided you fix the issues surfaced by CI.

majewsky:
CI looks happy now.

PR #502, opened 2020-01-11
Add Watson block

barskern:
This adds a simple block to display the current time tracking status for [watson](http://tailordev.github.io/Watson/).

The module does not add any dependencies and is rather simplistic.


ammgws:
@atheriel LGTM, no new deps either which is good.

PR #506, opened 2020-01-16
Remove superfluous docstring (same info in blocks.md)

ammgws:
Info is already in blocks.md, and it doesn't provide any extra information that would be useful for devs. Also seems outdated as it refers to a "config.json" file.


atheriel:
I would ideally like to keep some API-level documentation for developers in here, but agree that in this case the information is quite outdated.

PR #508, opened 2020-01-16
added mapped-volume option for correct volume

### LINKED ISSUE #422: blocks/sound: volume differs
On my machine the volume differs from alsamixer and volume applet (also won't go down to 0%) as discussed here: https://bbs.archlinux.org/viewtopic.php?pid=1114261#p1114261.
To get the correct volume i needed to use the `mapped-volume` option of amixer: https://github.com/rsteube/i3status-rust/commit/f56d6a15eb24c85a907c366fd8f25267351aae1e

Maybe add an option to activate this?

rsteube:
fixes #422

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -85,7 +85,7 @@ impl SoundDevice for AlsaSoundDevice {
 
     fn get_info(&mut self) -> Result<()> {
         let output = Command::new("amixer")
-            .args(&["get", &self.name])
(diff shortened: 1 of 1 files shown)


ohk2kt3t4:
@atheriel LGTM since this [was also an issue in i3blocks](https://github.com/vivien/i3blocks/pull/254), and the solution was the same.

ammgws:
Thanks!

PR #511, opened 2020-01-17
Chore: Order deps in alphabetical order

ammgws:
Since have been working with dependencies a lot lately this had been getting on my mind.

Had a gander at some other crates like `tokio` and they do this there too.


ammgws:
https://github.com/rust-lang/rfcs/blob/master/style-guide/cargo.md

> Sort key names alphabetically within each section

PR #513, opened 2020-01-17
Use serde_json to get ip address instead of sed

ammgws:
This is possible as `ip` has a JSON output option.

TODO
- [ ] Check what our oldest supported system is

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -262,28 +262,28 @@ impl NetworkDevice {
         if !self.is_up()? {
             return Ok(None);
         }
-        let mut ip_output = Command::new("sh")
+        let output = Command::new("sh")
             .args(&[
                 "-c",
-                &format!(
-                    "ip -oneline -family inet address show {} | sed -rn -e \"s/.*inet ([\\.0-9/]+).*/\\1/; G; s/\\n/ /;h\" -e \"$ P;\"",
-                    self.device
-                ),
+                &format!("ip -json -family inet address show {}", self.device),
             ])
             .output()
-            .block_error("net", "Failed to execute IP address query.")?
-            .stdout;
+            .block_error("net", "Failed to execute IP address query.")
+            .and_then(|raw_output| {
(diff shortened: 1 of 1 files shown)


ammgws:
Hmm, the `-json` flag might not be available on some older systems:
https://lwn.net/Articles/738897/

flying7eleven:
Is there any commitment on compatibility with distributions? And if there is one, how many version / years should it be tested on? So far I didn't knew about a commitment and I would be interested so I could tests my own pull-request against it.

ammgws:
I don't think there's an official commitment, but Ubuntu 18.04LTS users seem to pop up quite often in my experience so from 18.04 onwards might be a good reference point in general.

flying7eleven:
Thanks @ammgws, I try to keep that in mind while testing :+1: For this pull request this would seem to be fine then I guess. Ubuntu 18.04 seems to ship with an compatible version (https://packages.ubuntu.com/bionic/iproute2).

ammgws:
Thanks for that, when I checked last time it seemed like they were still on an older version.

ammgws:
Hmm from a quick look I can't find any commits where something like that was fixed: https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/log/?qt=grep&q=json

Wonder how prevalent this is

ammgws:
Aside from reverting this commit we could:

1. Fallback to parsing `ip` output using `sed` if cannot parse address from JSON. However this would mean the `ip` command would be run twice, plus having to keep around the previous code. Not really something I want to do.

2. Check the first member of the JSON array, and if it is empty then try the next member. This will lead to some code duplication but it would solve this issue.

3. Do nothing and file bug with upstream (or git bisect to see when this bug gets fixed)

hlmtre:
Hey, thank you! I will have some time and energy this weekend and intend to tackle parsing the JSON, including the malformed one. I can submit a fix using the `ip -json`  version when complete.

Issue #130, opened 2018-01-02
awesome icons broken after ttf-font-awesome update on Arch


bluehood:
As per the pacman logs below, `ttf-font-awesome` was updated by pacman on the 29th of December,
which resulted in i3status-rust looking like this:

![image](https://user-images.githubusercontent.com/10999034/34487256-db2f7bcc-efd3-11e7-832a-33cbad77449b.png)

As you can see icons do not display properly anymore. removing `i3status-rust-git` and re-installing from AUR did not fix the issue.

Relevant snippet of pacman logs:
```
[2017-12-29 17:12] [PACMAN] Running 'pacman -Syu'                                                                       
[2017-12-29 17:12] [PACMAN] synchronizing package lists                                                                 
[2017-12-29 17:13] [PACMAN] starting full system upgrade                                                                
[2017-12-29 17:22] [ALPM] transaction started                                                                           
[2017-12-29 17:22] [ALPM] upgraded docker (1:17.11.0-1 -> 1:17.12.0-1)                                                  
[2017-12-29 17:22] [ALPM] upgraded mesa (17.3.0-2 -> 17.3.1-2)                                                          
[2017-12-29 17:22] [ALPM] upgraded firefox (57.0.2-1 -> 57.0.3-1)                                                       
[2017-12-29 17:22] [ALPM] upgraded lib32-mesa (17.3.0-2 -> 17.3.1-2)                                                    
[2017-12-29 17:22] [ALPM] upgraded numactl (2.0.11-1 -> 2.0.11-2)                                                       
[2017-12-29 17:22] [ALPM] upgraded ttf-font-awesome (4.7.0-5 -> 5.0.2-1)                                                
[2017-12-29 17:22] [ALPM] upgraded xorg-server-common (1.19.5-1 -> 1.19.6-2)                                            
[2017-12-29 17:22] [ALPM] upgraded xorg-server (1.19.5-1 -> 1.19.6-2)                                                   
[2017-12-29 17:22] [ALPM] transaction completed                                                                         
[2017-12-29 17:22] [ALPM] running 'fontconfig.hook'...                                                                  
[2017-12-29 17:22] [ALPM] running 'gtk-update-icon-cache.hook'...                                                       
[2017-12-29 17:22] [ALPM] running 'systemd-sysusers.hook'...                                                            
[2017-12-29 17:22] [ALPM] running 'systemd-update.hook'...                                                              
[...]

PR #517, opened 2020-01-20
Simplify volume_muted icons

carloabelli:
Maybe I'm missing something, but font awesome includes a volume mute icon so we should just use that instead of attempting to build it out of multiple icons. It is also possible to simplify the code around creating custom mute icons. Not sure if there is a default volume mute icon for material icons, but would be happy to amend.

### DIFF
--- a/src/icons.rs
+++ b/src/icons.rs
@@ -26,8 +26,7 @@ lazy_static! {
         "volume_full" => " VOL ",
         "volume_half" => " VOL ",
         "volume_empty" => " VOL ",
-        // This icon has no spaces around it because it is manually set as text. (sound.rs)
-        "volume_muted" => "MUTED",
+        "volume_muted" => " VOL MUTED ",
         "thermometer" => " TEMP ",
         "xrandr" => " SCREEN ",
         "net_up" => " UP ",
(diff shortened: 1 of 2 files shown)


carloabelli:
Let me know if there's anything I can do to help move this along

src/icons.rs
```diff
@@ -26,8 +26,7 @@ lazy_static! {
         "volume_full" => " VOL ",
         "volume_half" => " VOL ",
         "volume_empty" => " VOL ",
-        // This icon has no spaces around it because it is manually set as text. (sound.rs)
-        "volume_muted" => "MUTED",
+        "volume_muted" => " VOL MUTED ",
```
ammgws: Was VOL added just for consistency?

carloabelli: It was to keep the output consistent with the previous output.

ammgws: Ah, I see what's going on now. The output is still the same as before.

carloabelli:
Ah I see now that Font Awesome 4 is still being used (per #130). Though it does appear that the underlying pango issue causing many of the hex box problems has been fixed [here](https://gitlab.gnome.org/GNOME/pango/-/merge_requests/67). Perhaps it would be worth considering an update to the Font Awesome 5 at this point? I can either wait on this commit until a decision is made there, or I can revert it to the old glyph.

I am using `ttf-font-awesome 5.13.0-1`:

screenshot_2020-04-21T14:26:54,656963612-04:00

carloabelli:
Changes now included in #619.

PR #519, opened 2020-01-21
added notes about additional nixos packages

fkarg:


atheriel:
Normally I would expect this to be part of the package documentation on NixOS -- is there a reason that won't be the case here?

fkarg:
Not sure about that, since I'm pretty unfamiliar, but It took me some time to figure out, so I'd like to help others with similar needs

ammgws:
Perhaps they could also be mentioned or added as optional dependencies to [the NixOS package](https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/window-managers/i3/status-rust.nix)?

fkarg:
Well I have no clue on how to do that, alas I just wanted to help people in a situation similar to mine, through the way I knew. I wanted to look into this at some point, but I do not have time for this currently, so this might take a while.

atheriel:
I have filed an upstream bug, let's see how that goes.

ammgws:
https://github.com/NixOS/nixpkgs/issues/79140 for reference

ammgws:
Looks like this has been resolved upstream (the `sysfsutils` part anyway).

PR #521, opened 2020-01-23
Convert to new Cargo.lock V2 format

ammgws:
Ran `cargo lock translate` utility to translate our Cargo.lock files to the [new merge-friendly V2 lockfile format](https://github.com/rust-lang/cargo/pull/7070).


atheriel:
Can't get this tool to work for me, unfortunately. Feel free to do this against `master`.

yvt:
By any chance did you convert the lock file by `cargo lock translate > Cargo.lock`? This won't work because `Cargo.lock` gets overwritten before `cargo lock translate` runs.

ammgws:
I was on mobile and I think I did exactly that. Just redid it with an intermediate step so it should be good now.

PR #526, opened 2020-01-24
Update bare trait objects

ammgws:
A couple that were missed in 6b602784f8758631c274c1f7013189695ad5bcb7.

Is it trivial to get our Travis CI to test our optional features as well? It doesn't seem to be doing that at the moment, just the standard build with no options.


atheriel:
It should be trivial, yes. I believe there is a flag to cargo to build all features; we'd just have to ensure that Travis has the system dependencies for those blocks.

PR #528, opened 2020-01-25
added labels to bluetooth devices

jeffw387:
I have a couple different bluetooth devices that I connect to at different times, and I realized that it would be a lot easier to tell them apart with labels. 

It's optional, and if included it shows up to the right of the icon and left of the connected indicator.


jeffw387:
status-bar-bluetooth-labels
Here's my current bar with the labels and two bluetooth devices (work and rowkin).

atheriel:
Nice little feature, thanks!

Issue #486, opened 2019-12-15
[Feature Request] focused_window displays the current mark


travankor:
The `focused_window` block does not change when I send a `title_format` command to a window on sway (tested via `swaymsg` or the config file). The GUI header that sway keeps on top of the window does however change as expected.

`i3status-rs` sticks with `%title`.

ammgws:
Does it happen on `i3` too?

Looks like the `focused_window` block is using `name` to get the window title, but I checked the output of `swaymsg -t get_tree` and it doesn't look like this field is updated by sway (and i3?) when using title_format.

travankor:
I'm not sure since I have barely used i3. Although I think you're right, in which case, there is no bug in `i3status-rs`. (This sort of feels like a glaring design flaw in i3 tho.)

I guess I could work around this by using mark command. Perhaps `i3status-rs` can add a block that shows the mark of the focused window?

atheriel:
I'm open to expanding the functionality of the `focused_window` block in this manner.

ammgws:
Looks like marks aren't shown in the get_tree output so might not be trivial to implement.

Looks like can get marks using get_marks, though.

So add an option to the block to display the value of mark instead of title_format, and then when focus changes call get_tree and see if the container has a mark, and if so used that.

ohk2kt3t4:
@travankor How are you using marks to get around this?

travankor:
It's not really a replacement for updating title_format, but more of a separate request I guess.
Both the title and the mark are shown on the header bar, but I usually hide the header bar by default.

Feel free to close if it's out of scope for the project.

ammgws:
It might be worth considering if it's not too hard to implement. Could you please show a screenshot of the header bar when they are both showing? I don't use marks so not sure it looks like.

travankor:
>It might be worth considering if it's not too hard to implement. Could you please show a screenshot of the header bar when they are both showing? I don't use marks so not sure it looks like.

Something like [this](https://files.catbox.moe/d360ia.png) although this is just an example. Marks are unique for each window, but you can have several marks per window in theory.

ammgws:
@travankor Implemented this in #532, however need to wait for the i3ipc crate to merge a PR first before we can add the feature. I have tested it and it seems to work OK, would you like to try #532 and see if it works for you?

ammgws:
Implemented via #532

PR #532, opened 2020-01-29
Add option to show marks instead of title

ammgws:
This can be merged once tmerr/i3ipc-rs/pull/51 has been implemented.

Solves #486

### DIFF
--- a/src/blocks/focused_window.rs
+++ b/src/blocks/focused_window.rs
@@ -20,6 +20,8 @@ use crate::widgets::text::TextWidget;
 pub struct FocusedWindow {
     text: TextWidget,
     title: Arc<Mutex<String>>,
+    marks: Arc<Mutex<String>>,
+    show_marks: bool,
     max_width: usize,
     id: String,
     update_interval: Duration,
@@ -37,6 +39,10 @@ pub struct FocusedWindowConfig {
         deserialize_with = "deserialize_duration"
     )]
     pub interval: Duration,
+
+    /// Show marks in place of title (if exist)
+    #[serde(default = "FocusedWindowConfig::default_show_marks")]
+    pub show_marks: bool,
 }
 
 impl FocusedWindowConfig {
@@ -46,6 +52,10 @@ impl FocusedWindowConfig {
 
     fn default_interval() -> Duration {
         Duration::from_secs(30)
+    }
+
+    fn default_show_marks() -> bool {
+        false
     }
 }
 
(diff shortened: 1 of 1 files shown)


travankor:
@ammgws 
Sorry for not getting to this earlier, but this works well so far.

Something to consider: I only use 1 mark per window, but sway lets you have several marks per window if you do `swaymsg mark [mark1]` and `swaymsg -- mark --add [mark2]` . So maybe multiple marks should get enclosed in brackets?

Also some interesting info from the man page:
```
show_marks yes|no
           If show_marks is yes, marks will be displayed in the window
           borders.  Any mark that starts with an underscore will not be drawn
           even if show_marks is yes. The default is yes.
```
I don't think that i3status-rust should follow this rule. I think there would be too many edge cases from this and marks are not shown by default so not much benefit with dealing with this.

ammgws:
I can look into the multiple marks thing and an option to hide the underscored ones. If the i3ipc crate is updated before that then we can just merge this simple implementation now and add those features later.

ammgws:
@travankor this is now implemented using the swayipc-rs crate. It works fine for me, please let me know if there are any issues.

travankor:
@ammgws seems like it works well but the issues raised [here](https://github.com/greshake/i3status-rust/pull/532#issuecomment-581854186) remain.

Also the binary size has gone up a significant bit since the last release -- from 6M to 11M.

ammgws:
So with one mark set, e.g. "abc", the bar should display: "abc"

And with multiple marks set, e.g. "abc" and "def", the bar should display: "[abc] [def]"

?

If the binary size is a problem then I will have to look into later. Perhaps enabling LTO could help. Do you mean last release as in the last tagged version vs git master?

travankor:
>So with one mark set, e.g. "abc", the bar should display: "abc"
And with multiple marks set, e.g. "abc" and "def", the bar should display: "[abc] [def]"

1)Currently multiple marks get concatenated, and it looks like one giant mark. `swaymsg -- mark --add anothermark' for example.

2)Marks that begin with underscores are considered hidden.

> Do you mean last release as in the last tagged version vs git master?

Yes. It's not a problem for me personally; I just noticed this and reported it.

ammgws:
> Currently multiple marks get concatenated, and it looks like one giant mark. swaymsg -- mark --add anothermark for example.

I don't use marks so I wasn't sure what it looked like in the first place. This is what one mark and two marks looks like for me: 
image

I was thinking about making i3status-rust show multiple marks with brackets around each one like sway does, but looks like sway does it for the singular case as well so I guess I will change that too.

ammgws:
> Also the binary size has gone up a significant bit since the last release -- from 6M to 11M.

FYI enabling link time optimizations brings it down to 8.6MB

PR #533, opened 2020-01-29
Name threads for easier debugging

ammgws:
This should make debugging at least a little easier/faster.
(Have been investigating a couple of different panics and this has come in handy,)

The unwraps are necessary because thread::Builder::spawn returns a Result, however the behaviour should be the same as before because thread::spawn panics on failure anyway.


ammgws:
Rebased and resolved conflicts

PR #538, opened 2020-01-31
Improve handling of thread panics

ammgws:
- Join the threads we create so that we can throw block_errors for any panics that occur instead of just crashing the whole program

- Clarify panic messages using `expect` where necessary

EDIT;
Need to do some more work on this since can't join them at the current place...


ammgws:
Will need to rethink this.

PR #541, opened 2020-02-01
Read config from XDG_CONFIG_HOME by default

### LINKED ISSUE #465: Provide a default config file paths
`i3status-rs` could look in default paths if no configuration file is given in CLI arguments.

Suggested paths:
- `.config/i3status-rs/config.toml`
- `$XDG_CONFIG_HOME/i3status-rs/config.toml`
- `.i3status-rs.toml`

ammgws:
Closes #465 by adding a `--config` option flag.

If the flag is omitted then we default to reading from:

1. $XDG_CONFIG_HOME/i3status-rust/config.toml
2. $HOME/.config/i3status-rust/config.toml if XDG_CONFIG_HOME is not set.

NOTE: This is a breaking change.
For users who happen to be using a different path to the above for their config files, they will have to update their invocation of i3status-rs to use the `--config` option. 
For other users they will need to remove the argument from their i3status-rs command.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,16 +83,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "clap"
-version = "2.31.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "bitflags 1.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "textwrap 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "unicode-width 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
 name = "cpuprofiler"
 version = "0.0.4"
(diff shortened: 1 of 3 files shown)


atheriel:
I'm a big fan of support for `XDG_CONFIG_HOME`, but I don't think this should be a breaking change. Wouldn't it be better to simply check if the user supplies a file argument, and then fall back on the standard config locations if it is missing?

ammgws:
Good idea, I'll see if it's possible using `clap`

ammgws:
New implementation:
- If user supplies config file path, then we use it (and fail if it's invalid etc)
- If no config file path is supplied then we default to XDG_CONFIG_HOME (and fail if it doesn't exist)

Other changes:
- Changed from `clap` to `getopts` which yields us less dependencies and is enough for our trivial needs.
- Read version info from Cargo.toml which means we only need to maintain it in one place

atheriel:
Looks great.

PR #544, opened 2020-02-03
feat(load): possible to configure thresholds

### LINKED ISSUE #509: load block: not possible to customize range used for set state
Would be nice to be able to customize the load block similarly to how the CPU block can be customized with `info`, `warning` and `critical`.

The load on my system is generally high, so the load block is constantly blue or yellow, with no way to change it.

Also, if load goes above `0.9 * logical_cores`, state is set to `Idle` as it is the default. Load can go beyond `1 * logical_cores`, so the logic should probably be changed to something similar to the CPU block.

https://github.com/greshake/i3status-rust/blob/840ceb23230ee22a42c14d3faeab67610b6aea46/src/blocks/load.rs#L112-L117

MiniGod:
Added possibility to customize thresholds for the states in the Load block. Mostly just pasta from cpu.rs.

closes #509

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -319,6 +319,9 @@ interval = 1
 
 Key | Values | Required | Default
 ----|--------|----------|--------
+`info` | Minimum load, where state is set to info. | No | `0.3`
+`warning` | Minimum load, where state is set to warning. | No | `0.6`
+`critical` | Minimum load, where state is set to critical. | No | `0.9`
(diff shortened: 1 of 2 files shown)


MiniGod:
FYI, the defaults are the same as the hard-coded values were before, and load over 0.9 is now `Critical` (was `Idle`, as mentioned in #509).

atheriel:
Looks good to me, thanks!

PR #548, opened 2020-02-05
Check HTTP response codes from curl in API calls

sashomasho:
When connected to public wifi networks which require additional interaction
from the user to provide access, curl receives either 307 or 511, which
results in block error. Actually I think that only 2XX and 4XX status codes
should be further processed and JSON output parsed, everything else indicates
errors which will probably be recovered once the network access issues are
resolved.


atheriel:
Looks like a nice little QoL improvement, thanks.

Issue #552, opened 2020-02-07
tip branch of i3status-rust panics


AzeemJiva:
The tip branch built panics with the following:

```
RUST_BACKTRACE=full i3status-rs
thread 'main' panicked at 'No option 'profile' defined', /home/azeemj/.cargo/registry/src/github.com-1ecc6299db9ec823/getopts-0.2.21/src/lib.rs:799:21
stack backtrace:
   0:     0x55e1bd69ad84 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h7b11b2a23cd910b3
   1:     0x55e1bd6bd72c - core::fmt::write::hca7efce6c0791439
   2:     0x55e1bd697a07 - std::io::Write::write_fmt::he56b7f6fb86cf3b5
   3:     0x55e1bd69d05e - std::panicking::default_hook::{{closure}}::hf51e7f0a7424f5cd
   4:     0x55e1bd69cd51 - std::panicking::default_hook::he8ba9c8182908cf7
   5:     0x55e1bd69d73b - std::panicking::rust_panic_with_hook::hf9a08396e191b616
   6:     0x55e1bd69d2ee - rust_begin_unwind
   7:     0x55e1bd69d230 - std::panicking::begin_panic_fmt::hfd104289df6728bf
   8:     0x55e1bd5d5517 - getopts::Matches::opt_vals::hd89edab5bd562e5f
   9:     0x55e1bd5d572b - getopts::Matches::opt_str::h8bada4d85a6db8c9
  10:     0x55e1bd4d1a09 - i3status_rs::main::hc54016647dc77074
  11:     0x55e1bd4d66d3 - std::rt::lang_start::{{closure}}::ha9b5bd5695f93204
  12:     0x55e1bd69d183 - std::panicking::try::do_call::hd0ef6b95899a3d3b
  13:     0x55e1bd6a55aa - __rust_maybe_catch_panic
  14:     0x55e1bd69dc90 - std::rt::lang_start_internal::hd59341531b4fe1e8
  15:     0x55e1bd4d66c2 - main
  16:     0x7fb404714153 - __libc_start_main
  17:     0x55e1bd4630ce - _start
  18:                0x0 - <unknown>
```

atheriel:
Do you have a configuration?

AzeemJiva:
Yup, it was working just fine with the build from ~1 week ago

ammgws:
[...]

PR #553, opened 2020-02-07
Solves #552

ammgws:
Needed to guard profiler options since they won't be available when running in a release build.

No panics with either cargo build or cargo build --release


ammgws:
Strange, wonder why it works locally...

ammgws:
Going to have to come back to this.

PR #556, opened 2020-02-08
Adds a man page

### LINKED ISSUE #380: Man page

atheriel:
The final man page is currently generated from the `blocks.md` and `themes.md` files, plus a custom preface and postface. You can run `man/generate.sh` to regenerate the man page after changes to those files.

I have also added a check to Travis CI that ~will fail if~ a commit updates the block/theme documentation but does not regenerate the man page in this way, which we should probably document somewhere.

This should close #380 and supersedes #391.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ addons:
       - libdbus-1-dev
       - libgoogle-perftools-dev
       - libnotmuch-dev
+      - pandoc
 cache: cargo
 os:
   - linux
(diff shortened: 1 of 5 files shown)


atheriel:
Unfortunately it turns out that different versions of pandoc produce slightly different output, so this can't be an actual check failing the build just yet.

PR #558, opened 2020-02-09
[WIP] Cull superfluous dependencies introduced after #540

ammgws:
We went from 66 to 100 dependencies, so there's a chance there are some newly introduced crate features that we aren't using that can be disabled.

So have have found some in `inotify` but still need to further investigate whether it's OK to disable them (compiles and runs fine with my config though).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,11 +49,6 @@ version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
-name = "bytes"
-version = "0.5.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
 name = "c2-chacha"
 version = "0.2.3"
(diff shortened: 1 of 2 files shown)


ammgws:
Seems like inotify is only used by the backlight block, however I don't have a laptop to test with yet.

atheriel:
There appears not to be a D-Bus interface that we could replace the `inotify` usage with, for now.

ammgws:
> There appears not to be a D-Bus interface that we could replace the `inotify` usage with, for now.

@atheriel Ah, I meant that I wanted to test whether there are any side effects from this PR but since `inotify` is only used in the backlight block and I don't have a laptop at the moment then I cannot personally test it.

johnae:
@ammgws This pull request deleted the lock file, was that intentional?

ammgws:
No it wasn't

Issue #555, opened 2020-02-07
Add some tests to Travis CI


ammgws:
Is there a trivial way to run the compiled binary against a simple test config after a successful build?

atheriel:
It should be possible with `--exit-on-error`, yes.

ammgws:
Though if there is no error it would keep running wouldn't it?

atheriel:
Oh, right.

atheriel:
We could maybe add a `--oneshot` mode to render the blocks exactly once.

ammgws:
Or perhaps this could be done using `cargo test`, that way it could be used locally too. There is some discussion [here](https://users.rust-lang.org/t/how-do-you-test-binaries-not-libraries/9554/11) that might shed some light on how to do it.

PR #559, opened 2020-02-09
Add "one-shot" mode and add some tests

ammgws:
As discussed in #555 

Also added a simple config to be tested with the compiled binary. It can be run locally as well using `cargo test`, and is tested by Travis CI as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,4 +33,5 @@ script:
   - rustup component add rustfmt
   - cargo fmt -- --check
   - cargo build --all-features --all-targets --verbose
+  - cargo build --all-features --all-targets --verbose --release
   - cargo test --all-features --all-targets --verbose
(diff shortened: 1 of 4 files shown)


ammgws:
Looks like it's working well

atheriel:
I'm not sure I'm a huge fan of running external commands in a test harness -- wouldn't a better alternative be to make `run()` possible from a test?

We could do this by hoisting config parsing (and profiling) into `main()` and then pass a `config: Config, one_shot: bool` to `run()`.

ammgws:
The reason I thought of this in the first place was because a bug in the config parser (when attempting to switch to getopts) went unnoticed because the profiling args are only compiled for debug releases. I saw a comment from one of the Rust/Cargo maintainers mentioning testing this way which is why I opted for it.

Either way I'd be happy with something that can test both parts.

ammgws:
Since this is just used for tests I will merge for now, and improvements can be made later.

PR #560, opened 2020-02-09
Use `clap` macros for version info etc

ammgws:


atheriel:
1. What's the advantage of these over the current environment variables?
2. Do you know why this includes `Cargo.lock` changes?

ammgws:
1. Either is fine, just less boilerplate this way since clap abstracts it away for us.

2. It looks like the new merge friendly Cargo.lock V2 format. Seems like conversion that was merged the other day didn't actually end up converting the file. Sigh. 
I can remove it from this PR if necessary.

atheriel:
Better to break out the lockfile changes into a separate commit, then.

ammgws:
@greshake Done

ammgws:
@atheriel Done

PR #566, opened 2020-02-15
NVIDIA GPU Block - Fixed Utilization Width (Issue #564)

### LINKED ISSUE #564: NVIDIA GPU Block - Fixed Utilization Width
I noticed that the gpu utilization does not have a prefixed 0. It is therefore moving blocks that are to the left of it when switching between the 0-9 and 10-99 ranges.

I propose to make it two digits wide by default so it matches the CPU utilization: https://github.com/TheJP/i3status-rust/commit/a72fd8ddf0bb38c823160915f0ed97e70457af58

TheJP:
Creating pull-request for #564

Edit: Closes #564.


atheriel:
Thanks!

PR #570, opened 2020-02-21
feat(net): support ppp vpn

MiniGod:
I'm using Ubuntu with a PPTP vpn. (configured via `gnome-control-center`).
The device gets the name `ppp0`. With `device = "ppp0"` in the net block, it showed up as inactive / missing.

The uevent file has `DEVTYPE=ppp`. I'm assuming that's how `wireguard` is done, so I copied the same logic for `ppp`, which works.


atheriel:
Thanks!

PR #571, opened 2020-02-21
chore: Update maildir to 0.4

ignatenkobrain:
This removes nix dependency, though adds gethostname one. That one does
exactly one thing and the size of sources (downloaded by cargo) is
significantly smaller: 0.2 MB vs 8.99 kB.

Size of i3status-rs is not really changed due to this.

Signed-off-by: Igor Raits <i.gnatenko.brain@gmail.com>


ammgws:
Can you rebase?

ammgws:
Merged in #729

PR #452, opened 2019-10-12
Fix inverted icons in backlight block.

marc0der:
The icons for the backlight block seem to be inverted. The icons show black when at it's brightest, yellow when at it's dimmest. This simply inverts them to the correct phase.

Issue #183, opened 2018-05-05
Brightness: Emoji disambiguity


GladOSkar:
Hi there, long time fan of the project.

My distribution recently introduced color emoji, which is very cool.

This led to a discovery though: The brightness-emoji (moon?) i3status-rs uses is actually used the other way around:

![image](https://user-images.githubusercontent.com/9497832/39660923-5e73f9a6-5049-11e8-8776-cbc22ea38861.png)

![image](https://user-images.githubusercontent.com/9497832/39660925-72d9c6aa-5049-11e8-8109-b619c9c09e81.png)

I know it would not make sense to invert this for everyone since it does make sense in black and white, but maybe an option to invert would be good?

Thank you!

(I'm learning rust currently, so if nobody wants to do it i may)

matthiaskrgr:
For the record, in black and white mode it looks like this:

![brightness2](https://user-images.githubusercontent.com/476013/39665995-e52958c6-509c-11e8-9f22-7268e647aee4.png)
![brightness1](https://user-images.githubusercontent.com/476013/39665996-e54713f2-509c-11e8-800e-8a2c17040a5e.png)

atheriel:
You can actually do this already! The `icon.overrides` section in your configuration allows you to set whatever icons you'd like. In this case, set `"backlight_empty"`, `"backlight_partial1"`, `"backlight_partial2"`, `"backlight_partial3"`, and `"backlight_full"` to the appropriate characters for your font.

GladOSkar:
Cool! Thank you :)

atheriel:
Glad to be of help.

Kabouik:
> For the record, in black and white mode it looks like this:
> 
> ![brightness2](https://user-images.githubusercontent.com/476013/39665995-e52958c6-509c-11e8-9f22-7268e647aee4.png) ![brightness1](https://user-images.githubusercontent.com/476013/39665996-e54713f2-509c-11e8-800e-8a2c17040a5e.png)

I'm sorry for asking that question 4 years later, but how do you enable black and white mode? I'd really like do use that for that block.

GladOSkar:
I think you have to choose a font or font system that doesn't include color emoji, but I'm not sure. Sometimes you can also just uninstall your distro's color emoji package. But that's something you'll have to look up for your distro yourself, it's nothing `i3status-rust` can control

PR #573, opened 2020-02-26
Fix reversed symbols in backlight module

rumpelsepp:
The icons for the backlight module were in the wrong order. This patch reverses them.


GladOSkar:
(See #452 and #183)

This is a problem with color emoji. Black and white ones look correct. I use this in my config to invert the icons:

```
[icons]
name = "awesome"
[icons.overrides]
backlight_empty = " 🌑 "
backlight_partial1 = " 🌘 "
backlight_partial2 = " 🌗 "
backlight_partial3 = " 🌖 "
backlight_full = " 🌕 "
```

rumpelsepp:
If I understand this correctly, then the problem is the other way round. Full moon means "full light".

GladOSkar:
If you go by the emoji definitions, yes. But since most linux desktop users are using black-and-white emojis, for them it looks the other way around when printed bright-on-dark.

Plus, we don't want to break user's existing configurations with changes like these afaik.

rumpelsepp:
Unfortunate. I personally consider this an upstream (= font) bug which should be fixed upstream.

GladOSkar:
Well technically most fonts are designed for dark-on-white as when printed on paper, and then the b&w fonts are correct again. It's just a limitation of the single color font system ¯\\\_(ツ)\_/¯

atheriel:
Indeed, this is an unfortunate ambiguity for these icons.

Thra11:
All the rest of the icons on my status bar are monochrome, but the brightness one has recently (since 0.14.0 perhaps?) become full-colour (I have `icons = "awesome5"` in my config), which makes it "backwards" as reported in #845. Perhaps that's the issue that needs to be fixed instead of reordering them?

GladOSkar:
The color of the icons does not depend on `i3status-rust` but rather on your system (font/pango/cairo) as far as i know. This project doesn't have control over that. More and more distros are shipping color-capable font systems so this "issue" is bound to occur for more and more people.

The only thing that could probably be done to mitigate is add a `flip-order` option or similar to make it easier to override the icon order.

As i mentioned earlier, the only reason this happens is because the monochrome icons are designed for dark-on-bright text (as with paper) and not bright-on-dark (as with most status bar themes), so the real issue is that monochrome fonts happen to adapt to text color, which never makes sense for an icon that is supposed to imply brightness.

Thra11:
Ah. I must confess I don't really know how these icon fonts work. I assumed that the output from i3status-rust specified the font and the character to draw, in theory giving complete control over the glyph that appears, so long as the requested font is installed and covers the character in question. However, as far as I can tell, font awesome doesn't actually contain any glyphs for the phases of the moon. I'm guessing that as a result, it's falling back to a font which _does_ cover the moon phase symbols, which presumably varies from system to system, sometimes resulting in monochrome moon icons, sometimes full-colour ones.

GladOSkar:
Yes. The `i3bar` protocol is just text, the font is specified in the bar/wm config, not by `i3status`. And afaik these icons are regular unicode emoji, not fontawesome icons. So it entirely depends on your font and the font rendering system used.

Issue #568, opened 2020-02-17
Feature: Hide blocks with missing resources


shazow:
I'm trying to create some shared configuration setup between multiple devices. Sometimes a battery is available, sometimes it's not. Sometimes there is a wifi device, sometimes it's an ethernet device, etc.

By default, the i3status-rust aborts when a block's resource is missing. Would be nice to be able to override a block to hide when the resource is missing (e.g. have the battery block be skipped if there's no battery available).

Related to https://github.com/greshake/i3status-rust/issues/567

Happy to take a crack at it, let me know if there's any objections.

wucke13:
I absolutely support this. I started writing my own rust program for an i3/sway status bar and this was the first thing I took care of. Once I found out about i3status-rust, I was quite surprised to find out that the config is not easily shareable!

shazow:
I haven't had a chance to give it a try lately, so if anyone wants to beat me to it please go ahead. :)

wucke13:
Where in the code would one have to ignore the error?
Would it be feasible to output the error message to stderr, and all other blocks to stdout, instead of only putting the error message and nothing else?

ammgws:
Closing this as the battery block and net block both support hiding the block when the specified device is missing.

If this support is required for another block then please re-open/create another issue.

PR #585, opened 2020-04-06
Do not fail if the battery is not available (sysfs driver)

flying7eleven:
Hi,

right now I have a Linux installation on a thumb-drive which I use on a laptop as well as a desktop computer. Obviously, the desktop computer does not have a battery which makes using `i3stats-rs` not very feasible. There are several open issues (for example #568 and #85) with a similar problem.

This patch introduces the `allow_unavailable_battery` option for the `battery`-Block (right now just for the sysfs backend) to solve this issue. If this option is set to `true`, a non-existent battery will not raise an error but will instead show a non-available battery icon and text. If the option is set to `false` (which is the default), the behavior is the same as in the current release version.

Unfortunately I do not have an laptop with an removable battery so I cannot test the behavior of the change if the battery is removed during runtime, but I think it should fix this issue as well.

I am still working on the UPower backend but right now I am not sure how to solve it in a adequate way.

Any suggestions or comments? :)

Best,
Tim


PhilipConte:
I would love for this to work, and especially for the block to not show at all when there's no battery. 

I gave a crack at a slightly different approach

https://github.com/PhilipConte/i3status-rust/commit/cfacf92cff1f305cb877ce5b445da9d5c80b2dcd

but my Rust experience is both very limited and stale, so there's a variety of errors in it.

Hopefully it can inspire you.

flying7eleven:
I just saw that there would be merge issues with the pull request, just fixed them real quick.

@PhilipConte I haven't had time so far to look into your changes but I do so as soon as I have some spare time.

ammgws:
@flying7eleven Personally I am leaning towards merging this, but IIRC there were some differing opinions when these battery related bar failures were brought up in past issues. Could you please summarise the situation from those past issues?

flying7eleven:
Hey @ammgws, yeah, there was a quite long discussion about different approaches and pros and cons. One of the solutions I also would like and prefer was a profile-based options where one could load the blocks based on a profile for the current computer. This would result in a lot of refactoring of the current code-base and would not be a short-term solution. Another solution was - and still is - to use the UPower blocks `DisplayDevice` option (which seems not to work when live-removing a battery) (don't know how my solution would handle that, btw.). The here suggested patch was more and quick (and hopefully not so dirty) solution so I could use `i3status-rs` on my quite weird setup. I could only find one other user with a similar setup who could use it (at least based on use issues) and several others with live-unplugging the battery where I do not know if this would be a solution.

The most controversial discussions were about the fact if the block should be still be displayed or completely hidden if the batteries where not detected. I've gone for the fist solution so it might behave better with live-unplugging.

I hope this helps, if not, let me know!

Nukesor:
Hiding this behind a configuration flag seems like a very sensible thing to do. Would love to see this added to the project.

Feel free to ping me, if you need someone to test this PR.

flying7eleven:
If there is still interest in this pull request, I could bring it to a state where it could be merged into the master in the next few days 👍🏻

Nukesor:
@flying7eleven Hey :)

I just took the opportunity and created a follow-up PR with a few more features and hot-swap support.
Your PR has been the basis, I hope this isn't a problem :)

PR can be found over here: https://github.com/greshake/i3status-rust/pull/835

flying7eleven:
Why should it be a problem, @Nukesor? That's the advantage of open source development. Anything can be used and improved :)

flying7eleven:
Closing this PR in favor of #835

PR #591, opened 2020-04-12
battery: only display remaining time when it is useful

debugloop:
Hi, I did not like the battery module displaying `0:00` most of the time. I am not sure about the space left behind, i.e. in `{percentage}% {time}`. It seems unwise to have a formatter that generates a space. which would allow doing `{percentage}%{time}` eliminating that trailing space, but looks weird in configs and leads to double spaces with existing configs.

It might be best to introduce a new formatter, but I'm not sure.

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -513,7 +513,10 @@ impl Block for Battery {
                 Err(_) => "×".into(),
             };
             let time = match self.device.time_remaining() {
-                Ok(time) => format!("{}:{:02}", time / 60, time % 60),
+                Ok(time) => match time {
+                    0 => "".into(),
+                    _ => format!("{}:{:02}", time / 60, time % 60),
+                },
                 Err(_) => "×".into(),
             };
             let power = match self.device.power_consumption() {


ammgws:
What does 0:00 mean in the first place? Is it when it's being charged/on AC power? (I don't use a battery powered device.)

debugloop:
It always indicates the remaining time, i.e. both charging or draining the battery. The meaning depends on the icon, color, and wether there's a cable protruding from the laptop :)

I was mostly adressing the case where it is sitting at 0:00 when fully charged and plugged, although this MR will theoretically also hide the time a moment before powering off due to an empty battery.

gabm:
I would love to see that merged.. I think the downside that it won't show charging when its approaching "empty" doesn't matter too much. Anyway, one could always try to estimate whether the battery is charging using either of these. 

```
/sys/class/power_supply/AC/online
/sys/class/power_supply/BAT0/power_now
/sys/class/power_supply/BAT0/status
```

Currently on my system (on AC) `AC/online` is 1,  `BAT0/power_now` is 0, but `BAT0/status` is "Unknown (threshold may prevent charging)` ... so it might not be straight forward... 

In my opinion its too complicated for a small glitch...

ammgws:
@flying7eleven, @JohnAZoidberg, @GladOSkar, @Ma27, @Thra11 

As recent contributors and users of the battery block, do you have any comments regarding this PR?

GladOSkar:
I only ever see `0:00` for a short time while the battery is almost fully charged and actually fully charged (when the time disappears) and I don't mind. My battery never gets so empty that it would occur while discharging anyway.

If @debugloop or @gabm have a weird system where the battery percentage in  `upower` or `sysfs` doesn't ever _fully_ reach 100% but rather 99.9% and thereby always shows `0:00`, I could imagine that being annoying. Not sure what the case is here though.

debugloop:
@GladOSkar I can confirm your theory, my Lenovo T460p never reaches a flat 100%, instead it sits as 99.9% permanently.

Thra11:
This works well for me. I doubt the hidden time when empty is a big issue, but `status` is in scope, so you could always check its value (I assume it's `"Discharging"` when it's nearly empty, but haven't actually confirmed it. As soon as you plug in the charger and it changes to "Charging", it should change to showing the time remaining to full, so it would not hide the time unless fully charged.)

ammgws:
Thanks!

flying7eleven:
Sorry, just saw the mention. Even if it is already merged, I just wanted to confirm I have the same issue with an T460p as well as an T480... so I really appreciate the change :+1:

PR #593, opened 2020-04-12
keyboard_layout: Add `format` argument

thiagokokada:
This PR add `format` argument to `keyboard_layout` block. For now, it only supports `{layout}` placeholder.

The reason for this addition is that I want to add an icon to my keyboard block, something like this:

```toml
[[block]]
block = "keyboard_layout"
driver = "localebus"
interval = 5
format = " {layout}"
```

I could have also added an icon to the keyboard block, however I prefer this approach since it is more flexible, and allow in the future support for more information in `keyboard_layout` block (like keyboard variants).

This is my first code in Rust so if there is any issues, please report it and I will try to fix it :+1: .

### DIFF
--- a/src/blocks/keyboard_layout.rs
+++ b/src/blocks/keyboard_layout.rs
@@ -18,6 +18,7 @@ use crate::config::Config;
 use crate::de::deserialize_duration;
 use crate::errors::*;
 use crate::scheduler::Task;
+use crate::util::FormatTemplate;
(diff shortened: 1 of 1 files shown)


ammgws:
Works for me and could possibly be expanded in the future.

Thanks!

atheriel:
This should be added to the documentation as well.

PR #530, opened 2020-01-26
Ignore stdin and stdout for `spawn_child_async`

Celti:
If stdout from child processes is not deliberately ignored and/or
quieted via external means, it can interfere with the i3bar protocol
stream. Stdin is less dangerous, but gets closed almost immediately by
`wait()`ing.

Connect both of them to `Stdio::null()` to avoid any problems. Stderr
is safe and can remain as-is for debugging purposes.

Issue #420, opened 2019-08-11
Sound block start pavucontrol with errors sway/wayland


vvrein:
Version: i3status-rs 0.9 //latest from upstream
System: Archlinux

i3status-rs backtrace
```
1
thread 'main' panicked at 'failed printing to stdout: Broken pipe (os error 32)', src/libstd/io/stdio.rs:792:9
stack backtrace:
   0:     0x55ccc5b005f3 - std::sys::unix::backtrace::tracing::imp::unwind_backtrace::h32cdc43507ada68c
   1:     0x55ccc5afbe7b - std::sys_common::backtrace::_print::h5e0f5e9cbf257358
   2:     0x55ccc5aff2a6 - std::panicking::default_hook::{{closure}}::hbafdebc07929bbe6
   3:     0x55ccc5aff039 - std::panicking::default_hook::h91c151aff85ca479
   4:     0x55ccc5aff9af - std::panicking::rust_panic_with_hook::h0d3d26cb2543ef46
   5:     0x55ccc5aff531 - std::panicking::continue_panic_fmt::h31b7e7fa8e3eb727
   6:     0x55ccc5aff47e - std::panicking::begin_panic_fmt::h2d0f9f4f044c2cfc
   7:     0x55ccc5af7873 - std::io::stdio::_print::hbcac8d94a14d3963
   8:     0x55ccc595e33f - i3status_rs::util::print_blocks::he19e549396fc2b43
   9:     0x55ccc59349b3 - i3status_rs::main::ha3775e1684a207e1
  10:     0x55ccc58e71b2 - std::rt::lang_start::{{closure}}::h6c45b36e4b1e1dca
  11:     0x55ccc5aff3b2 - std::panicking::try::do_call::h58bfd138a962b46f
  12:     0x55ccc5b07959 - __rust_maybe_catch_panic
  13:     0x55ccc5afff7c - std::rt::lang_start_internal::h4d084069bbddd485
  14:     0x55ccc5935e51 - main
  15:     0x7f4619835ee2 - __libc_start_main
  16:     0x55ccc585a0bd - _start
  17:                0x0 - <unknown>
thread '<unnamed>' panicked at 'cannot access stdin during shutdown', src/libcore/option.rs:1036:5
```

sway config 
```
bar {
[...]

PR #594, opened 2020-04-13
Unify use of `spawn_child_async` in click handlers

Celti:
This follows up #530 by using `crate::subprocess::spawn_child_async` for
all click handlers that spawn such processes, to unify handling of stdin
and stdout and prevent broken pipe errors such as #420 in all cases.

This also uses spawns such commands with `sh -c` to making handling
environment and quoting more flexible.


ammgws:
LGTM

Celti:
Yes, I did — specifically the sound block as referenced in #420 but I also deliberately searched for and attempted to trigger the bug in other blocks.

ammgws:
Thanks!

PR #601, opened 2020-04-22
taskwarrior: Force rc.gc=off

jubnzv:
Garbage Collection is a taskwarrior configuration option that is descibed here: https://taskwarrior.org/docs/ids.html.

It should be always disabled when calling `task` from scripts. Otherwise gc will modify ID numbers of the pending tasks when we complete or delete tasks from the another script.

This will also increase the speed of update for this widget (see "Configuration" section here: https://taskwarrior.org/docs/ids.html).


flying7eleven:
LGTM :+1: (didn't know that, thanks for the info :+1: )

ammgws:
Nice catch

PR #605, opened 2020-04-23
Use /sys/class/net/<device>/carrier instead of operstate in is_up()

happycoder97:
When using android usb tethering in Arch Linux (kernel: Linux 5.6.5-zen4-1-zen),
`operstate` file shows `unknown`. But the `carrier` file correctly shows `1`.
  `ethtool` also correctly says `Link detected: yes`.

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -72,8 +72,8 @@ impl NetworkDevice {
     /// Check whether this network device is in the `up` state. Note that a
     /// device that is not `up` is not necessarily `down`.
     pub fn is_up(&self) -> Result<bool> {
-        let carrier_file = self.device_path.join("carrier");
-        if !carrier_file.exists() {
+        let operstate_file = self.device_path.join("operstate");
+        if !operstate_file.exists() {
             // It seems more reasonable to treat these as inactive networks as
             // opposed to erroring out the entire block.
             Ok(false)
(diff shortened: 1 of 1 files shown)


ammgws:
It seems like there isn't 100% overlap between the two as operstate seems to offer some more context. Unless you're sure that switching to just checking carrier is fine, perhaps we could check for both?

/sys/class/net/xxx/carrier
> Indicates the current physical link state of the interface.
> 		Posssible values are:
> 		0: physical link is down
> 		1: physical link is up

/sys/class/net/xxx/operstate
> Indicates the interface RFC2863 operational state as a string.
> 		Possible values are:
> 		"unknown", "notpresent", "down", "lowerlayerdown", "testing",
> 		"dormant", "up".

> IF_OPER_UNKNOWN (0):
>  Interface is in unknown state, neither driver nor userspace has set
>  operational state. Interface must be considered for user data as
>  setting operational state has not been implemented in every driver.
> IF_OPER_DOWN (2):
>  Interface is unable to transfer data on L1, f.e. ethernet is not
>  plugged or interface is ADMIN down.
> IF_OPER_UP (6):
>  Interface is operational up and can be used.

https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-net
https://www.kernel.org/doc/Documentation/networking/operstates.txt

atheriel:
I'd go with the official docs here and say we should check both. It's also a good idea to check the source of `ethtool` to figure out what it's doing here, if possible.

happycoder97:
Yeah it is a better idea to check both. I will re-add the code.
But by the way, I have been using this patch for a few days and it shows carrier file is not found error when I disconnect and reconnect the device. It only happens sometimes and goes away if I reload i3. This problem was not there when using operstate file.
What could be happening here?

ammgws:
Seems like the file doesn't exist when the interface is down:
https://linuxconfig.org/how-to-detect-whether-a-physical-cable-is-connected-to-network-card-slot-on-linux

Checking both could be like:
1. check operstate
2. if not 'up', then check carrier
3a. if carrier is 1 then return 'up'
3b if carrier is 0/doesn't exist then return the original operstate result

ammgws:
Ended up adding it myself. Thanks!

Issue #603, opened 2020-04-22
Display usage% per CPU core in the cpu block


grim7reaper:
Hello,

I'm currently moving from XMonad/xmobar to i3 and I'm trying to have a status bar similar to what I had with xmobar.
I've managed to get almost all I needed with your cool project here :slightly_smiling_face: , but I'm still missing a few features.

Regarding the block `cpu` I'm missing two things:
1. the ability to display the utilization of each core in %. IIUC, the `{utilization}` format only gives a global average.
2. ideally, the color would be applied to each % indivually, according to the threshold.

In picture, here is what I'm trying to get:

![cpu](https://user-images.githubusercontent.com/6479907/80019889-80fba800-84d8-11ea-9c77-8887de332d25.png)

I've had a quick look at `src/blocks/cpu.rs` and I think that the first point should be easy to do because we already fetch the info for `{barchart}`: we "simply" have to display it differently.
I can probably provide up with a PR for that :wink:

The second point (independant color inside a single block), looks trickier with the current implementation :confused:

ammgws:
Give it a go and see what you can come up with

grim7reaper:
Ok, I'll give it a try :slightly_smiling_face: 
I'll let the color issue aside for now: it seems related to `widget.rs` and may have a wider impact.
Anyway, it can be addressed later.

Should I go with:
- a new format, I'll need to find a good name for it (`{utilizations}` sounds way too similar to the existing one)
- a new bool in the block config that would toggle "show average/show all" for the existing format `{utilization}`
?

PR #606, opened 2020-04-23
Allow to display per-core CPU metrics

grim7reaper:
This PR introduce a new config parameter for the block `cpu`: `per_core`.

When set, the block will display the metrics for each core instead of computing an average.
This work for both frequency and utilization percentage.

Closes: #603

### DIFF
--- a/src/blocks/cpu.rs
+++ b/src/blocks/cpu.rs
@@ -133,8 +133,8 @@ impl Block for Cpu {
             .block_error("cpu", "Your system doesn't support /proc/stat")?;
         let f = BufReader::new(f);
 
+        let mut cpu_freqs: [f32; MAX_CPUS] = [0.0; MAX_CPUS];
         let mut n_cpu = 0;
-        let mut freq: f32 = 0.0;
         if self.has_frequency {
             let freq_file =
                 File::open("/proc/cpuinfo").block_error("cpu", "failed to read /proc/cpuinfo")?;
@@ -149,12 +149,10 @@ impl Block for Cpu {
                     let numb = last
                         .parse::<f32>()
                         .expect("failed to parse String to f32 while getting cpu frequency");
-                    freq += numb;
+                    cpu_freqs[n_cpu] = numb;
                     n_cpu += 1;
                 }
             }
-            // get the average
-            freq = (freq / (n_cpu as f32) / 1000.0) as f32;
         }
 
         let mut cpu_utilizations: [f64; MAX_CPUS] = [0.0; MAX_CPUS];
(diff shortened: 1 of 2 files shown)


src/blocks/cpu.rs
```diff
@@ -239,3 +242,33 @@ impl Block for Cpu {
...
+    }
+}
+
+#[inline]
+fn format_frequency(cpu_freqs: &[f32], count: usize, per_core: bool) -> String {
+    if per_core {
+        cpu_freqs
+            .iter()
+            .take(count)
+            .map(|v| format!("{0:.1}GHz", v / 1000.0))
```
ammgws, line 267: Is there a reason only `per_core` displayed "Ghz"?

src/blocks/cpu.rs
```diff
@@ -239,3 +242,33 @@ impl Block for Cpu {
...
 }
+
+#[inline]
+fn format_utilization(values: &[f64], count: usize, per_core: bool) -> String {
+    if per_core {
+        values
+            .iter()
+            .take(count)
+            .skip(1) // The first value is a global one.
+            .map(|v| format!("{:02.0}%", 100.0 * v))
```
ammgws, line 253: Is there a reason why only `per_core` appends the percentage sign?

src/blocks/cpu.rs
grim7reaper: From what I understood, the current code wasn't adding the "unit": it's up to the user to add it in the format string (as in `format = "{barchart} {utilization}% {frequency}GHz" ` from the documentation).

Now, with N values I find it clearer to add the unit to each value, but you're right: that's no longer consistent with the single-value case.

Maybe I can also add the "unit" in the single-value case and update the example format in the doc (and the default one in the code) to remove the `%`, what do you think?

src/blocks/cpu.rs
grim7reaper: Same reason as above :slightly_smiling_face:

src/blocks/cpu.rs
grim7reaper: I've added the unit even when `per_core=false`, to be consistent.

But as mentioned in the commit message, I think it may surprise some users when they update, if they already added %/GHz in their format string (maybe we can mention it in the changelog though).

Anyway, it's in a dedicated commit so I can easily remove/rework it if need be :slightly_smiling_face:

atheriel: I'm also concerned about breaking existing configurations here.

grim7reaper: Yeah, that can be an issue and that's why I only added the "unit" when `per_core` is `true` (it's `false` by default, so no breakage on upgrade), but as pointed out it's a bit surprising when reading the code/inconsistent.

Possible solutions:
- add a comment explaining why we don't add unit in the else branch (backward compat)
- don't add unit even with `per_core=true` (but I think the result is less legible)
- add another flag, `add_unit` (may be overkill, and I think we don't want to multiply the number of boolean flags)
- something I haven't thought of?

grim7reaper: @ammgws or @atheriel : Anything I could do to make this move forward :slightly_smiling_face:?
Have you a preferred option/solution?

ammgws: Still thinking about whether there's another way

ammgws: Perhaps revert the last commit for now

grim7reaper: Done :slightly_smiling_face:

ammgws: I think we can figure this out later

PR #607, opened 2020-04-24
Fix #592

### LINKED ISSUE #592: fakeroot required for pacman block
Hello,

The pacman block is always indicating "system is up to date" on my fresh Arch install.

After some investigation it appears that the pacman block relies on `fakeroot` (which wasn't installed on my machine).

Maybe it would be worth mentioning that `fakeroot` is required by the pacman block ?

IMO the following snippet is a bit misleading: I understand that if `fakeroot` is not found
 on the system, the pacman block will report 0 update...

https://github.com/greshake/i3status-rust/blob/4049a6a5a91bee026e1cf175476a0f7217b6e2ab/src/blocks/pacman.rs#L116-L119

I'm not too sure about this check: 

https://github.com/greshake/i3status-rust/blob/4049a6a5a91bee026e1cf175476a0f7217b6e2ab/src/blocks/pacman.rs#L103-L106

`type` does not have a `-P` option, maybe it should be `type -p fakeroot` instead?

What do you think?

Thanks for this project =)

PicoJr:
Fix #592 

add unit testsfor `has_command`.

### DIFF
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -111,17 +111,27 @@ fn run_command(var: &str) -> Result<()> {
         .map(|_| ())
 }
 
-fn has_fake_root() -> Result<bool> {
+fn has_command(command: &str) -> Result<bool> {
     Ok(String::from_utf8(
         Command::new("sh")
-            .args(&["-c", "type -P fakeroot"])
+            .args(&["-c", format!("type -p {}", command).as_ref()])
             .output()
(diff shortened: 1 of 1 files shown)


PicoJr:
CI is KO, I'll investigate and fix my PR.

PicoJr:
Unit tests display `/usr/bin/sh`, output should be redirected to /dev/null

ammgws:
Thanks!

Perhaps at some point the has_command could be moved to util.rs since it seems like it could be handy for other blocks.

PR #611, opened 2020-04-25
Allow for reading theme information from standalone files (cf PR #366)

PicoJr:
Allow for reading theme information from standalone files (cf PR #366)

This PR contains the original 2 commits from [atheriel](https://github.com/atheriel) (rebased).

Thanks to @atheriel who did the heavy-lifting =)

It should be compatible with previous config file format: see unit tests in `src/config.rs`.

I took the opportunity to move `has_command` from `pacman.rs` to `utils.rs`.

I also added some unit tests for `color_from_rgba`, I changed the code a little to fix a _panic_ when input is too short: `color_from_rgba("rust")` would panic...


src/themes.rs
```diff
@@ -166,51 +169,123 @@ lazy_static! {
...
+            "gruvbox-dark" => Some(GRUVBOX_DARK.clone()),
+            "space-villain" => Some(SPACE_VILLAIN.clone()),
+            _ => None,
+        }
+    }
+
+    pub fn from_file(file: &str) -> Option<Theme> {
+        let full_path = Path::new(file);
+        let xdg_path = util::xdg_config_home()
+            .join("i3status-rs/themes")
```
atheriel, line 217: We should be consistent with the existing config location here and use `i3status-rust/themes` (here an elsewhere).

src/util.rs
```diff
@@ -9,16 +9,19 @@ use std::fmt::Display;
 use std::fs::{File, OpenOptions};
 use std::io::prelude::*;
 use std::io::BufReader;
-use std::num::ParseIntError;
 use std::path::{Path, PathBuf};
 use std::prelude::v1::String;
+use std::process::Command;
+use toml;
+
+pub const USR_SHARE_PATH: &str = "/usr/share/i3status-rs";
```
atheriel, line 17: Should be `i3status-rust`.

atheriel:
This looks good to me in general, thanks for taking up the cause. Two concerns:

1. As decided in #465, we should use `i3status-rust` for config paths, not `i3status-rs`.
2. This introduces about a dozen new dependencies to run the unit tests -- is that necessary?

src/themes.rs
PicoJr: Thanks for your feedback, I'll fix that.

src/util.rs
PicoJr: Thanks I'll fix that too.

PicoJr:
> This looks good to me in general, thanks for taking up the cause. Two concerns:
> 
>     1. As decided in #465, we should use `i3status-rust` for config paths, not `i3status-rs`.

Config path fixed in latest commit.

>     2. This introduces about a dozen new dependencies to run the unit tests -- is that necessary?

I wouldn't say it is necessary: It might be possible to adapt the tests so that no temporary files are required (we wouldn't depend on `assert_fs`)

I would argue it's only [dev-dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies): `assert_fs` is not required to build the project, only to run the tests.  
Nevertheless it still negatively impacts CI validation time (by how much I don't know).

If this new dependency is an issue don't hesitate to reach me out :)

ammgws:
> I would argue it's only [dev-dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies): `assert_fs` is not required to build the project, only to run the tests.

I think it's fine if it's just a dev dep. @atheriel ?

PR #617, opened 2020-04-27
sound block: use -M flag for set too

### LINKED ISSUE #614: Mouse wheel volume control broke on sound block.
**Manjaro Linux
i3status-rust-git
Driver: Alsa**

[[block]]
block = "sound"
driver = "alsa"
step_width = 3

For some reason the mouse wheel volume controls do not work anymore. This is a recent occurrence. When I try to scroll on the block the volume adjusts erratically until it finally ends up at "0".  I have to go into alsamixer to correct. 

I do not know if ;
**added mapped-volume option for correct volume #508** 
(*Note: I do not think this is it as it was working and sound was sinking better than before this commit.)
or 
**Unify use of `spawn_child_async` in click handlers**
could be the culprit

ammgws:
Possibly fixes #614

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -120,7 +120,7 @@ impl SoundDevice for AlsaSoundDevice {
         let volume = max(0, self.volume as i32 + step) as u32;
 
         Command::new("amixer")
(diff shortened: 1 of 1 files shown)


ammgws:
Closing for now since not 100% fixed yet

Issue #247, opened 2018-09-03
SSID with wpa_cli


concatime:
Hi. I want to display only my SSID, but without installing `iw`. I have wpa_cli, and with `wpa_cli status -iwlp3s0`, I'm able to get my ssid. Can you implement this as an alternate way if iw is absent?
Thank you.

atheriel:
This seems like a reasonable request.

ammgws:
@concatime Sorry for the delay but does #625 solve this for you?

ammgws:
Confirmed it works.

Closing via #625

concatime:
Sorry, I was quite busy. And to be honest with you, I’m currently using Waybar.

PR #625, opened 2020-05-01
net block: add fallback to `wpa_cli` for retrieving SSID

ammgws:
Solves #247


ammgws:
OK I have confirmed this works with both iw and nmcli unavailable.

Issue #637, opened 2020-05-03
'cpu' module breaks for more than 32 CPUs


snicket2100:
It looks like https://github.com/greshake/i3status-rust/commit/3bd6bde6c09017fc45a39ba0230203b0bc99f481#diff-add6d8cceff79e8da5607fd20c1137fd has introduced a bug which with more than 32 CPUs present causes `i3status-rust` to crash due to the `cpu` block. The error message is as follows:

```
thread 'main' panicked at 'index out of bounds: the len is 32 but the index is 32', src/blocks/cpu.rs:153:21
```

my `cpu` block config is:
```
[[block]]
block = "cpu"
interval = 1
format = "{utilization}% @{frequency}GHz"
```

I guess this code fragment could help around line 153:
```
if cpu_i >= MAX_CPUS {
    break;
};
```
the same way it is handled in line 195 for example.

snicket2100:
I have tested it with the fix and it doesn't crash for me anymore.

ammgws:
So you actually have more than 32 on your machine?

snicket2100:
Yeah, an 18-core CPU with hyper-threading counts as 36 and was enough to trigger this bug.

PR #639, opened 2020-05-03
bugfix: cpu block used to crash if >32 CPUs present

snicket2100:


snicket2100:
Fixes https://github.com/greshake/i3status-rust/issues/637

PR #645, opened 2020-05-06
s/use_nag/enable_i3nagbar/g

kAworu:
To match the documentation.

### DIFF
--- a/src/blocks/pomodoro.rs
+++ b/src/blocks/pomodoro.rs
@@ -74,8 +74,8 @@ pub struct PomodoroConfig {
     pub message: String,
     #[serde(default = "PomodoroConfig::default_break_message")]
     pub break_message: String,
-    #[serde(default = "PomodoroConfig::default_use_nag")]
-    pub use_nag: bool,
+    #[serde(default = "PomodoroConfig::default_enable_i3nagbar")]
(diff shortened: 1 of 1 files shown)


kAworu:
Before this patch, the documented `pomodoro` block `enable_i3nagbar` option does not work:
```toml
[[block]]
block = "pomodoro"
length = 25
break_length = 5
message = "Take a break!"
break_message = "Back to work!"
enable_i3nagbar = true
```
```
% ./target/debug/i3status-rs pomo-enable_i3nagbar.toml 
[[{"background":"#000000","color":"#dc322f","full_text":" Configuration error: Failed to deserialize block config..\nCause: unknown field `enable_i3nagbar`, expected one of `length`, `break_length`, `message`, `break_message`, `use_nag` ","separator":false,"separator_block_width":0}]


Configuration error: Failed to deserialize block config..
Cause: unknown field `enable_i3nagbar`, expected one of `length`, `break_length`, `message`, `break_message`, `use_nag`
```

ammgws:
Probably better to update the docs instead, since the original author might be using `use_nag` already.

kAworu:
I think `enable_i3nagbar` is clearer but I agree that as it stand the patch is a breaking change. I'll change the PR as you suggested.

PR #649, opened 2020-05-09
Add profiling.md

PicoJr:
Add a short document explaining how to use the `profiling` feature to profile a block.


ammgws:
Thanks!

ammgws:
I just realised this was already mentioned at the end of the block developer guide in the Wiki

PicoJr:
You are right, my bad, `profiling.md` can be removed then.

PR #653, opened 2020-05-09
battery block: fallback for determining power consumption

### LINKED ISSUE #429: Battery: sysfs driver can't get the power draw
When I use the sysfs driver, I can't get the power consumption to show on the bar. However, the upower driver does work.

Worth mentioning is that my laptop lacks a  `/sys/class/power_supply/BAT0/power_now` file but it does have `current_now` and `voltage_now`.

ammgws:
Resolves #429 by implementing the simple approach mentioned here: https://github.com/greshake/i3status-rust/issues/429#issuecomment-626132440. This could be worked on in the future for better hysteresis.

I don't have a battery operated device to test with at the moment so would appreciate if @travankor, @GladOSkar could check this out if they have time.

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -198,12 +198,25 @@ impl BatteryDevice for PowerSupplyDevice {
     }
 
     fn power_consumption(&self) -> Result<u64> {
+        // power consumption in µWh
         let power_path = self.device_path.join("power_now");
+        // current consumption in µA
+        let current_path = self.device_path.join("current_now");
+        // voltage in µV
+        let voltage_path = self.device_path.join("voltage_now");
 
         if power_path.exists() {
             Ok(read_file("battery", &power_path)?
                 .parse::<u64>()
                 .block_error("battery", "failed to parse power_now")?)
+        } else if current_path.exists() && voltage_path.exists() {
+            let current = read_file("battery", &current_path)?
+                .parse::<u64>()
+                .block_error("battery", "failed to parse current_now")?;
+            let voltage = read_file("battery", &voltage_path)?
+                .parse::<u64>()
+                .block_error("battery", "failed to parse voltage_now")?;
+            Ok(current * voltage)
         } else {
             Err(BlockError(
                 "battery".to_string(),
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -182,6 +182,10 @@ impl BatteryDevice for PowerSupplyDevice {
             ));
         };
 
+        // If the device driver uses the combination of energy_full, energy_now and power_now,
+        // all values (full, fill, and usage) are in Watts, while if it uses charge_full, charge_now
+        // and current_now, they're in Amps. In all 3 equations below the units cancel out and
+        // we're left with a time value.
         let status = self.status()?;
         match status.as_str() {
             "Full" => Ok(((full as f64 / usage) * 60.0) as u64),
@@ -573,6 +577,7 @@ impl Block for Battery {
                 Ok(time) => format!("{}:{:02}", time / 60, time % 60),
                 Err(_) => "×".into(),
             };
+            // convert µW to W for display
             let power = match self.device.power_consumption() {
                 Ok(power) => format!("{:.2}", power as f64 / 1000.0 / 1000.0),
                 Err(_) => "×".into(),
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -64,7 +64,7 @@ impl PowerSupplyDevice {
             ));
         }
 
-        // Read charge_full exactly once, if it exists.
+        // Read charge_full exactly once, if it exists, units are µAh
         let charge_full = if device_path.join("charge_full").exists() {
             Some(
                 read_file("battery", &device_path.join("charge_full"))?
@@ -75,7 +75,7 @@ impl PowerSupplyDevice {
             None
         };
 
-        // Read energy_full exactly once, if it exists.
+        // Read energy_full exactly once, if it exists. Units are µWh.
         let energy_full = if device_path.join("energy_full").exists() {
             Some(
                 read_file("battery", &device_path.join("energy_full"))?
@@ -135,6 +135,7 @@ impl BatteryDevice for PowerSupplyDevice {
     }
 
     fn time_remaining(&self) -> Result<u64> {
+        // Units are µWh
         let full = if self.energy_full.is_some() {
             self.energy_full.unwrap()
         } else if self.charge_full.is_some() {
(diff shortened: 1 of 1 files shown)


ammgws:
One thing I'm not sure about is how these changes affect this block:

```rust
       let status = self.status()?;
        match status.as_str() {
            "Full" => Ok(((full as f64 / usage) * 60.0) as u64),
            "Discharging" => Ok(((fill / usage) * 60.0) as u64),
            "Charging" => Ok((((full as f64 - fill) / usage) * 60.0) as u64),
            _ => {
                // TODO: What should we return in this case? It seems that under
                // some conditions sysfs will return 0 for some readings (energy
                // or power), so perhaps the most natural thing to do is emulate
                // that.
                Ok(0)
            }
        }
```

Since it seems that depending on the context, `usage` has two different values (power or current). Might need to add another if guard block if so.

GladOSkar:
Will test in ~4-5 hrs

GladOSkar:
Works for me.

I had never used the `{power}` placeholder but when i looked it up, my system actually uses `current_now` and `voltage_now` as well instead of `power_now`, which doesn't exist. (Latitude E7440).

`13.7W` is a reasonable value and it changes according to load, so i assume it's correct.

image

It does not work with `upower = false` though:

image

Lmk if you need anything else.

Cheers!

GladOSkar:
> One thing I'm not sure about is how these changes affect this block:
> 
> ```rust
>        let status = self.status()?;
>         match status.as_str() {
>             "Full" => Ok(((full as f64 / usage) * 60.0) as u64),
>             "Discharging" => Ok(((fill / usage) * 60.0) as u64),
>             "Charging" => Ok((((full as f64 - fill) / usage) * 60.0) as u64),
>             _ => {
>                 // TODO: What should we return in this case? It seems that under
>                 // some conditions sysfs will return 0 for some readings (energy
>                 // or power), so perhaps the most natural thing to do is emulate
>                 // that.
>                 Ok(0)
>             }
>         }
> ```
> 
> Since it seems that depending on the context, `usage` has two different values (power or current). Might need to add another if guard to check the current status before setting usage.

Oh wait - i remember writing that part of the code now exactly because my system doesn't have `energy`[Wh] and `power`[W] but `charge`[Ah] and `current`[A].

The idea was that `usage` always represents the usage in terms of the maximum value, independent of what units the device driver was using (Amps or Watts), since apparently the device driver always uses either one or the other for both the `now` and the `max` value.

For displaying the power drawn, this is not useful of course.

GladOSkar:
For reference, my original PR: #214

travankor:
Seems like power is returned in μW, while the upower driver returns the a value in W.

travankor:
> It does not work with upower = false though:

Are you using the sysfs driver?

GladOSkar:
Hey @travankor i revised my original test because i missed a couple of things. I usually use `upower` but now focused on `sysfs` because that's what this PR is for. I saw the μW issue as well.

src/blocks/battery.rs
```diff
@@ -167,10 +167,19 @@ impl BatteryDevice for PowerSupplyDevice {
...
                 .parse::<f64>()
                 .block_error("battery", "failed to parse power_now")?
+        } else if current_path.exists() && voltage_path.exists() {
+            let current = read_file("battery", &current_path)?
+                .parse::<f64>()
+                .block_error("battery", "failed to parse current_now")?;
+            let voltage = read_file("battery", &voltage_path)?
+                .parse::<f64>()
+                .block_error("battery", "failed to parse voltage_now")?;
+            current * voltage
```
GladOSkar, line 182: This is the part i would suggest to drop because it breaks ERT calculation on Amp-based device drivers

src/blocks/battery.rs
```diff
@@ -198,12 +207,25 @@ impl BatteryDevice for PowerSupplyDevice {
...
                 .parse::<u64>()
                 .block_error("battery", "failed to parse power_now")?)
+        } else if current_path.exists() && voltage_path.exists() {
+            let current = read_file("battery", &current_path)?
+                .parse::<u64>()
+                .block_error("battery", "failed to parse current_now")?;
+            let voltage = read_file("battery", &voltage_path)?
+                .parse::<u64>()
+                .block_error("battery", "failed to parse voltage_now")?;
+            Ok(current * voltage)
```
GladOSkar, line 228:
```suggestion
            Ok((current * voltage) / 1000000)
```
Because the result is in microwatts otherwise.

(This doesn't actually fully make sense, since microamps * microvolts should be picowatts, not microwatts. But the actual value suggests it's actually microwatts, so maybe current is in milliamps and voltage in millivolts?)

src/blocks/battery.rs
```diff
@@ -167,10 +167,19 @@ impl BatteryDevice for PowerSupplyDevice {
...
+            let current = read_file("battery", &current_path)?
+                .parse::<f64>()
+                .block_error("battery", "failed to parse current_now")?;
+            let voltage = read_file("battery", &voltage_path)?
+                .parse::<f64>()
+                .block_error("battery", "failed to parse voltage_now")?;
+            current * voltage
         } else if current_path.exists() {
             read_file("battery", &current_path)?
                 .parse::<f64>()
```
GladOSkar, line 185: This is the part i would suggest to drop because it breaks ERT calculation on Amp-based device drivers:

```suggestion
        let power_path = self.device_path.join("power_now");
        let current_path = self.device_path.join("current_now");
        let usage = if power_path.exists() {
            read_file("battery", &power_path)?
                .parse::<f64>()
                .block_error("battery", "failed to parse power_now")?
        } else if current_path.exists() {
            read_file("battery", &current_path)?
                .parse::<f64>()
```

src/blocks/battery.rs
```diff
@@ -167,10 +167,19 @@ impl BatteryDevice for PowerSupplyDevice {
 
         let power_path = self.device_path.join("power_now");
         let current_path = self.device_path.join("current_now");
+        let voltage_path = self.device_path.join("voltage_now");
```
GladOSkar, line 170:
```suggestion
```

src/blocks/battery.rs
GladOSkar, line 182: This is the part i would suggest to drop because it breaks ERT calculation on Amp-based device drivers:

```suggestion
```

src/blocks/battery.rs
```diff
@@ -198,12 +207,25 @@ impl BatteryDevice for PowerSupplyDevice {
...
         let power_path = self.device_path.join("power_now");
+        // current consumption in µA
+        let current_path = self.device_path.join("current_now");
+        // voltage in µV
+        let voltage_path = self.device_path.join("voltage_now");
 
         if power_path.exists() {
             Ok(read_file("battery", &power_path)?
                 .parse::<u64>()
                 .block_error("battery", "failed to parse power_now")?)
```
GladOSkar, line 220: I wonder if this is in the wrong unit dimension as well?

ammgws:
OK so I believe it should be good to go now.

@travankor Can you please test again?

GladOSkar:
FYI, works fine on my end now.

ammgws:
Thanks!

PR #659, opened 2020-05-10
battery: read remaining time from time_to_empty_now if it exists

Thra11:
My battery (cw2015-battery) does not support any of the current properties used to calculate the remaining time (doesn't have `energy_now` or `charge_now` in sysfs). However, it does have its own `time_to_empty_now`. This PR changes the `time_remaining` function to simply read from `time_to_empty_now` if it exists, and otherwise falls back to the existing calculation.

### Things to consider:
* Is `time_to_empty_now` commonly available? Is it a "standard" name? Are there similar files with slightly different names but the same purpose, which could be tried on other devices?
* When charging, `time_to_empty_now` is 0. There is no corresponding `time_to_full_now` for my device: Are there other devices which have both and should conditionally read one or the other based on charging status? 
* Are there any situations where `time_to_empty_now` exists alongside the files required for the original calculation and we would rather calculate the remaining time ourselves for some reason?

I have tested that this change works for my specific device. I am unable to test that currently working devices are unaffected.

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -135,68 +135,75 @@ impl BatteryDevice for PowerSupplyDevice {
     }
 
     fn time_remaining(&self) -> Result<u64> {
-        // Units are µWh
-        let full = if self.energy_full.is_some() {
-            self.energy_full.unwrap()
-        } else if self.charge_full.is_some() {
-            self.charge_full.unwrap()
+        let time_to_empty_now_path = self.device_path.join("time_to_empty_now");
+        if time_to_empty_now_path.exists() {
+            read_file("battery", &time_to_empty_now_path)?
+                .parse::<u64>()
+                .block_error("battery", "failed to parse time to empty")
         } else {
-            return Err(BlockError(
-                "battery".to_string(),
-                "Device does not support reading energy".to_string(),
-            ));
-        };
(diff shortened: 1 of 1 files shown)


ammgws:
> Is time_to_empty_now commonly available? Is it a "standard" name? 

Listed here so it's standard at least: https://github.com/torvalds/linux/blob/e99332e7b4cda6e60f5b5916cf9943a79dbef902/include/linux/power_supply.h#L151

> There is no corresponding time_to_full_now for my device: Are there other devices which have both and should conditionally read one or the other based on charging status?

Looks like it's at least possible for it to exist (POWER_SUPPLY_PROP_TIME_TO_FULL_NOW), so might be worth doing.

@GladOSkar, @travankor  Sorry to call you in again, but if you have any info it would be great to hear.

GladOSkar:
My system has neither `time_to_empty_now` nor `time_to_full_now`.

But since apparently devices with only these endpoints exist, this PR would probably make sense to include.

Tested and still works fine on my machine, code seems reasonable as well.

ammgws:
@Thra11 Could you please implement `time_to_full_now` as well?

Thra11:
I've created #665 which handles `time_to_full_now` as well, though I think the code probably needs some tidying.

PR #661, opened 2020-05-10
Fix clippy lints

carloabelli:
This fixes all but 1 clippy lint. Most of the changes are simple. Biggest change is breaking the net block update function up into smaller functions.

Perhaps it might be worth adding automatic clippy linting to CI?

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -709,7 +709,7 @@ impl Sound {
                 if self.bar {
                     self.text.set_text(format_percent_bar(volume as f32));
                 } else {
-                    self.text.set_text(format!("{}", text));
+                    self.text.set_text(text);
                 }
             } else {
                 self.text.set_text("");
@@ -724,7 +724,7 @@ impl Sound {
             self.text.set_text(if self.bar {
                 format_percent_bar(volume as f32)
             } else {
-                text.to_string()
+                text
             });
             self.text.set_state(State::Idle);
         }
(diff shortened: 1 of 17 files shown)


ammgws:
So all those `clone()`s were safe to remove in the end?
https://rust-lang.github.io/rust-clippy/master/#redundant_clone

Also, can you submit the net block code refactor as another PR?

src/blocks/sound.rs
```diff
@@ -724,7 +724,7 @@ impl Sound {
             self.text.set_text(if self.bar {
                 format_percent_bar(volume as f32)
             } else {
-                text.to_string()
+                text
```
ammgws, line 727: What lint was this one? Surprised I didn't see it before as I thought I had got all the trivial ones like this already.

carloabelli: this was also a redundant clone

carloabelli:
> So all those `clone()`s were safe to remove in the end?

They should be safe to remove. For most of them the function takes ownership and it can just move it into the next function.

> Also, can you submit the net block code refactor as another PR?

Done.

ammgws:
> Perhaps it might be worth adding automatic clippy linting to CI?

Yeah, though perhaps we should start with simple/conservative lints so new contributors don't get too discouraged at first.

carloabelli:
> Yeah, though perhaps we should start with simple/conservative lints so new contributors don't get too discouraged at first.

I see your point here. If clippy is enabled in CI, it would be good to limit lints to balance this.

I will say that in my experience the recommended fix for this lint has been correct. It can sometimes require some reasoning about how all the borrowing works to really understand why, but for a new contributor just going with the recommended fix, it should work almost every time.

ammgws:
Send a PR for enabling CI clippy lints?

Issue #663, opened 2020-05-11
Enable clippy linting in travis


carloabelli:
This is a tracking issue for enabling clippy lints in travis CI.

My current plan is to enable basic clippy linting so that `cargo clippy` passes. There is one remaining lint warning, but for now I will simply have clippy ignore it since it looks like it could be a more involved change.

After that it would be nice to:
- get `cargo clippy --all-targets --all-features` working
- remove the workaround described above

Before this is merged in, there should also be a discussion if any of the default lints should be omitted.

carloabelli:
Currently the only lints that will cause a failure are those marked deny by default in the correctness, style, complexity, and perf groups.

It could be a good idea to also disallow any warnings in the defaults groups as well.

The various lints can be explored here: https://rust-lang.github.io/rust-clippy/master/index.html

The various lints can be explored here: https://rust-lang.github.io/rust-clippy/master/index.html.

PR #664, opened 2020-05-11
Enable clippy linting in travis

carloabelli:
See #663.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,10 +28,13 @@ matrix:
     - rust: nightly
 env:
   - RUSTFLAGS="-D warnings"
+before_script:
+  - rustup component add rustfmt
+  - rustup component add clippy || cargo install --git https://github.com/rust-lang/rust-clippy/ --force clippy
 script:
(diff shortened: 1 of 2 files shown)


ammgws:
Can you move those code changes to another PR so we can keep this one just about the CI itself?

carloabelli:
@ammgws Sounds good. Once this is merged, I can put up another PR enabling the checks for --all-targets and --all-features (this requires some code changes to pass). Unless you would prefer to merge the code changes in first and then update this PR to include --all-targets and --all-features?

carloabelli:
One other thing to consider is how we want to handle clippy warnings. One possibility (the one currently implemented) is to fail on any warning (which will also fail for rustc warnings). The current script does not fail on rustc warnings so perhaps that is undesirable? Another solution is to make all the default clippy warnings errors, which will allow rustc warnings through, but fail for clippy warnings. Finally we can just ignore all warnings altogether.

ammgws:
Now that all the lints have been fixed, perhaps we could go for your suggestion as it would help keep future code in check as you mentioned.

carloabelli:
Okay in that case this PR should be good to go then

PR #665, opened 2020-05-11
battery: read remaining time from time_to_empty_now and time_to_full_now if they exist

Thra11:
Otherwise, fall back to original calculations for time_remaining.

w.r.t. the code, I think the general idea is ok (collect possible sources of information as `Options`/`Result`s, then calculate the time based on the charging status and available info sources), but I suspect the code could be cleaner. There is some potential to optimise by not reading more values once we've found a suitable source, but due to the fact there are two calculations (time to empty vs. time to full) which sometimes use the same values (`fill`, `full` & `usage`) it would add some complexity, or wouldn't handle situations where e.g. `time_to_empty_now` can be read directly but `time_to_full_now` cannot and needs to be calculated.

Also, I'm not sure what this line is supposed to do, so I've basically left it alone. Is there a situation where the status is "Full", but the remaining time (until what?) isn't zero?
```
"Full" => Ok(((full as f64 / usage) * 60.0) as u64),
```

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -135,16 +135,36 @@ impl BatteryDevice for PowerSupplyDevice {
     }
 
     fn time_remaining(&self) -> Result<u64> {
+        let time_to_empty_now_path = self.device_path.join("time_to_empty_now");
+        let time_to_empty = if time_to_empty_now_path.exists() {
+            read_file("battery", &time_to_empty_now_path)?
+                .parse::<u64>()
+                .block_error("battery", "failed to parse time to empty")
+        } else {
+            Err(BlockError(
+                "battery".to_string(),
+                "Device does not support reading time to empty directly".to_string(),
+            ))
+        };
+        let time_to_full_now_path = self.device_path.join("time_to_full_now");
+        let time_to_full = if time_to_full_now_path.exists() {
+            read_file("battery", &time_to_full_now_path)?
+                .parse::<u64>()
+                .block_error("battery", "failed to parse time to full")
+        } else {
+            Err(BlockError(
+                "battery".to_string(),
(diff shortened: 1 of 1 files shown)


GladOSkar:
Good point, i have no idea why that line is written that way

Thra11:
Assuming "Full" means the charger is connected but the charge/capacity is at 100%, then the two possible values for the remaining time are 0 (time to full, i.e. it will be full in 0 seconds) or infinity (time to empty, i.e. based on the current state, it will never be empty as it's not discharging). I can't see any benefit to trying to represent infinity, so I suggest we remove the line, and let the `"Full"` state be handled in the `_` bit, which returns 0. The 0 value will be hidden thanks to the other recent PR, so it shouldn't look weird or confusing.

GladOSkar:
Sounds good to me

ammgws:
I was hoping to setup my new laptop in order to test this but not sure when I can get around to it so hopefully some other laptop users could test run this!

GladOSkar:
Just tested, works fine

ammgws:
Out of interest which of the following does your laptop support?

time_to_empty
time_to_full
energy_full
charge_full
energy_now
charge_now
power_now
current_now

GladOSkar:
```
% ls -1 /sys/class/power_supply/BAT0/ 
hwmon0/
power/
alarm
capacity
capacity_level
charge_full
charge_full_design
charge_now
current_now
cycle_count
device -> ../../../PNP0C0A:00
manufacturer
model_name
present
serial_number
status
subsystem -> ../../../../../../class/power_supply
technology
type
uevent
voltage_min_design
voltage_now
```

ammgws:
Thanks!

PR #667, opened 2020-05-11
Replace i3ipc-rs with swayipc-rs

ammgws:
This allows future work such as adding marks support to the focused_window block,
and a sway driver for the keyboard_layout block. sway IPC is built on top of i3 IPC,
so anything that works in i3 also works in sway. However with this crate we can
access sway-only parts of the IPC such as the keyboard layout. Also, the maintainer
for i3ipc-rs seems to be AWOL at the moment.


euri10:
since this I get `failed to acquire connect to IPC`

debian buster
i3 version 4.16.1 (2019-01-27) © 2009 Michael Stapelberg and contributors

new to rust,  but it seems to me the way swayipc get_path tries to detect the i3 socket could use the same logic with `i3 --get-socketpath` as it does with `sway -get-socketpath` but I've no idea where to report this

ammgws:
Report here: https://github.com/JayceFayne/swayipc-rs

For now you could call i3status-rs by something like
I3SOCK=$(i3 --get-socketpath) i3status-rs

ammgws:
@euri10 This has been fixed upstream, please rebuild and try again.

euri10:
> @euri10 This has been fixed upstream, please rebuild and try again.

works fine, thanks for the very fast fix !

PR #672, opened 2020-05-13
focused_window block: show marks with brackets like sway does

ammgws:
Resolves https://github.com/greshake/i3status-rust/pull/532#issuecomment-581854186

@travankor

### DIFF
--- a/src/blocks/focused_window.rs
+++ b/src/blocks/focused_window.rs
@@ -92,8 +92,12 @@ impl ConfigBlock for FocusedWindow {
                                     }
                                     let mut marks = marks_original.lock().unwrap();
                                     if !e.container.marks.is_empty() {
-                                        *marks =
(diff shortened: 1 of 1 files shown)


travankor:
I don't know rust, but I think there might be a regression here. `focused_window` stopped updating after this.

travankor:
Seems like if you focus a marked window, the whole bar freezes and stops updating.

ammgws:
Argh I thought I had tested it already. I'll check it out

ammgws:
OK so I could reproduce that the focused_window block stops updating, but the rest of the bar/blocks still keep updating.

travankor:
To get the whole bar to freeze: 

1) open two terminals
2) `swaymsg mark 123` on one terminal
3) keep changing focus and the bar will freeze

ammgws:
Yeah that's what I did. Tried it again and still the same - only the focused_window block stops updating.

PR #680, opened 2020-05-14
focused_window block: allow show_marks to hide underscored marks

ammgws:
@travankor this should complete your feature request


travankor:
Seems like everything is working 👍

travankor:
@ammgws not sure if this is related or entirely separate, but seems like Xwayland windows causes  this block to freeze

ammgws:
Interesting, is that regardless of whether show_marks is set?

travankor:
seems like it is regardless of what value you have for `show_marks`

PR #684, opened 2020-05-14
Fix #89 support running a block only once

### LINKED ISSUE #89: make custom block command run once

PicoJr:
Fix #89 

```toml
interval = 2  # every 2 seconds
```
```toml
interval = "once" # only update once
```
```toml
# interval not provided, use default value
```

This PR modifies the `Block` trait and adds an `Update` enum:

```diff
    /// Forces an update of the internal state of the block.
-    fn update(&mut self) -> Result<Option<Duration>> {
+    fn update(&mut self) -> Result<Option<Update>> {
```

```rust
pub enum Update {
    Every(Duration), // update every <duration> seconds
    Once, // only update once
}
```

What do you think?

## How to try this PR

status.toml:

```toml
[[block]]
block = "custom"
command = "date"
interval = 2 # updated every 2 seconds

[[block]]
block = "custom"
command = "date"
interval = "once" # only updated once

[[block]]
block = "custom"
command = "date"
# no interval provided, updated every 10 seconds
```

`./target/debug/i3status-rs test.toml`

Then check the first block is updated every 2 seconds, the 2nd block only once and 3rd block every 10 seconds.

## TODO before merge

- [x] rebase
- [x] rename `Refresh` to `Update` ?
- [x] add doc
- [x] does it make sense to support running once for other blocks?

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -191,11 +191,15 @@ For further customisation, use the `json` option and have the shell command outp
 
 ### Examples
 
+Display temperature, update every 10 seconds:
+
 ```toml
 [[block]]
 block = "custom"
 command = ''' cat /sys/class/thermal/thermal_zone0/temp | awk '{printf("%.1f\n",$1/1000)}' '''
 ```
+
+Display temperature, update every 1 second, run `<command` when block is clicked:
 
 ```toml
 [[block]]
@@ -205,11 +209,22 @@ on_click = "<command>"
 interval = 1
 ```
 
+Use JSON output:
+
 ```toml
 [[block]]
 block = "custom"
 command = "echo '{\"icon\":\"weather_thunder\",\"state\":\"Critical\", \"text\": \"Danger!\"}'"
 json = true
+```
+
+Display kernel, update the block only once:
+
+```toml
+[[block]]
+block = "custom"
+command = "uname -r"
+interval = "once"
 ```
 
 ### Options
@@ -221,7 +236,7 @@ Key | Values | Required | Default
 `command` | Shell command to execute & display. | No | None
 `on_click` | Command to execute when the button is clicked. The command will be passed to whatever is specified in your `$SHELL` variable and - if not set - fallback to `sh`. | No | None
 `cycle` | Commands to execute and change when the button is clicked. | No | None
-`interval` | Update interval, in seconds. | No | `10`
+`interval` | Update interval, in seconds (or `"once"` or `"Once"` to update only once). | No | `10`
 `json` | Use JSON from command output to format the block. If the JSON is not valid, the block will error out. | No | `false`
 
 
--- a/blocks.md
+++ b/blocks.md
@@ -199,7 +199,7 @@ block = "custom"
 command = ''' cat /sys/class/thermal/thermal_zone0/temp | awk '{printf("%.1f\n",$1/1000)}' '''
 ```
 
-Display temperature, update every 1 second, run `<command` when block is clicked:
+Cycle between "ON" and "OFF", update every 1 second, run `<command>` when block is clicked:
 
 ```toml
 [[block]]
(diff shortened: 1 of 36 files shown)


src/blocks.rs
```diff
@@ -76,6 +76,24 @@ use crate::input::I3BarEvent;
 use crate::scheduler::Task;
 use crate::widget::I3BarWidget;
 
+#[derive(Clone, Debug, PartialEq)]
+pub enum Refresh {
```
ammgws, line 80: If we go with this then perhaps Update might be better, as you mentioned.

PicoJr: I'll rename `Refresh` to `Update` then =)

blocks.md
```diff
@@ -218,7 +233,7 @@ Key | Values | Required | Default
 `command` | Shell command to execute & display. | No | None
 `on_click` | Command to execute when the button is clicked. The command will be passed to whatever is specified in your `$SHELL` variable and - if not set - fallback to `sh`. | No | None
 `cycle` | Commands to execute and change when the button is clicked. | No | None
-`interval` | Update interval, in seconds. | No | `10`
+`interval` | Update interval, in seconds (or `"once"` or `"Once"` to update only once). | No | `10`
```
ammgws, line 236: Let's just go with "once".

PicoJr: Sure =)

ammgws:
> does it make sense to support running once for other blocks?

| Block | Has update interval? | Run once sensible? |
| ---- | ----| ----|
|Backlight | no | - |
|Battery|yes| no |
|Bluetooth|no| -|
|CPU Utilization|yes|no|
|Custom| yes | yes |
|Custom DBus|no | - |
|Disk Space| yes | maybe |
|Focused Window|no | - |
|IBus| no | - |
|Keyboard Layout| yes | maybe |
|Load| yes | no |
|Maildir|yes | no |
|Memory| yes | no |
|Music| no | - |
|Net| yes | no |
|NetworkManager| no | - |
|Notmuch| yes | no |
|Nvidia Gpu| yes | maybe |
|Pacman| yes | no |
|Pomodoro |yes | no |
|Sound| yes | no |
|Speed Test| yes | maybe |
|Taskwarrior|yes | no |
|Temperature| yes | no |
|Time| yes | no |
|Toggle|yes | yes |
| Uptime | yes | no |
|Watson| yes | no |
|Weather|yes | maybe |
|Xrandr| yes | maybe |

ammgws:
I was mulling whether we should go with a mixed keyword/numeric option like you've proposed here or just purely numeric based with `-1` representing "run once". However I just checked i3blocks and they use mixed so I guess it's OK. If we ever introduce any other interval modes, having a named option would be better than trying to remember what `-2` means.

i3blocks interval options:
- \<non-zero number\>
- `-1` or `once`
- `-2` or `repeat`
- `-3` or `persist`

PicoJr:
Thanks for your feedback: I'll rebase and resolve conflicts.

PR #688, opened 2020-05-17
Fix CPU format when per_core is true

nalysius:
When the option "per_core" was true the last core
had two percent signs because one was added for each core
and another was added after the "utilization" placeholder.

When "per_core" is true the last percent sign after "utilization"
is no longer added.

### DIFF
--- a/src/blocks/cpu.rs
+++ b/src/blocks/cpu.rs
@@ -105,6 +105,8 @@ impl ConfigBlock for Cpu {
     ) -> Result<Self> {
         let format = if block_config.frequency {
             "{utilization}% {frequency}GHz".into()
+        } else if block_config.per_core {
+            "{utilization}".to_owned()
         } else {
             block_config.format
         };


ammgws:
Wait what does your config look like? This is mine and there's no double percent signs:

image

```toml
[[block]]
block = "cpu"
format = "{utilization}"
per_core = true
```

nalysius:
Thank you for this quick answer. My config looks like this:

```toml
[[block]]
block = "cpu"
interval = 5
per_core = true
```

I use the default format, wich adds the percent sign after the utilization. It's ok when per_cpu is false because utilization is only the number, but when per_cpu is true the percent signs are added after each number so the last percent sign added by `default_format()` should not be there.

ammgws:
OK that makes sense. Can you fix the build failure by collapsing the if block as shown by clippy?

ammgws:
Now just needs rustfmt/cargo fmt

ammgws:
Thanks!

nalysius:
Thank you for the tips! I got confused with the Travis messages.

PR #690, opened 2020-05-17
weather block: add autolocate option

### LINKED ISSUE #367: Automatically keep Weather block location up-to-date
At the moment you need to hard-code your City ID to use the OpenWeatherMap API, so when you're traveling the weather information will obviously be incorrect.

As a possible solution, the `GeoClue2` program does provide [a D-Bus API](https://www.freedesktop.org/software/geoclue/docs/ref-dbus.html) for reporting/updating the user's location. Not everyone has GeoClue installed, but we could make use of it if present. The location information provided by GeoClue is in the form of longitude/latititude, which the OpenWeatherMap API [supports](https://openweathermap.org/current).

ammgws:
Resolves #367 

This uses the geolocation service mentioned in https://github.com/greshake/i3status-rust/issues/367#issuecomment-489889913 to automatically change the location used for retrieving weather info

@kflak If you have some time could you please test this? It works fine for me but more QA wouldn't hurt.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -1039,6 +1039,8 @@ Creates a block which displays local weather and temperature information. In ord
 
 Configuring the Weather block requires configuring a weather service, which may require API keys and other parameters.
 
+If using the `autolocate` feature, set the block update interval such that you do not exceed ipapi.co's free daily limit of 1000 hits.
+
 ### Examples
 
 Show detailed weather in San Francisco through the OpenWeatherMap service:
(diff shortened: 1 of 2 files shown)


kflak:
Hey there, this seems to do the trick from my location! Haven't been able to travel around for the last couple of days to battletest it properly, but it's delivering the right kind of info. Thanks a lot for this :-)

ammgws:
Thanks for the feedback!

Issue #676, opened 2020-05-13
Add per widget wide/narrow switching behavior 


matclab:
My bar is quite full and some time has not enough space to display everything.

Thus, it would be useful to enable switching between a full width widget and between a narrower widget (with less information).
For example, the net block may display IP addresses only in the wide mode.
The time block may have two format string, on for the wide mode and on for the narrow mode.
…

I propose using the right or middle mouse button to switch between mode.

What do you think about this proposal ?

ammgws:
Haha, I just thought of this tonight as well. The i3bar JSON schema actually has a field called `short_text` which is used for this exact situation. I was going to test it out with a random block to see if/how it works. For reference we only use the `full_text` field at the moment:

https://i3wm.org/docs/i3bar-protocol.html
> full_text
> 
>     The full_text will be displayed by i3bar on the status line. This is the only required key. If full_text is an empty string, the block will be skipped.
> 
> short_text
> 
>     Where appropriate, the short_text (string) entry should also be provided. It will be used in case the status line needs to be shortened because it uses more space than your screen provides. For example, when displaying an IPv6 address, the prefix is usually (!) more relevant than the suffix, because the latter stays constant when using autoconf, while the prefix changes. When displaying the date, the time is more important than the date (it is more likely that you know which day it is than what time it is).

travankor:
Some blocks already have a collapsed mode that can be collapesed or expanded on click. I wonder how this approach compares to short_text?

matclab:
Oh ! I didn't used any of this block (music & temperature), so I didn't know.
I guess it is the same feature which just as to be ported to other blocks… 
I'll submit a PR using the same terminology.

ammgws:
Let's investigate `short_text` first, since it means collapsing would be done automatically, and would free up a click action for the blocks. Some blocks might already be using all clicks, or might in the future (such as the `custom` block), for other functionality.

matclab:
Ok. But as a user, I'd like to be able to switch back to `full_text` in order to see masked information ("what is my IP, what is the date?") if possible (if I have not mapped the click to other functionality in the configuration or if the block has not already mapped all clicks).
I may give a try, but with no promise about delivery :-) 
@ammgws do not hesitate to give a try !

GladOSkar:
I agree that some sort of hybrid would be most desirable - Both automatic (when screen width is reached) and manual (on click) truncation/shortening. No idea how that would be possible though.

ammgws:
[...]

PR #693, opened 2020-05-17
Allow `short_text` to be set for Custom block

ammgws:
One of the things mentioned here: https://github.com/greshake/i3status-rust/issues/676

When using the `json` option with the `custom` block, you can now also set `short_text` so that i3/sway can automatically shrink the block if it would otherwise overflow the status bar.


ammgws:
Closing in favour of #703

PR #694, opened 2020-05-17
pomodoro: store datetime instead of using ticks, add nag config

GChicha:
I tried to use the pomodoro block with the hide mode enabled in sway, however I notice that it didn't update while the bar is not visible. I've made some fixes to the block, changed the behavior to not update a counter, now it stores the `DateTime` when the pomodoro starts.

I still have some problems calling `swaynag` or `i3-nagbar` when the bar is not visible.

### DIFF
--- a/src/blocks/pomodoro.rs
+++ b/src/blocks/pomodoro.rs
@@ -1,4 +1,5 @@
-use std::time::Duration;
+use std::fmt;
+use std::time::{Duration, Instant};
 
 use crossbeam_channel::Sender;
 use serde_derive::Deserialize;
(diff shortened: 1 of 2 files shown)


ammgws:
@ghedamat Could you please confirm?

ghedamat:
Yup. I'll review and test soon and report back!

GChicha:
I realize now that I forgot to document the new config I added to the block.
As I understand i3status is not running when it's in hide mode. Therefore, it is not possible to call the nagbar in the meantime. Is there any workaround to do this?

ammgws:
> Is there any workaround to do this?

https://github.com/i3/i3/issues/3242#issuecomment-504565232

src/blocks/pomodoro.rs
```diff
@@ -15,76 +17,107 @@ use crate::widget::I3BarWidget;
...
+            ),
+            State::OnBreak(_) => write!(
+                f,
+                "\u{2615} {}:{:02}",
+                self.elapsed().as_secs() / 60,
+                self.elapsed().as_secs() % 60
+            ),
+            State::Paused(duration) => write!(
+                f,
+                "\u{2615} {}:{:02}",
```
ghedamat, line 61: I would have kept this to the pause icon `\u{f04c}`

GChicha:
I made a small change, instead of using `DateTime` now it uses `Instant`. I don't expect any differences from datetime behavior.

src/blocks/pomodoro.rs
GChicha: Had copied the wrong one. fixed

ammgws:
Thanks!

PR #700, opened 2020-05-19
watson: allow to configure show_time and interval

schelmo:


ammgws:
Do docs need updating?

schelmo:
show_time was already there, but i added interval into the docs

PR #701, opened 2020-05-20
Prevent bar being stopped by using signal other than SIGSTOP

ammgws:
@GChicha Can you see if this makes any difference with regards to the bar being stopped while hidden?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -109,7 +109,13 @@ fn main() {
 #[allow(unused_mut)] // TODO: Remove when fixed in chan_select
 fn run(matches: &ArgMatches) -> Result<()> {
     // Now we can start to run the i3bar protocol
-    print!("{{\"version\": 1, \"click_events\": true}}\n[");
+    print!(
+        "{}",
+        format!(
+            "{{\"version\": 1, \"click_events\": true, \"stop_signal\": {}}}\n[",
+            nix::sys::signal::Signal::SIGUSR2
+        )
+    );
 
     // Read & parse the config file
     let config_path = match matches.value_of("config") {
--- a/src/main.rs
+++ b/src/main.rs
@@ -118,7 +118,7 @@ fn run(matches: &ArgMatches) -> Result<()> {
     let initialise = if matches.is_present("never-pause") {
         format!(
             "\"version\": 1, \"click_events\": true, \"stop_signal\": {}",
-            nix::sys::signal::Signal::SIGUSR2 as i8
+            nix::sys::signal::Signal::SIGCONT as i8
         )
     } else {
         "\"version\": 1, \"click_events\": true".to_string()
--- a/src/main.rs
+++ b/src/main.rs
@@ -55,6 +55,12 @@ fn main() {
                 .takes_value(false),
         )
         .arg(
+            Arg::with_name("never-pause")
+                .help("Ignore any attempts by i3 to pause the bar when hidden/fullscreen")
+                .long("never-pause")
+                .takes_value(false),
+        )
+        .arg(
             Arg::with_name("one-shot")
                 .help("Print blocks once and exit")
                 .long("one-shot")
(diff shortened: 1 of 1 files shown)


GChicha:
I just checked and it seems to work perfectly. The bar now runs even when hidden.

ammgws:
Now to think of any pitfalls of doing this...

Perhaps we could make it opt-in via a command line argument?

GChicha:
I was wondering about it too. Maybe it's only bad when someone was expecting that stopping behavior (like a great computational task when the bar is visible). In my use case, I just hide the status bar for appearance reasons

I was thinking about a flag in the `config.toml`, but a opt-in via command line seems more versatile.

ammgws:
@GChicha Could you please try with the updated code?

GChicha:
Didn't work for me. Maybe some differences between i3 and sway. I'm getting `error reading from status command` at the status bar but the i3status-rs is running.

If I run without your argument I'm getting only a raw json at status bar. I have no clue what's going on.

src/main.rs
```diff
@@ -109,7 +115,15 @@ fn main() {
...
 fn run(matches: &ArgMatches) -> Result<()> {
     // Now we can start to run the i3bar protocol
-    print!("{{\"version\": 1, \"click_events\": true}}\n[");
+    let initialise = if matches.is_present("never-pause") {
+        format!(
+            "{{\"version\": 1, \"click_events\": true, \"stop_signal\": {}}}\n[",
+            nix::sys::signal::Signal::SIGUSR2 as i8
+        )
+    } else {
+        "{{\"version\": 1, \"click_events\": true}}\n[".to_string()
```
GChicha, line 124:
```suggestion
        "{\"version\": 1, \"click_events\": true}\n[".to_string()
```
I think you don't need to escape `{` here

src/main.rs
```diff
@@ -109,7 +115,15 @@ fn main() {
 #[allow(unused_mut)] // TODO: Remove when fixed in chan_select
 fn run(matches: &ArgMatches) -> Result<()> {
     // Now we can start to run the i3bar protocol
-    print!("{{\"version\": 1, \"click_events\": true}}\n[");
+    let initialise = if matches.is_present("never-pause") {
+        format!(
+            "{{\"version\": 1, \"click_events\": true, \"stop_signal\": {}}}\n[",
+            nix::sys::signal::Signal::SIGUSR2 as i8
```
GChicha, line 121:
```suggestion
            nix::sys::signal::Signal::SIGSTOP as i8
```
I thought this should be SIGSTOP

GChicha: Even with SIGSTOP it does not prevent the pause
I remember you asked me to test with signal 17 it seems to be associated with SIGCHLD

ammgws: @GChicha SIGSTOP is what i3 sends by default and is what causes i3status-rust to pause, and we should not ignore that signal if it is sent. However i3 lets us customising which signal it sends, so we can set it to something other than SIGSTOP to effectively prevent pausing from happening. I plan on using SIGUSR1 for reloading the config so using SIGUSR2 here for this one. I meant to use it in the original commit but accidently put 17 instead of 12.

src/main.rs
ammgws: It chucks a sad if you don't escape it, but then it somehow ends up printing both... no idea why but I changed the code around to avoid that.

ammgws:
Should run now

GChicha:
Oh sorry, I forgot to answer you. Unfortunately I only could run using SIGCHLD, other signals didn't work.

GChicha:
I checked the `ps auxf` output and seems that even when the bar is visible `i3status-rs` does not leave the Stopped state (T)
```
guilher+   53710  0.0  0.0  18684  7476 tty1     S+   19:47   0:00 swaybar -b bar-0
guilher+   53776  0.0  0.0 151824  9356 tty1     Tl+  19:47   0:00  \_ /home/guilherme/.cargo/bin/i3status-rs --never-pause /home/guilherme/.config/i3status/co
```

ammgws:
OK so TIL the default action for SIGUSR1/2 is to terminate the process...
I will need to either capture the signal or change it to something else that can be used.

>     The SIGUSR1 and SIGUSR2 signals are set aside for you to use any way you want. They’re useful for simple interprocess communication, if you write a signal handler for them in the program that receives the signal.
> 
>     The default action is to terminate the process.

ammgws:
@GChicha I just changed it to use SIGCONT instead. When SIGCONT is sent to a running process it is ignored by default, which is what we want. Could you please try again?

In the meantime I may try talk to the i3 devs to at least let us turn this feature off so we don't need to come up with hacks like this.

ammgws:
It appears to be working for me.

GChicha:
It looks good now. Thank you for your help. And sorry for the late reply.

ammgws:
I also mentioned this over at [i3](https://github.com/i3/i3/issues/4110) and it looks like this PR will end up being the proper solution so I will merge this later on. (Just have to change the signal to '0')

Issue #697, opened 2020-05-18
Feature request: Enhanced on_click, on_right_click, on_scroll etc commands for every block


eater:
It would be very helpful to be able to add custom shell commands to every block. I'm thinking in particular of the music block. I'd love to be able to open the music player on_click, and adjust its volume with scrolling commands.

ammgws:
If you have the time could you please go through our current blocks list and make a table like the one below? So we can get an overview of the current situation with mapped click actions.

| block | left click | middle click | right click | scroll
| --- | --- | --- | --- | --- 
| foo | in use | free | free | in use | free

eater:
Happy to help! I just searched through the docs so hopefully I got everything:

| block             | left click       | middle click  | right click  | scroll 	           |
|-----------        |------------      |-------------- |------------- |--------	           |
| Backlight         |      X           |      X        |      X       |   brightness       |
| Battery           |      X           |      X        |      X       |      X             |
| Bluetooth         |      X           |      X        |      connect |      X             |
| CPU Utilization   |      X           |      X        |      X       |      X             |
| Custom            |  command         |      X        |      X       |      X             |
| Custom DBus       |      X           |      X        |      X       |      X             |
| Disk Space        |      X           |      X        |      X       |      X             |
| Docker            |      X           |      X        |      X       |      X             |
| Focused Window    |      X           |      X        |      X       |      X             |
| IBus              |      X           |      X        |      X       |      X             |
| Keyboard Layout   |      X           |      X        |      X       |      X             |
| Load              |      X           |      X        |      X       |      X             |
| Maildir           |      X           |      X        |      X       |      X             |
| Memory            | switch swap/mem  |      X        |      X       |      X             |
| Music             | collapsed cmd    |      X        |      X       |      X             |
| Net               |      X           |      X        |      X       |      X             |
| NetworkManager    | command          |      X        |      X       |      X             |
| Notmuch           |      X           |      X        |      X       |      X             |
| Nvidia Gpu        | change output    |      X        |      X       | fan speed          |
| Pacman            |      X           |      X        |      X       |      X             |
| Pomodoro          |      X           |      X        |      X       |      X             |
| Sound             |      X           |      X        | mute         |      volume        |
| Speed Test        |      X           |      X        |      X       |      X             |
| Taskwarrior       | update number    |      X        | toggle filter|      X             |
| Temperature       |      X           |      X        |      X       |      X             |
| Time              | command          |      X        |      X       |      X             |
[...]

PR #703, opened 2020-05-21
Add short_text management + example for time and net block

matclab:
This PR propose to add short_text management as a response to #676. 

The features are:
- a widget may set full_text and short_text which are then automatically used by i3bar depending upon the available space in the bar (this is the default behavior) — this is similar to PR #693 ,
- the user may right or middle click on the block to switch from the default behavior to a widget where both full_text and shot_text are set to either the wide text or the narrow text.

This allow the user to enforce a widget view (full, short or default) by clicking or by configuration.

Does this behavior suites you ?

NB. I'm a rust and i3 beginner, so, do not hesitate to review and suggest improvement.

### DIFF
--- a/src/blocks/time.rs
+++ b/src/blocks/time.rs
@@ -24,7 +24,6 @@ use crate::widgets::button::ButtonWidget;
 
 pub struct Time {
     widget: ButtonWidget,
-    time: chrono::DateTime<Local>,
     id: String,
     update_interval: Duration,
     format: String,
@@ -105,26 +104,26 @@ impl TimeConfig {
 }
 
 impl Time {
-    fn format<'a>(& self, format: &'a str) -> Result<DelayedFormat<StrftimeItems<'a>>> {
-        let time = match &self.locale {
+    fn format<'a>(& self, time: &chrono::DateTime<Local>, format: &'a str) -> Result<DelayedFormat<StrftimeItems<'a>>> {
+        let formatted_time = match &self.locale {
             Some(l) => {
                 let locale: Locale = l
                     .as_str()
                     .try_into()
                     .block_error("time", "invalid locale")?;
                 match self.timezone {
-                    Some(tz) => self.time
+                    Some(tz) => time
                         .with_timezone(&tz)
                         .format_localized(format, locale),
-                    None => self.time.format_localized(format, locale),
+                    None => time.format_localized(format, locale),
                 }
             }
             None => match self.timezone {
-                Some(tz) => self.time.with_timezone(&tz).format(format),
-                None => self.time.format(format),
+                Some(tz) => time.with_timezone(&tz).format(format),
+                None => time.format(format),
             }
         };
-        Ok(time)
+        Ok(formatted_time)
     }
 }
 
@@ -138,7 +137,6 @@ impl ConfigBlock for Time {
     ) -> Result<Self> {
         let i = pseudo_uuid();
         Ok(Time {
-            time: Local::now(),
             id: i.clone(),
             format: block_config.format,
             format_short: block_config.format_short,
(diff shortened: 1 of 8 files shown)


src/blocks/time.rs
```diff
@@ -78,44 +97,59 @@ impl ConfigBlock for Time {
...
     }
 }
 
 impl Block for Time {
     fn update(&mut self) -> Result<Option<Update>> {
-        let time = match self.timezone {
-            Some(tz) => Utc::now().with_timezone(&tz).format(&self.format),
-            None => Local::now().format(&self.format),
+        self.time = match self.timezone {
+            Some(tz) => Local.from_utc_datetime(&Utc::now().with_timezone(&tz).naive_local()),
```
ammgws, line 119: Unrelated change? If so please open a new PR.

ammgws:
Middle click might be the best bet for this as no block uses it:
https://github.com/greshake/i3status-rust/issues/697#issuecomment-632134634

src/blocks/time.rs
matclab: Oups, it was a leftover of a first try. I'll push a commit that remove self.time. However I need to store the datetime instead of the formated string as the datetime is then formatted differently for short_text and full_text

src/blocks/time.rs
```diff
@@ -78,44 +97,59 @@ impl ConfigBlock for Time {
...
+                match e.button {
+                    MouseButton::Left => {
+                        if let Some(ref cmd) = self.on_click {
+                            spawn_child_async("sh", &["-c", cmd])
+                                .block_error("time", "could not spawn child")?;
+                        }
+                    }
+                    MouseButton::Right | MouseButton::Middle => {
+                        self.width = self.width.next();
+                        let _ = self.update();
```
matclab, line 142: @ammgws Is it better to do the update via a `Sender` like it is done in custom block? 
```
            self.tx_update_request.send(Task {
                id: self.id.clone(),
                update_time: Instant::now(),
            })?;
```

src/widgets/button.rs
```diff
@@ -60,7 +81,16 @@ impl ButtonWidget {
     }
 
     pub fn set_text<S: Into<String>>(&mut self, content: S) {
-        self.content = Some(content.into());
+        let content = Some(content.into());
```
matclab, line 84: I guess that this function should probably always set both `content` and `short_content` to keep existing blocks working as they do now (else, we may have blocks that get displayed with the `short_text` being the `full_text` that was given at initialisation instead of the actual `full_text`)

ammgws:
Sorry I am not sure if I can review this soon. Have you been using it so far?

matclab:
Hello,

There's no hurry.

Yes I use it daily. It works quite well but the three state clicking
behaviour is a bit frustrating (the user doesn't know on which state is the
block, so the user shall sometimes click twice to see a change). For now I
don't see how I can do better and I think it is better than having some
blocks not displayed at all.

I'm also a bit frustrated by i3bar behaviour which is : full_text for every
blocks or short_text for every blocks. I would have prefer that it used
short_text only for the minimum of needed blocks.

Anyway I think the PR need some documentation and perhaps I can also port
the behaviour to other blocks.

ammgws:
> I'm also a bit frustrated by i3bar behaviour which is : full_text for every
> blocks or short_text for every blocks. I would have prefer that it used
> short_text only for the minimum of needed blocks.

It might be worth looking at the i3 source to see how it is handling this.
It might be able to be fixed trivially.

matclab:
Some progress information : I'm still working on this, but struggling a little with GraphWidget not displaying shorttext. Some more investigations needed.

ammgws:
I opened a feature request over at i3: https://github.com/i3/i3/issues/4113

Feel free to add to the (lack of) discussion over there since it might lead to something!

ammgws:
Looks like our feature request was accepted by i3

matclab:
close in favor of #1207

PR #704, opened 2020-05-23
net block: improve throughput formatting

### LINKED ISSUE #240: Net format
Roboto Mono
Noto Sans

The net block has a few problems:

 - Takes too much space
 - Varies in width (particularly with prettier proportional fonts like `Noto Sans Regular`)

The point of a status bar is to get a quick, rough estimate of usage. This means 2-3 digits of precision, e.g. 

    4.20G
    0.42G
    42.0M
    4.20M
    0.42M
    42.0K
    4.20K
    0.42K
    4.20B
    0.42B
    0.04B
    0.00B

The aforementioned representations do not (significantly) vary in width with proportional fonts.

----

I propose a format specification:

    [[block]]
    name = "net"
    format = "{M.3}"

where:

 - `M` represents the limiting precision. (Anything below `0.01M` is labeled `0.00M`.)
 - `.3` represents 3 digits of precision.

That is:

    4.20G  <- Not really necessary...
    0.42G  <- Not really necessary... 
    42.0M
    4.20M
    0.42M
    0.04M
    0.00M

ammgws:
(Mostly?) Resolves #240. 

@GladOSkar Thanks for doing the legwork on this one!

This also uncovered a calculation error with `use_bits`, which is now fixed.

Comparing the new default format (max 3 digits, minimum unit K):
```
new format: '46.8M'	 with use_bits: '0.37Gb'
cur format: '   46M'	 with use_bits: '  368M'

new format: '6.81M'	 with use_bits: '54.5Mb'
cur format: '    6M'	 with use_bits: '   48M'

new format: '0.81M'	 with use_bits: '6.51Mb'
cur format: '  813k'	 with use_bits: '    0M'

new format: '13.6K'	 with use_bits: '0.11Mb'
cur format: '   13k'	 with use_bits: '  104k'
```

In my opinion this is now way better than before. The only caveat may be when using `use_bits` as in the first example above - perhaps we should display in megabits instead of gigabits for that case. However I will leave that for a future PR.

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -1,3 +1,4 @@
+use std::fmt;
 use std::fs::read_to_string;
 use std::fs::OpenOptions;
 use std::io::prelude::*;
@@ -376,11 +377,34 @@ pub struct Net {
     tx_bytes: u64,
     rx_bytes: u64,
     use_bits: bool,
+    speed_min_unit: Unit,
+    speed_digits: usize,
     active: bool,
     hide_inactive: bool,
     hide_missing: bool,
     last_update: Instant,
     on_click: Option<String>,
+}
+
+#[derive(Copy, Clone, Debug, Deserialize)]
+pub enum Unit {
+    B,
+    K,
+    M,
+    G,
+    T,
+}
+
+impl Default for Unit {
+    fn default() -> Self {
+        Unit::K
+    }
+}
+
+impl fmt::Display for Unit {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
 }
 
 #[derive(Deserialize, Debug, Default, Clone)]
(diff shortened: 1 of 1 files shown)


GladOSkar:
Wow, i wrote that almost 2 years ago :D Not quite what i'd call beautiful code but it does the job

ammgws:
If you're up for beautifying it then please go ahead!

GladOSkar:
I played around with it for a while and it doesn't actually get a lot better so... meh ¯\_(ツ)_/¯

GladOSkar:
I wonder if this could be adapted into the speedtest block as well, because the formatting of that one annoys me to the point that i always patch it before updating `i3status-rs`... But the values are obtained in a very different way there

ammgws:
I created a new issue for that.

By the way do you have any other improvements you'd like to share? (or maybe put your forked repo online?)

GladOSkar:
No, it's really just a hardcoded removal of everything after the decimal points in the speedtest block

PR #707, opened 2020-05-23
speedtest block: improve throughput formatting

### LINKED ISSUE #706: speedtest block: possible improvements to speed format?
From https://github.com/greshake/i3status-rust/pull/704#issuecomment-633022351
> I wonder if (the net block speed formatting) could be adapted into the speedtest block as well, because the formatting of that one annoys me to the point that i always patch it before updating i3status-rs... But the values are obtained in a very different way there

These are the values `speedtest-cli` outputs and i3status-rs then extracts and displays as-is:

```
>speedtest-cli --simple
Ping: 17.058 ms
Download: 977.85 Mbit/s
Upload: 876.94 Mbit/s

>speedtest-cli --simple --bytes
Ping: 16.492 ms
Download: 91.13 Mbyte/s
Upload: 98.73 Mbyte/s
```

@GladOSkar I'll see what we can do

ammgws:
Resolves #706 

@GladOSkar Just a quick mockup for now, is this what you had in mind?

TODO
- move common code to util.rs
- update docs

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -18,7 +18,7 @@ use crate::errors::*;
 use crate::input::{I3BarEvent, MouseButton};
 use crate::scheduler::Task;
 use crate::subprocess::spawn_child_async;
-use crate::util::{escape_pango_text, format_percent_bar};
+use crate::util::{escape_pango_text, format_percent_bar, format_speed};
 use crate::widget::I3BarWidget;
 use crate::widgets::button::ButtonWidget;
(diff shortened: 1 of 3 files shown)


GladOSkar:
Yeah pretty much. I wonder if it would make sense to provide an option for a format string here as well instead of always showing all 3 values? But this is already way better

ammgws:
Yeah format string would be nice. I think all blocks should be done that way.

GladOSkar:
@ammgws i think the block documentation is still missing the new options?

PR #714, opened 2020-05-30
Add icons option to Disk Space block

jamesmcm:
Adds icons option to use hard disk icon instead of alias in the DiskSpace block.

Tested with FontAwesome4 (but I looked up glyph values for the others, to find comparable icons).

This is my first PR, please let me know if there is anything else to check!


ammgws:
Maybe we could add a `format` string instead and have qualifiers `{icon}`, `{alias}`, `{path}`, `{available}` etc.

jamesmcm:
Sure, is there another block which has an example of this?

ammgws:
Search blocks.md for "format", there are a few that could be used as a reference.

ammgws:
Let me know if you need some help

jamesmcm:
@ammgws done :slightly_smiling_face: 

Example:
example

I also made some changes to the way the alert state is triggered too, to allow for triggering on an absolute amount used, etc.

jamesmcm:
@ammgws the only thing I dislike is that the 2 decimal places is hardcoded in, rather than being part of FormatTemplate, at the moment I see it only pushes the pre-formatted values matching keys: https://github.com/greshake/i3status-rust/blob/735fc75dc1daf678b68feac4b6548ae1ec4e67e3/src/util.rs#L434

But maybe that could be extended to allow for stuff like `{free:.0}` etc. in the config.

GladOSkar:
I was also wondering if decimal control could be integrated into the format string.

When thinking about things like the `net` or the `speedtest` block, it would be neat to be able to define the precision separately for the different displayed values. Ideally the syntax would allow to define all 3 arguments of the `format_speed()` function added in #704 / #707, like for example

`Free: {free:3:G:B}` for `Free: 21.9GB`

or

`Uplink: {up:2:M:b}/s` for `Uplink: 50Mb/s`

blocks.md
```diff
@@ -288,12 +290,27 @@ Key | Values | Required | Default
 ----|--------|----------|--------
 `path` | Path to collect information from | No | `"/"`
 `alias` | Alias that is displayed for path | No | `"/"`
-`info_type` | Currently supported options are `"available"`, `"free"`, `"total"` and `"used"` | No | `"available"`
+`icon` | Whether to prepend the block with the hard disk icon | No | `false`
```
ammgws, line 293: We can remove this config option by allowing `icon` to be set in the format string.
See the KDEConnect block for something similar.

src/blocks/disk_space.rs
```diff
@@ -97,14 +108,6 @@ pub struct DiskSpaceConfig {
     /// Diskspace alert in GiB (red)
     #[serde(default = "DiskSpaceConfig::default_alert")]
     pub alert: f64,
-
-    /// Show percentage
-    #[serde(default = "DiskSpaceConfig::default_show_percentage")]
```
ammgws, line 102: Is it possible to keep this and show_bar so we don't break existing configs? We can mark it as deprecated and they can be removed in a later release. The CPU and battery blocks do something like that.

blocks.md
jamesmcm: I originally did it that way, but does it make sense since we have the `.with_icon()` method and the icon would always be at the front anyway?

It's a simple change in any case, I just didn't see much of a use case in putting the icon anywhere but the front (in which case we can use the existing method, and do it outside the `update()` call.

src/blocks/disk_space.rs
jamesmcm: Sure, I guess we could override the default format string in this case, unless it is manually set.

ammgws: Yeah that work work

blocks.md
ammgws: I think I'd rather it go in the format string especially since it was originally proposed as an alternative to the alias.

jamesmcm:
@ammgws the above issues are fixed :slightly_smiling_face:

ammgws:
What's going on with that diff for blocks.md?

jamesmcm:
Oops, looks like it changed it to UNIX line endings, I can switch it back tonight, or we could just swap it to that like the rest of the files are.

ammgws:
Actually could you change it back for now? After merging this submit another PR to fix the line endings.

jamesmcm:
@ammgws done :)

blocks.md
```diff
@@ -1001,7 +1017,7 @@ Clicking the left mouse button on the icon updates the number of pending tasks i
 
 Clicking the right mouse button on the icon toggles the view of the block between filtered (default) and non-filtered
 tasks. If there are no filters configured, the number of tasks stays the same and both modes are behaving
-equally.  
+equally.
```
ammgws, line 1020: Sorry to be pedantic but can we get this of this change from this PR?

jamesmcm: Done!

ammgws:
Thanks!

PR #717, opened 2020-06-01
Add kdeconnect block

ammgws:
Testers welcome!

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -11,6 +11,7 @@
 - [Focused Window](#focused-window)
 - [Github](#github)
 - [IBus](#ibus)
+- [KDEConnect](#kdeconnect)
 - [Keyboard Layout](#keyboard-layout)
 - [Load](#load)
 - [Maildir](#maildir)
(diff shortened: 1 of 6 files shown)


GladOSkar:
Works beautifully! Awesome idea :)

image

Although personally, i don't really care about my phone's charge state (it always lasts the day) but _only_ the notification count. Could we have the icon and state/color criteria changeable?

ammgws:
Yeah I'll change it so icons are specified via the format string instead, I think that should work. How should the state/colour work with notifications? Also does yours ever show anything but 0?

GladOSkar:
I'd probably set it to `info` whenever there are any notifications, and default otherwise.

Actually, now that you mention it, the notification counter doesn't seem to update in place. When i reload the bar though, it does show:

image

ammgws:
Oh I think I forgot to add the handler for notifications. Oops! I'll do it sometime this week

ammgws:
Updated.

- notifications now update
- changed block icon to a phone
- made battery and notification icons selectable as qualifiers for the format string
- block state can be set to act on notifications instead of battery level by setting all the thresholds to 0

GladOSkar:
Works great! Thanks!

> block state can be set to act on notifications instead of battery by setting all the thresholds to 0
This feels a little hacky but if it works ¯\_(ツ)_/¯. I'd recommend adding it to the block documentation.

ammgws:
Yeah it's what I went up with now since the other thing was to separate the block into widgets which would then mean multiple format strings are needed.

GladOSkar:
Ok yeah then that makes sense. I originally thought of another config option but that's probably way more bloat than necessary

GladOSkar:
OK i found another Problem: The bar crashes when my device isn't connected.

image

ammgws:
Ah I was waiting for that to come up. Is that when you start the bar before connecting your phone, or does it happen if you disconnect your phone after successfully starting the bar? Somehow I haven't managed to trigger it but I guess my VPN has kept it's connection this whole time.

GladOSkar:
Well my usecase is that i take this computer lots of places, and only sometimes (at home) are both devices on the same subnet. At university, work, or a friend's place, either

a) The network does not allow communication between devices or
b) My phone isn't on wifi and therefore not on the same network

I only saw it while starting the bar as of yet, but when i'm home i can test what happens when i disconnect the phone from wifi

ammgws:
I turned my VPN and WiFi off before heading home and found that the bar was still alive, albeit the kdeconnect block was static with the last seen values. As soon as I connected my phone to the network the block updated again.

Things to try:
1. Show phone as disconnected when it goes away (I think there are some signals on the bus I can watch for this)
2. Figure out what do to with the startup case

PR #718, opened 2020-06-03
kdeconnect: detect disconnections

ammgws:
See https://github.com/greshake/i3status-rust/pull/717#issuecomment-638152064

Haven't tested yet but it should work assuming the "reachable" flag supplied by kdeconnect is accurate.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -426,6 +426,7 @@ Key | Values | Required | Default
 ----|--------|----------|--------
 `device_id` | Device ID as per the output of `kdeconnect --list-devices\`. | No | Chooses the first found device, if any.
 `format` | Format string. Available qualifiers are `"bat_icon"`, `"bat_charge"`, `"bat_state"`, `"notif_icon"`, `"notif_count"`, `"name"`, `"id"`. `"bat_icon"` will automatically change between the various battery icons depending on the current charge state. | No | `"{name} {bat_icon}{bat_charge}% {notif_icon}{notif_count}"`
+`format_disconnected` | Format string for when the phone is disconnected/unreachable. Same qualifiers as above, however they will be fixed at the last known value until the phone comes back online. | No | `"{name}"`
 `bat_info` | Min battery level below which state is set to info. | No | `60`
 `bat_good` | Min battery level below which state is set to good. | No | `60`
 `bat_warning` | Min battery level below which state is set to warning. | No | `30`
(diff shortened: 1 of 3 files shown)


GladOSkar:
Doesn't work on my machine:

image

ammgws:
Yeah I haven't dealt with that situation yet. This just changes the icon when the phone becomes 'unreachable' while the bar is running.

GladOSkar:
Ah alright. Yeah as someone who shuts down and starts his computer many times a day when not at home, this is not really usable then

ammgws:
Try setting your `device_id` in the config.

I disconnected my phone from the network and started i3status-rust, and it somehow returned the last known values for the device. Then when I reconnected the phone it started updating as usual. Need to test with a fully cold start as well but at least that's a step forward I guess?

ammgws:
OK so as a test I just pushed some changes so that it doesn't crash the bar when the device supplied in the config is not found at startup.

image

However it's not updating even after I connected the phone so I need to look into it a bit more

GladOSkar:
> I disconnected my phone from the network and started i3status-rust, and it somehow returned the last known values for the device.

Yeah in my case it also took about a minute until `kdeconnectd` even realized the phone was gone. During that period, the bar would still start up successfully, afterwards not anymore. Actually, sometimes `kdeconnectd` doesn't seem to notice the disconnect at all.

With the device ID, startup detection works, but if i then connect my phone to the network, it doesn't show when kdeconnect connects to the device.

GladOSkar:
Also, when disconnected, can we just not show device information and instead _just_ the "Not connected" icon? Might be less confusing

ammgws:
I was hoping there would be some activity on the bus when the phone reconnects but somehow there's nothing... Checked both `busctl --user monitor org.kde.kdeconnect` and `busctl --user monitor org.kde.kdeconnect.daemon`. Would hate to have to add a manual update interval for this.

GladOSkar:
Oh i guess i got confused by the icon not updating then. Will test again now

GladOSkar:
BTW, missing semicolon on line 146 :P

ammgws:
I ended up adding another format string to choose what to display when the phone is disconnected.

ammgws:
> Works perfectly

Hmm, it still doesn't recover if you disconnect the phone while the bar is running. I'll merge this for now as it's better than what we have now, and in the meantime I will see what else I can do.

GladOSkar:
> Hmm, it still doesn't recover if you disconnect the phone while the bar is running. I'll merge this for now as it's better than what we have now, and in the meantime I will see what else I can do.

In my case it still sometimes does detect it and sometimes doesn't, but that's just because kdeconnect apparently doesn't even notice

PR #719, opened 2020-06-03
Fixes missing "icon" for `bat_half` in case of the none theme

varunkashyap:
The icon text for `bat_half` was missing for the none theme. This PR adds that and I also took this oppurtunity to arrange the strings in ascending order of the battery capacity!


ammgws:
Did it just output "" when the battery was in the half range?

varunkashyap:
Yes! The percentage sign could be used to infer it's battery percentage, but it irked my OCD :)

ammgws:
Thanks!

PR #723, opened 2020-06-05
kdeconnect: detect when phone comes back online

ammgws:
Continued from here: https://github.com/greshake/i3status-rust/pull/718#issuecomment-639554000

@GladOSkar This is working for me when disconnecting phone after bar is already running.


GladOSkar:
Seems to work reliably for me, connection and disconnection, with and without bar restarts in between. Wonderful :)

PR #734, opened 2020-06-08
kdeconnect: automatically discover even unreachable devices

### LINKED ISSUE #733: kdeconnect block breaks if no device is connected on startup
d193388cc purportedly fixed this, but if you don't have a device ID specified in the config file you'll still see an error on startup if there aren't any devices connected.

Applying this change fixes the problem, though I don't know if it's the _correct_ solution (hence this being an issue rather than a PR):

```patch
diff --git a/src/blocks/kdeconnect.rs b/src/blocks/kdeconnect.rs
index 888d0db15..1e30ad5d4 100644
--- a/src/blocks/kdeconnect.rs
+++ b/src/blocks/kdeconnect.rs
@@ -131,9 +131,9 @@ impl ConfigBlock for KDEConnect {
                 "/modules/kdeconnect",
                 Duration::from_millis(5000),
             );
-            // method call opts: only_reachable=true, only_paired=true
+            // method call opts: only_reachable=false, only_paired=true
             let (devices,): (Vec<String>,) = p1
-                .method_call("org.kde.kdeconnect.daemon", "devices", (true, true))
+                .method_call("org.kde.kdeconnect.daemon", "devices", (false, true))
                 .unwrap();
             if devices.is_empty() {
                 return Err(BlockError(
```

sersorrel:
Closes #733.

### DIFF
--- a/src/blocks/kdeconnect.rs
+++ b/src/blocks/kdeconnect.rs
@@ -131,9 +131,9 @@ impl ConfigBlock for KDEConnect {
                 "/modules/kdeconnect",
                 Duration::from_millis(5000),
             );
-            // method call opts: only_reachable=true, only_paired=true
+            // method call opts: only_reachable=false, only_paired=true
             let (devices,): (Vec<String>,) = p1
-                .method_call("org.kde.kdeconnect.daemon", "devices", (true, true))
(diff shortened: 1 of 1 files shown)


ammgws:
Thanks!

Issue #692, opened 2020-05-17
[net block] add `format` option


ammgws:
As mentioned here https://github.com/greshake/i3status-rust/issues/237#issuecomment-570277470

gurditsbedi:
I read the mentioned comment (and thread), not much able to understand, what config option are we trying to change here? Any list of proposed config options ?

ammgws:
Basically instead of having a million options like "ipv6", "ssid" etc, we just have one called "format" that can be set by the user to a string such as "{my IP: {ip}, connected to {ssid}". Check out the docs for examples of other blocks that have a format string

gurditsbedi:
Got the thing here, I think I will be able to do this task.

So the following options will stay out of the format string:
`device`
`interval`
`hide_missing`
`hide_inactive`

I don't think these will be able to become a part of the format string
so these will be out also? Any alternative idea about putting them inside?
`speed_min_unit`
`speed_digits`

Rest of all the options listed in blocks.md will go to become placeholder's in
the format string.

What do you say?

ammgws:
Yep that's what I was going for. Regarding speed_digits/min_unit, if you can figure out a way to do it then sure otherwise I don't have any ideas at the moment.

Also we might need to keep around the old ones and list them as deprecated. There's another block that does that, forgot which one.

gurditsbedi:
Hi, I was working on the issue. So, I am having difficulty to figure out that how should I manage both the legacy option formatting and new format option based formatting.

Should I opt for a "rule based" checking of the options in the config provided by the user? and accordingly modify the format string or is their a better way to do it.

gurditsbedi:
I think I have found a way out. I would request to look at [my commits](https://github.com/gurditsbedi/i3status-rust/commits/dev-net-block-fmt-one). This is a rough work. Let me know if there is a better way out.

If everything is fine. I am thinking of doing the following.
1. I will be convert the all the variable `Widgets` in the Net Struct into `string` and display the whole output as a `ButtonWidget`. This will create another issue of handling the icons.
2. Create an `icon` option to show/hide all the icons used in the block.

gurditsbedi:
I also experimented in the following way. 
```rust
enum ButtonOrGraphWidget {
    ButtonWidget(ButtonWidget),
    GraphWidget(GraphWidget),
}

impl ButtonOrGraphWidget {
    fn set_icon(&mut self, name: &str) {
        match self {
[...]

PR #738, opened 2020-06-10
Add format option to net block

gurditsbedi:
PR for the issue #692 . 
The format option is added to the `net` block by this PR.
If a user uses a deprecated option, the `format` option will not be used. This helps users into transitioning to the new options.


ammgws:
Might need to check the logic.

Using this config on my wired desktop:
```toml
[[block]]
block = "net"
```

Get this: `Error in block 'net': Signal strength is only available for connected wireless devices.`

gurditsbedi:
Confirming the error. Will look into it.

ammgws:
Did you figure it out?

gurditsbedi:
> Did you figure it out?

I think 83a8311f956ad2efb09bf2d1888b20a0988d7bed fixed it. Is the issue still not resolved?

gurditsbedi:
Looks like, I did a silent commit and I didn't announce the update in the thread.

ammgws:
Can you fix the blocks.md conflict?

gurditsbedi:
I think its done. 
Sorry I am new to this, so I thought that the "merger to master" resolve the conflicts, that to for no reason at all.

gurditsbedi:
I see that the last issue was relating to line endings. "\n" and "\r\n". 
I think we should decide upon the type of line ending to be used in the project.

Also we can do the following
1. checking of line endings in the CI pipeline.
OR
2. add a git hook commit script.

ammgws:
They should all be Unix line endings. I fixed the conflict for now.

src/util.rs
```diff
@@ -334,6 +335,36 @@ pub fn format_percent_bar(percent: f32) -> String {
         .collect()
 }
 
+pub fn format_vec_to_bar_graph<T>(content: &[T], min: Option<T>, max: Option<T>) -> String
+where
+    T: Ord + ToPrimitive,
+{
+    let bars = [
+        '\u{2581}', '\u{2582}', '\u{2583}', '\u{2584}', '\u{2585}', '\u{2586}', '\u{2587}',
```
ammgws, line 343: Can you add a comment saying what these unicode chars are meant to be?

src/blocks/net.rs
```diff
@@ -854,41 +867,52 @@ impl Block for Net {
...
+            "{speed_down}" => &s_dn,
+            "{graph_up}" =>  self.graph_tx.as_ref().unwrap_or(&empty_string),
+            "{graph_down}" =>  self.graph_rx.as_ref().unwrap_or(&empty_string)
+        );
+
+        self.output
+            .set_text(if let Ok(s) = self.format.render_static_str(&values) {
+                s
+            } else {
+                "[invalid connection format string]".to_string()
```
ammgws, line 907: I believe we usually return a BlockError here. Is there a reason for deviating from that pattern?

gurditsbedi: I was refering to https://github.com/greshake/i3status-rust/blob/master/src/blocks/networkmanager.rs file while doing. After you mentioned this, I realized that similar problem exists in that file also.  Also I have update the file as per this comment.

src/util.rs
gurditsbedi: ` (x * one eighth block) https://en.wikipedia.org/wiki/Block_Elements`
Added to the file also.

ammgws:
Thanks!

PR #371, opened 2019-04-22
Add direction field to sound block to enable microphone control

fedario:
This is the second attempt of this: #352

This pull request:

- direction field to sound block to enable microphone control
- adds new icons for microphone control
- renames the "volume_muted" icon to "crossed" to better reflect its new purpose

Edit: Closes #277.

PR #352, opened 2019-03-28
Implements #277 + minor stuff and fixes

fedario:
**This pull request contains the following**:

- Implementation for #277 through a "direction" field in the sound block. This field can be either "input" or "output" and defines whether the block is used as a sink/output/sound or source/input/microphone.

- Added support for Font Awesome 5 Icons. The icons.name field now accepts "awesome 5" as a value.

- Added microphone icons.

- Fixed a few icon bugs in the sound block.

- Fixed the icons in the xrandr block.

**Things that require testing**:

- Do the Material icons work correctly? I couldn't get the to work correctly even without this PR.

- Does the sound block as input work correctly with alsa? I wasn't able to test this.

PR #740, opened 2020-06-11
microphone support

remi-dupre:
This is (apparently) the third attempt for https://github.com/greshake/i3status-rust/pull/352.

Note that this is a duplicate of https://github.com/greshake/i3status-rust/pull/371, which I didn't use since there was quite a lot of conflicts and some changes had to be done anyway. I took a similar approach so this may not be an issue.

Should I update the manpages? Running `generate.sh` seems to introduce ~1600 changes to the file right now :(

*NOTE*: The sound module is quite big, should we consider an attempt of splitting it in separate modules (eg: alsa, pulsoaudio, and generic module)? I could attempt to do this in this PR if you think this may be relevant.


ammgws:
I haven't had a chance to look at the code yet, but does this address the comments from the previous PR about duplicating the client?

remi-dupre:
> I haven't had a chance to look at the code yet, but does this address the comments from the previous PR about duplicating the client?

Yes, it's using a single client to take track of states for both sources and sinks.

ammgws:
Can you update the docs too?

remi-dupre:
> Can you update the docs too?

Done!

Also, it just made me realize that I totally forgot to support ALSA in this PR. In https://github.com/greshake/i3status-rust/pull/371 it seems pretty short and straightforward, however this part seems to have slightly changed since. I currently don't have any setup to test/experiment with ALSA but I'll try to dive into it :)

ammgws:
Pretty sure ALSA supports mic already - all you have to do is set `name = "Capture"` or whatever your mic shows up as in the output of `amixer`

remi-dupre:
Oh indeed, you're right!

ammgws:
Can you fix the blocks.md conflict?

> the sound module is quite big, should we consider an attempt of splitting it in separate modules (eg: alsa, pulsoaudio, and generic module)? I could attempt to do this in this PR if you think this may be relevant.

We can leave this for another PR, though I think some other blocks are similarly large.

remi-dupre:
> Can you fix the blocks.md conflict?

:heavy_check_mark:
It seems that some encoding issue appeared, I'm not sure how I introduced that

> We can leave this for another PR, though I think some other blocks are similarly large.

Do you think it would be relevant to spend some time on it? I'd love to spend some time on this project and sometime (I found it a few days ago, it's awesome, thanks for your work btw!). Plus, I'm kind of a tidying nerd ^^'

Updating the manpages still generates a huge diff, am I missing something? Maybe this is just a pandoc version issue? (`pandoc -v`: 2.9.2.1)

ammgws:
The encoding issue was due to another PR. 

I think it would be better to spend time on end user features such as new blocks, but if you're interested in tidying up then it can't hurt to give it a go and see what we can come up with.

I'm not sure about the manpages as they were added by atheriel, perhaps you could check some past PRs and see if it was the same diff volume for them?

remi-dupre:
> I'm not sure about the manpages as they were added by atheriel, perhaps you could check some past PRs and see if it was the same diff volume for them?

I think I was overthinking, it was just not udpated for [some time](https://github.com/greshake/i3status-rust/commits/master/man/i3status-rs.1) and git seems to have a hard time dealing with this big update, many lines counted in the diff actually didn't change.

ammgws:
Thanks!

PR #743, opened 2020-06-14
Make sure KDE Connect properly errors when it can't connect to the daemon

v0idifier:
This would make i3status-rs crash for me, now it at least gives an error. Although maybe it would be better to wait until it can connect to the daemon? idk

### DIFF
--- a/src/blocks/kdeconnect.rs
+++ b/src/blocks/kdeconnect.rs
@@ -134,7 +134,10 @@ impl ConfigBlock for KDEConnect {
             // method call opts: only_reachable=false, only_paired=true
             let (devices,): (Vec<String>,) = p1
                 .method_call("org.kde.kdeconnect.daemon", "devices", (false, true))
-                .unwrap();
+                .block_error(
+                    "kdeconnect",
+                    &"Couldn't connect to KDE Connect daemon".to_string(),
+                )?;
             if devices.is_empty() {
                 return Err(BlockError(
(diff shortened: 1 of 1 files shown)


ammgws:
So I take it your kdeconnect daemon is started after the bar?

v0idifier:
It seems like it :|

ammgws:
OK, thanks!

> Although maybe it would be better to wait until it can connect to the daemon?

It would be nice but would need to come up with a good way to do that. Other blocks using DBus etc also fail the same way.

marzzzello:
How do you autostart kdeconnectd? 
On every boot I get the error that kdeconnectd is not available (yet). So I manually restart i3 and then it works fine.
i3 and the status bar are just faster than kdeconnectd. Waiting for kdeconnectd to be ready would be a nice solution.

ammgws:
I use a systemd user service (placed in /home/username/.config/systemd/user/kdeconnectd.service)
```ini
[Unit]
Description=Connect your computer to your smartphone or tablet

[Service]
ExecStart=/usr/lib/kdeconnectd
ExecStop=/usr/bin/kquitapp5 kdeconnectd
BusName=org.kde.kdeconnect
Environment=QT_QPA_PLATFORM=offscreen

[Install]
WantedBy=default.target
```

marzzzello:
Thanks, it works fine now 👍. I also tried a user service but without BusName or Environment.

PR #748, opened 2020-06-17
load: delegate counting cpus to cpuinfo crate

Thra11:
/proc/cpuinfo is meant to be human readable, and its format is subject to change. On my system (linux kernel 5.7.0), it does not contain the lines which load.rs attempts to read. This PR simplifies things by getting another crate to do the hard work. Works correctly on my system.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,6 +136,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cpuinfo"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


ammgws:
https://docs.rs/cpuinfo/0.1.1/src/cpuinfo/count/logical.rs.html#4-6

Would copying the logic of the last function be enough?

Thra11:
I guess it would

ammgws:
Perhaps send another PR using that method. If it doesn't work out we can revisit this one if it makes a difference

Issue #724, opened 2020-06-05
Add ability to customise number formatting when using `format` strings


ammgws:
> I was also wondering if decimal control could be integrated into the format string.
> 
> When thinking about things like the `net` or the `speedtest` block, it would be neat to be able to define the precision separately for the different displayed values. Ideally the syntax would allow to define all 3 arguments of the `format_speed()` function added in #704 / #707, like for example
> 
> `Free: {free:3:G:B}` for `Free: 21.9GB`
> 
> or
> 
> `Uplink: {up:2:M:b}/s` for `Uplink: 50Mb/s`
> 

_Originally posted by @GladOSkar in https://github.com/greshake/i3status-rust/pull/714#issuecomment-639379159_

GladOSkar:
I would like to work on this but really don't have the time currently, so if somebody else wants to i don't mind

ammgws:
Covered by #1108

PR #750, opened 2020-06-18
Rich formatting for bytes values

remi-dupre:
This PR attempts to address issue https://github.com/greshake/i3status-rust/issues/724. Note that this is based on PR https://github.com/greshake/i3status-rust/pull/738 since it brings more meaning to this change, [last commit](https://github.com/greshake/i3status-rust/commit/9c19618281a92a726bf15f946eef741fbe4e942e) is probably a better way to check the diff.

Also note that this is still WIP as some things are left unfinished, I think it may be relevant to discuss this "syntax proposal" quite early.

This extends current "custom format strings" to allow some formatting parameters for variables:

 - `{var}` stills works as before;
 - `{var:unit}` can be used to specify a unit, by default the unit is only displayed as a suffix. The variable represents bytes, this can be used to specify the minimal prefix (`K`, `M`, ..., `-` for any prefix) and if this should be converted to bits (`b`, `-b`, `Kb`, ...);
 - `{var:.precision}` can be used to specify precision of the displayed value, behaving like `std`'s `format!`;
 - `{var:unit.precision}` can be used to combine both.

For example `{speed:-b.2}/s` with `{"speed": 10000}` as input will be rendered as `"80.00Kb/s"`.

The "unit" specifier could probably be extended manipulate other kind of data, for example `{var:mm:ss}` to specify how to format a duration / time / etc...?

**Side note**: by merging different widgets into a fully customizable format string, it makes it harder to integrate icons in the output, what would you think about adding a template syntax to insert icons, say `<icon_name>`?

I think the following is still required before this is no longer WIP:

 - [x] merge of https://github.com/greshake/i3status-rust/pull/738
 - [ ] validation of this syntax
 - [ ] user documentation
 - [ ] better integration in concerned modules (net / speedtest for now)


GladOSkar:
Nice!

But if i read you correctly this doesn't seem to cover the newly included "always 3 digits" mode that was introduced in #704/#707, right?

I.e. automatically choosing between `6.81M`, `46.8M` and `0.25G` based on magnitude in order to allow for a neat, human readable, consistently-sized value. See #240.

That's what my [original idea](https://github.com/greshake/i3status-rust/pull/714#issuecomment-639379159) was based on, because that's how it's implemented for `net` and `speedtest` via the `util::format_speed()` function.

Any idea how that could be incorporated into the syntax?

remi-dupre:
Yeah you're right, plus after testing I noticed that this "always XX digits" mode is much more intuitive and convenient (it's pretty close to an actual accuracy in the sense of scientific notations).

I think there would be 3 ways to address that:

 1. completely replace the behavior of `{:.2}`, downside is that we loose consistency with how `std::fmt` works plus maybe in some case the standard behavior could be preferred?
 2. add a separate syntax (uninspired suggestions: `{:.=2}`, `{:..2}`, `{:,2}`, ...), downside is that it would probably make the feature even harder to discover.
 3. behave differently with Bytes, downside is that creating exceptions is probably a very bad idea.

I would guess that I am overthinking it (as I always do, just tell me if I do... :p), in which case we should probably go for *1*.

GladOSkar:
I do somewhat agree that re-using the `std::fmt` syntax for a different result feels improper and unintuitive.

I honestly don't expect users to have preconceptions about how string formatting is supposed to work anyway, many might not even know rust syntax, so i think using a slightly different/extended one would be fine.

How about 
```
{:=3,G,b}
```
to optionally include the minimum unit and specify bits instead of bytes?

Keep in mind i'm not a maintainer here so one of them might have differing views.

remi-dupre:
> I honestly don't expect users to have preconceptions about how string formatting is supposed to work anyway, many might not even know rust syntax, so i think using a slightly different/extended one would be fine.

That's a pretty nice argument, I'll stick to the previous more pleasant behavior X)

> How about
> 
> ```
> {:=3,G,b}
> ```


You got me thinking: it could be nice to have a generic way to identify special formatter. For example:

- `{:B=[...params...]}` would be our formatter for bytes
- `{:t=[...]}` could be a future formatter for time
- `{:d=[...]}` could be a future formatter for dates
- `{:D=[...]}` could be a future formatter for duration
- ...

We could even enforce some kind of homogeneity: a time formatter could only be used for time values otherwise an error is raised.

What would you think about that ?

GladOSkar:
Sounds pretty neat to be honest. I wouldn't mind either way, but maybe @ammgws or @atheriel have an opinion

remi-dupre:
Hi ! Figured out I should keep you updated with what I've been doing, I think I'll try to polish this PR now to mark it as not WIP.

 - I updated the syntax to something close to what we discussed: `{var_name:B[digits=2,min_prefix=K]}`, I think this verbose use of parameters may make it easier to understand, do you think that would be a good idea.
 - I made it easy to add formatter with a similar syntax in the future: `{var_name:formatter_name[param1=val1,...]}`
 - I reimplemented the old way of dealing with precision for speeds
 - I added a syntax for icons: `<icon_name>`, the motivation behind this is that merging big set of blocks like in the old `net` module into a single format string make it harder to deal with icons. You can of course add icons by using directly its unicode id `\uXXXX`, but that's not really user friendly, and it means it doesn't adapt to your selected set of icon.

As an example, the following format string:

`<net_up> {speed_up:B[digits=2,min_prefix=K,unit=b]}b/s <net_down> {speed_down:B[digits=2,min_prefix=K]}B/s`

Compiles to:

```
FormatTemplate([
    Str("⬆"),
    Str(" "),
    Var {
        name: "speed_up",
        formatter: Bytes {
            unit: Bits,
            digits: 2,
            min_prefix: K,
        },
    },
    Str("B/s "),
    Str("⬇"),
    Str(" "),
    Var {
        name: "speed_down",
        formatter: Bytes {
            unit: Bytes,
            digits: 2,
            min_prefix: K,
        },
    },
    Str("B/s"),
])
```
Which renders like this (net block): Screenshot_2020-06-28_13-10-44

ammgws:
Hmm, I'm not too sure about the verbosity. What would the alternative be using your latest syntax?

ammgws:
@remi-dupre Are you able to rebase this?

remi-dupre:
> @remi-dupre Are you able to rebase this?

I don't think I'll have time to spend on this on the next two weeks but I'll love to finish this PR as soon as I have some time for myself! (including the rebase ;p)

BourgondAries:
Any work on this or is this abandoned?

ammgws:
@BourgondAries Are you thinking about working on this?

ammgws:
Closing in favour of #1112 which is nearing completion.

Thanks for the initial work on this!

PR #751, opened 2020-06-19
Add instructions on how to install with cargo

bluec0re:
Until this is published to crates.io, a cargo install --git command can be used as well.


atheriel:
I didn't know about this feature of `cargo`. Neat!

PR #754, opened 2020-06-22
Add clickable and show_gpu_name options to nvidia_gpu.

maximeplante:
`clickable` is useful when the user is not interested in having the nvidia_gpu block react to clicks. `show_gpu_name` is to let the user decide if the gpu name is shown by default. This got added because the block can be non-clickable.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -851,6 +851,8 @@ Key | Values | Required | Default
 `show_temperature` | Display gpu temperature. | No | `true`
 `show_fan_speed` | Display fan speed. | No | `false`
 `show_clocks` | Display gpu clocks. | No | `false`
+`show_gpu_name` | Display gpu name. | No | `false`
+`clickable` | Respond to click. | No | `true`
 
 ## Pacman
 
--- a/blocks.md
+++ b/blocks.md
@@ -190,10 +190,10 @@ Key | Values | Required | Default
 
 Creates a block that display the output of custom shell commands.
 
-For further customisation, use the `json` option and have the shell command output valid JSON in the schema below:  
-`{"icon": "ICON", "state": "STATE", "text": "YOURTEXT"}`  
-`icon` is optional, it may be an icon name from `icons.rs` (default "")  
-`state` is optional, it may be Idle, Info, Good, Warning, Critical (default Idle)  
+For further customisation, use the `json` option and have the shell command output valid JSON in the schema below:
+`{"icon": "ICON", "state": "STATE", "text": "YOURTEXT"}`
+`icon` is optional, it may be an icon name from `icons.rs` (default "")
+`state` is optional, it may be Idle, Info, Good, Warning, Critical (default Idle)
 
 ### Examples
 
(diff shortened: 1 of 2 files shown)


src/blocks/nvidia_gpu.rs
```diff
@@ -324,115 +342,117 @@ impl Block for NvidiaGpu {
...
-                                "-a",
-                                &format!(
-                                    "[fan:{}]/GPUTargetFanSpeed={}",
-                                    self.gpu_id, new_fan_speed
-                                ),
-                            ])
-                            .output()
-                            .block_error("gpu", "Failed to execute nvidia-settings.")?;
-                        self.fan_speed = new_fan_speed;
-                        fan_widget.set_text(format!("{:02}%", new_fan_speed));
```
maximeplante, line 435: This huge diff is caused by enclosing the code inside a if statement to check self.clickable. I did not change anything in that code but I had to increase the indentation of the whole block.

ammgws:
Instead of adding a new config option for showing the GPU name, we could instead implement a `format` string and have `name` be one of the qualifiers. Grep blocks.md for other blocks that are using format strings.

maximeplante:
This would break the current behavior of that block. Since the user can click to toggle the display of some information, we need to think about how we could reconcile this with the use of the `format` parameter. One option could be to deactivate the toggling behavior when `format` is set. What are you thoughts on that?

maximeplante:
To me, it seems like this would break the current behavior of that block. Since the user can click to toggle the display of some information, I would need to reconcile this with the use of the `format` parameter. One option could be to deactivate the toggling behavior when `format` is set. What are you thoughts on that?

ammgws:
Oh I didn't know it was clickable already. Perhaps two format strings then?

ammgws:
The memory block does something like that

maximeplante:
Ah yes, using the memory block design could fix that.

Another problem is that by clicking on the fan speed indicator, you can manually set the speed of the fan by scrolling. To go around this issue we can make it that when you scroll on any part of the block it will change the fan speed. However, this might change the component's behavior too much. What do you think? (I do not use this feature but I assume some users are relying on it).

blocks.md
```diff
@@ -696,7 +696,7 @@ Key | Values | Required | Default
 
 ## Net
 
-Creates a block which displays the upload and download throughput for a network interface. Units are by default in bytes per second (kB/s, MB/s, etc), 
+Creates a block which displays the upload and download throughput for a network interface. Units are by default in bytes per second (kB/s, MB/s, etc),
```
ammgws, line 699: Can you remove these unrelated changes? (Submit another PR if there's an issue with the formatting in the sections apart from the Nvidia block)

ammgws:
> Another problem is that by clicking on the fan speed indicator, you can manually set the speed of the fan by scrolling.

Hmm OK, perhaps this can't be avoided then so I guess we can forget the format string for now.

ammgws:
> To go around this issue we can make it that when you scroll on any part of the block it will change the fan speed

Actually I think this might be OK.

ammgws:
@maximeplante I think I understand the intent behind your PR now. Currently if one sets `label` in their config, the block would still show the default gpu name unless you clicked on the block to change the text to `label`, which is counter-intuitive to what one would expect.

#856 makes it so the label is shown by default if it is provided in the config, so that should cover your use-case.

Please re-open otherwise.

PR #757, opened 2020-06-24
Allow setting the icon and state in custom_dbus

jmgrosen:
This is a fairly straightforward feature for the relatively new custom_dbus block.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -251,7 +251,7 @@ Key | Values | Required | Default
 
 Creates a block that can be updated asynchronously using DBus.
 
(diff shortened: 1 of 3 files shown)


ammgws:
Thanks!

Out of interest, what are you using the CustomDBus block for?

jmgrosen:
I use org-mode in Emacs with its "clocking" ability to track what task
I'm currently working on. I wanted to put my current task in my status
bar so that I can easily remember what I'm working on if I get
distracted. I could have just used the Custom block, but Emacs has
easy-to-use DBus integration, so it's just as easy (if not easier) and
this way the change happens instantly.

PR #758, opened 2020-06-26
Sed awk grep removal

### LINKED ISSUE #725: Replace uses of sed/etc with Rust native regex
There's a bunch of places in the code-base such as [this](https://github.com/greshake/i3status-rust/blob/22d53b44219a3bbed3cbf00d395143ec3a92ab0b/src/blocks/net.rs#L81) where we shell out to an external program, which is OK, however we are using sed/awk/etc to parse the output when we could possibly use Rust code instead. We already depend on `regex` so it's not an issue to convert these over, and it would bring us closer to our tagline "...written in pure Rust".

themadprofessor:
Replace calls to sed, awk, grep, etc with the regex crate where possible.
Currently, don't have access to a machine with Wi-Fi, so those calls have been left in for now.

Closes #725

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -29,6 +29,7 @@ use crate::widgets::button::ButtonWidget;
 lazy_static! {
     static ref DEFAULT_DEV_REGEX: Regex = Regex::new("default.*dev (\\w*).*").unwrap();
     static ref WHITESPACE_REGEX: Regex = Regex::new("\\s+").unwrap();
+    static ref ETHTOOL_SPEED_REGEX: Regex = Regex::new("Speed: (\\d+\\w\\w/s)").unwrap();
 }
 
 pub struct NetworkDevice {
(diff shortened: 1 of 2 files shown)


themadprofessor:
That's the remaining calls replaced.
The regexs for `iwctl` and `nmctl` haven't been tested as I don't have a machine with those commands, but they should work.
If someone can test those, that would be great.

ammgws:
The fallback logic is broken. I only have `iwctl` installed so I get:

> Error in block 'net': Failed to execute SSID query using iw.

However the previous behaviour is so that it fallbacks to the next possible command.

themadprofessor:
Whoops, good spot. Should be fixed now.

src/blocks/net.rs
```diff
@@ -936,3 +908,150 @@ impl Block for Net {
...
+
+#[inline]
+/// Attempt to get the SSID the given device is connected to from iw.
+/// Returns Err if:
+///     - `iw` is not a valid command
+///     - failed to spawn an `iw` command
+///     - `iw` failed to produce a valid UTF-8 SSID
+/// Returns Ok(None) if `iw` failed to produce a SSID.
+fn get_iw_ssid(dev: &NetworkDevice) -> Result<Option<String>> {
+    let raw = Command::new("iw")
```
ammgws, line 940: Do you think it's worth DRYing this block that's also used in the other get_*_ssid functions?

src/blocks/net.rs
```diff
@@ -312,27 +271,30 @@ impl NetworkDevice {
...
+        } else if let Some(ip_bytes) = WHITESPACE_REGEX.splitn(&ip_output, 3).nth(2) {
+            let ip = String::from_utf8(ip_bytes.to_vec())
                 .block_error("net", "Non-UTF8 IP address.")?
                 .trim()
                 .to_string();
             Ok(Some(ip))
+        } else {
+            Ok(None)
         }
     }
```
ammgws, line 300: Can you please update or remove this outdated comment?
> /// Queries the bitrate of this device (using `iwlist`)

src/blocks/net.rs
```diff
@@ -345,30 +307,40 @@ impl NetworkDevice {
...
-        let mut bitrate_output = Command::new("sh")
-            .args(&["-c", &command])
-            .output()
-            .block_error("net", "Failed to execute bitrate query.")?
-            .stdout;
+        if self.wireless {
+            let bitrate_output = Command::new("iw")
+                .args(&["dev", &self.device, "link"])
+                .output()
+                .block_error("net", "Failed to execute bitrate query.")?
```
ammgws, line 314: Can you please add the command name to the error like the other ones? Makes it easier to debug later.

src/blocks/net.rs
```diff
@@ -345,30 +307,40 @@ impl NetworkDevice {
...
 
-        if bitrate_output.is_empty() {
-            Ok(None)
+            if let Some(rate) = IW_BITRATE_REGEX
+                .captures_iter(&bitrate_output)
+                .next()
+                .and_then(|x| x.get(1))
+            {
+                String::from_utf8(rate.as_bytes().to_vec())
+                    .block_error("net", "Non-UTf8 bitrate")
```
ammgws, line 323: nitpick: UTf -> UTF

src/blocks/net.rs
themadprofessor: Which block? The lines shown are the start of a function.
Those functions are quite repetitive in general, so pulling some of it out into a helper function might be a good idea.

ammgws: Sorry, was on mobile.

This part that's common across the different functions:
 ```
let raw... 
... 
return Err(err).block_error("net", "Failed to execute SSID query using xx.");
```

Comment (ammgws) [reply src/blocks/net.rs:940]: You can leave it for another time if you want

Comment (themadprofessor) [reply src/blocks/net.rs]: No worries, it does make sense to pull that section out.

**Comment by ammgws** [comment]
Hmm, current master and the below config works fine but pulling in this branch causes it to fail. It shouldn't fail since `bitrate` isn't specified in the config.

```
[[block]]
block = "net"
format = "{ssid}"
```

```
>./target/debug/i3status-rs test.toml
[[{"background":"#000000","color":"#dc322f","full_text":" Error in block 'net': Failed to execute bitrate query with iw. ","separator":false,"separator_block_width":0}]

Error in block 'net': Failed to execute bitrate query with iw.
```

ammgws:
OK I see what's going on. Before it would run, for example, `sh -c "ethtool ... | awk ..." and due to the command being piped to awk/sed/etc the exit code was 0 even if ethtool itself was not installed/failed to run. Now that all those pipes to sed/awk are gone and the command is called directly it is failing as it should have done before.

Issue #766, opened 2020-07-07
net block shows speeds regardless of config


ammgws:
> After updating my current config shows fields that it didn't before. Even though i have set
> 
> ```
> [[block]]
> block = "net"
> device = "ppp0"
> hide_inactive = true
> hide_missing = true
> interval = 10
> speed_up = false
> speed_down = false
> ip = false
> 
> [[block]]
> block = "net"
> device = "eno1"
> hide_inactive = true
> hide_missing = true
> interval = 10
> speed_up = false
> speed_down = false
> ip = false
> 
> [[block]]
> block = "net"
> device = "wlp2s0"
> signal_strength = true
> interval = 10
> speed_up = false
> speed_down = false
> ip = false
> ```
> 
> The blocks show some speeds:
> ![image](https://user-images.githubusercontent.com/9497832/86793146-78445800-c06b-11ea-9dc9-1b02dbf876c0.png)
> 
> This is probably unintended and should be fixed?
> 
> _Originally posted by @GladOSkar in https://github.com/greshake/i3status-rust/issues/467#issuecomment-654886404_

ammgws:
Hmm I can't seem to reproduce with just:
```
[[block]]
block = "net"
speed_up = false
speed_down = false
```

GladOSkar:
I just had a look again. When i change `format` back to:

```
[[block]]
block = "net"
device = "eno1"
speed_up = false
speed_down = false
```
It shows the speed again:

![Peek 2020-07-07 16-16](https://user-images.githubusercontent.com/9497832/86794866-446a3200-c06d-11ea-97ab-2388ce1f099b.gif)

GladOSkar:
When i unset `device` as well. I built from the `master` i _just_ pulled.

ammgws:
Actually my bad, I was testing on another branch. Able to repro now.

@gurditsbedi  Are you able to have a look?

gabm:
I have the same... running on Arch: 0.14.1.r1225.gc9d5f3e-1,that means I'm on this commit and the problem exists: https://github.com/greshake/i3status-rust/commit/c9d5f3e710ac5ade11956012672591ffc33e9f71

gurditsbedi:
Confirming the issue. Will send a fix soon.

ammgws:
Closing via #767 

Tested fine for me, please re-open if anyone still has issues.

GladOSkar:
Works. Thanks!

PR #767, opened 2020-07-08
fix for issues-766

gurditsbedi:
https://github.com/greshake/i3status-rust/issues/766
Defining the deprecated option was taken as true though it being false. This has been fixed.
Logic for checking boolean/format style has been changed.


ammgws:
Tested with a couple of different configs, LGTM.

Thanks for the quick response!

PR #770, opened 2020-07-12
fix: correct block config and enable default format

DCjanus:


src/blocks/net.rs
```diff
@@ -620,34 +620,21 @@ impl ConfigBlock for Net {
...
-            "ipv6",
-            "speed_up",
-            "speed_down",
-            "graph_up",
-            "graph_down",
-        ];
-
-        // if "format" option is present it will be preferred
-        let format = if config.blocks[0].1.get("format").is_some() {
+        let (_, net_config) = config
```
ammgws, line 623: What is going on here, why not just use `block_config`?
You just need to `clone()` it.

DCjanus: In previous code, we were trying to compatible with options that have been deprecated, which means, it could be:

+ using `format` option
+ using deprecated options only
+ nothing

options in `block_config` would be given default values, which means `format` would not be empty.

src/blocks/net.rs
```diff
@@ -699,6 +686,35 @@ impl ConfigBlock for Net {
...
+        "bitrate",
+        "ip",
+        "ipv6",
+        "speed_up",
+        "speed_down",
+        "graph_up",
+        "graph_down",
+    ];
+
+    let mut old_format = Vec::new();
```
ammgws, line 704: In the old config `speed_up` and `speed_down` defaulted to `true`, so we we need to change the logic here to account for that.

i.e.
In 0.13 this would show `ip`, `speed_up` and `speed_down`.
```
[[block]]
block = "net"
ip = true
```

In 0.13 this would show just `ip`.
```
[[block]]
block = "net"
ip = true
speed_down = false
speed_up = false
```

DCjanus: Should be resolved in this commit: https://github.com/greshake/i3status-rust/pull/770/commits/8dbc8f512159ec5f9633fedac3698c6e8cb6d986

ammgws:
LGTM

Thanks!

PR #771, opened 2020-07-12
Add check for valid nvidia_smi output

themadprofessor:
When the nvidia driver is updated and the new module hasn't been loaded, `nvidia-smi` can produce invalid output such as "Failed to initialize NVML: Driver/library version mismatch", which can cause a panic.
This adds a check to ensure the output from `nvidia-smi` contains at least 2 commas.


ammgws:
Is the exit status non-zero in that situation? Perhaps you could catch that instead.

themadprofessor:
That it does, shall change it to check for that.

ammgws:
I don't have a way to test this so I hope it's OK.

Thanks!

PR #778, opened 2020-07-18
fix crashing thread in window block

remi-dupre:
My window block is systematically crashing after a while:

```
thread 'focused_window' panicked at 'called Result::unwrap() on an Err value: Custom { kind: UnexpectedEof, error: "failed to fill whole buffer" }', src/blocks/focused_window.rs:81:27
```

While I'm not exactly sure what swayipc occasional errors mean, I think it is not critical enough to crash the subscribed thread. I changed that behavior to log the error in `stderr` and continue running normally. Also, it seems to show that my error might be that swayipc crate receive wrongly formatted data somewhere (?):

```
[warn] error in `window` block: missing field `instance` at line 1 column 619
```

While I was on it, I tried to replace all `unwrap` calls by at least an `except`, this doesn't change much of the behavior but I think it is recommended since it is easier to catch what could happen. I also tried to factorize some of the code that was changed since it would accentuate the repetitions :)

### DIFF
--- a/src/blocks/focused_window.rs
+++ b/src/blocks/focused_window.rs
@@ -134,15 +134,10 @@ impl ConfigBlock for FocusedWindow {
 
                 let events = conn
                     .subscribe(&[EventType::Window, EventType::Workspace])
-                    .expect("could not subscribe to window events")
-                    .filter_map(|event| {
-                        event
-                            .map_err(|err| eprintln!("[warn] error in `window` block: {}", err))
-                            .ok()
-                    });
+                    .expect("could not subscribe to window events");
 
                 for event in events {
-                    let updated = match event {
+                    let updated = match event.expect("could not read event in `window` block") {
                         Event::Window(e) => match (e.change, e.container) {
                             (WindowChange::Mark, Node { marks, .. }) => update_marks(marks),
                             (WindowChange::Focus, Node { name, marks, .. }) => {
--- a/src/blocks/focused_window.rs
+++ b/src/blocks/focused_window.rs
@@ -73,7 +73,7 @@ impl ConfigBlock for FocusedWindow {
                     .lock()
                     .expect("lock has been poisoned in `window` block");
 
-                let changed = *title == new_title;
+                let changed = *title != new_title;
                 *title = new_title;
                 changed
             }
@@ -118,7 +118,7 @@ impl ConfigBlock for FocusedWindow {
                     .lock()
                     .expect("lock has been poisoned in `window` block");
 
-                let changed = *marks == new_marks_str;
+                let changed = *marks != new_marks_str;
                 *marks = new_marks_str;
                 changed
             }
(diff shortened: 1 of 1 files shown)


travankor:
The block crashes whenever I try to load specific Xwayland programs, like `xev` or nwjs program. I will try testing with this PR soon.

travankor:
Confirmed that this fixes the crash. I have the same `[warn] error` although the column value is slightly different.

remi-dupre:
> Confirmed that this fixes the crash. I have the same `[warn] error` although the column value is slightly different.

Nice!

> And adding a new mark does not update the block anymore.

Oh right, the reason for this is obviously that I'm too dumb to check if a string changed: https://github.com/greshake/i3status-rust/pull/778/files#diff-90d876d1135b14deb9ddb2a61ea0f692R121

```rust
let changed = *marks == new_marks_str; 
```

I'll fix it this evening, I'll also have a look at the hanging title issue, I hope it's not much more complicated.

remi-dupre:
@travankor it should ve fixed now, I'm not sure about the marks since I'm running i3, I'll setup sway if the issue persists :/

travankor:
I think you forgot to push the new commit.

remi-dupre:
> I think you forgot to push the new commit.

Right, I just pushed  it(and apparently I forgot about it every evening for 8 days, sorry :/)

travankor:
This seems to fix the issue

ammgws:
Is this reliably reproducible? I want to check out what's actually going on.

ammgws:
Submitted a fix upstream: https://github.com/JayceFayne/swayipc-rs/pull/12

ammgws:
Try git master

remi-dupre:
> Try git master

I'm not sure what you mean?

> Submitted a fix upstream: [JayceFayne/swayipc-rs#12](https://github.com/JayceFayne/swayipc-rs/pull/12)

I would probably not invalidate this change? I think it is still relevant to have a more robust wrapper around swayipc?

ammgws:
The cause of the block crashes have been identified and fixed, but aren't in a tagged release yet so you will need to use the latest git code for the fixes.

We can replace those unwraps as you've done for better error messages but I'd rather you revert the other part that logs to stderr instead of crashing.

remi-dupre:
> The cause of the block crashes have been identified and fixed, but aren't in a tagged release yet so you will need to use the latest git code for the fixes.

Thanks for the patch and the update on master, I didn't realize this was just an issue with swayipc-rs :)

> We can replace those unwraps as you've done for better error messages but I'd rather you revert the other part that logs to stderr instead of crashing.

I just reverted this specific change, sorry for the delay. I've been quite busy looking for an apartment and doing associative stuff lately but I'll soon have more time to spend on my other PR ;)

ammgws:
Thanks!

PR #787, opened 2020-07-31
Add parameter dynamic_width for music block

UnkwUsr:
By default, the music block has a static width (= `max_width`), and when the name of the music is too short, it looks very ugly: just unnecessary free space inside the bar.

When `dynamic_width` is set to true and the name of the music is longer than `max_width`, block width will be set to `max_width` (marquee option remains available), but when name is less than `max_width`, block width will be set equal to the name of the music, so no free space left.
P.S. I have the music block in the first place (counting from the left), so dynamic width looks very cool. But if people have it in the middle or at the end, then music block will affect next blocks after music and jumpy text can be annoying, so we have this behavior as a config parameter.


Alexeyan:
+1 from me. Would like this feature.

PR #788, opened 2020-08-02
Fix: TaskWarrior - Don't count deleted items

HPrivakos:
The task command count the deleted tasks as pending , adding `-DELETED` to the task command give the correct count of current pending tasks


ammgws:
Thanks!

Issue #715, opened 2020-05-31
Future block ideas


ammgws:
From [bumblebee-status](https://bumblebee-status.readthedocs.io/en/master/modules.html):
block | description | comment |
--- | --- | --- |
apt | Displays APT package update information | we habe pacman, so I guess we could have this too |
arandr | Enables handy interaction with arandr for display management | 
caffeine | Enable/disable automatic screen locking | 
currency | Displays currency exchange rates |
dnf | Displays DNF package update information |
dunst | Toggle dunst notifications |
getcrypto | Displays the price of a cryptocurrency |
git | Print branch and git status for currently focused window |
hddtemp | Fetch hard drive temeperature data from a hddtemp daemon |
hostname | Displays the system hostname |
http_status | Display HTTP status code |
indicator | Displays the indicator status, for numlock, scrolllock and capslock |
kernel | Shows Linux kernel version information |
libvirtvms | Displays count of running libvirt VMs |
mocp | Displays information about the current song in mocp |
octoprint | Displays the Octorpint status and the printer’s bed/tools temperature in the status bar |
pihole | Displays the pi-hole status (up/down) together with the number of ads that were blocked today |
ping | Periodically checks the RTT of a configurable host using ICMP echos |
prime | Displays and changes the current selected prime video card |
progress | Show progress for cp, mv, dd, … | this seems useful to me |
publicip | Displays public IP address | could be added to net block |
redshift | Displays the current color temperature of redshift |
rotation | Shows a widget for each connected screen and allows the user to loop through different orientations |
rss | RSS news ticker |
smartstatus | Displays HDD smart status of different drives or all drives |
spaceapi | Displays the state of a Space API endpoint Space API is an API for hackspaces based on JSON |
sun | Displays sunrise and sunset times |
system | Adds the possibility to shutdown or reboot the system |
twmn | Toggle twmn notifications |
vault | Copy passwords from a password store into the clipboard (currently supports only ‘pass’) |
vpn | Displays the VPN profile that is currently in use |
xkcd | Opens a random xkcd comic in the browser |
yubikey | Shows yubikey information |
zpool | Displays info about zpools present on the system |

From [py3status](https://py3status.readthedocs.io/en/latest/modules.html):
block | description | comment
--- | --- | --- |
air_quality | Display air quality polluting in a given location | could be added to weather block
aws_bill | Display bill for Amazon Web Services |
bitcoin_price | Display bitcoin using bitcoincharts.com |
 |  |
 |  |
[...]

PR #789, opened 2020-08-03
Add dunst notification toggle

thomaspeklak:
This adds a bell icon that can turn on or off [Dunst](https://dunst-project.org/) notifications. It mirrors the functionality of the corresponding [bumblebee module](https://github.com/tobi-wan-kenobi/bumblebee-status/blob/main/bumblebee_status/modules/contrib/dunst.py)

I don't know any Rust and probably did some stuff wrong. It compiles and works as I have been using it for some days now. Please feel free to point out stuff that I missed.


ammgws:
As it is now I believe this would be better off as a custom/toggle block (see dicussion here: https://github.com/greshake/i3status-rust/issues/715#issuecomment-640951152).

Also, I don't see an update function so the block wouldn't update if dunst was paused via another method.

However if this PR could be rewritten to use DBus to control/monitor dunst then I think it would be a good addition as a native block. It looks like you can get and also set dunst's current state through dbus -  see the code for `is-paused` and `set-paused` in dunstctl: https://github.com/dunst-project/dunst/blob/fb2ffd425e2cfe42de0c633bcdf3a0203bfef90d/dunstctl#L67. You can probably monitor DBus for changes to that property as well, which means we can make the block update instantly if dunst was paused/unpaused by another method.

thomaspeklak:
I have not

thomaspeklak:
I have not seen the other ticket. If it makes more sense in your architectural concept to do this in another way, you can just decline the issue for now.

Otherwise I will look into dunstctl and figure out how to do the dbus stuff.

I am interested to take over dunst and dnf.

ammgws:
What's dnf?

thomaspeklak:
> What's dnf?
dnf ist the package manager for Fedora

ammgws:
#851 should cover this if you are able to use the latest git version of dunst

Issue #801, opened 2020-08-09
Running into multiple errors about JSON while trying to implement a new block


AkechiShiro:
I was writing a block to use redshift or sct directly from a block (and adjusting the temperature using the mouse scrolling, had a look into the Xrandr block to see how to properly do this) basically an equivalent to the hueshift block from py3status, after a lot of fixes and a successful compilation of i3status-rs I got this error on the i3bar when I explicitly add the new block (when I remove it the error goes away so it must be the block that's is the problem I suppose) I created : 
Error: Could not parse JSON (lexical error: probable comment found in input text
Looking into `.xsession-errors` I see this error precisely:
```[i3bar] Could not parse JSON input (code = 2, message = lexical error: probable comment found in input text, comments are not enabled.):```
Is that having any link with Rust comments (I think that's not the case here) I left in the new code block ?
This error comes from i3bar
I suppose the best course of action from here should perhaps be to create a PR with the block and figure out what's going on from there ?

ammgws:
Feel free to share the code via PR

AkechiShiro:
I've shared the code via PR, I'd like a bit of feedback or guidance on the issue I'm facing as I don't really understand what's going on and why I'm getting an error related to JSON. Since my code doesn't really do anything with JSON, I suppose something in i3status-rust codebase organizes the data using the JSON format and somewhere an error is raised.

ammgws:
Does the error in i3bar occur straight away, or does it happen randomly after the i3status-rs is already up and showing in the bar?

Can you run i3status-rs (with a minimal config) in a terminal and copy the output here?
(I am running sway so don't have a X11 environment to test with myself)

AkechiShiro:
The error occurs straight away when the block hueshift is present in the toml config file, the bar is never shown.
I'm using i3wm as my Windows manager currently.

Using a minimal config with only the block hueshift and using the compiled version this is what I get it seems the error is not happening.
```JSON 
i3status-rust/target/release/i3status-rs ~/.config/i3/minimal.toml    
{"version": 1, "click_events": true}
/bin/sct 
/bin/redshift
[[{"background":"#000000","color":"#a9a9a9","full_text":"| ","markup":"pango","separator":false,"separator_block_width":0},{"background":"#000000","color":"#93a1a1","full_text":" 6500 ","separator":false,"separator_block_width":0}],
[...]

Issue #669, opened 2020-05-12
Question: How to debug blocks?


xxxserxxx:
Is there an env var to enable debugging? I don't see a command line option to enable debugging.

I have a toggle block with a `command-state` that, when executed on the command line, outputs "1" on STDOUT and a success exit code.  The toggle remains "off" in the bar.  Manually toggling it triggers something, but it flips back almost immediately. I'd like to debug this.  I don't see any output in `.xsession-errors` that looks like it comes from i3status-rust, although when I created a simple test that should always have been "true," whenever I toggle the widget I get a "1" in the `.xsession-errors` log.

What's the correct way of debugging widgets like this?

ammgws:
There's no debug output at the moment.

What version are you using?
Can you give me a minimal reproducible example that I could test with?

xxxserxxx:
`-V` says 0.13.0, but it's cloned and built from git; `rev-list` HEAD gives me 5e6e1ed4e47ca38b21cb5268251995ed23fd9708.

**Any** command produces the same behavior.  `ls`. `yes | head -1`.  Here's a version using "echo":
```
[[block]]
block = "toggle"
text = "Test"
command_state = "echo true"
command_on = "echo true"
command_off = "echo false"
interval = 2
```
It all behaves the same way.  A couple of weeks ago, I had something working, but I changed what I was toggling about a week ago and it stopped working; I thought it was my script, but now I can't get toggle to work at all.  With the above block, the toggle is off.  If I manually toggle it on, it turns itself off.
```

I'm sure it's something I'm doing, but I can't figure it out; hence the request for debugging.

**Comment by xxxserxxx** [comment]
I just pulled and compiled, and now i3bar is **really** unhappy, because i3status-rust is putting a "," at the end of every response.  I *think* that's what it's complaining about, anyway -- the `.xsession-errors` is filling up with these errors:
```
[i3bar] Could not parse JSON input (code = 2, message = parse error: after array element, I expect ',' or ']'): 
```

xxxserxxx:
Reverting commit 88bfba7efbd2a4832857f337cbcf815e23ad7932 doesn't make Toggle work for me, but it stops the block from crashing i3bar.

ammgws:
> because i3status-rust is putting a "," at the end of every response

It's meant to do that, so it's probably something else. I wonder if it's due to the output of your commands. Does it still give those errors with your `echo` example from above?

This is your example block running with i3status-rs git master:
[...]

PR #802, opened 2020-08-09
Add hueshift block & updated list of blocks

AkechiShiro:
* Scrolling up or down updates the screen temperature by a "step"
  amount (max_step 2000K).
* Supports either Xsct (X11 set screen color)  or Redshift to change screen color temperature.
* Redshift is preferred as it as a much larger range of temperature
  color changes (between 1000K and 25_000K).
* Currently for testing purpose only redshift is supported.

I'm open to any feedback or suggestion.
I'm a fairly new to Rust and willing to learn if there is anything I can make a little nicer, just say so.
Also I've used u16 because the temperature of the color should never be negative nor go over 65535 since the highest limit I've seen is 25 000K that Redshift can set.

I'm also planning to add a MouseButton event later perhaps to toogle between different values instantly by clicking on the text.
But first I have to find out what's going on with the JSON error #801 I'm getting from i3bar.


omertuc:
You forgot to update the documentation to include your block and its configuration options

src/blocks/hueshift.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    fn click(&mut self, event: &I3BarEvent) -> Result<()> {
+        if event.name.is_some() {
+            match event.button {
+                mb => {
+                    use LogicalDirection::*;
+                    match self.config.scrolling.to_logical_direction(mb) {
+                        Some(Up) => {
+                            let new_temp: u16 = self.current_temp + self.step;
+                            if new_temp < self.max_temp {
```
omertuc, line 148: <=

ammgws:
Does your output still look like this when scrolling?

```
[[{"background":"#000000","color":"#a9a9a9","full_text":"| ","markup":"pango","separator":false,"separator_block_width":0},{"background":"#000000","color":"#93a1a1","full_text":" 6500 ","markup":"pango","name":"8e354b15488f46a9bbbf97e89ccd7574","separator":false,"separator_block_width":0}],
[{"background":"#000000","color":"#a9a9a9","full_text":"| ","markup":"pango","separator":false,"separator_block_width":0},{"background":"#000000","color":"#93a1a1","full_text":" 6500 ","markup":"pango","name":"8e354b15488f46a9bbbf97e89ccd7574","separator":false,"separator_block_width":0}],
5
[{"background":"#000000","color":"#a9a9a9","full_text":"| ","markup":"pango","separator":false,"separator_block_width":0},{"background":"#000000","color":"#93a1a1","full_text":" 6500 ","markup":"pango","name":"8e354b15488f46a9bbbf97e89ccd7574","separator":false,"separator_block_width":0}],
5
[{"background":"#000000","color":"#a9a9a9","full_text":"| ","markup":"pango","separator":false,"separator_block_width":0},{"background":"#000000","color":"#93a1a1","full_text":" 6500 ","markup":"pango","name":"8e354b15488f46a9bbbf97e89ccd7574","separator":false,"separator_block_width":0}],
5
[{"background":"#000000","color":"#a9a9a9","full_text":"| ","markup":"pango","separator":false,"separator_block_width":0},{"background":"#000000","color":"#93a1a1","full_text":" 6500 ","markup":"pango","name":"8e354b15488f46a9bbbf97e89ccd7574","separator":false,"separator_block_width":0}],
5
[{"background":"#000000","color":"#a9a9a9","full_text":"| ","markup":"pango","separator":false,"separator_block_width":0},{"background":"#000000","color":"#93a1a1","full_text":" 6500 ","markup":"pango","name":"8e354b15488f46a9bbbf97e89ccd7574","separator":false,"separator_block_width":0}],
[{"background":"#000000","color":"#a9a9a9","full_text":"| ","markup":"pango","separator":false,"separator_block_width":0},{"background":"#000000","color":"#93a1a1","full_text":" 6100 ","markup":"pango","name":"8e354b15488f46a9bbbf97e89ccd7574","separator":false,"separator_block_width":0}],
```

As a sanity check can you change your scroll commands to "echo test" and see what the output looks like?

AkechiShiro:
I did what you told me, and here's is what I'm getting now.
It seems the 5 or the 4 are related to whether I scroll up or down.
The command I wrote inside the code is the following : `echo test >/tmp/debugi3rs 2>&1`
```
[{"background":"#000000","color":"#a9a9a9","full_text":"| ","markup":"pango","separator":false,"separator_block_width":0},{"background":"#000000","color":"#93a1a1","full_text":" 5700 ","markup":"pango","name":"90ac1a68fe8d44578291852aad88f6a9","separator":false,"separator_block_width":0}],
tail: /tmp/debugi3rs: file truncated
test
4
```
When I scroll up a 4 get printed, when I scroll down a 5 shows up, finally when I click onto the block a 1 shows up and the JSON error shows up as well.

AkechiShiro:
I will also add that a 3 is printed when I right-click on it and a 2 is printed when I use the `mouse3` button on it.
To me this seems like some kind of debug for the mouse events that is currently getting printed.

AkechiShiro:
I've got even more interesting now, even without the new block the JSON error appears with the numbers as well in the debug.
I just need to scroll anywhere on the status bar.
This bug seems to happen (I'm running i3-gaps, and the status bar having this error is the bottom one, I have an upper one too on which scrolling works).
I've tested the hueshift block on the upper one and the same JSON error does not appear at all.
On the bottom status bar there is also the i3 workspace numbering 1 to 10 showing and if depending on the direction on which I'm scrolling I can switch to other workspaces.

So i3 seems to be somewhat interfering with the i3status-rs JSON output I suppose, how that is happening is a little beyond me to be honest but the link is that scrolling does two actions on the bottom status bar.

ammgws:
What's the bare minimum config you can reproduce this with?

AkechiShiro:
This is the bare minimum config I can reproduce this with : 
```toml
[[block]]
block = "hueshift"
```
Or with any other config that doesn't have this block either, it happens even if I choose another block than the custom one I've made.
However this issue seems to be intermittent, right now when I scroll on the i3 bottom status bar, the JSON error does not get triggered anymore, this could be harder to reproduce since it doesn't seem to be related to the i3status-rs config file at all.

ammgws:
OK I can reproduce the numbers in the output, however they are only logged to stderr so it shouldn't have an affect on i3/swaybar. When using tee command from before it was redirecting stderr to stdout which explains why it was causing those JSON errors.

I think the stderr output was added by accident in https://github.com/greshake/i3status-rust/commit/fe517af58a5c10d7596cb7d4d6096cd7d5739052. I'm thinking about added a `--debug` flag since it has been [mentioned](https://github.com/greshake/i3status-rust/issues/669) before, but until then I'll just remove it for now.

AkechiShiro:
I can confirm that without the debug the sliding works with no errors on the bottom status bar. Seems the redirection of the stderr in stdout was indeed the culprit.
Right now, there is just one more issue I'm having, it's that when I scroll the color temperature does change. But then when an update of the bar happens the screen goes back to the default color temperature (6500K) instead of the number that is shown updated in the bar.
Any idea why that is happening ?

AkechiShiro:
The block is done and it's documentation is too, I did not test however changing the temperature when setting the hue_shifter to `sct` but with redshift the block seems to work as intended.

My issue came from a custom block I had made to change my color temperature every time the i3bar updated the command with the default 6500K would change back conflicting with the new `hueshift` block.
There are no more issue I believe, if any changes are required or any more documentation needed feel free to ask me and I'll make the required changes.
This PR I believe is now ready to be merged with master.
Thanks for all the help given here.

blocks.md
```diff
@@ -404,6 +405,34 @@ It requires a Github [personal access token](https://github.com/settings/tokens/
 
 For more information about reasons, please see the [API documentation](https://developer.github.com/v3/activity/notifications/#notification-reasons).
 
+## Hueshift
+
+Creates a block which display the current color temperature in Kelvin. When scrolling upon the block the color temperature is changed.
+A left click on the block sets the color temperature to the mean value from the `max_temp` and `min_temp`.
```
omertuc, line 411: AFAIK the default, no-adjustment temperature is 6500k, so ideally the defaults for min and max should have 6500 as the middle value? Right now it's 5500k. Or even better, set the left click value to something configurable, instead of the middle value. Because as it is right now, since 10000 is the current maximum, it forces the user to set the minimum to 3000 (which is subjectively pretty high for a minimum) in order for the middle value to be 6500.

Maybe I'm just misunderstanding the intended use case you chose the middle value for left click though

omertuc:
A bug I found while using it throughout the day: Seems like left click anywhere in the bar (for example, pausing/unpausing music) triggers the left click action on the hueshift block

blocks.md
AkechiShiro: I think you're right it's better to set it configurable for the left click.

The reason I choose the middle value was because I wanted to always have a `current_temp` between the range of `min_temp` and `max_temp`, but by default I think the middle value should be 6500K, it's much more natural.

In case, the `max_temp` is lower than 6500, then I'll just (by default) set the left click value to the `max_temp` in that case, I'm not too sure it's a good idea, but basically the key idea here is that we're trying to get as close to the default value as we can.

Concerning the bug, I've been having somewhat of a bug too, when I had the `backlight` block and this block in the same status bar, when scrolling on the backlight, the hueshift block changes color temperature as well, I'm not too sure how I would go around limitting the interaction to only the mouse hovering over the specific block text. I'll have a look at already implemented blocks to try and fix it.

Thank you for the useful feedback!

omertuc:
> A bug I found while using it throughout the day: Seems like left click anywhere in the bar (for example, pausing/unpausing music) triggers the left click action on the hueshift block

Figured out what causes it, you need to add a check that the click event was on your block ID, like this:
image

AkechiShiro:
Thanks for the help to fix the bug, I believe that the block is now fully ready, if you wish to test it some more and provide some feedback however I think there won't be any issues anymore.
I've also updated the doc describing the left click behavior and its config argument `click_temp`.
Maybe a neat upgrade would be to manage some kind of state, and when the `left_click` is pressed once the state switch and  the `click_temp` value is set then if another `left_click` is triggered a reset to a `default_temp` happens, but currently it is possible to do this by doing a left click then a right click so I'm not too sure it's really necessary.

ammgws:
LGTM and looks like you already have a happy user in @omertuc 

I checked some other status bars and apart from geolocation/GPS (which I don't think should be handled by us), it looks like we are feature complete.
https://bumblebee-status.readthedocs.io/en/latest/modules.html?highlight=redshift#redshift
https://py3status.readthedocs.io/en/latest/modules.html#hueshift

Thanks!

AkechiShiro:
Also, I'm thinking of opening another PR or continuing this one if possible? 

To improve the `hueshift` block at the moment.

The current implementation does not work under Wayland (which I was going to transition to), even if `i3status-rust` works on Wayland. 

Redshift  has a specific package which works well under Wayland provided you use the `-m wayland` method.

However, I'm thinking that to be broader, I should try to include more Wayland compatible hue shifting programs to the `hueshift` block such as [gammastep](https://gitlab.com/chinstrap/gammastep) unfortunately, there is no easy way to distinguish the `redshift` that is patched to work with Wayland and the one that isn't (the one that isn't just spits an error when given the `-m wayland` option).

aloispichler:
Thanks for this widget. It recognizes the `hue_shifter` automatically (_redshift_, on my machine) and changes the temperature upon scrolling.
But the shifter does not autostart, and it does not change the temperature displayed, if changed. I would expect that the temperature

MaxVerevkin:
> and it does not change the temperature displayed, if it changes during the day

I might be wrong, but AFAIK there is no way to query current temperature from `redshift`. This is not a limitation of this block because it is able to monitor changes when `wl_gammarelay(_rs)` is used.

> But the shifter does not autostart

Can you describe current behavior and expected behavior?

> Further, an icon is missing

Covered in #1465

naveg:
> I might be wrong, but AFAIK there is no way to query current temperature from redshift.

I believe `redshift -p` does this.

I agree that it would be a good improvement for the block to display the current temperature, as set by redshift itself outside of the status bar.

MaxVerevkin:
> I believe redshift -p does this.

Good to know! I did some initial work in 34e2b45ceffa32b072e283ba5dea2d3590d3063b, PRs implementing temperature updates for concrete back-ends are welcome.

PR #805, opened 2020-08-12
Docs: Update docs with xprofile envs

benmezger:
Add missing steps on adding env for i3 to be able to read from.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -383,7 +383,7 @@ Key | Values | Required | Default
 `format` | A format string. See below for available placeholders. | No | `"{total}"`
 `api_server`| API Server URL to use to fetch notifications. | No | `https://api.github.com`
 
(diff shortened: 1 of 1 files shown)


ammgws:
Hmm I'm not too sure about this, especially since there are several ways of achieving this aside from using .xprofile.

benmezger:
Trying to set in `.xprofile` here didn't work :(

benmezger:
Which ways? I was only able to get it working with =.xprofile=, but I would be happy to update this documentation if that's the case

ammgws:
Consider the simplest scenario in which you boot straight into a login shell and run i3 via `startx` from the command line. All you need is for the environment variables to be exported in the shell before you run `startx`, and then i3 should inherit them when it starts. In this case, there are many ways of setting the env vars: in your shell's init file (e.g. `~/.bash_profile`), `~/.xinitrc`, or probably even `~/pam_environment`. In your case I take it you are using some kind of display manager instead of booting into a shell, and it seems like `~/.xprofile` is required for the one you're using.

PR #806, opened 2020-08-13
 Memory block: add integer values for unit GiB

DCsunset:
It might be useful to support integer values for unit GiB for saving some space. The format strings are similar to the integer values of percentage, like `{MTgi}`, `{MAgi}` and etc.


GladOSkar:
This functionality would be included in the implementation of #724 / #750 i think

DCsunset:
Sounds good! That is a more general solution. I'll close this now.

PR #818, opened 2020-08-28
Add RSS icon

nikaro:
I need it for my custom block, so it would be nice to have it.


ammgws:
It's not currently used by any block, and if we add this there are probably many more which could be argued for instead. Is there a problem with just manually using the icon in your custom block config?

nikaro:
On 2020-09-02 09:58, Jason wrote:
>Is there a problem with just manually using the icon in your custom block config?

How would you do that? I did not find it in the manpage.

-- 
Cordialement,
Nicolas Karolak

ammgws:
Ah I guess you are trying to specify the icon using the `icon` field of the JSON payload? I am suggesting to just add the icon to the `text` field since it will achieve the same purpose. Can you show me your custom block config?

PR #820, opened 2020-08-28
Add warning_updates_regex to pacman block

brownjohnf:
This does pretty much what you expect: allows you to specify a warning
regex using the same syntax as the critical regex, and if any packages
in the warning regex have updates, the block will have the warning style
applied to it. If a critical regex package matches, the block will have
the critical formatting applied.

The critical regex will take priority over the warning; if both regexes
match, the block will be styled as critical. There are optimizations to
the code that could be made to short-circuit checking the warning regex
if the critical regex matches, but I think that's better left to a
different PR.


brownjohnf:
Something about my changes appears to break the GitHub checks (I get timeouts, when other PRs aren't having those issues). If anyone has any suggestions, they'd be welcome.

ammgws:
Remove the test config and it will fix the tests

brownjohnf:
Is there a way to test it with config? I guess in this case it's probably trying to call pacman and failing.

ammgws:
You'll just have to test it locally for now. We only have an Ubuntu based test setup now, and it's probably better to add unit tests with some predetermined pacman test output instead.

brownjohnf:
Okay, sounds good. I've been running it locally since opening this PR, and all appears to be good, but I'll add some unit tests. Thanks!

brownjohnf:
I started looking at adding tests here here, but the code as-is doesn't have a good way of testing this functionality.

ammgws:
In that case just forego them.

ammgws:
Rebase to remove this commit: https://github.com/greshake/i3status-rust/pull/820/commits/fb9539ea00e5eff56f4501e0c0c34efeb01fea4c

brownjohnf:
@ammgws Rebased, and also amended the commit to update the docs in blocks.md.

ammgws:
Thanks!

Issue #302, opened 2018-11-09
[feature request] reload config file on USR1 signal


faulesocke:
I have found no way to reload the configuration without restarting the whole process (which is only possible through restarting i3). However, this is inconvenient and from time to time causes weird display bugs in my system.

greshake:
What issues do you experience exactly when restarting i3 in-place (i.e. without restarting the X server)?

faulesocke:
For example sometimes i3 does not come up again, for unknown reasons. Or the UI hangs during i3 restart until I kill the compositor.

GladOSkar:
I use geany, which crashes on about every 5th i3 restart. But i think since the config isn't changed that often this is not very important.

atheriel:
I didn't realize people had such unstable systems... Given the current structure of the code, adding async configuration updates would be a pretty big undertaking, even if this is a desirable feature.

faulesocke:
Wouldn't this be easily achieveable by just calling exec() and restarting the program? If  that is not a preferred solution, I would agree with @GladOSkar and close this issue since it's not so important.

Also, another use for a signal would be updating all blocks like the classic i3status does on SIGURS1.

atheriel:
> Wouldn't this be easily achieveable by just calling exec() and restarting the program?

I hadn't thought of that. Sneaky. Any possible side effects?

faulesocke:
Maybe. Some connections/file descriptors could stay open when the close-on-exec flag isn't set there. However this can easily be checked by just trying it and then looking at /proc.

Another, probably more idiomatic solution would be to wrap the main function inside a `loop` and each time SIGUSR1 occours, exit from the main function and restart it. So, in other words, implementing restarting manually.

atheriel:
> Another, probably more idiomatic solution would be to wrap the main function inside a loop and each time SIGUSR1 occours, exit from the main function and restart it.

I do not believe that this is possible, due to the way we currently use per-block threads.

If someone whips up a proof-of-concept using the `exec()` approach we'd happily review it, but lacking a champion this feature is unlikely to be implemented. I'm marking this closed in the meantime.

sharethewisdom:
I think it is worth reopening this, considering one might like to use SIGUSR1 to change just the theme. Isn't that easier to implement? At the moment I'm reloading the window manager to mach the new `config.toml` symlinked to `config_dark.toml` or `config_light.toml`. The only changes are idle_bg and idle_fg in `[theme.overrides]`.

[...]

PR #822, opened 2020-08-29
Signal support

Gelox:
Add a **status** field to the custom block that takes a value between 0 and 30. These correspond to SIGRTMIN+0 to SIGRTMIN+30.
This causes the block to be updated by sending the corresponding SIGRT signal to i3status-rs. This can be done via
`pkill -SIGRTMIN+1 i3status-rs` in order to send a signal to blocks listening for signal 1.

This feature allows custom blocks to not need to be updated on an interval, but instead can be updated via scripts sending a signal when they need to update.

A possible flaw is that I was not able to find a good way of pulling in the signal range dynamically from the users computer, instead the range is hard-coded to listen to signals between 34 and 64 (where this is the range on my computer).
The relevant constants can be found in **src/signals.rs**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,6 +25,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arc-swap"
+version = "0.4.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4d25d88fd6b8041580a654f9d0c581a047baee2b3efee13275f2fc392fc75034"
+
+[[package]]
 name = "assert_fs"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -331,6 +337,7 @@ dependencies = [
  "serde",
  "serde_derive",
  "serde_json",
+ "signal-hook",
  "swayipc",
  "toml",
  "uuid",
@@ -729,6 +736,26 @@ dependencies = [
  "itoa",
  "ryu",
  "serde",
+]
+
+[[package]]
+name = "signal-hook"
+version = "0.1.16"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "604508c1418b99dfe1925ca9224829bb2a8a9a04dda655cc01fcad46f4ab05ed"
+dependencies = [
+ "libc",
+ "signal-hook-registry",
+]
+
+[[package]]
+name = "signal-hook-registry"
+version = "1.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a3e12110bc539e657a646068aaf5eb5b63af9d0c1f7b29c97113fad80e15f035"
+dependencies = [
+ "arc-swap",
+ "libc",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ serde_derive = "1.0"
 serde_json = "1.0"
 swayipc = "2.7"
 toml = "0.5"
+signal-hook = "0.1.16"
 uuid = { version = "0.8", features = ["v4"] }
 # Optional features/blocks
 libpulse-binding = { optional = true, version = "2.15.0", default-features = false }
--- a/blocks.md
+++ b/blocks.md
@@ -234,6 +234,16 @@ command = "uname -r"
 interval = "once"
 ```
 
+Display the screen brightness on an intel machine and update this only when `pkill -SIGRTMIN+4 i3status-rs` is called:
+
+```toml
+[[block]]
+block = "custom"
+command = ''' cat /sys/class/backlight/intel_backlight/brightness | awk '{print $1}' '''
+signal = 4
+interval = "once"
+```
+
 ### Options
 
 Note that `command` and `cycle` are mutually exclusive.
@@ -245,6 +255,7 @@ Key | Values | Required | Default
 `cycle` | Commands to execute and change when the button is clicked. | No | None
 `interval` | Update interval, in seconds (or `"once"` to update only once). | No | `10`
 `json` | Use JSON from command output to format the block. If the JSON is not valid, the block will error out. | No | `false`
+`signal` | Signal value that causes an update for this block with 0 corresponding to `-SIGRTMIN+0` and the largest value being `-SIGRTMAX` | No | None
 
 
 
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -112,6 +112,17 @@ pub trait Block {
         Ok(None)
     }
 
+    ///Sends a signal event with the provided signal, this function is called on every block
+    ///for every signal event
+    fn signal(&mut self, _signal: i32) -> Result<()> {
+        Ok(())
+    }
+
+    /// Returns the signal that this block is listening to, if any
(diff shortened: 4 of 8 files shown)


ammgws:
> A possible flaw is that I was not able to find a good way of pulling in the signal range dynamically from the users computer, instead the range is hard-coded to listen to signals between 34 and 64 (where this is the range on my computer).

I guess we just have to wait for this: https://github.com/nix-rust/nix/issues/1143

ammgws:
I haven't had time to test it yet, how has it been going for you so far?

Gelox:
I have used this feature on my own computer, it works fine. I updated the relevant blocks.md parts.
However the signal value constants I don't have an idea for how to make progress on. I think I started an issue in libc asking what they thought about exposing SIGRTMIN and SIGRTMAX but I haven't got a response nor do I think one will be given for a long time. I think the issue is that these specific constants change depending on your system and in libc they expose constants by hard-coding them. One would have to MAYBE create a function, but that entails its own troubles.

So far I don't know how to get the relevant constants in a dynamic way (perhaps we could add C-bindings).

The options that I see available are:
* Provide this feature behind a feature flag and keep the constants as they are
* Recreate the update mechanism without signals in some other way
* Do not merge this feature
* Keep the constants as they are, it will work on any system that has allocated 2 RT signals for system purposes (which is all the systems that I am familiar with, debian, arch, ..)
* Explore adding C-bindings
* Figure out another solution

I'll let the community (or you) decide with what direction they want to go, I do not mind helping with the decided way assuming it's not too much work.

Gelox:
I'll add on to that list of possible solutions.
If there would be a way to get the SIGRTMIN and SIGRTMAX values from a shell then we could run a shell command from rust to get it.
However as far as I can tell there is no such possibility. Getting this value seems to be exceedingly hard if you are not in C.
kill -L and kill -l seem to be out of date for me, so that's not a possibility to finding these values for all systems.

ammgws:
On mobile now so can't review the code but what happens in the case a user sets an invalid signal value?

Gelox:
```rust
//In ConfigBlock::new()...
if let Some(signal) = block_config.signal {
         // If the signal is not in the valid range we return an error
         if signal < 0 || signal > SIGMAX - SIGMIN {
            return Err(Error::BlockError(
                    custom.id,
                    format!("The provided signal was out of bounds. An allowed signal needs to be between {} and {}", 0, SIGMAX - SIGMIN)
                    ));
        } else {
            custom.signal = Some(signal + SIGMIN)
        }
    };
}
```

It returns an error from the CustomBlock::new() method.

ammgws:
Sorry I meant something like this: For example say we hardcoded the range as 34 to 64, what would happen in the two situations below.

1) SIGRTMIN starts at 35 on someone's system, and they choose signal 34 for one of their blocks.

2) Someone only has 20 realtime signals on their system and they choose signal 64 for one of their blocks.

https://man7.org/linux/man-pages/man7/signal.7.html

>        Starting with version 2.2, Linux supports real-time signals as
>        originally defined in the POSIX.1b real-time extensions (and now
>        included in POSIX.1-2001).  The range of supported real-time signals
>        is defined by the macros SIGRTMIN and SIGRTMAX.  POSIX.1-2001
>        requires that an implementation support at least _POSIX_RTSIG_MAX (8)
>        real-time signals.
> 
>        The Linux kernel supports a range of 33 different real-time signals,
>        numbered 32 to 64.  However, the glibc POSIX threads implementation
>        internally uses two (for NPTL) or three (for LinuxThreads) real-time
>        signals (see pthreads(7)), and adjusts the value of SIGRTMIN suitably
>        (to 34 or 35).  Because the range of available real-time signals
>        varies according to the glibc threading implementation (and this
>        variation can occur at run time according to the available kernel and
>        glibc), and indeed the range of real-time signals varies across UNIX
>        systems, programs should never refer to real-time signals using hard-
>        coded numbers, but instead should always refer to real-time signals
>        using the notation SIGRTMIN+n, and include suitable (run-time) checks
>        that SIGRTMIN+n does not exceed SIGRTMAX.

Or realistically, if SIGRTMIN and SIGRTMAX only vary on non-Linux kernels then we could hardcode the range without any any worries?

Gelox:
>SIGRTMIN starts at 35 on someone's system, and they choose signal 34 for one of their blocks.

Well this would cause incorrect behavior but not necessarily break anything. If their SIGRTMIN is 35 and they choose to listen to signal 0 (which corresponds to 34) then the result would be that for them to update i3status-rs they'd have to send SIGRTMIN-1 which would be an occupied signal.
However realistically they wouldn't send SIGRTMIN-1 signals, rather they'd try to send SIGRTMIN+0 and see that it does not update (as they'd be sending signal 35 rather than 34).
However I guess you would also be listening to potential 34 signals from the OS, which could potentially create problems?

>Someone only has 20 realtime signals on their system and they choose signal 64 for one of their blocks.

As far as I know SIGRTMAX does not vary, someone having 20 realtime signals would have to be a result of SIGRTMIN being 44 which would be the same problem as the previous one.

> Or realistically, if SIGRTMIN and SIGRTMAX only vary on non-Linux kernels then we could hardcode the range without any any worries?

Well, if we are only targeting Linux kernels and SIGRTMIN and SIGRTMAX do not vary on Linux kernels then no issue would appear. I have not seen a guarantee that they do not vary however. In fact in the man-page you referenced it seems to indicate that it can sometimes be 35 on Linux kernels.

It seems that libc is open to adding SIGRTMIN support to their library, I'm thinking I might try to add it myself when I get around to it. This means that eventually we should be able to add this feature correctly regardless of merging it right now, but if we do merge it right now we can always push a fix later.
:)

ammgws:
> it can sometimes be 35 on Linux kernels.

Yeah then we could just make our range start a bit higher like 40 - 64 to be safe, at least until libc is enhanced.

Gelox:
So, I just checked how difficult it is to bring the SIGRTMIN into rust through C bindings, super simple. I didn't realize this before so I learned my lesson.
Either way this should solve all of our problems, it does require we do 1 temporary C binding but this can be removed when libc updates their API and this propagates to the nix crate.

I'll try to update this right now. How does that sound?

ammgws:
Sounds reasonable to me. Hopefully nix/libc get updated by the next time we want to do a release.

Do you think we could also add a support for a signal that schedules an update for the whole bar (every block) at once? Maybe could use SIGUSR1 or SIGUSR2 for that. 

In addition to the above another use case we could hopefully cover at some time is reloading the config (see #302) but this one probably isn't trivial enough to cover with these PR.

Gelox:
I added the C-bindings, there shouldn't be an issue now.
As for adding support for an entire bar update that isn't added in this PR, I think that's best left to another PR.

I could also look a bit at the reloading thing. But again better left to another PR. :)

ammgws:
LGTM!

I suppose in the future the implementation could be made block agnostic so it can apply to any block, but it's not a cause for delaying this PR.

Thanks for working on this!

Gelox:
Fantastic! Currently no documentation exists that SIGUSR1 updates all blocks, I don't know where this fits. Not sure if it is necessary but just giving a heads up.

Thanks for letting me work on this! 💯

ammgws:
Good catch. Not really sure where to put it either. I think the README needs a rewrite.

PR #835, opened 2020-09-28
Support for hot-swapping batteries and allowing missing batteries without errors

Nukesor:
Hi!

First of all, thanks for providing this project! Looks great and it really has a super low memory/processing power footprint.

This PR is a follow-up to the other two existing battery PRs.
https://github.com/greshake/i3status-rust/pull/233
https://github.com/greshake/i3status-rust/pull/585

The idea of this PR is to allow several new things, but all visible changes will be hidden behind configurable feature flags.

The PR is based on https://github.com/greshake/i3status-rust/pull/585. Thanks to @flying7eleven for providing an easy entry point to this whole issue.

There are three issues, each addressed in it's own commit:

1. Users should decide, whether they want to ignore if a battery is missing or not. The current error message kills the whole bar and forces the user to reload i3.
This issue has been solved by introducing a new configuration flag `allow_missing_battery`. If set to `true`, there will be a simple placeholder text instead of crashing the whole bar.
2. Users should be able to completely hide the widget, if a battery goes missing.
This is something that comes from people using the same configuration accross several devices.
I myself, for instance, have two laptops (one at work, one at home) and one of them has a single battery, the other one has two.
This change simply enhances the ergonomics for users with multiple devices.
3. Hot-swapping batteries.
This needed a few more changes, since some values of the battery were previously only read once during initialization.
Since hot-swapping can change the specs of a battery device, these values now have to be read continuously.

GIF of a hot-swap in action.
Peek 2020-09-29 12-33

GIF of a hot-swap with `hide_missing = true`
Peek 2020-09-30 15-31
`


Brod8362:
This is great! I was just thinking of this the other day, having a battery that may not always be present shouldn't break the entire bar.

ammgws:
Yeah I think this is the best we can do, thank you @Nukesor and @flying7eleven for the clear overview and work on this issue. Perhaps it would be nice to add another format string to use when `allow_missing_battery` is turned on so one could have it display something other than "X%" or "xx:xx" when the battery is missing (like the KDEConnect block's `format_disconnected` option).

blocks.md
```diff
@@ -115,6 +115,8 @@ Key | Values | Required | Default
 `driver` | One of `"sysfs"` or `"upower"`. | No | `"sysfs"`
 `interval` | Update interval, in seconds. Only relevant for `driver = "sysfs"`. | No | `10`
 `format` | A format string. See below for available placeholders. | No | `"{percentage}%"`
+`allow_missing_battery` | Don't display errors when the battery cannot be found. Only works with the `sysfs` driver. | No | `false`
```
ammgws, line 118: Perhaps we could shorten this to `allow_missing`?

blocks.md
```diff
@@ -115,6 +115,8 @@ Key | Values | Required | Default
...
+`hide_missing_battery` | Completely hide this block, if the battery cannot be found. Only works in combination with `allow_missing_battery`. | No | `false`
```
ammgws, line 119: Perhaps we could shorten this to `hide_missing`?

blocks.md
flying7eleven: Yep, I agree. The battery is redundant in this context

ammgws:
Also could you share a gif or image of what the bar looks like with `hide_missing` is used?

Nukesor:
Hey :)
I renamed the parameters from `*_missing_battery` to `*_missing`. This is indeed a better name!

On top of that a new `missing_format` string has been added. This has been tested and works as intended.

I also added a few missing parameters to `blocks.md` in this commit.

ammgws:
LGTM

Thanks!

PR #836, opened 2020-09-29
Add mpd block 

Brod8362:
Add mpd block. Already documented in blocks.md.
This is written primarily because the music block is a little bit hack-y to work with mpd and also doesn't have too many customization options (when it comes to layout) 

The mpd block has support for playback flags (repeat, random, etc) which isn't supported by the music block. It also does not require any additional software to be installed to work with mpd.

Transport controls available with left, middle, and right click, as well as mouse wheel to control volume.


ammgws:
I'm a bit wary of adding just a block for mpd because it introduces yet another dependency, and opens the flood gates for other music players which would result in more maintenance burden versus the current approach of using the MPRIS interface.

> music block is a little bit hack-y to work with mpd 

What in particular felt hacky to you?

> music block also doesn't have too many customization options (when it comes to layout)

By layout do you mean lack of a `format` string? If so then perhaps this could be worked into the music block, albeit the info we can get over MPRIS may not be as abundant as your pure mpd approach.

Brod8362:
> I'm a bit wary of adding just a block for mpd because it introduces yet another dependency, and also opens the flood gates for other music players which would result in more maintenance burden versus the current unified approach of only using the MPRIS interface.

This a totally valid concern, and while I'm not very familiar with features that other MPRIS compatible media players have, it kind of limits just how much functionality you can have since it's a "one size fits all" solution.

> What in particular felt hacky to you?

Just having to install mpDris2 and having it running, its not the end of the world but also not ideal when a direction connection to mpd is possible.

> By layout do you mean lack of a `format` string?

Kind of, but not exactly. While you can customize which buttons you have, you can't change what information you want displayed, or how you want it displayed. I also prefer having the block itself act as the button instead of having tiny buttons that can be difficult to click on, in my implementation the entire block is the button and left/middle/right click controls the various transport controls. I was thinking about adding a customization option for this too, but I haven't gotten around to it yet.

If you don't want to merge it, I can totally understand, but I think that while MPRIS is a convenient solution, it's not the best solution, because it's missing on a lot of potential features that an individual music player could offer.

Seirdy:
> Just having to install mpDris2 and having it running

An alternative, easier-to-install MPRIS bridge is
[`mpd-mpris`](https://github.com/natsukagami/mpd-mpris) since it's a single binary
that can be statically-linked.

PR #838, opened 2020-09-30
[Bug] Fix sct hueshift and remove warning

Nukesor:
This PR fixes the `unreachable pattern` warning for the hueshift module.

```
warning: unreachable pattern
   --> src/blocks/hueshift.rs:226:9
    |
226 |         Some(_sct) => {
    |         ^^^^^^^^^^
    |
    = note: `#[warn(unreachable_patterns)]` on by default

warning: unreachable pattern
   --> src/blocks/hueshift.rs:245:9
    |
245 |         Some(_sct) => {
    |         ^^^^^^^^^^
```

Previously there were some match blocks that were trying to match on specific strings. Since that's not how Rust's match works, the block always used `redshift` and never `sct`.

This PR introduces a proper Enum for configuration parameter parsing and matching.

This has been tested with both `redshift` and `sct`. Scrolling and clicking both work as intended.

### DIFF
--- a/src/blocks/hueshift.rs
+++ b/src/blocks/hueshift.rs
@@ -23,7 +23,7 @@ pub struct Hueshift {
     current_temp: u16,
     max_temp: u16,
     min_temp: u16,
-    hue_shifter: Option<String>,
+    hue_shifter: Option<HueShifter>,
     click_temp: u16,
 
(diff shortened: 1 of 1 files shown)


ammgws:
Cheers

PR #839, opened 2020-09-30
[Code cleanup] Fix warning and clippy issues

Nukesor:
This removes all current clippy issues and compiler warnings except the one for `hueshift`, which is addressed in the other PR.


ammgws:
Thanks!

PR #842, opened 2020-10-01
Add theme examples

Nukesor:
This PR adds a simple example image for every theme there is.

The images aren't super fancy, but they should give users a basic feel for each theme.
Currently, the only other way to find out how stuff looks, is to manually cycle through each theme :D

I also deleted two other images, since I couldn't find any usages for those.


ammgws:
Thanks!

PR #844, opened 2020-10-04
Add custom actions to the battery block

Triton171:
This PR adds custom actions as mentioned in  #697 to the battery block. Due to #703 planning to use middle-click on all blocks, I've left out the middle-click, but feel free to suggest changes.

Edit: Apparently I'm too stupid to run rustfmt before committing...


ammgws:
I think in the case of #697 the solution should be a generic implementation on the base block that applies to every block, rather than having to add the code to each block one by one.

Triton171:
Yeah, that makes sense. Would it be a good idea to create a struct for storing the click actions and adding it as a member to each block?
I'm new to Rust so I'm not sure what's the best way to add the generic data to all blocks (as far as I'm aware, Traits can only contain functions and not member variables).

PR #846, opened 2020-10-07
nvidia block: don't crash when gpu is disabled

### LINKED ISSUE #826: nvidia_gpu block: hybrid graphics compability
`nvidia_gpu `block dont work on notebooks properly:

When nvidia card not using - nvidia-smi returns: 
`NVIDIA-SMI has failed because it couldn't communicate with the NVIDIA driver. Make sure that the latest NVIDIA driver is installed and running.` (return code: 9)

When nvidia card using - nvidia-smi returns as expected.

If disable nvidia card during `i3status-rs` - It failing all status string.

Suggestion:
hide block/disable output to block(Like a unconnected WLAN) when nvidia-smi cannot connect to dirver.

ammgws:
Resolves #826 (hopefully)

### DIFF
--- a/src/blocks/nvidia_gpu.rs
+++ b/src/blocks/nvidia_gpu.rs
@@ -18,18 +18,17 @@ use crate::widgets::text::TextWidget;
 
 pub struct NvidiaGpu {
     gpu_widget: ButtonWidget,
+    gpu_enabled: bool,
     id: String,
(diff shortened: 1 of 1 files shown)


ammgws:
Can confirm it works as intended after testing with a mock nvidia-smi script.

Issue #837, opened 2020-09-29
Conditional display of blocks


bascht:
Hi @greshake. I hope it's okay if I open up an issue for a request, but I couldn't find anything in the code, docs, or other related issues:

I would like i3status-rs to have some kind of conditional functionality that would let me show / hide different blocks based on their value / status, e.g.

- only display the *load block* if load is > 2
- hide the battery block if the battery is fully charged

I know some blocks provide that functionality but I think it would be a great addition to i3status-rs itself. Would you consider this a feature?

Brod8362:
I think these features would be best implements on a per-block level as opposed to a universal switch. The battery block already collapses when it's fully charged, but the load block does not have a similar option. If you know a bit of rust, it wouldn't be too difficult to jump in an add them yourself.

Regardless, I do think the load block could do with a bit more customization, although "load > 2" is pretty unclear since it could be 1m, 5m, or 15m load.

bascht:
@Brod8362 You might be right. Thanks for the extensive answer! I just scrolled through #847 and this looks way easier than I thought. Maybe I get to write some rust after all… :D 

You can close this issue at will! :raised_hands:

Seirdy:
Yeah, conditional display is great; +1 on that.

> I think these features would be best implemented on a per-block level as opposed to a universal switch.

@Brod8362 I see where you’re coming from; however, there exist way too many possible use-cases and conditionals to bake into i3status-rust. Some examples:

- Only display some blocks on hostnames matching a regex. My desktop doesn’t need to display battery status or disk usage (since it has a big disk).
- Complex battery conditionals:
  - If this computer is my laptop with the main battery, display the battery block if:
    - connected to AC *and* when the battery is above 50%, *or*:
    - when the battery is above 80% and discharging
  - If this computer is my laptop with a small spare battery, always display the battery block
- Only display the mail block for my work inbox during working hours.
- If my screen is small, save space by only displaying my UTC time block if the music block isn’t also displayed.

Most people are part of a unique edge case, but implementing each unique edge would take hundreds of hours and make i3status-rs 10x more complex.

The two easiest ways to handle this:

1.  Provide a mechanism for re-sourcing the configs, or document a simple way to restart i3status-rs without a nuclear option like reloading sway/i3 (reloading sway takes 5 seconds of full CPU utilization on my 2013 Elitebook). That way, other user-made scripts can alter the i3status-rs configs for specific edge-cases.
[...]

PR #847, opened 2020-10-08
Load block: Add option for conditional display

Jmc18134:
Adds an option to the Load block as suggested in #837 which hides the block unless the load exceeds given values.


src/blocks/load.rs
```diff
@@ -136,10 +148,23 @@ impl Block for Load {
...
-            / self.logical_cores as f32;
+        /* Parse all the loadavg values. The block is
+         * visible if they are all greater than their min
+         * in self.minimum_visible */
+        let mut parsed = [0.0; 3];
+        for (e, v) in parsed.iter_mut().zip(split.iter()) {
+            *e = v
+                .parse::<f32>()
+                .block_error("load", "failed to parse load value as float")?;
+        }
```
dancek, line 159: This could be written without a `mut`, e.g.
```suggestion
        /* Parse all the loadavg values. The block is
         * visible if they are all greater than their min
         * in self.minimum_visible */
        let parsed: Vec<f32> = split.iter().take(3).map(|v| v.parse::<f32>().unwrap_or(f32::INFINITY)).collect();
```

src/blocks/load.rs
```diff
@@ -136,10 +148,23 @@ impl Block for Load {
...
+
+        self.visible = self
+            .minimum_visible
+            .iter()
+            .zip(parsed.iter())
+            .all(|(min, v)| v > min);
```
dancek, line 165: It would be more intuitive to have `>=` IMHO. Consider minimum_visible 1.0: should 1.00 be displayed?
```suggestion
            .all(|(min, v)| v >= min);
```

ammgws:
Past comment which is relevant: https://github.com/greshake/i3status-rust/pull/243#issuecomment-417859077

ammgws:
Closing until a consensus is reached in #837

PR #851, opened 2020-10-10
Add new Notify block

ammgws:
Supersedes #789  

New block akin to [dunst](https://github.com/vivien/i3blocks-contrib/tree/master/dunst) from i3blocks, however with a couple of differences:

- all control/monitoring of dunst is done using D-Bus, so updates are instantaneous*
- block can be expanded in the future to support `mako` and other notification daemons (though not necessarily using D-Bus if they do not support it)

*This requires running the latest build of dunst which includes https://github.com/dunst-project/dunst/pull/766.

Leaving the block undocumented for now.

TODO: 
- introduce `driver` option and add support for `mako`
- officially document block
- add format specifiers {enabled} {disabled} to use for overriding the icon


Silipwn:
Hi, 
Just a small suggestion, I thought it would be cool to see the number of notifications waiting when we suspend the dunst. 
I think `dunstctl counter` exposes some stats we could use.
Of course, could be done via a custom script, but I thought might be useful to have in `notify`

PR #874, opened 2020-10-17
music block: reset title when player is gone from bus

### LINKED ISSUE #791: music block does not reset title on MPRIS closed
So it looks like music block does not listen or react to unregister of DBus service and it does not remove the tile of playing song.
This should be easy to fix.

I might have a look but for now lets mark it as unassigned.

ammgws:
Resolves #791

... or at least attempts to


ammgws:
Will include this in the upcoming music block refactor

PR #894, opened 2020-10-22
[theme-plain] have seperator without extra space

Stunkymonkey:
currently the plain theme has a weird spacing with the separator. I expect this is not intentional.

Screenshot-2020-10-22_150211


ammgws:
Strange indeed.

Thanks!

PR #895, opened 2020-10-22
temperature: Add Fahrenheit support.

rjframe:
Allow using the Fahrenheit temperature scale in the Temperature block's display and thresholds.

In order to adjust the default thresholds based on the scale used, I've used Options to provide delayed initialization in `Temperature`'s `new()`, rather than when serde deserializes the configuration. Alternatively I think we could manually deserialize the struct.


ammgws:
Cheers

PR #896, opened 2020-10-22
WIP: Fix net hiding

GladOSkar:
Fix net block issue where both `hide_inactive` and `hide_missing` had to be enabled to ever hide the block on either case.

This PR implements the functionality as i would expect it:

- active = !exists || !up
- show full block if active
- hide block if inactive && hide_inactive || !exists && hide_missing
- otherwise, show only icon


GladOSkar:
Sorry have some errors in there

Issue #891, opened 2020-10-22
net: support switching between ethernet and wireless


mvdan:
This block works fine for my wireless device:

```
[[block]]
block = "net"
device = "wlan0"
speed_up = true
interval = 2
```

However, when I connect ethernet and disable wifi, as expected the block then does nothing. It would be nice to instead see the traffic for the wired device, in my case `ens8u2`.

Here are two ideas to potentially solve this:

1) Default to using the device for the first default route in `ip route`; most setups will only have one such device. If the wrong one is detected, the user can always specify one via `device = "..."`.

2) Allow using a list of fallback devices, where the first one with non-zero traffic is used. In my case, I could use something like `devices = ["wlan0", "ens8u2"]`.

Both options would require extra work, but I see no alternative if we want to automatically switch over between two devices.

ammgws:
How about defining two blocks in your config and using `allow_missing` and `hide_missing`?

mvdan:
I thought about that, but that takes up more space. Right now, the "empty" wlan0 net block contains two icons: the wireless signal icon, and an "x" presumably to say that the device isn't enabled.

mvdan:
Perhaps option 3 is to make the block be omitted when the input (device) is disabled.

ammgws:
IIRC isn't that what `hide_missing` does? (hiding the block when the device can't be found)

mvdan:
Perhaps. I have to admit that the docs don't clarify the difference between missing and inactive, and the code in `net.rs` seems to use both in the same way as far as I can see.

I tried both, and I still see the wlan0 widget:

```
$ grep -C3 wlan0 .config/i3status-rust/config.toml 

[[block]]
block = "net"
device = "wlan0"
speed_up = true
interval = 2
hide_inactive = true
$ ip l | grep wlan0
4: wlan0: <BROADCAST,MULTICAST> mtu 1500 qdisc noqueue state DOWN mode DORMANT group default qlen 1000
```

I completely reloaded sway after changing the config each time, so that fully restarts i3status-rust. I did add the wired net block, which now shows up.

If this worked I'd be fine with it :)

GladOSkar:
I actually use both `hide_missing = true` + `hide_inactive = true` on each interface for your exact purpose, only showing those interfaces that i'm connected to.

Only issue is with alternating color schemes (see #349) but that's less important

ammgws:
@mvdan You need both `hide_inactive` and `hide_missing` in your config for it to hide the block

[...]

PR #897, opened 2020-10-22
Fix net block hiding issue

GladOSkar:
Fix net block issue where both `hide_inactive` and `hide_missing` had to be enabled to ever hide the block on either case. (#891)

This PR implements the functionality as i would expect it:

- active = !exists || !up
- show full block if active
- hide block if inactive && hide_inactive || !exists && hide_missing
- otherwise, show only icon

I tested all cases and it works as described.

Does that sound correct to you, @ammgws @mvdan ?

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -367,6 +367,7 @@ pub struct Net {
     speed_min_unit: Unit,
     speed_digits: usize,
     active: bool,
+    exists: bool,
     hide_inactive: bool,
     hide_missing: bool,
     last_update: Instant,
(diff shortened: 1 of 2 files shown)


src/blocks/net.rs
```diff
@@ -855,6 +857,8 @@ impl Block for Net {
         let is_up = self.device.is_up()?;
         if !exists || !is_up {
             self.active = false;
+            self.exists = exists;
```
ammgws, line 860: Maybe better rewritten as:
```rust
let self.exists = self.device.exists()?;
let self.active = self.device.is_up()?;
if !self.exists || !self.is_up {
...
```
?

ammgws (changes requested):
LGTM,  now `hide_missing` actually does what it was supposed to do.

src/blocks/net.rs
GladOSkar: Alright, i did it a little differently than you suggested because it's even cleaner to clarify that `inactive` includes `missing`

GladOSkar:
Alright i updated the block docs and rewrote the logic there a bit

GladOSkar:
No idea why the CI check is failing :/

ammgws:
Just need to run rustfmt, but otherwise it builds fine locally.

Thanks!

PR #920, opened 2020-11-08
Block khal

theamma:
I just created a new block for people using the standards based CLI and terminal calendar program [_khal_](https://github.com/pimutils/khal). It shows the amount of events of the current day and setting state in case thresholds are reached.


theamma:
I'm gonna close the PR in order to get feedback first.

ammgws:
I don't use khal so I can't offer feedback in terms how to improve it's usability/etc :(

theamma:
I decided to write a program that outputs JSON for use in a custom block. This is probably a better choice to not have dependencies to external programs like the pimutils khal and todolist.

Issue #900, opened 2020-10-23
equal amount of whitespaces as seperators in a single block


Stunkymonkey:
when using mono spaced fonts this catches my eye

some blocks have more spaces, then needed:

- wifi but no ethernet
![Screenshot-2020-10-24_005148](https://user-images.githubusercontent.com/1315818/97060269-1f092180-1593-11eb-9ccc-ba89ad4f1f0b.png)
- battery if fully charged
![Screenshot-2020-10-24_005121](https://user-images.githubusercontent.com/1315818/97060255-144e8c80-1593-11eb-9075-e46f5b0b4722.png)
- when connecting ethernet and wifi there is some weird spacing inbetween

is this intentional? I expect that is was programmed this way, because it was easy.

It would be more pleasing if the spacing is even on both sides.

I cloud help implementing a fix, but first I want some feedback and suggestions on how to do it or maybe why not to do it.

ammgws:
I doubt it's intentional to have uneven spacing. Can you post your actual config?

Stunkymonkey:
https://pastebin.com/x14cjkGe

ammgws:
So for the battery block your config is `"{percentage}%<space>{time}"` and here in one of the cases `time` is set to an empty string:

https://github.com/greshake/i3status-rust/blob/db9b567e31f497b3a35f25451b7286a31f5e6e59/src/blocks/battery.rs#L726-L732

Which means the output turns into `"{percentage}%<space>"`.

One way to fix it might be to include the space in `{time}` itself so you can set your config to `"{percentage}%{time}"`, but it would introduce an extra space for people who still have their configs set to `"{percentage}%<space>{time}"`...

Stunkymonkey:
maybe we could have a regex searching for more then two spaces and replace it with a single space?

Stunkymonkey:
this would then have some bad behavior if the user has a format string `{percentage}%<space><space>{time}`

Stunkymonkey:
> include the space in {time}

would prevent people to not have any space at all. Maybe that is something users want.

ammgws:
As you can see from the links below fixing spacing has been an on going battle, and likely the issues left now are not trivial to solve. For blocks that use format strings perhaps each format specifier such as `time` etc could each have a format string itself such as `{time_format}` but things might get out of hand

https://github.com/greshake/i3status-rust/pull/866
https://github.com/greshake/i3status-rust/pull/894
[...]

PR #923, opened 2020-11-10
[blocks] fix equal spacing in battery & sound & networkmanager

Stunkymonkey:
related to #900 .

If the battery if fully charged, then the percentage does not have an additonal space.

Same if the volume if muted (and percentage not printed)

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -22,7 +22,7 @@ use crate::scheduler::Task;
 use crate::util::{
     battery_level_to_icon, format_percent_bar, pseudo_uuid, read_file, FormatTemplate,
 };
-use crate::widget::{I3BarWidget, State};
+use crate::widget::{I3BarWidget, Spacing, State};
 use crate::widgets::text::TextWidget;
 
 /// A battery device can be queried for a few properties relevant to the user.
@@ -746,6 +746,7 @@ impl Block for Battery {
             self.output
                 .set_text(self.full_format.render_static_str(&values)?);
             self.output.set_state(State::Good);
+            self.output.set_spacing(Spacing::Hidden);
         } else {
             self.output
                 .set_text(self.format.render_static_str(&values)?);
@@ -781,6 +782,7 @@ impl Block for Battery {
                 "Charging" => "bat_charging",
                 _ => battery_level_to_icon(capacity),
             });
+            self.output.set_spacing(Spacing::Normal);
         }
 
         match self.driver {
--- a/src/blocks/networkmanager.rs
+++ b/src/blocks/networkmanager.rs
@@ -22,7 +22,7 @@ use crate::input::{I3BarEvent, MouseButton};
 use crate::scheduler::Task;
 use crate::subprocess::spawn_child_async;
 use crate::util::{pseudo_uuid, FormatTemplate};
-use crate::widget::{I3BarWidget, State};
+use crate::widget::{I3BarWidget, Spacing, State};
 use crate::widgets::button::ButtonWidget;
 
 enum NetworkState {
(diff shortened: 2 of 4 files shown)


Stunkymonkey:
@ammgws any thoughts?

ammgws:
On mobile now so hard to review, however if your battery config was "{time}%<space>{percentage}" instead of "{percentage}%<space>{time}" as mentioned in #900, does it still behave as expected?

Stunkymonkey:
> On mobile now so hard to review, however if your battery config was `"{time}%<space>{percentage}"` instead of `"{percentage}%<space>{time}"` as mentioned in #900, does it still behave as expected?

this does not touch the formatting directly. In #900 I wrote, that rewriting the formatting would be ideal. But this is a lot of work. Maybe it is worth it, but the solution i found is quiet good I think.

So i decided to look at the current code and find out how to fix it, with the already given functions. So I adjusted the `Spacing` of the blocks, that I use and are wrongly aligned.

ammgws:
Ah ok, IIRC you wrote that it solved all your issues so I was trying to figure out how this PR solved that particular one.

Stunkymonkey:
the fix is done in a different way, then discussed in the issue. Sorry for the confusion.

I discovered, that we would had to rewrite all blocks, which would definitely introduce bugs. Using the `Spacing`-Enum is fine for me, and all the blocks i am using are only displaying a single space at the start & end (what is wanted by default).

The only uneven spacing **in-between** happens if ethernet and wifi are used at the same time. But I consider this a minor issue.

ammgws:
I tested with the below config but the output was the same before and after this PR. What am I not seeing? 
```toml
[[block]]
block = "battery"
driver = "sysfs"
device = "CMB1"
format = "{percentage}% {time}"
allow_missing = true
hide_missing = true
```

**after PR**
with font awesome
full_text":"  43% 1:49 "
full_text":"  43% 1:14 "

no icons
"full_text":" BAT 44% 99:15 "
"full_text":" CHG 44% 1:17 "

**before PR**
with font awesome
"full_text":"  44% 2:57 "
"full_text":"  44% 1:23 "

no icons
"full_text":" BAT 46% 99:15 "
"full_text":" CHG 46% 1:11 "

ammgws:
The `sound` block one I was able to confirm:
after PR
"full_text":" MIC MUTED "

before PR
"full_text":" MIC MUTED  "

ammgws (approved):
Sound block one confirmed fixed

ammgws (approved):

ammgws:
OK I see why I couldn't reproduce the battery block one. Apparently "Not charging" is different to "Discharging" but I thought they were the same

```    /// Query the device status. One of `"Full"`, `"Charging"`, `"Discharging"`,
    /// or `"Unknown"`. Thinkpad batteries also report "`Not charging`", which
    /// for our purposes should be treated as equivalent to full.
```

ammgws:
My battery is taking too long to charge so I hardcoded it to 100% and confirmed the spacing bug is fixed:

With this PR: `"full_text":" FULL<space>"`
Before this PR: `"full_text":" FULL<space><space>"`

ammgws:
I'll take your word for the networkmanager one.

Thanks for your work!

Stunkymonkey:
by interest: when is the next release planned?

GladOSkar:
@Stunkymonkey From my experience a release happens about every 3-4 months depending on the amount of changes. Since the last one was just 2 weeks ago that might still be a while.

ammgws:
I've starting putting together the release notes but I want to give it at least another couple of weeks or so (some people run the -git version so if there are any issues with recent changes then we should find out soon).

Stunkymonkey:
no problem at all. Testing is always good. I was just curious. Thanks for all your awesome work :+1:

Issue #349, opened 2019-03-22
`hide_missing` looks weird with alternating color scheme


GladOSkar:
When a block is hidden, it is not removed from the alternating color sequence.

![screenshot](https://user-images.githubusercontent.com/9497832/54833063-a4449a00-4cbd-11e9-956c-22c9acdd545f.png)

Ideally when a block is hidden, it should be removed.

atheriel:
This might be tricky to fix, but it is for sure a bug. Alternating colours are handled in a very different place than the missing block code.

GladOSkar:
@atheriel 

The easiest solution would imo be to not push empty blocks in `main.rs:180`.

The clean solution would of course be to offer blocks a way to remove themselves entirely, but that seems overly complicated and would probably break a few assumptions.

GladOSkar:
@atheriel  The way to do it is probably yo tint the blocks after they have been rendered and not before. But that would of course require more computation because the tint is calculated on every update, not just when creating the blocks.

ammgws:
If it's not that big of a computation then I think we should go for it

PR #924, opened 2020-11-11
Fix alternating tint logic when blocks are hidden

ammgws:
Fixes https://github.com/greshake/i3status-rust/issues/349

@GladOSkar would you like to give it a go and see if it's fixed? Testing with the below config showed the colours match the current behaviour so I assume it's working:

```toml
[theme]
name = "slick"
[theme.overrides]
separator = "SEPARATOR"

[icons]
name = "awesome"

[[block]]
block = "time"

[[block]]
block = "time"

[[block]]
block = "time"

[[block]]
block = "time"
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -159,59 +159,15 @@ fn run(matches: &ArgMatches) -> Result<()> {
         return Ok(());
     }
 
-    let mut config_alternating_tint = config.clone();
-    {
-        let tint_bg = &config.theme.alternating_tint_bg;
-        config_alternating_tint.theme.idle_bg =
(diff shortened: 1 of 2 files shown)


GladOSkar:
Yes it works beautifully, thank you!

GladOSkar:
I wonder if it would make more sense to do the alternating from right to left than from left to right, since the right is basically the anchor of the status bar.

Currently, whenever a network device shows up or disappears on my system, the alternating color stays the same on the left of that block and flips on the right. I'd rather have it the other way around

Issue #925, opened 2020-11-11
Change alternating tint start point


ammgws:
I wonder if it would make more sense to do the alternating from right to left than from left to right, since the right is basically the anchor of the status bar.

Currently, whenever a network device shows up or disappears on my system, the alternating color stays the same on the left of that block and flips on the right. I'd rather have it the other way around

_Originally posted by @GladOSkar in https://github.com/greshake/i3status-rust/pull/924#issuecomment-725403106_

ammgws:
I guess it makes sense since the bar itself is filled from right to left. Only problem is that the blocks are ordered left-to-right in the JSON output so we need to reverse the order to calculate the tints and then reverse it again to print the blocks.

GladOSkar:
Yeah seems a bit clunky to implement. I always found it confusing that the first block in the config was the leftmost and not the rightmost one on the bar, but that can't be changed anymore.

Still, i'll have a look

GladOSkar:
I played around with it and have a solution in #927. A review would be appreciated :)

PR #927, opened 2020-11-11
Layout + tinting rework

GladOSkar:
In response to #925 

I restructured and cleaned up a lot which turned out not to be necessary for the final solution but might be nice anyways. I don't think i introduced any regressions but that should be tested.

Alternating tint is now anchored to the right by counting the visible blocks beforehand and starting the alternating sequence accordingly on the left.

I barely have any rust experience, please lmk if something could be done better :)

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -11,7 +11,6 @@ use std::process::Command;
 use getrandom;
 use regex::Regex;
 use serde::de::DeserializeOwned;
-use serde_json::value::Value;
 
 use crate::blocks::Block;
 use crate::config::Config;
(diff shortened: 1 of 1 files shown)


src/util.rs
```diff
@@ -199,30 +198,40 @@ macro_rules! map_to_owned (
 
 struct PrintState {
```
ammgws, line 199: Since has_predecessor no longer exists we can get rid of this struct

GladOSkar: I guess we can, i already wondered why that was a struct anyway

GladOSkar:
There ya go

ammgws:
Thanks for adding the comments as well!

PR #928, opened 2020-11-12
Highlight "net" module when tx/rx speed is above thresholds

jypma:
I've been using `i3status-rust` for a while with great pleasure, and felt that it would be nice if the "net" module highlights itself when the network is used (just like the CPU or Load modules).

This PR adds the ability for the "net" module to highlight itself in info/warning/error colors, depending on the rx or tx speed exceeding certain thresholds.

There are a few things that aren't quite done yet, though:
- [ ] Thresholds should be configurable (I can add this shortly)
- [x] ~Right now, only the text section of the "net" module lights up. The icon itself doesn't. This is different from the CPU or Temperature modules, where both the icon and text light up. I'm not sure what causes this difference.~

Any feedback / help is appreciated :-)

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeMap;
+use std::cmp;
 use std::ffi::OsStr;
 use std::fmt;
 use std::fs::{read_to_string, OpenOptions};
@@ -23,7 +24,7 @@ use crate::util::{
     escape_pango_text, format_number, format_percent_bar, format_vec_to_bar_graph, pseudo_uuid,
     FormatTemplate,
 };
-use crate::widget::{I3BarWidget, Spacing};
+use crate::widget::{I3BarWidget, Spacing, State};
 use crate::widgets::button::ButtonWidget;
 
 lazy_static! {
@@ -382,6 +383,8 @@ pub struct Net {
     rx_buff: Vec<u64>,
     tx_bytes: u64,
     rx_bytes: u64,
+    tx_bytespersec: u64,
+    rx_bytespersec: u64,
     use_bits: bool,
     speed_min_unit: Unit,
     speed_digits: usize,
@@ -687,6 +690,8 @@ impl ConfigBlock for Net {
             tx_buff: vec![0; 10],
             rx_bytes: init_rx_bytes,
             tx_bytes: init_tx_bytes,
+            tx_bytespersec: 0,
+            rx_bytespersec: 0,
             active: true,
             exists: true,
             hide_inactive: block_config.hide_inactive,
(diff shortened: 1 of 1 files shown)


GladOSkar:
The second point is because the net block has the icon on a separate widget from the mail text. Colors are only applied to a single widget in your case. Solutions would include a) moving the icon into the main widget (with format string placeholder?) or b) applying the same color to the icon widget.

ammgws:
There's also this request: https://github.com/greshake/i3status-rust/issues/397 (which makes more sense to me)

GladOSkar:
> There's also this request: #397 (which makes more sense to me)

That can probably be integrated into this one as an option

jypma:
Thanks for the feedback! Both widgets are now updated. As to incorporating #397, I have a few ideas:
- Have the proposed configuration options `info_speed`, `warn_speed` and `critical_speed` with the bytes/sec _above_ which the status would be set (defaulting to `u32::MAX`).
- Have additional options `info_signal_strength`, `warn_signal_strength` and `critical_signal_strength` with the percentage _below_ which the status would be set (defaulting to `0`).
- The actual status picked would be higher one of the above.

This would cover most cases, except being able to display a wired connection link being up or down. For that, we could add a third option:
- Have a setting `inactive_status`, which can be `idle`, `info`, `warn` or `critical`, which is enabled when that `[[net]]` section is not active (defaulting to `idle`).

There's potentially a lot of configuration options in here... do we go for all of them?

jypma:
If there's no further feedback, I'll try and implement all configuration options when I get some time.

ammgws:
I wonder if it's possible to keep the format string option but reintroduce the widgets that were removed when the format string was implemented and then have per widget state/colours instead.

ammgws:
Looks like we should be able to do this by doing something like https://github.com/greshake/i3status-rust/commit/714b29f1f0244c976780a0bf9c35c8fe9bbd6149

GladOSkar:
@MaxVerevkin already mentioned that he doesn't really like the `render_widgets()` function from that commit and i think i agree, it's not that nice and barely any blocks need/use it. I actually removed it in #1207 after we decided `speedtest` doesn't need it anymore.

So I'd say just including the icon in the main widget is the best solution here IMO.

MaxVerevkin:
@GladOSkar
> But we could also reintroduce that function if it's worth the extra effort

We can reintroduce but we will lose "short text" functionality for some blocks that would use it. Auto-hiding is much more valuable than highlighting if you ask me (in particular for net block).

So I suppose the only way to have "per widget" highlighting is to use pango, but as discussed earlier, i3 and sway renders it differently.

MaxVerevkin:

PR #932, opened 2020-11-15
Music block: Fix `hide_when_empty` not working

GladOSkar:
So for some reason my `music` block wasn't hiding with `hide_when_empty = true` even though i had no players open.

This change fixed that, although i don't _really_ know why or what the different variables actually mean.

Hope this is okay?

### DIFF
--- a/src/blocks/music.rs
+++ b/src/blocks/music.rs
@@ -677,7 +677,7 @@ impl Block for Music {
             .players
             .lock()
             .expect("failed to acquire lock for `players`");
-        if players.len() == 0 && self.hide_when_empty {
+        if self.current_song_widget.is_empty() && self.hide_when_empty {
             vec![]
         } else if players.len() > 0 && !self.current_song_widget.is_empty() {
             let mut elements: Vec<&dyn I3BarWidget> = Vec::new();


ammgws:
Which variables in particular aren't you sure about?

GladOSkar:
Well i don't know why `players.len()` wasn't `0` in my case since i didn't have any players open.

My config is just

```
[[block]]
block = "music"
marquee_interval = 8
marquee_speed = 0.4
hide_when_empty = true
```

ammgws:
How easily/often can you reproduce it? Perhaps we could try debugging to see the players array wasn't empty.

(One thing it might be: Videos/etc playing in Firefox and Chrome also show up as players)

GladOSkar:
Honestly i'm not sure. I'll switch back to `master` and see how often this occurs. Right now it has disappeared again.

I checked that i didn't have any players open in Chrome either at the time. But it also wasn't showing the "full" playback mode music block but only the icon when i had the issue.

What do you recommend for checking for running players?

ammgws:
`busctl --user list | grep mpris` should list all currently registered players

ammgws:
and `busctl --user get-property org.mpris.MediaPlayer2.xxx /org/mpris/MediaPlayer2 org.mpris.MediaPlayer2.Player Metadata` to see what its current status is

ammgws:
There's at least one bug I've found when cycling through players and the next one happens to have no metadata: the block gets stuck in the collapsed state.

GladOSkar:
I think this occurs when i had a player (like youtube) open in chrome and then close that tab. The player stays in the bus:

```
busctl --user list | grep mpris
org.mpris.MediaPlayer2.chrome.instance1140                                1140 chrome          oskar :1.64         session-c1.scope  c1      -          
org.mpris.MediaPlayer2.playerctld                                            - -               -     (activatable) -                 -       -
```

```
busctl --user get-property org.mpris.MediaPlayer2.chrome.instance1140 /org/mpris/MediaPlayer2 org.mpris.MediaPlayer2.Player Metadata
a{sv} 3 "xesam:album" s "" "xesam:artist" as 1 "" "xesam:title" s ""
```

ammgws:
FaIr enough. Might need to revisit this later

PR #938, opened 2020-11-17
Add native theme

### LINKED ISSUE #182: Native theme
In other words no theme at all.
Let i3bar itself choose the default colors and draw separator.

I.e. do not produce `color`, `background` values and use native i3bar separators instead of `"| "` blocks.

GladOSkar:
Intended to resolve #182 

This PR introduces a `native` theme. If this theme is selected,

* No background colors are rendered (transparent -> bar color)
* Native (i3/sway) separators are shown between blocks
* Separator blocks are not produced.

Implementation decisions:

* I've kept the foreground (text) color similar to the plain theme. Makes sense to me because we don't want to lose the block states
* Native separators stay hidden between widgets

It's a lot of special case code which is not nice, but it works.


ammgws:
I would have expected this to respect the colours set in the i3/sway bar config, i.e. have no `background` or `color` fields in our output at all.

Example bar config:
```
bar {
    position top
    status_command while date +'%Y-%m-%d %l:%M:%S %p'; do sleep 1; done
    
    colors {
        statusline #00ff00
        background #676767
        inactive_workspace #32323200 #32323200 #5c5c5c
    }
}
```

Default sway status bar which takes colours from the bar config:
image

i3status-rust using this PR:
image

GladOSkar:
Yes as i said above, i felt like it made sense to keep the text color to indicate the block state. I could remove that as well but thought it would remove a lot of valuable information a user might still want to see

ammgws:
@xzfc seemed fine with losing those colours based on his initial issue

Perhaps there could be `semi-native` and `native` to address both?

GladOSkar:
Hm yeah. I'll have a look

GladOSkar:
Alright i came up with this. Every color value is now an `Option`, which is nice because you could do stuff like "Transparent background when `Idle` but normal background when other state".

I feel like i had to fight the type system a lot with the different string types and probably didn't do this right? But i also couldn't figure out another way...

ammgws:
image
image
Something funky going on with the spacing near the separators? Looks like less spacing on the trailing end of the blocks.

GladOSkar:
image
Seems quite even to me

GladOSkar:
> Something funky going on with the spacing near the separators? Looks like less spacing on the trailing end of the blocks.

Oh sorry you're talking about the screenshot for the docs i see? Yeah that's not actually my screenshot, i just used the `plain` one and changed the background color because i didn't want to recreate the config

src/util.rs
```diff
@@ -300,22 +318,39 @@ pub fn print_blocks(
     Ok(())
 }
 
-pub fn color_from_rgba(
+pub fn color_from_rgb(
```
ammgws, line 321: Removed alpha since it wasn't being used?

GladOSkar: Yes i always was confused about it because the `i3bar` protocol doesn't even support alpha. It was never used.

ammgws: I think `i3-gaps` (https://github.com/Airblader/i3/blob/22059fc6f580462a81be0fbbc28821344362ce88/i3bar/src/main.c#L63) and `sway` both support the alpha channel though.

ammgws: Looks like vanilla i3 also has that option: https://github.com/i3/i3/blob/eb83b37936d5356291f49fdc6c1d23ac02fd1c5a/i3bar/src/main.c#L63

ammgws: https://classicforum.manjaro.org/index.php?topic=27138.0

GladOSkar: Huh I didn't know that, will revert later

src/util.rs
```diff
@@ -230,49 +230,67 @@ pub fn print_blocks(
...
             .iter()
             .map(|widget| {
+                let mut w_json: serde_json::Value = widget.get_rendered().to_owned();
                 if alternator {
                     // Apply tint for all widgets of every second block
-                    let mut w_json: serde_json::Value = widget.get_rendered().to_owned();
                     *w_json.get_mut("background").unwrap() = json!(add_colors(
-                        &w_json["background"].to_string()[1..],
-                        &config.theme.alternating_tint_bg
+                        w_json["background"].as_str(),
```
ammgws, line 240: Does this still work without trimming the leading "#" from the colour string?

GladOSkar: Curiously, yes as far as I can see. It's later trimmed again when it's split into RGB

src/util.rs
GladOSkar: Reverted

src/util.rs
GladOSkar: Just checked the actual color values in the output and they are correct without this trim. I don't know why this worked before, probably the reference didn't have it trimmed?

ammgws: Weird, when I did the alternating tint PR I remember having to trim it otherwise it would panic

GladOSkar: Well it works very reliably for me now

src/themes.rs
```diff
@@ -8,186 +8,223 @@ use crate::util;
...
-        idle_fg: "#ffffff".to_owned(),
-        info_bg: "#2196f3".to_owned(),
-        info_fg: "#ffffff".to_owned(),
-        good_bg: "#8bc34a".to_owned(),
-        good_fg: "#000000".to_owned(),
-        warning_bg: "#ffc107".to_owned(),
-        warning_fg: "#000000".to_owned(),
-        critical_bg: "#f44336".to_owned(),
-        critical_fg: "#ffffff".to_owned(),
+        native_separators: false.to_owned(),
```
ammgws, line 11: Pretty sure you can get rid of all these `to_owned`s
Try running cargo clippy.

ammgws: This should do the trick for 95% of it:
```
rustup component add clippy --toolchain nightly-x86_64-unknown-linux-gnu
cargo +nightly clippy --fix -Z unstable-options
```

GladOSkar: Hm yeah i left them there because i didn't know if they might be important.

Running your `clippy` command changed a lot of things in a lot of files, that's maybe for another PR. For now i'll only apply it to `themes.rs` and `util.rs`

ammgws: The other files were all fixed in master an hour or so ago so don't worry about it

src/util.rs
```diff
@@ -230,49 +229,67 @@ pub fn print_blocks(
...
+                        config.theme.alternating_tint_bg.as_ref()
```
ammgws, line 240: Try changing this to `.as_deref()`

ammgws, line 240: Try changing this to `.as_deref()`

src/util.rs
```diff
@@ -318,6 +335,28 @@ pub fn color_to_rgba(color: (u8, u8, u8, u8)) -> String {
...
+pub fn add_colors(
+    a: Option<&str>,
+    b: Option<&String>,
+) -> ::std::result::Result<Option<String>, Box<dyn std::error::Error>> {
+    match (a, b) {
+        (None, _) => Ok(None),
+        (Some(a), None) => Ok(Some(a.to_string())),
+        (Some(a), Some(b)) => {
+            let (r_a, g_a, b_a, a_a) = color_from_rgba(a)?;
+            let (r_b, g_b, b_b, a_b) = color_from_rgba(b.as_str())?;
```
ammgws, line 348: and then this can be changed back to `b`

src/util.rs
```diff
@@ -318,6 +335,28 @@ pub fn color_to_rgba(color: (u8, u8, u8, u8)) -> String {
     )
 }
 
+// TODO: Allow for other non-additive tints
+pub fn add_colors(
+    a: Option<&str>,
+    b: Option<&String>,
```
ammgws, line 341: and then this can be changed back to `&str`

ammgws (changes requested):
I think this should do it

src/util.rs
GladOSkar: Neat, i did not know this was a possibility :)

ammgws:
Big thanks!

ammgws (approved):

PR #943, opened 2020-11-21
Add new Apt block

ammgws:
From limiting testing it seems to work... I will leave it undocumented for now.

@GladOSkar  You wouldn't happen to be running Ubuntu or something?


GladOSkar:
I do not, sorry

PR #946, opened 2020-11-25
Document availability in home-manager

workflow:
`i3status-rust` is now availble as a [Home Manager](https://github.com/nix-community/home-manager) module for installation!

[Home Manager](https://github.com/nix-community/home-manager) provides for a declarative way to set up userland programs on all Linux distributions and MacOS (using [nix-darwin](https://github.com/LnL7/nix-darwin)) and is also a popular choice on NixOS.

Checkout [the list of available module options for i3status-rust here](https://nix-community.github.io/home-manager/options.html#opt-programs.i3status-rust.enable).

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,8 @@ Stable releases are packaged on some distributions:
 
 * On NixOS: `nix-env -iA nixos.i3status-rust`
 
+* With [Home Manager](https://github.com/nix-community/home-manager): `programs.i3status-rust.enable = true` [see available options](https://nix-community.github.io/home-manager/options.html#opt-programs.i3status-rust.enable) 
+
 Otherwise, you can install from source:
 
 ```shell


ammgws:
I don't know much about NixOS but does that mean https://github.com/greshake/i3status-rust/pull/519 is no longer an issue?

workflow:
@ammgws That's pretty much unrelated, and I believe the issue you linked was already resolved by https://github.com/NixOS/nixpkgs/issues/79140

Home Manager is a separate way to install user packages that uses nix, but is not necessarily confined to NixOS.
It runs on Mac and other *nixes as well, but many people in the NixOS ecosystem use it to get a more declarative configuration of their programs.
Home manager actually uses the Nix package (the one that was fixed in https://github.com/NixOS/nixpkgs/issues/79140) under the hood, but also gives you a way to declaratively manage your *configuration* as well.

Here's my `i3status-rust` PR against Home Manager as a reference: https://github.com/nix-community/home-manager/pull/1470

Thanks again for this awesome i3 bar by the way :)

ammgws:
Thanks for the explanation!

PR #947, opened 2020-11-25
Implement per-block colour overrides

ammgws:
This will allow you to override colours on a per-block basis:
image
```toml
icons = "awesome5"

[[block]]
block = "uptime"
[block.color_overrides]
idle_fg = "#00ff00"
idle_bg = "#da11ff"

[[block]]
block = "time"
[block.color_overrides]
idle_fg = "#ff0000ff"
idle_bg = "#00aabbff"

[[block]]
block = "music"
[block.color_overrides]
info_fg = "#0000ff"
info_bg = "#aabbcc"

[[block]]
block = "time"
[block.color_overrides]
idle_fg = "#ff0000ff"
idle_bg = "#deadbeef"
```

The implementation may be a bit hacky but there doesn't seem to be a better way to achieve this, and I don't foresee the config changing regardless of the implementation in code (so any future rewrites should not break existing configs using block.color_overrides). I would rather have this feature out sooner than keeping users waiting (possibly forever) for a rewrite of the codebase.

It's possible some more work may be required so anyone willing to test this would be greatly appreciated!

### DIFF
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -149,7 +149,34 @@ macro_rules! block {
         let block_config: <$block_type as ConfigBlock>::Config =
             <$block_type as ConfigBlock>::Config::deserialize($block_config)
                 .configuration_error("Failed to deserialize block config.")?;
-        Ok(Box::new($block_type::new(block_config, $config, $update_request)?) as Box<dyn Block>)
+        let mut main_config = $config;
+        if let Some(ref overrides) = block_config.color_overrides {
+            for entry in overrides {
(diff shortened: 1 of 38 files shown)


elliotf:
This works for me. My use case was to be able to quickly distinguish between the four clocks I have in different timezones: ```[[block]]              
block = "time"         
interval = 60          
format = "%T%z"        
timezone = "UTC"       
#colour = "#AAFF88"    
[block.color_overrides]
idle_fg = "#aaffaa"    

# repeat with different colors for different timezones
```

PR #953, opened 2020-11-28
Re-add `alternating_tint_fg`

### LINKED ISSUE #952: Configuration parse error
Today I updated to the latest master and a configuration error started to come up:

```
Configuration error: failed to parse TOML from file contents.
Cause: invalid type: map, expected a string for key `theme` at line 139 column 1
```

My [config](https://up1.secretalgorithm.com/#55rkQW6Cy-N0aVqfTEiUrg) for reference.

I didn't found a break change in the commit logs, so I thought it was all ok to update. Do you know the reason for the error?

ammgws:
Resolves #952

@GladOSkar﻿ I'm guessing removing it was unintentional? Or was it removed because after #924/#927 we found that it wasn't actually being used anywhere?

### DIFF
--- a/src/themes.rs
+++ b/src/themes.rs
@@ -23,6 +23,7 @@ lazy_static! {
         separator_bg: Some(String::from("auto")),
         separator_fg: Some(String::from("auto")),
         alternating_tint_bg: Some(String::from("#111111")),
+        alternating_tint_fg: Some(String::from("#111111")),
     };
 
     pub static ref SOLARIZED_DARK: Theme = Theme {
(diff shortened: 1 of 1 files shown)


GladOSkar:
Yeah i removed it after i searched the entire source and didn't find it used anywhere. Forgot that it might break configs, sorry

ammgws:
No worries! I think I may have messed up the logic by accidentally leaving it out when implementing #924, since before that PR it looks like tint_fg was actually being used.

GladOSkar:
Ahh that would make sense. Not sure if it really makes sense as an option though but it doesn't hurt to have it i guess

PR #957, opened 2020-11-30
Add a calendar icon

theamma:
With Reference to issue 956, please add a calendar icon.


src/icons.rs
```diff
@@ -94,6 +95,7 @@ lazy_static! {
         "bell" => " \u{f0f3} ",
         "bell-slash" => " \u{f1f7} ",
         "bluetooth" => " \u{f294}",
+        "calendar" => " \u{f073} ",
```
ammgws, line 98: AWESOME5 and MATERIAL need updating too

theamma: Sorry. I Just added both lines.

ammgws:
Thanks!

PR #968, opened 2020-12-05
Bugfix for issue 965

### LINKED ISSUE #965: 0.14.3 breaks custom theme configuration
Upgraded to 0.14.3, but found that I now get an error in the status bar (ran directly in terminal to get the output):

```
$ /usr/bin/i3status-rs ~/.config/i3status-rust/config.toml

[[{"background":"#000000","color":"#dc322f","full_text":" Configuration error: failed to parse TOML from file contents.\nCause: invalid type: map, expected a string for key `theme` at line 5 column 1 ","separator":false,"separator_block_width":0}]


Configuration error: failed to parse TOML from file contents.
Cause: invalid type: map, expected a string for key `theme` at line 5 column 1
```

`~/.config/i3status-rust/config.toml`:

```toml
[theme]
file = "onedark.toml"


[[block]]
block = "time"
format = "%a %d/%m %R"
```

`~/.config/i3status-rust/themes/onedark.toml`:

```toml
idle_bg = "#282c34"
idle_fg = "#abb2bf"
info_bg = "#61afef"
info_fg = "#282c34"
good_bg = "#282c34"
good_fg = "#abb2bf"
warning_bg = "#e5c07b"
warning_fg = "#282c34"
critical_bg = "#e06c75"
critical_fg = "#282c34"
separator = ""
separator_bg = "auto"
separator_fg = "auto"
alternating_tint_bg = "#282c34"
alternating_tint_fg = "#abb2bf"
```

Downgrading to 0.14.2 sees everything working normally. The error would suggest that I need to use a different style of configuration to specify a custom theme, though I'm not sure what that would be.

ammgws:
This should resolve #965, however I wonder if it could be made better by dropping the `native_separator` key and somehow using `separator` itself (by special casing "None")

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -138,7 +138,7 @@ where
     D: Deserializer<'de>,
 {
     map_type!(Icons, String;
-              s => Ok(Icons(icons::get_icons(s).ok_or("cannot find specified icons")?)));
+              s => Ok(Icons(icons::get_icons(s).ok_or(format!("cannot find icon set called '{}'", s))?)));
 
     deserializer.deserialize_any(MapType::<Icons, String>(PhantomData, PhantomData))
 }
--- a/src/themes.rs
+++ b/src/themes.rs
@@ -8,7 +8,7 @@ use crate::util;
 
 lazy_static! {
     pub static ref SLICK: Theme = Theme {
-        native_separators: false.to_owned(),
+        native_separators: Some(false),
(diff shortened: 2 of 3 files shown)


GladOSkar:
Whoops that was my bad i guess, totally forgot about config-defined themes. I agree making `separator` ans `Option` would probably be a neater solution

PR #975, opened 2020-12-11
kdeconnect: update for new D-Bus structure

### LINKED ISSUE #974: Battery info broken in the latest version of kdeconnect
D-Bus structure changed: https://github.com/KDE/kdeconnect-kde/commit/f90d6a019ff288c4a16f91c2d7baa617ce793077

ammgws:
Things that changed in the latest kdeconnect ver: battery (https://github.com/KDE/kdeconnect-kde/commit/f90d6a019ff288c4a16f91c2d7baa617ce793077) and notification plugins (https://github.com/KDE/kdeconnect-kde/commit/4439e7c662c7829ba930efd0cbc504b533e02a26)

Resolves #974

- [x] fix initial method calls
- [x] fix signal handlers
- [x] figure out how to support both the old and new versions

kdeconnect v20.11.80 new
kdeconnect v20.08.03 old

### DIFF
--- a/src/blocks/kdeconnect.rs
+++ b/src/blocks/kdeconnect.rs
@@ -161,20 +161,19 @@ impl ConfigBlock for KDEConnect {
             format!("/modules/kdeconnect/devices/{}", device_id),
             Duration::from_millis(5000),
         );
-
         let initial_name: String = p2
             .get("org.kde.kdeconnect.device", "name")
             .unwrap_or_else(|_| String::from(""));
+        let initial_reachable: bool = p2
+            .get("org.kde.kdeconnect.device", "isReachable")
+            .unwrap_or(false);
 
-        let (initial_charge,): (i32,) = p2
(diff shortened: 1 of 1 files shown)


ammgws:
Signal handlers should be able to be fixed if https://invent.kde.org/network/kdeconnect-kde/-/merge_requests/365 is merged.

ammgws:
OK so now battery is fixed, not sure about notifications yet. Might still be broken - need to find a better way to test it.

So far this has only updated the block to use the new D-Bus structure; work still needed to detect what version of kdeconnectd is running and then adjust our handlers appropriately. Important since it's likely users on Ubuntu and the like will still be on the old version for a while.

ammgws:
As far as I can tell notifications are now working again too.

ammgws:
Manually downgraded to the old kdeconnect and looks like it's working.

Code is ugly as hell but it'll do for now

PR #984, opened 2020-12-17
Add universal formatter

GladOSkar:
I was annoyed by the fact that the ping in the `speedtest` block always has 2-3 decimals, not adhering to the `speed_digits` option that the up/down values use.

So i changed the `format_speed` function to `format_number` and made it unit-agnostic so it can be used for any SI-scaled units.

What do you think @ammgws?

(Maybe technically there should be a separate `ping_digits` but i figured that'd be overkill)

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -1295,7 +1295,7 @@ Key | Values | Required | Default
 ----|--------|----------|--------
 `bytes` | Whether to use bytes or bits in the display (true for bytes, false for bits). | No | `false`
 `interval` | Update interval in seconds. | No | `1800`
-`speed_digits` | Number of digits to use when displaying speeds. | No | `3`
+`speed_digits` | Number of digits to use when displaying speeds and latencies. | No | `3`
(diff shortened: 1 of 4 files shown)


ammgws:
Would #750 supercede this?

GladOSkar:
No it's unrelated but might make the undertaking of #750 a little bit easier

ammgws:
OK I see now. Perhaps it should be mentioned in the block docs that the ping format will also be affected?

GladOSkar:
> Also from memory I thought #750 would allow you to set the precision (number of digits) for anything that uses a format string hence why I asked.

Yes exactly that was the idea. What i'm saying is #750 would have required some modifications similar to these ones as well (although they did them very differently) but also doesn't include support for <1 SI prefixes i also don't know if/how they kept the "always N digits" system

ammgws:
OK I see what you mean now. I guess it is a reasonable compromise to have ping also follow `speed_digits`.

Issue #293, opened 2018-10-18
[net] emojis in SSID are not forwarded; hex code shown instead


sebastianst:
I'm sometimes connected to a wireless network with emojis in its name. Those are not shown in the title, but their hex code, including `\x`. Like so `\x0f0\x9f\x8d\x8d`.
It would be great if i3status-rs would just forward unicode chars to i3bar.
Thanks!

atheriel:
It might worth checking if your bar font supports emojis as well.

sebastianst:
It does, I use emojis for my keyboard layout indicator and the CPU symbol. I use
```
bar {
    font pango:DejaVu Sans Mono, FontAwesome 10
...
```
Running `i3status-rust` in a terminal doesn't print the emojis, but escaped slashes and hex codes, so you see `\\x0f\\x9f\\x8d\\x8d` where the ssid is. Which means, `i3status-rs` already internally converts the emojis into multi byte hex codes.

sebastianst:
It would probably also just print weird boxed glyphs otherwise...

sebastianst:
I found out the reason. The SSID is queried by calling `iw`, which returns the ssid with all unicode chars as escaped hex codes, see https://github.com/greshake/i3status-rust/blob/192274fe392f9658ca824d57166c7a2a801840b1/src/blocks/net.rs#L101
The good news is, this line also has `nmcli -g general.connection device show {}` as a backup should the first call fail and nmcli returns the SSID as is.

I propose to change the order of commands to `nmcli ... || iw ...` because it is better anyways to first query `nmcli` instead of `iw | grep | sed`. Also note here that the `grep | sed` call could be combined into a single sed call. I also think that this parsing should rather be done in rust using regexp than calling `sh -c 'iw | grep | sed'` ;)

jheyens:
nmcli is part of NetworkManager, though, which most i3 (and thus i3status-rust) users don't even install on their machines. 

I spent the last hour trying to find a clean linux way of obtaining the ESSID of the currently connected network, but unfortunately, there are no good news. 
The ESSID attribute is neither exposed to `/proc/net/` not to `/sys/class/net` filesystems. Additionally, due to Linux' history of wifi interfaces, we won't ever be able to get a generic way of obtaining this information. 
The `iw` tool links against the `nl80211` kernel library and is probably the most widespread choice. 
[...]

PR #995, opened 2020-12-23
Decode escaped unicode in SSID

2m:
It so happens that my home SSID name has some Lithuanian characters. That inspired me to implement escaped unicode decoding for the SSID names which was discussed here https://github.com/greshake/i3status-rust/issues/293#issuecomment-442554796 some time ago.

I am a beginner in Rust so any feedback is greatly appreciated and will be quickly addressed. :)

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -1207,4 +1207,12 @@ mod tests {
             Some(r"\x backslash".to_string())
         );
     }
+
+    #[test]
+    fn test_ssid_decode_surrounded_by_spaces() {
+        assert_eq!(
+            maybe_ssid_convert(Some(r"\x20surrounded by spaces\x20".as_bytes())).unwrap(),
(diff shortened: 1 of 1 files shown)


ammgws:
Do we need any of the other logic used in the C function mentioned in https://github.com/greshake/i3status-rust/issues/293#issuecomment-431582708?

2m:
All of the testcases were derived empirically. I set my wifi name to the expected test value and used the output from `iw` as the input for the testcases.

ammgws:
Yeah I just don't understand what that linked C function is doing by checking for and special casing `" "` and whether it is something we need to check as well.

2m:
Aha! It checks for a space at the beginning or the end and then encodes that. For example `\x20surrounded by spaces\x20`. I will add a test case for that.

ammgws:
Cheers! Don't know why it does that but I guess we may as well copy the behaviour.

PR #1006, opened 2021-01-17
Implement on_click for all blocks

### LINKED ISSUE #697: Feature request: Enhanced on_click, on_right_click, on_scroll etc commands for every block

edwin0cheng:
This PR allow allow all blocks to support `on_click` properties by implementing following things:

1. Add `BaseBlock` for common properties for blocks (right now only 'on_click')
2. Allow child block override this `on_click` behavior by adding a trait method in `Block` trait. It is because some blocks maybe have a complicated flow for clicking.
3. Added id for `TextWidget` ,`GraphWidget` and RotatingTextWidget` for letting `I3BarEvent` name available.

It partially fixes #697, other common properties could be added in the same fashion later.

Document is lacking, but will be submitted later if needed.

### DIFF
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -144,11 +144,15 @@ pub trait ConfigBlock: Block {
     ) -> Result<Self>
     where
         Self: Sized;
+
+    fn override_on_click(&mut self) -> Option<&mut Option<String>> {
+        None
+    }
 }
 
 macro_rules! block {
     ($block_type:ident, $block_config:expr, $config:expr, $update_request:expr) => {{
-        let block_config: BaseBlockConfig<<$block_type as ConfigBlock>::Config> =
+        let mut block_config: BaseBlockConfig<<$block_type as ConfigBlock>::Config> =
             BaseBlockConfig::<<$block_type as ConfigBlock>::Config>::deserialize($block_config)
                 .configuration_error("Failed to deserialize block config.")?;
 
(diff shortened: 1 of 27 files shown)


ammgws:
This LGTM, any possible issues would be internal and can be dealt with later.

Things to figure out: 
- the best way to expose it via config if/when implementing support for right, middle click and scroll up/down

- whether we need more context for config errors
- 
 with v0.14.3:
> Configuration error: Failed to deserialize block config..
> Cause: unknown field `this_doesnt_exist`, expected one of `interval`, `collapsed`, `scale`, `good`, `idle`, `info`, `warning`, `format`, `chip`, `inputs`, `color_overrides`

 with this PR:
> Configuration error: Failed to deserialize block config..
> Cause: unknown field `this_doesnt_exist`

when testing with
```toml
[[block]]
block = "temperature"
this_doesnt_exist = "alacritty"
```

Might be best to leave this undocumented for now.

edwin0cheng:
I think the differences between both version is actually a bug of `serde` , so I filed https://github.com/serde-rs/serde/issues/1957

> This LGTM, any possible issues would be internal and can be dealt with later.

So what's my next step ?

ammgws:
Perhaps just this: https://github.com/serde-rs/serde-rs.github.io/pull/120

Otherwise I'd be glad to merge and leave further work for another PR

edwin0cheng:
@ammgws I reworked the serialization part to paper over that issue, and the output error is much better now: 

```
Configuration error: Failed to deserialize block config..
Cause: unknown field `this_doesnt_exist`, expected one of `interval`, `collapsed`, `scale`, `good`, `idle`, `info`, `warning`, `format`, `chip`, `inputs`, `color_overrides`
```

But you could see there is no "on_click" in there, it is because that error is generated by serde, and that is just a `std::io::Error::custom(str)` , which is hard to modify except patching the string directly.

If you think it is okay, I think we could merge it and lets further work for another PR.

ammgws:
Cheers!

PR #1007, opened 2021-01-19
Add PiHole block

EugeneKorshenko:
New block for monitoring PiHole server (https://pi-hole.net project) as suggested here https://github.com/greshake/i3status-rust/issues/715. Clones appropriate _bumblebee-status_ extension.

P.S.: This is my first Rust code, so please don't throw tomatoes a lot :smile:

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -25,6 +25,7 @@
 - [Notmuch](#notmuch)
 - [Nvidia Gpu](#nvidia-gpu)
 - [Pacman](#pacman)
+- [PiHole](#pihole)
 - [Pomodoro](#pomodoro)
 - [Sound](#sound)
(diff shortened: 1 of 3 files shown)


ammgws:
> as suggested here #715.

Sorry if there was any confusion, however that was only a list of blocks from other status programs. As discussed over there, for blocks such as this one I believe they are trivial enough to be implemented using a Custom or Toggle block.

ammgws:
Thanks anyway!

EugeneKorshenko:
@ammgws Thanks. Sorry, generally agree, that this block can be relatively easily implemented via Custom block.

EugeneKorshenko:
@ammgws Sure. Here is a PR: #1226

Issue #905, opened 2020-11-05
Taskwarrior block filter is overly restrictive


matt-snider:
The current implementation of the taskwarrior block only allows specifying a filter of the form `+tag1 +tag2 +tag3 ...` which does not allow for negating tags (`-tag1`), or specifying `or` instead of the implicit `and` ( `+tag1 +tag2 or +tag2`). An example of a query I'd like to do would be `+PENDING +OVERDUE or +DUETODAY`. Unless I have missed something, this isn't possible.

For reference, here is a minimal configuration:

```toml
[[block]]
block = "taskwarrior"
filter_tags = ["tag1", "tag2"]
```

The cause of the restrictiveness is that the block already prepends the `+`: 

```rust
fn tags_to_filter(tags: &[String]) -> String {
    tags.iter()
        .map(|element| format!("+{}", element))
        .collect::<Vec<String>>()
        .join(" ")
}
```

This seems unnecessary because the effort for the user to specify `+` themselves is negligible, and it reduces the block's flexibility significantly.

I have two different ideas for how we could improve the block:

1. Prefix with `+` only if the tag does not already start with `+` or `-`  
    * Allows more complex queries, but still does not allow for specifying `or` 
    * Is backward compatible and a minimal change in the code
2. Allow specifying a `filter` string instead of `filter_tags` which is passed to the `task` command as is
    * Is very flexible
    * Allows an easy to switch from `filter_tags` to `filter`
    * Is more work for users to change - we could deprecate the option initially, and simply convert `filter_tags` to `filter` at the parse stage using the existing `tags_to_filter()` function.

I would be grateful to get some feedback from those who have worked on and use this block - @flying7eleven @HPrivakos @jubnzv let me know what you think

flying7eleven:
The implementation - as it is right now - was due to my own use-case. Additionally I wanted to keep it as simple as possible for the user. I can completely understand that this is not flexible enough for everyone. I would go for the second option since even the backward-compatibility can be easily achieved with your posted approach. 👍🏻

matt-snider:
@flying7eleven Thanks for the feedback. I'll give it a shot! :-)

ammgws:
Closed via @1008

PR #1008, opened 2021-01-20
Support multiple filters in Taskwarrior block

matt-snider:
Hi there!

First of all, thanks for the work on this great project. It's something that I use every single day, and is always visible on my monitor  and working very well :-)

Secondly, this is my first contribution to this project, so apologies in advance if I've missed something.

------

This PR addresses #905 and generalizes the Taskwarrior block by supporting an arbitrary number of filters.

Multiple filters can be specified as follows:

```toml
[[block.filters]]
name = "today"
filter = "+PENDING +OVERDUE or +DUETODAY"
[[block.filters]]
name = "work"
filter = "proj:work +PENDING +OVERDUE or +DUETODAY"
```

Deprecations:

* `filter_tags` are deprecated in favour of `filters`
* `TaskwarriorBlockMode` is removed in favour of `filter_index` since we can now cycle through all filters
* If `filter_tags` are specified, two filters are created to emulate existing `TaskwarriorBlockMode` behaviour:
  * all: `-COMPLETED -DELETED`
  * filtered: `-COMPLETED -DELETED <tags>`
* New-style `filters` do not prepend the `-COMPLETED -DELETED` criteria, since a user may want to specify something other than pending tasks. The user is completely free in specifying this value.

Just an additional note: the first commit in this PR (49336c0f3c86dc3c7e22e8dd27ab1453e269d1b6) was my initial solution, where only a single, more flexible filter can be specified. I built on that to a solution which I personally think is better, but if you prefer the first solution, you can take a look and let me know.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -1303,13 +1303,12 @@ Key | Values | Required | Default
 
 ## Taskwarrior
 
-Creates a block which displays number of pending and started tasks of the current users taskwarrior list.
+Creates a block which displays the number of tasks matching user-defined filters from the current user's taskwarrior list.
 
-Clicking the left mouse button on the icon updates the number of pending tasks immediately.
+Clicking the left mouse button on the icon updates the number of tasks immediately.
 
-Clicking the right mouse button on the icon toggles the view of the block between filtered (default) and non-filtered
-tasks. If there are no filters configured, the number of tasks stays the same and both modes are behaving
-equally.  
+Clicking the right mouse button on the icon cycles the view of the block through the user's filters.
+
 
 #### Examples
 
@@ -1317,12 +1316,17 @@ equally.
 [[block]]
 block = "taskwarrior"
 interval = 60
-format = "{count} open tasks"
-format_singular = "{count} open task"
+format = "{count} open tasks ({filter_name})"
+format_singular = "{count} open task ({filter_name})"
 format_everything_done = "nothing to do!"
 warning_threshold = 10
 critical_threshold = 20
-filter = "+work +important"
+[[block.filters]]
+name = "today"
+filter = "+PENDING +OVERDUE or +DUETODAY"
+[[block.filters]]
+name = "some-project"
+filter = "project:some-project +PENDING"
 ```
 
 #### Options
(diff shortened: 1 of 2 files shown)


matt-snider:
@flying7eleven @HPrivakos @jubnzv tagging you guys as contributors to this block.

flying7eleven:
LGTM 👍🏼

jubnzv:
@matt-snider Hi,
I didn't run it since I don't use i3status-rust anymore, but I got the idea. These changes look good to me.

I would like to suggest to test one more thing. Try to create a task with a _really long description_, say, 300 symbols. Will the task counters work correctly? In this case output of `task` command can be wrapped to multiple lines (default wrap is 80 symbols), so you will probably need to check it in the output parser. I ran into a similar problem while working on [my tool](https://github.com/jubnzv/qtask) and solve this passing `rc.defaultwidth=0` to `task` arguments [here](https://github.com/jubnzv/qtask/tree/dfbc2cf592f38a84daeff1b5e7aef374e390820b/src/taskwarrior.cpp#L13). Hope this helps.

matt-snider:
@flying7eleven @jubnzv Thank you both for taking a look.

> Try to create a task with a really long description, say, 300 symbols. Will the task counters work correctly? In this case output of task command can be wrapped to multiple lines (default wrap is 80 symbols), so you will probably need to check it in the output parser.

I don't think this will be a problem because the `count` command is used here, which doesn't return any details of the task. Does that make sense or have I misunderstood your suggestion?

jubnzv:
> I don't think this will be a problem because the count command is used here

Okay, in this case, everything will work fine.

blocks.md
```diff
@@ -1332,7 +1336,8 @@ Key | Values | Required | Default
 `interval` | Update interval, in seconds. | No | `600` (10min)
 `warning_threshold` | The threshold of pending (or started) tasks when the block turns into a warning state. | No | `10`
 `critical_threshold` | The threshold of pending (or started) tasks when the block turns into a critical state. | No | `20`
-`filter_tags` | A list of tags a task has to have before its counted as a pending task. | No | ```<empty>```
+`filter_tags` | Deprecated in favour of `filter`. A list of tags a task has to have before its counted as a pending task. The list of tags will be appended to the base filter `-COMPLETED -DELETED`. | No | ```<empty>```
```
ammgws, line 1339: I'm guessing you meant to say `filters` here?

matt-snider: Good catch - fixed!

ammgws:
Thanks!

matt-snider:
@ammgws I solved the conflicts and now I'm just building and testing it locally

matt-snider:
Okay I tested it and it seems to be working as it did before

ammgws:
Cheers!

PR #1011, opened 2021-01-25
Refactored weather block to use curl crate

simao:
This commit makes the weather block use the curl bindings through the
`curl` crate rather than start a sub process to execute the curl binary.

This has a few advantages:

- Is faster and uses less resources as no shell is started

- We can use simpler code to parse the errors, status codes and output

- The user does not need to depend on the `curl` binary.  The `curl`
crate depends on `curl-sys` which can be configured to link to libcurl statically or dynamically.

Change the `docker` and `github` blocks to use these bindings as well.

Additionally, some improvements to the weather block were implemented:

Allow the user to configure the block with coordinates in addition to `city_id` or `place`.

Implement new format specifier `wind_kmh` with the value of wind speed in km/h. This is a feature forecast.io used to provide, so I believe many users are used to this.

Refactored code to make it easier to understand and read.

Handle different `Err` values and do not crash the entire bar if an error occurs. Show an error to the user or a simple `x` depending on the severity of the error, but keep the other blocks visible to the user.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -221,6 +221,36 @@ dependencies = [
 ]
 
 [[package]]
+name = "curl"
+version = "0.4.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e268162af1a5fe89917ae25ba3b0a77c8da752bdc58e7dbb4f15b91fbd33756e"
+dependencies = [
+ "curl-sys",
+ "libc",
+ "openssl-probe",
+ "openssl-sys",
+ "schannel",
+ "socket2",
+ "winapi",
+]
+
+[[package]]
+name = "curl-sys"
+version = "0.4.39+curl-7.74.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "07a8ce861e7b68a0b394e814d7ee9f1b2750ff8bd10372c6ad3bacc10e86f874"
+dependencies = [
+ "cc",
+ "libc",
+ "libz-sys",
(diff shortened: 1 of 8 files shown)


simao:
I read the contributing guidelines and the part about dependencies, but I think there are good reasons to add this dependency, especially because the user already needs to have the `curl` binary installed.

ammgws:
I had a look the other night and even though the dep count went up by around 10 or so, compile time only increased by 5 seconds on my machine. Since that is the only negative I can think of I am inclined to accept this change.

A couple of requests:
1) Update the other blocks as well.
2) If possible have the weather block improvements in a separate commit.
3) Perhaps change `wind_ca` to be a format specifier instead. That is, `{wind_ca}`.

simao:
Yes that makes sense, I can work on it.

simao:
Hi @ammgws. I implemented those changes, and a few more. I split my changes into different commits, I could open different pull requests if that makes it easier for you.

ammgws:
GitHub block is not working after the refactor. From some quick debugging it looks like it's exiting too early here: https://github.com/greshake/i3status-rust/blob/7a71a37ec40f67c8fb16bb771a8c1c48bc16dcc0/src/blocks/github.rs#L197 which means it's not getting the data it should be. Perhaps the auth header fields aren't being set correctly?

simao:
Thanks for checking that. Looks like github requires us to send a user agent. it should be fixed now.

ammgws:
That did it. Thanks for your work!

PR #1013, opened 2021-01-26
Fixes #1012

### LINKED ISSUE #1012: Pacman Block: Add "on_click" option
Hello i3status-rust developers,
I'm a big fan of the status bar and wondered whether it was possible to add an `on_click` option to some more blocks (especially the `pacman` block).
A cool use-case would be something like this:
```
on_click: i3-sensible-terminal -e "sudo pacman -Syu"
```

to run the update command in a new terminal window when clicking on the status bar icon/text.

Cheers
Jan

edwin0cheng:
I forget to check various blocks to use id for the widget events name in #1006 such that some blocks 'on_click' event do not work.

It should fix #1012.


ammgws:
Cheers

PR #1017, opened 2021-02-04
fix microphone icons in awesome5

MaxVerevkin:
Change microphone icon to [alt](https://fontawesome.com/icons/microphone-alt?style=solid) and [alt-slash](https://fontawesome.com/icons/microphone-alt-slash?style=solid), because normal ones do not work.


ammgws:
Thanks!

PR #1018, opened 2021-02-05
refactoring: fix all clippy warnings

MaxVerevkin:
Just some refactoring. Make clippy happy.

PR #1019, opened 2021-02-05
fix speedtest

MaxVerevkin:
In #1018 I accidentally assumed that `ping` is stored in milliseconds whereas it is in seconds.


ammgws:
No worries

PR #1020, opened 2021-02-05
Add a stockmarket block

hirschenberger:
It displays a stockmarket symbol from the yahoo finance API

This is a first work-in progress, open for discussion.

Todo: 

* offline handling

Ideas are:

* Colors based on pricechange per timeunit
* Simple chart display


ammgws:
I'm not too sold on the idea given that it could probably be trivially implemented with a Custom block, and that it requires a Yahoo specific dependency to be added.

PR #1021, opened 2021-02-05
Fix format_number function

MaxVerevkin:
When I fist read the code I thought that `format_number` should format numbers to [engineering notation](https://en.wikipedia.org/wiki/Engineering_notation), but in fact it produces outputs like "0.43MB" instead of "430KB", even if the minimal suffix is set to "K". For me, it doesn't make much sense. Please correct me if the current behavior is intentional.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -50,7 +50,9 @@ pub fn format_number(raw_value: f64, total_digits: usize, min_suffix: &str, unit
         _ => -4,
     };
 
-    let exp_level = (raw_value.log10().div_euclid(3.) as i32).max(min_exp_level);
+    let exp_level = (raw_value.log10().div_euclid(3.) as i32)
+        .max(min_exp_level)
+        .min(4);
     let value = raw_value / (10f64).powi(exp_level * 3);
 
     let suffix = match exp_level {
--- a/src/util.rs
+++ b/src/util.rs
@@ -503,12 +503,12 @@ mod tests {
 
     #[test]
     fn test_format_number() {
-        //assert_eq!(format_number(1.0, 3, "", "s"), "1.00s");
-        //assert_eq!(format_number(1.007, 3, "", "s"), "1.01s");
-        //assert_eq!(format_number(1.007, 4, "K", "s"), "0.001Ks");
-        //assert_eq!(format_number(1007., 3, "K", "s"), "1.01Ks");
-        //assert_eq!(format_number(107_000., 3, "", "s"), "107Ks");
-        //assert_eq!(format_number(107., 3, "", "s"), "107s");
+        assert_eq!(format_number(1.0, 3, "", "s"), "1.00s");
+        assert_eq!(format_number(1.007, 3, "", "s"), "1.01s");
+        assert_eq!(format_number(1.007, 4, "K", "s"), "0.001Ks");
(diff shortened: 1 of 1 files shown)


ammgws (commented):
For the sake of comparison, what are each of these tests currently returning with the current code?

MaxVerevkin:
@ammgws 
```
[src/util.rs:562] format_number(1.0, 3, "", "s") = "1.00s"
[src/util.rs:563] format_number(1.007, 3, "", "s") = "1.01s"
[src/util.rs:564] format_number(1.007, 4, "K", "s") = "0.001Ks"
[src/util.rs:565] format_number(1007., 3, "K", "s") = "1.01Ks"
[src/util.rs:566] format_number(107_000., 3, "", "s") = "0.11Ms"
[src/util.rs:567] format_number(107., 3, "", "s") = "0.11Ks"
[src/util.rs:568] format_number(0.000_123_123, 3, "", "N") = "0.12mN"
```

ammgws:
@GladOSkar What do you think?

GladOSkar:
When i implemented this i based it off the suggested format from #240, which suggests `0.42K` instead of `420K`. I believe the idea was to always have a decimal point in there such that the entire string always has the same length (given a monospaced font) so the bar doesn't jump around when the value changes.

That said, i don't mind either way. But this could be considered a regression by some.

MaxVerevkin:
One simple solution I see is to count characters, not digits. For example, if I have `total_digits=3` (which should be renamed then), I would get "420K" and "5.3M".

MaxVerevkin:
And maybe make this configurable, because switching between these two behaviors is quite easy:
```
-    let exp_level = (raw_value.log10().div_euclid(3.) as i32)
+    let exp_level = ((raw_value.log10() + 1.).div_euclid(3.) as i32)
```

GladOSkar:
Sounds reasonable, but that decision is not up to me @ammgws

MaxVerevkin:
Well, with non-monospace fonts the bar will change its size either way. @ammgws what do you think?

MaxVerevkin:
I will leave it as it is, maybe just refactor it later.

ammgws:
> it produces outputs like "0.43MB" instead of "430KB", even if the minimal suffix is set to "K"

This does seem like it should be fixed though

MaxVerevkin:
If "430KB" *is* the expected behavior, then if user wants the bar not to change it's size, he may set `speed_min_unit = "M"`.

ammgws:
Yeah that's my thought too.

src/util.rs
```diff
@@ -36,42 +36,26 @@ pub fn escape_pango_text(text: String) -> String {
...
-    //println!("Magnitude: ({}, {})", magnitude_value, magnitude_level);
-
-    let (value, level) = if magnitude_level < min_unit_level {
-        (min_unit_value, min_unit_level)
-    } else {
-        (magnitude_value, magnitude_level)
-    };
+    let exp_level = (raw_value.log10().div_euclid(3.) as i32)
+        .max(min_exp_level)
+        .min(4);
```
MaxVerevkin, line 55: By the way, `clamp()` [is stabilized](https://github.com/rust-lang/rust/pull/77872) in [rust 1.50.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md) which releases soon.

ammgws:
Cheers!

PR #1023, opened 2021-02-06
github: add an option to hide when no notifications are available

MaxVerevkin:
Similar to #982. Add an option to hide the block if no notifications are available. Please comment on this PR so that I can check if it works.


ammgws:
Test

MaxVerevkin:
It works

ammgws:
Seems reasonable enough

PR #1025, opened 2021-02-07
Refactoring: use `block_error()` and handble DBus errors in battery.rs

MaxVerevkin:
Use `block_error()` if possible since it makes code mode readable. Also handle some DBus errors instead of doing `unwrap()`.


ammgws:
Thanks, does read better now

PR #1027, opened 2021-02-07
Hueshift: add support for gammastep

### LINKED ISSUE #1026: Add gammastep support in hueshift block
Since [redshift](https://github.com/jonls/redshift) does not work on Wayland, would it be possible to support [gammastep](https://gitlab.com/chinstrap/gammastep) in `hueshift` block?

MaxVerevkin:
Resolves #1026. Also now the block shows an error if it couldn't find a program to use, because having a block that behaves like it's working whereas it doesn't doesn't feel right.

In addition make the internal structure more like in `keyboard_layout` block.

Note that `gammastep` does not allow to to change it's state smoothly (or I just couldn't find a way to do so), current implementation `killall gammastep`s before it sets a new temperature. This results in flickering.

If I get I right, `gammastep` has to be running in order to work, it cannot just exit like `redshift`. Maybe it's a limitation of wayland?

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -528,10 +528,19 @@ click_temp = 3500
 Key | Values | Required | Default
 ----|--------|----------|--------
 `step`        | The step color temperature is in/decreased in Kelvin. | No | `100`
-`hue_shifter` | Program used to control screen color, either `"redshift"` or `"sct"`. | No | `"redshift"`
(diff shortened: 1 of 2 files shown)


ammgws:
@quantizzed Can you verify the gammastep behaviour?

ammgws:
It seems to be how gammastep works.

Merging for now, thanks!

PR #1028, opened 2021-02-08
keyboard_layout: split layout name and variant in sway driver

MaxVerevkin:
If I'm right, sway's keyboard layout is either `layout (variant)` or `layout` if there is no variant. If so, then it is possible to implement [this TODO](https://github.com/greshake/i3status-rust/blob/aaec57f85e150eaa0b13c5a4238df16b41649e41/src/blocks/keyboard_layout.rs#L365).

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -674,7 +674,7 @@ Key | Values | Required | Default
   Key    | Value
 ---------|-------
 `{layout}` | Keyboard layout name
-`{variant}` | Keyboard variant (only `localebus` is supported so far)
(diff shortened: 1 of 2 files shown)


ammgws:
Yeah this makes sense as sway is just reporting the value from XKB, and XKB defines the string as "Layout (Variant)" so the thing in brackets should always be the variant.

ref: `man xkeyboard-config` and sway codebase

ammgws:
Thanks!

PR #1031, opened 2021-02-08
cpu: fix "{frequency}" format in per-core mode

### LINKED ISSUE #1029: cpu: per-core frequency
At the moment it is not possible to show per-core frequency, even though documentation clearly has it:
> Display CPU frequencies and utilization per core.

This behavior was introduced in #688 as an attempt to solve incorrect *default* formatting. What I can see from this, is that no one actually use this feature if it was broken for almost a year.

Now, is it reasonable to try to fix it or just remove this feature and simplify the code?

MaxVerevkin:
An attempt to solve an issue discussed in #688 by moving the percent symbol inside `{utilization}` and include "GHz" directly in `{frequency}`, because per-core code assumes that anyway. This allows me to revert changes introduced in #688 and thus make the block customizable again.

Closes #1029

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -231,7 +231,7 @@ Update CPU usage every second:
 [[block]]
 block = "cpu"
 interval = 1
-format = "{barchart} {utilization}% {frequency}GHz"
+format = "{barchart} {utilization} {frequency}"
 ```
 
 #### Options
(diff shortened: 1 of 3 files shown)


MaxVerevkin:
Well, apparently there already was a [discussion](https://github.com/greshake/i3status-rust/pull/606) on this topic, but there were no solution found:
> I think we can figure this out later

 because of this reason in particular:
> But as mentioned in the commit message, I think it may surprise some users when they update, if they already added %/GHz in their format string (maybe we can mention it in the changelog though).

> I'm also concerned about breaking existing configurations here.

Any ideas?

> Still thinking about whether there's another way

ammgws:
I think we should be OK with mentioning the change in the changelog since it is easy enough to fix.

PR #1032, opened 2021-02-09
README: fix scrolling default value

MaxVerevkin:
[Source](https://github.com/greshake/i3status-rust/blob/f4ee3e7ac97cead50f17b247feb9be7198bed507/src/config.rs#L115)


ammgws:
Nice catch

PR #1033, opened 2021-02-09
Change IDs to be `usize` instead of `String`

MaxVerevkin:
### Why?
This change may make some performance improvements, since it removes [order vector](https://github.com/greshake/i3status-rust/blob/f4ee3e7ac97cead50f17b247feb9be7198bed507/src/main.rs#L174) and [hash map](https://github.com/greshake/i3status-rust/blob/f4ee3e7ac97cead50f17b247feb9be7198bed507/src/main.rs#L183). Also, it removes code like [this](https://github.com/greshake/i3status-rust/blob/f4ee3e7ac97cead50f17b247feb9be7198bed507/src/blocks/kdeconnect.rs#L115) and thus it decreases heap allocation.

Also, it removes `getrandom` dependency.

### How?
~~`pseudo_uuid` have been repurposed to generate IDs starting from zero, so those IDs can be used to actually index blocks' vector. This has one major issue - each block *must* use `pseudo_uuid` one and *only* once.~~

When the block is created it is now explicitly given an ID. `pseudo_uuid` instead if randomly generating a sting now generates `usize` counting from `usize::MAX`.
### TODOs
- [x] Confirm that all blocks that rely on IDs are working
- - [x] Music
- - [x] Notify
- - [x] Nvidia GPU
- [x] Don't use `pseudo_uuid` to create IDs
- [x] Once preveous is done, use `pseudo_uuid` instead of [this](https://github.com/MaxVerevkin/i3status-rust/blob/5675d2eea8ae8b027c0e546f1a283b6056d7f35a/src/blocks/music.rs#L290)
- [x] Use `usize` instead of `u64` (previously used because of hash function)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -356,17 +356,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "getrandom"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee8025cf36f917e6a52cce185b7c7177689b838b7ec138364e50cc2277a56cf4"
-dependencies = [
- "cfg-if 0.1.10",
- "libc",
- "wasi 0.9.0+wasi-snapshot-preview1",
-]
-
-[[package]]
 name = "gimli"
 version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -407,7 +396,6 @@ dependencies = [
  "crossbeam-channel",
  "curl",
  "dbus",
- "getrandom 0.2.0",
  "inotify",
  "lazy_static",
  "libpulse-binding",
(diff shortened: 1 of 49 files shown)


ammgws:
Seems like a good idea!

MaxVerevkin:
If someone uses notify block or has nvidia GPU, please check if it still works on this branch.

ammgws:
Nvidia block seems to work fine. Partial test using a simulated command with this config:

```toml
[[block]]
block = "nvidia_gpu"
label = "mygpu"
show_memory = true
show_clocks = false
show_utilization = false
show_temperature = false
show_fan_speed = false
interval = 1
```

and overrided nvidia-smi with a script in ~/.local/bin that just outputs: "GTX1080, 4196, <random number>".

Clicking on the "name" widget part of the block switches between the label set in the config, as expected. Clicking on the "memory" widget part of the block switches between the total memory (4196) and the used memory (random number generated by script), so it is working as expected.

I suspect you were worried about the clicks not registering correctly?

MaxVerevkin:
>Nvidia block seems to work fine.

Great!

> I suspect you were worried about the clicks not registering correctly after the ID changes?

Yeah, that's right.

ammgws:
Notify block is working fine too.

ammgws:
Music block is behaving as expected too. That should more or less prove there's no issues.

ammgws:
Thanks for your work!

PR #1039, opened 2021-02-11
remove redundant ConfigLegacy structure

MaxVerevkin:
Manually `impl<'de> Deserialize<'de> for Theme` which gives more flexibility than having two `Config` structs and trying to deserialize one and then another. This solution is similar to what `icons` has.

 Full compatibility is preserved.

Future thoughts: don't copy the entire config to every block. Instead, give each block something like `Rc<Icons>` and `Rc<Theme>` (maybe create struct "appearance"?), and clone the theme only if block has theme overrides. As far as I know, blocks need global config only for theme and icons, please correct me if I'm wrong.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,24 +1,22 @@
 use std::collections::HashMap as Map;
 use std::marker::PhantomData;
 use std::ops::Deref;
-use std::path::Path;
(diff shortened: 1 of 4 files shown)


ammgws:
Thanks!

PR #1041, opened 2021-02-11
memory: fix `clickable`

### LINKED ISSUE #1040: No swap is not critical, but printed with "critical" style
# Environment
* Gentoo Linux, amd64
* i3status-rs 0.14.1

# Summary

I have a machine with no swap enabled, and total swap on this machine is (of course) zero bytes.
i3status-rs considers 0MB/0MB swap usage as 100% used, and tells it is "critical".

screenshot-2021-02-11-193613+0900

However, when no swap is enabled, 0MB free swap is no problem, as it is system admin's intention.

# Expected result

At least no swap is not treated as "critical".

I'm not sure what it actually should look like. `0MB/0MB (0%)` would be one option, `0MB/0MB (--%)` also seems good, or simply printing `N/A` or `no swap` would be possible.

# Actual result

Swap usage is treated as 100% used, and it is printed as "critical".

MaxVerevkin:
Closes #1040

### DIFF
--- a/src/blocks/memory.rs
+++ b/src/blocks/memory.rs
@@ -490,7 +490,7 @@ impl Block for Memory {
     }
 
     fn click(&mut self, event: &I3BarEvent) -> Result<()> {
-        if event.matches_id(self.id) {
+        if event.matches_id(self.id) && self.clickable {
             if let MouseButton::Left = event.button {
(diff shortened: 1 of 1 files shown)


ammgws:
Indeed

Thanks!

PR #1042, opened 2021-02-12
Do not copy the entire config to every widget

MaxVerevkin:
### Binary size
- before: 9.3MB
- after: 9.1MB
### Memory consumption
Result of running `ps -eo sz,vsz,rss | grep i3status-rs`
- before 139889 559556 15628
- after 139727 558908 14956 (700KB less)
### Design
`Appearance` structure has to fields: theme and icons, both inside `Rc`. If block has `color_overrides` (which was moved to `base_block` since every block had it), the theme is cloned.
### Issues
- I had to remove support for legacy options for net block. As the author of #738 mentioned, the legacy support is made to help users transition to a new config, but eight months have passed since then. It was removed because it used a somewhat hacky workaround which requires access to non-deserialized config.
- ~~Unfortunately, for now I had to move `scrolling` option from global config to every block that uses it. I think I'll move it back, but that means I have to rename "Appearance" to something like "SharedConfig" which is annoying (maybe just leave it to be "appearance"?).~~
### TODOs
- [x] Rename `Appearance` to `SharedConfig`
- [x] Move scrolling option back


MaxVerevkin:
Note that this branch requires `rustc 1.50.0` as it uses `clamp()`

ammgws:
> Note that this branch requires rustc 1.50.0 as it uses clamp() and apparently github doesn't support it.

Perhaps best leave that for another PR? Didn't it only get released yesterday :laughing: ?

MaxVerevkin:
> Perhaps best leave that for another PR? Didn't it only get released yesterday :laughing: ?
OK, I guess so

MaxVerevkin:
I also deleted `ButtonWidget` because it is the same thing as `TextWidget`

ammgws:
Yes that had always bugged me too

src/blocks/base_block.rs
```diff
@@ -54,10 +57,12 @@ impl<T: Block> Block for BaseBlock<T> {
 pub(super) struct BaseBlockConfig {
     /// Command to execute when the button is clicked
     pub on_click: Option<String>,
+
+    pub color_overrides: Option<HashMap<String, String>>,
 }
 
 impl BaseBlockConfig {
-    const FIELDS: &'static [&'static str] = &["on_click"];
+    const FIELDS: &'static [&'static str] = &["on_click", "color_overrides"];
```
ammgws, line 65: Let's just change this to `theme_overrides`.

src/config.rs
```diff
@@ -1,30 +1,105 @@
...
+                "info_bg" => theme.info_bg = Some(entry.1.to_string()),
+                "good_fg" => theme.good_fg = Some(entry.1.to_string()),
+                "good_bg" => theme.good_bg = Some(entry.1.to_string()),
+                "warning_fg" => theme.warning_fg = Some(entry.1.to_string()),
+                "warning_bg" => theme.warning_bg = Some(entry.1.to_string()),
+                "critical_fg" => theme.critical_fg = Some(entry.1.to_string()),
+                "critical_bg" => theme.critical_bg = Some(entry.1.to_string()),
+                x => {
+                    return Err(errors::ConfigurationError(
+                        format!("Theme element \"{}\" cannot be overriden", x),
```
ammgws, line 51: and allow everything to be overridden

ammgws (changes requested):

src/blocks/base_block.rs
MaxVerevkin: Yeah, I also thought that would be a better name.

src/config.rs
MaxVerevkin: OK, but that also needs some changes in `print_blocks()` to actually use those overrides.

MaxVerevkin: And right now there is no way to get `shared_config` outside of the block itself. I have to add a new function to block trait `get_shared_config() -> SharedConfig`.

MaxVerevkin: Do we really need to have overriding of separator for each block? Anyway, I think I'll leave it for a separate PR since it's a lot of work.

ammgws: That's fine

ammgws (approved):

ammgws:
LGTM

Thanks for your awesome work!

PR #1049, opened 2021-02-13
memory: Left-pad Mug

BourgondAries:
Most systems nowadays have more than 10 GB of memory. When using memory
in the bar while hovering around 9-10, it shifts the bar during
transitions, causing an annoying visual effect.

This patch fixes this by just left-padding Mug using spaces.


MaxVerevkin:
Memory block already has a lot of formatting options, why not just add a new one with this behavior?

ammgws:
If padding was somehow incorporated in https://github.com/greshake/i3status-rust/pull/750 then perhaps we could accomplish it that way. Since even if this is merged someone will always find another thing they want displayed differently.

ammgws:

PR #1050, opened 2021-02-13
music: Add position/duration to Music (incomplete)

BourgondAries:
These patches allow format arguments {position} and {duration} to the format configuration.

I'm posting these here to get some eyes on the position data, should we poll this? Afaict it's not included in the PropertiesChanged data, so it seems like we need a thread that polls this in regular intervals.

### DIFF
--- a/src/blocks/music.rs
+++ b/src/blocks/music.rs
@@ -36,6 +36,7 @@ struct Player {
     artist: Option<String>,
     title: Option<String>,
     position: i64,
+    duration: i64,
     //TODO
     //volume: u32,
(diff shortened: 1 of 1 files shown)


ammgws:
You shouldn't have to spin up a thread for that, just introduce an update interval (see https://github.com/greshake/i3status-rust/blob/master/src/blocks/template.rs). However if it were to be implemented, I'd prefer it so that the update interval is only activated when the user's format string contains a property that needs to be polled.

MaxVerevkin:

PR #1051, opened 2021-02-13
net: Some fixes to prevent bar shifting

BourgondAries:


src/blocks/net.rs
```diff
@@ -784,7 +784,7 @@ impl Net {
             let value = self.device.relative_signal_strength()?;
             if let Some(ref mut signal_strength_string) = self.signal_strength {
                 if let Some(v) = value {
-                    *signal_strength_string = format!("{}%", v);
+                    *signal_strength_string = format!("{:3}%", v);
```
MaxVerevkin, line 787: I don't think that's necessary. Having 100% signal strength is very rare and having two spaces looks weird.

MaxVerevkin: Also this doesn't look great if user uses participants around signal strength: `( 78%)`

BourgondAries: I have 100% signal strength because my router is close-by. This should be handled instead of shifting everything else.
If desirable it could be made configurable.

ammgws:
Same comment as https://github.com/greshake/i3status-rust/pull/1049#issuecomment-778724569

ammgws:

PR #1056, opened 2021-02-14
use stabilized clamp()

MaxVerevkin:
This requires rust 1.50.0


ammgws:
Might sit on this one for a bit

MaxVerevkin:
By the way, is there a way to request github actions to rerun?

ammgws:
image
You don't have that button near the top right?

MaxVerevkin:
That's what I have

image

ammgws:
Hmm not sure then. Will have to look into it. I'll manually trigger it for you now.

MaxVerevkin:
I guess only collaborators have is but this weird.

MaxVerevkin:
I did some research using `grep` and it turns out `GraphWidget` is not used at all.

Also remove `num-traits` dependency.

MaxVerevkin:
It compiles!

ammgws:
Also replace the clamp_root_scaling function?

MaxVerevkin:
> Also replace the clamp_root_scaling function?

This?
```rust
/// Clamp scale root to a safe range. Useful values are 1.0 to 3.0.
fn clamp_root_scaling(root_scaling: f64) -> f64 {
    if 0.1 < root_scaling && root_scaling < 10.0 {
        root_scaling
    } else {
        1.0
    }
}
```
I'm not sure how and why.

ammgws:
Don't worry about it!

Anyway I'll go ahead and merge this since the next release probably won't be out for a while and we do state that we require recent versions of rustc in the readme.

PR #1059, opened 2021-02-14
keyboard_layout: add {layout_short} format option

MaxVerevkin:
~~Add `{layout_short}` format option which translates "English" to "EN", "Greek" to "GR", "Russian" to "RU" and to on. Saves some space.~~

Add `mappings` option to map `layout (variant)` to user-defined short names.

Config example:
```toml
[[block]]
block = "keyboard_layout"
driver = "sway"
format = "{layout}"
mappings = {"English (Workman)"="EN", "Russian (N/A)"="RU"}
```

TODO:
- [x] Write docs for mappings.
- [ ] Per-layout colors?


ammgws:
Ideally I figure we should conform to the ISO 2 letter codename for each language but I guess that might be too much work? For example Polish is meant to be PL not PO.

Some other things that come to mind:
- Does the XKB layout string get affected by localisation or is it always in English (if the latter then we could implement the suggestion above)?
- If https://github.com/greshake/i3status-rust/pull/750 had an option to truncate strings, would that be a better overall approach?

MaxVerevkin:
>If #750 had an option to truncate strings, would that be a better overall approach?

Do you mean something like `"{layout:.2}"`?

>Does the XKB layout string get affected by localisation or is it always in English

Very good question! I don't know. I suspect it does not, but I'll check that.

>Ideally I figure we should conform to the ISO 2 letter codename for each language but I guess that might be too much work?

Yeah, that would be great, but I couldn't find a way to implement it without a giant hash map.

ammgws:
> Do you mean something like "{layout:.2}"?

Yeah something like that. I imagine that things like this will keep popping up from time to time, so it would be better than having to add a new option etc each time. In this case, however, if we can implement the ISO codenames than a new option is fine.

> I couldn't find a way to implement it without a giant hash map (or giant match construction).

HashMap seems fine to me?

MaxVerevkin:
> HashMap seems fine to me?

Well, yes, if layout is always in English then it will have 248 enties which seems OK, but if layout is affected by a locale then is's 61504 (probably less but still)!

> I imagine that things like this will keep popping up from time to time, so it would be better than having to add a new option etc each time

Yes, I agree.

MaxVerevkin:
>Does the XKB layout string get affected by localisation or is it always in English

I couldn't check that because setting a new locate on void linux with musl libc turns out to be not that easy :eyes:

ammgws:
Ok so I switched my locale to German and the layout names were still in English so I don't think it changes.

```
> swaymsg -t get_inputs | rg layout                                 
    "xkb_layout_names": [
    "xkb_active_layout_index": 0,
    "xkb_active_layout_name": "English (US)",
    "xkb_layout_names": [
    "xkb_active_layout_index": 0,
    "xkb_active_layout_name": "English (US)",
...
```

Crandel:
You could get all names here
```
/usr/share/X11/xkb/symbols
```

MaxVerevkin:
OK, so there are 141 layouts in `man xkeyboard-config`. [file](https://github.com/MaxVerevkin/i3status-rust/blob/layouts/layouts_uniq)

MaxVerevkin:
@Crandel We need full names.

MaxVerevkin:
So, what is more preferable a giant `HashMap` or giant `match`?

MaxVerevkin:
What should be the code for `Russian (Georgia)`?  It's in `/usr/share/X11/xkb/symbols/ge` but *layout name* is Russian for some reason.

MaxVerevkin:
Also it is not clear to me how it will work with drivers other than `sway` and I have no opportunity to test those.

ammgws:
> What should be the code for `Russian (Georgia)`? It's in `/usr/share/X11/xkb/symbols/ge` but _layout name_ is Russian for some reason.
> 
> There are lots of these.

If you're talking about ge(ru), doesn't the name show up as Russian (Georgia)? Seems like that is trivial to match?
ref https://www.mankier.com/7/xkeyboard-config

> Also it is not clear to me how it will work with drivers other than sway and I have no opportunity to test those.
An alternative to this could be something like the IBus block, where the user can specify the string themselves.
It would work with any of the drivers, and way less work to implement than the other option. 
e.g.
```toml
[[block]]
block = "keyboard_layout"
[block.mappings]
"Russian (Georgia)" = "GE"
"German (dead acute)" = "DE"
```

MaxVerevkin:
> If you're talking about ge(ru), doesn't the name show up as Russian (Georgia)? Seems like that is trivial to match?

Sure, but I hoped that it would be possible to match against the first part of the layout name. That would decrease the number of matches by about five times.

> An alternative to this could be something like the IBus block, where the user can specify the string themselves.

Yeah, that seems reasonable. Also, this can be extended to a point where the user can specify a color for each layout.

ammgws:
Ready to go?

MaxVerevkin:
> Ready to go?

Yes, I'll leave the colors for now.

ammgws:
Tested it with sway, cheers!

PR #1061, opened 2021-02-14
music: fix on_collapsed_click

### LINKED ISSUE #1060: music block on_collapsed_click doesn't work
``on_click`` works fine on the music block. I have tested multiple commands  for ``on_collapsed_click`` like ``firefox`` without any success.

My current config for music block:
image

Tested on Ubuntu 20.10 and version 0.14.5 of i3status-rs.
image

MaxVerevkin:
Resolves #1060


ammgws:
Thanks!

PR #1063, opened 2021-02-14
net: add `clickable` option and `format_alt` formating option

MaxVerevkin:
Adds an ability to have two different formats and switch between them by just clicking on the block. For example, there is no need to constantly monitor your ssid, but is is nice to be able to click and see what is it.


ammgws:
This seems reasonable.

Thanks!

PR #1067, opened 2021-02-16
Copy some docs from the wiki to the code

MaxVerevkin:
The wiki is not actively maintained, but is has some good documentation for `Block` and `BlockConfig` traits. Also, if I'm not mistaken, there is no way to contribute to the wiki.

### DIFF
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -108,31 +108,16 @@ impl Into<Update> for Duration {
     }
 }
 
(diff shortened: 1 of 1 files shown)


ammgws:
> Also, if I'm not mistaken, there is no way to contribute to the wiki.

Didn't know that was the case. Doesn't look like I can change the settings either.

PR #1068, opened 2021-02-16
net: refactoring

MaxVerevkin:
The `Net` struct has some fields that are marked as `Option` but they are always `Some`. Remove the option. Also print "N/A" when trying to get ssid or signal strength when wired connection is used instead of throwing an error.

TODO:
- [ ] Test on a machine with wired connection. (I don't have one)

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -373,10 +373,10 @@ pub struct Net {
     ip_addr: Option<String>,
     ipv6_addr: Option<String>,
     bitrate: Option<String>,
-    output_tx: Option<String>,
-    graph_tx: Option<String>,
-    output_rx: Option<String>,
-    graph_rx: Option<String>,
+    output_tx: String,
+    output_rx: String,
+    graph_tx: String,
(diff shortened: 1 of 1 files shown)


ammgws (approved):

ammgws:
> Maybe it's better to just print "N/A" if the connection is not wireless?

Sure, guess it's better than erroring the whole block/bar

src/blocks/net.rs
```diff
@@ -888,10 +852,14 @@ impl Block for Net {
         let mut s_dn = self.shared_config.get_icon("net_down").unwrap_or_default();
         s_dn.push_str(&self.output_rx);
 
+        //let ssid = self.ssid.as_ref().unwrap_or(
```
ammgws, line 855: Forgot to remove this?

MaxVerevkin: Yeah :laughing:

ammgws:
Works fine on my wired machine

ammgws:
Thanks!

Issue #1001, opened 2021-01-08
Uneven spaces around icons


quantizzed:
- **i3status-rust version:** 0.14.3

- **Configuration:**
```
[[block]]
block = "net"
device = "tun0"
format = ""
interval = 1
hide_inactive = true
hide_missing = true

[[block]]
block = "battery"
driver = "upower"
interval = 10
format = " {percentage}% {time}"
```
- **Description:**

I have uneven spaces around the VPN icon (`net` block) and the battery icon (`battery` block).

![2021-01-09_010823](https://user-images.githubusercontent.com/24410141/104070155-4331e180-5217-11eb-94f2-a1e3ac3ccd22.png)

The Wi-Fi icon of another `net` block is shown for reference, as it has even spaces around it.

harryNID:
I too noticed this recently as I use the latest git version. I have almost the same issue except for me it was NetworkManager icon and Battery icon.  It was first introduced in a commit sometime in December. I meant to report it then but I was having unrelated video issues with my computer and thought it might just be me. This is the first report I have noticed about this issue. I wound up having to reload linux a few days ago and noticed it is still there.
Here are some notes,observations and workarounds I discoverd.
For the battery block I discoverd if you use driver = "sysfs". It works. It shows a battery percentage after the icon but spacing is correct. As far as I can tell the spacing issues is in "upower" code possibly. I even tried to manually overide the icon but the spacing issue is still present with driver = "upower".
For me, with NetworkManager I used until recently in my config "device_format = "{icon}" which worked until the mentioned commit. It dropped the left spacing in front of the icon. Since the icon is exposed in string for this block all I had to do was add a manual space to the code like this, "device_format = " {icon}" for it to work.

ammgws:
Can confirm the net block example in the OP with this minimal config:

```toml
[[block]]
block = "net"
format = ""
```

```sh
>cargo run test1001.toml
{"version": 1, "click_events": true}
[[{"background":"#000000","color":"#a9a9a9","full_text":"|","markup":"pango","separator":false,"separator_block_width":0},{"background":"#000000","color":"#93a1a1","full_text":" ETH ","markup":"pango","name":"d8bfdc91354e476abb19053c5447ab89","separator":false,"separator_block_width":0},{"background":"#000000","color":"#93a1a1","full_text":" ","markup":"pango","name":"","separator":false,"separator_block_width":0}],
```

[...]

PR #1071, opened 2021-02-17
sound block: fix spacing for empty format strings

ammgws:
Resolves https://github.com/greshake/i3status-rust/issues/1001#issuecomment-780498034

Test run to see if this can be applied to all blocks using format strings


MaxVerevkin:
@ammgws Maybe do this check inside of `TextWigdet` on every `set_text`? Or maybe introduce `set_text_checked` function? I think that's more ergonomic.

PR #1072, opened 2021-02-17
net block: fix spacing for empty format strings

ammgws:


src/blocks/net.rs
```diff
@@ -598,6 +598,8 @@ impl ConfigBlock for Net {
             use_bits: block_config.use_bits,
             speed_min_unit: block_config.speed_min_unit,
             speed_digits: block_config.speed_digits,
+            // TODO: why are we using a separate widget for just the icon instead of
```
MaxVerevkin, line 601: My thoughts exactly. I plan to remove it.

src/blocks/net.rs
```diff
@@ -412,6 +411,7 @@ pub struct NetConfig {
     /// Which interface in /sys/class/net/ to read from.
     pub device: Option<String>,
 
+    // TODO: remove all deprecated options from the code
```
MaxVerevkin, line 414: Right now they do nothing, right?

PR #1073, opened 2021-02-17
net: remove deprecated options and redundant widget

MaxVerevkin:
Solve TODO from #1072. Those options don't do anything anyway. The "hide" functionality works as before.

Also add documentation for `max_ssid_width` option.

In addition this solves issue with double space.
### Before:
image

### After:
image


ammgws:
Thanks!

PR #1074, opened 2021-02-17
`set_text()`: check whether the content is empty

MaxVerevkin:
Improve #1071 and #1072. Fix issues like #1001 for every block (in theory).

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -787,13 +787,7 @@ impl Block for Net {
             &self.format
         };
 
-        let text = format.render_static_str(&values)?;
-        self.output.set_text(text.clone());
-        // If the format string is empty then this will lead to two spaces after the icon
-        // unless we disable the TextWidget spaces here
-        if text.is_empty() {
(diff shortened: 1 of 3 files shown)


ammgws:
> @ammgws Maybe do this check inside of TextWigdet on every set_text? Or maybe introduce set_text_checked function? I think that would be more ergonomic.

I think it's better to do it with `set_text`, otherwise someone writing a block might get confused as to which one is meant to be used. However, can we check what happens with blocks that have multiple widgets? e.g. networkmanager block

MaxVerevkin:
NetworkManager works fine for me.

ammgws:
Can you squash your commits?

MaxVerevkin:
I hope I didn't break everything with `get rebase`

MaxVerevkin:
I thought there is an option in github to squash commits while merging.

MaxVerevkin:
Ehhh, I did

MaxVerevkin:
Yay, did it.

ammgws:
Thanks!

PR #1076, opened 2021-02-18
Simplify `FormatTemplate`

MaxVerevkin:
Current implementation of formatting is somewhat messy. This is my attempt to rewrite it.

This implementation does not use recursive functions and the data structure is just a vector instead of custom linked list.

Probably needs some testing.

### DIFF
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -57,7 +57,7 @@ pub struct Weather {
     id: usize,
     weather: TextWidget,
     format: String,
-    weather_keys: HashMap<String, String>,
+    weather_keys: HashMap<&'static str, String>,
     service: WeatherService,
     update_interval: Duration,
     autolocate: bool,
@@ -272,7 +272,7 @@ impl Weather {
                 let apparent_temp =
                     australian_apparent_temp(raw_temp, raw_humidity, raw_wind_speed, *units);
 
-                self.weather_keys = map_to_owned!("{weather}" => raw_weather,
+                self.weather_keys = map!("{weather}" => raw_weather,
                                   "{temp}" => format!("{:.0}", raw_temp),
                                   "{humidity}" => format!("{:.0}", raw_humidity),
                                   "{apparent}" => format!("{:.0}",apparent_temp),
@@ -341,7 +341,8 @@ impl Block for Weather {
         match self.update_weather() {
             Ok(_) => {
                 let fmt = FormatTemplate::from_string(&self.format)?;
-                self.weather.set_text(fmt.render(&self.weather_keys));
+                self.weather
+                    .set_text(fmt.render_static_str(&self.weather_keys)?);
                 self.weather.set_state(State::Idle)
             }
             Err(BlockError(block, _)) | Err(InternalError(block, _, _)) if block == "curl" => {
--- a/src/util.rs
+++ b/src/util.rs
@@ -387,100 +387,51 @@ where
 }
 
 #[derive(Debug, Clone)]
-pub enum FormatTemplate {
-    Str(String, Option<Box<FormatTemplate>>),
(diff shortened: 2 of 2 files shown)


MaxVerevkin:
I'm also thinking of ways to implement "lazy formatting" - the value is computed only if it is present in the format string. Net block would benefint from this feature a lot. Actually, net block already does this in a hachy way: is simply checks if some options are presented using `.contains()`.

Unfortunately, [`std::lazy`](https://doc.rust-lang.org/std/lazy/index.html) is nightly olny.

I tried writing something like:
```rust
pub struct LazyString {
    inner: Option<Box<dyn FnOnce() -> String>>,
    value: Option<String>,
}
impl LazyString {
    pub fn new(x: Box<dyn FnOnce() -> String>) -> Self {
        Self {
            inner: Some(Box::new(x)),
            value: None,
        }
    }
    pub fn init(&mut self) {
        if let Some(inner) = self.inner.take() {
            self.value = Some(inner());
        }
    }
    pub fn val(&self) -> Option<String> {
        self.value.clone()
    }
}

...

    pub fn render_lazy(&self, vars: &mut HashMap<&str, LazyString>) -> Result<String> {
        let mut rendered = String::new();

        for token in &self.tokens {
            match token {
                FormatToken::Text(text) => rendered.push_str(&text),
                FormatToken::Var(ref key) => {
                    let s = vars.get_mut(&**key).unwrap();
                    s.init();
                    rendered.push_str(&format!("{}", s.val().unwrap()));
                }
            }
        }

        Ok(rendered)
    }

```

And it works for simple things like
```rust
        let layout = move || layout;
        let variant = move || variant;

        let mut values = map!(
            "{layout}" => crate::util::LazyString::new(Box::new(layout)),
            "{variant}" => crate::util::LazyString::new(Box::new(variant))
        );

        self.output.set_text(self.format.render_lazy(&mut values)?);

```

But I cannot do stuff like
```rust
        let layout = || self.layout();
        let variant = move || variant;

        let mut values = map!(
            "{layout}" => crate::util::LazyString::new(Box::new(layout)),
            "{variant}" => crate::util::LazyString::new(Box::new(variant))
        );

        self.output.set_text(self.format.render_lazy(&mut values)?);
```

Can someone with more rust experience give me some hints please? Compiler tells me that in this case `self` has to have `static` lifetime.

MaxVerevkin:
Although I don't quite know how exactly integrate it into net block because, for example, it updates ssid only once every 30 seconds.

ammgws:
Can you include the full keyboard_layout example? Might as well commit it.

MaxVerevkin:
@ammgws Done.

ammgws:
Okay, perhaps I'm missing something but couldn't you handle the net block issue you mentioned above by moving the below logic inside the function that returns the SSID?

```rs
let now = Instant::now();
if (now.duration_since(self.last_update).as_secs() > 30)
...
```

MaxVerevkin:
Hmm, well, yes. Why haven't I thought about that :laughing:?

MaxVerevkin:
Ohhh, `self` has to be mutable, that's why I cannot move it there.

> `self.last_update = now;`

MaxVerevkin:
I think I'll wait better wait for [`std::lazy`](https://doc.rust-lang.org/std/lazy/index.html) to be stabilized.

MaxVerevkin:
By the way, I haven't experienced any issues with this implementation yet.

ammgws:
> By the way, I haven't experienced any issues with this implementation yet.

Neither, though just with my personal config. I think it's time for this to get more action via master.

ammgws:
Thanks!

MaxVerevkin:
TODO: Fix the issue with trailing text in format string.

PR #1077, opened 2021-02-18
Fix version in Cargo.toml

crozbo:


ammgws:
Need Cargo.lock as well

ammgws:
Actually can you bump it to 0.15.0

ammgws:
Thanks!

PR #1079, opened 2021-02-19
net: remove explicit `clickable` option

### LINKED ISSUE #1078: net block: remove `clickable` option
```md
`clickable` | Whether to allow clicking on the block and swithching between `format` and `format_alt` | No | `false`
`format_alt` | The same as `format`. Shows when the block is toggled with a mouse click (if `clickable` is `true`). | No | `"{ssid}"`
```

Instead of `clickable`, just see if `format_alt` was specified in the config. If it was, then the user obviously wants the block to be clickable. If there's really some reason for disabling click events then it can be done once https://github.com/greshake/i3status-rust/issues/1038 is implemented by setting the on-click action to empty and overriding the block on-click handler.

MaxVerevkin:
Closes #1078


ammgws:
Thanks :)

PR #1080, opened 2021-02-19
Remove graph widget

MaxVerevkin:
I thought I should move this out of #1056 to a new PR because who knows when everyone will update to new a rustc.

About the change: `GraphWidget` is not used by any of the blocks. Every blocks uses `util::format_vec_to_bar_graph` instead if needed.


ammgws:
Yeah this became redundant after the blocks using it were switched to format strings.

Can you fix the conflicts?

MaxVerevkin:
Sure, done

PR #1084, opened 2021-02-20
fix clippy warnings; show profiling options only if profiling is enabled

MaxVerevkin:
Currently if built in debug mode, even if profiling is disabled, `i3status-rs -h` gives this:
```
--profile <profile>
--profile-runs <profile-runs>
```
which is not correct.

Also fix some clippy warnings and remove
```rust
#[macro_use]
extern crate serde_json;
```
because it isn't necessary since edition 2018


ammgws:
Thanks!

PR #1085, opened 2021-02-21
Introduce I3BlockData structure, do not call `click()` on every block

MaxVerevkin:
### Changes
- Introduce `I3BlockData` structure that represents every field described by the protocol here: https://i3wm.org/docs/i3bar-protocol.html
- Use (name, instance) pair to index every widget. Each block has a unique name/id, and inside of that block, each widget has an "instance". Now `click()` is called on the block that was clicked only.

### DIFF
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -165,13 +165,13 @@ pub trait Block {
         Ok(())
     }
 
-    /// Sends click events to the block. This function is called on every block for every click.
+    /// Sends click events to the block.
     ///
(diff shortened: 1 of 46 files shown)


ammgws:
Wow this is some nice work! On mobile now but this LGTM.

MaxVerevkin:
Interesting, if I run `i3status-rs --one-shot | wc -c` on master, I get `2906`, but on this branch with custom json serialization I get `2738`.

ammgws:
Has been fine for me so far.

Thanks! Great stuff.

PR #1087, opened 2021-02-22
I3BlockData: allow `min_width` in pixels; fix `align` rendering

MaxVerevkin:
A small bugfix for #1085:
- fix rendering of "align" option
- allow "min_width" option to be in pixels

Found those while trying to fix some spacing-related issues by moving each icon into it's own widget. Is this a reasonable idea and have someone tried this before? In some fonts each icon occupy two characters and overlaps with the next character. In some fonts this doesn't happen. If each icon had it's own widget, user would be able to set the size of the icon and it's align.

### DIFF
--- a/src/widgets/i3block_data.rs
+++ b/src/widgets/i3block_data.rs
@@ -11,7 +11,7 @@ pub struct I3BlockData {
     pub border_right: Option<usize>,
     pub border_bottom: Option<usize>,
     pub border_left: Option<usize>,
-    pub min_width: Option<String>,
+    pub min_width: Option<I3BlockMinWidth>,
     pub align: Option<I3BlockAlign>,
     pub name: Option<String>,
     pub instance: Option<String>,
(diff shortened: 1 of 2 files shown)


ammgws:
> moving each icon into it's own widget. Is this a reasonable idea and have someone tried this before?

I don't believe so.

> If each icon had it's own widget, user would be able to set the size of the icon and it's align.

How do you imagine it looking in the config?

MaxVerevkin:
> How do you imagine it looking in the config?

```toml
icon_width = 40
icon_align = "center"
```

ammgws:
I see. To be honest I haven't yet experienced spacing issues with the leading icon itself. I feel like it would be simpler resovled via `format` strings and adding the spaces there yourself when necessary. Anyway feel free to open up a new issue if it's something you feel strongly about so we can get discussion from others too.

MaxVerevkin:
> adding the spaces there yourself when necessary

If I get it right, this would mean that if I click between two icons the click will be ignored.

PR #1090, opened 2021-02-23
Remove trailing comma

### LINKED ISSUE #1089: Latest `master` produces JSON error
`rustc` says:
```
rustc 1.52.0-nightly (a15f484b9 2021-02-22)
```

`i3bar` says:
image

`i3status-rs` says:
```
% i3status-rs ~/.config/i3/status-rs-config.toml 
[[{"full_text":"  Tue 2021-02-23 12:13 ","name":"0","instance":"0","markup":"pango",}],
```

Note the trailing comma in the object. That's not allowed in JSON.

MaxVerevkin:
Closes #1089

### DIFF
--- a/src/widgets/i3block_data.rs
+++ b/src/widgets/i3block_data.rs
@@ -66,6 +66,7 @@ impl I3BlockData {
         json_add_val!(retval, self.separator_block_width, separator_block_width);
         json_add_str!(retval, self.markup, markup);
 
+        retval.pop();
         retval.push('}');
         retval
     }


ammgws:
Thanks!

GladOSkar:
Actually i think this might have removed the last character of some blocks? My `memory`, `volume` and `battery` blocks are now missing the last character. The `%` sign in the former 2 and the `)` in the last one.

MaxVerevkin:
Yeah, I see. But I think that was introduced in #1076. The issue is that in format string like `some_text {variable} text`, ` text` is dropped. This shouldn't be hard to fix.

GladOSkar:
Yeah i just saw the same thing, hence the deleted comment, sorry. I'll open another issue

PR #1094, opened 2021-02-23
rotatingtext widget: fix `min_width` calculation

MaxVerevkin:
With this fix it is possible to override music icon to something like that:
```toml
music = " <span size='medium' font_family='NotoSans Nerd Font'>\uf886</span> "
```
Currently, if this override is applied, music widget will get very wide.


ammgws:
Thanks!

Issue #1086, opened 2021-02-22
material icon set renders invalid icons


kgilmer:
## Environment:

Ubuntu 20.04

i3-gaps
```
$ i3 --version
i3 version 4.19 (2020-11-15) © 2009 Michael Stapelberg and contributors
```

i3status-rs
```
$ ~/bin/i3status-rs --version
i3status-rs 0.15.0 (commit 4318a37 2021-02-21)
```

Material font installed:
```
$ apt list --installed | grep material
fonts-materialdesignicons-webfont/focal,now 1.6.50-3regolith3 amd64 [installed]
```

i3status-rs config:
```
theme = "native"
icons = "material"

[[block]]
block = "time"
interval = 60
format = "%a %d/%m %R"
```

## Steps to reproduce:
1. build i3status-rs from source, install as directed on wiki.

Expected behavior: material icons show on bar
Actual behavior: 
![image](https://user-images.githubusercontent.com/49683/108644673-e0ae4100-7464-11eb-9a6e-da1d2dbcf288.png)

FWIW the FontAwesome icon setting does not produce the "unknown unicode" charaters but also doesn't seem to map to the intended glyphs.

## Other notes

```bash
$ fc-list | grep -i material
/usr/share/fonts/truetype/materialdesignicons-webfont/materialdesignicons-webfont.ttf: Material Design Icons:style=Regular
```

FWIW I believe that the mappings from character code to icon have changed in the material design icon font in version 1.6 of that package ([debian](https://packages.debian.org/sid/fonts/fonts-materialdesignicons-webfont)).

I checked [this issue](https://github.com/greshake/i3status-rust/issues/744) but nothing there resulted in working behavior on my system

GladOSkar:
I don't think your `materialdesignicons` font is FontAwesome/NerdFont compatible? Have you tried a FontAwesome or NerdFont font?

MaxVerevkin:
I tried both FontAwesome and NerdFont but I get the same result as the author. Also I checked `src/icons.rs` and some icons are just missing.

GladOSkar:
Interesting. I'm using the `FuraCode Nerd Font` and both `awesome` and `material` icon sets work fine with the clock icon:

![image](https://user-images.githubusercontent.com/9497832/108739868-597cce00-7535-11eb-9cb0-f6d0b4e2a71a.png)

But you're right that some of the networking icons are missing then, but that's another issue.

MaxVerevkin:
With this font config:
```
[...]

Issue #981, opened 2020-12-15
Add ability to define global pango markup wrappings


kneitinger:
Hello all, great project!  I'm in the process of porting my status from i3pystatus and having a wonderful time, but hit a small formatting roadblock that I wanted to get input on.

### Problem Statement
I use an italic monospace font for i3, and due to that am having some issues with icon rendering in i3status-rs.  All of my Font Awesome glyphs (both 4 and 5) are being italicized since my i3 bar config contains

![2020-12-15_14-29-48](https://user-images.githubusercontent.com/5200550/102282408-6a7de300-3ee5-11eb-9f5b-e9c0bdd89e94.png)

```
font pango:OperatorMonoMedium Italic 10
```

As far as I can tell, while i3 does support fallback fonts, it does not have the ability to specify individual styles or sizes for them versus the intended font.

It appears as though I could possibly use `fontconfig` to tell the renderer to never apply italics to `Font Awesome 5 Free`, but that is very cumbersome and I cannot imagine I'm the only one experiencing this issue.

One way around this would be to use pango markup to in my TOML, such as:
```
[icons.overrides]
bat_charging = "<span style=\"normal\">\uICON</span>"
```
if I want to keep my font the same, or

```
[[block]]
block = "net"
device = "wlp3s0"
interval = 5
format = "<span style=\"italic\">{ssid}[{ip}]</span>"
on_click = "nm-connection-editor"
```
as well as my i3 workspace labels, if I switch my font to normal and italicize everything I want.

In either of these options, there is a ton of manual override, which isn't a pleasant user experience.

### Proposed Solution

I was thinking that instead of having to manually adjust every icon, or every format string, there could be a top-level TOML setting that specifies the format that is applied to all icons and/or all text, such as:

```
theme = "native"
icons = "awesome5"
icon_fmt = "<span style=\"normal\">{}</span>"
text_fmt = "<b>{}</b>"

[[block]]
...
```
where every icon or text string substitutes the `{}` in their respective formats.

With something like this, I can largely rely on the wonderful defaults of the themes and blocks, but apply my own styling to a type of element, as opposed individual icons or block texts.

What are your thoughts?  If this sounds like a good idea to folks, I'm happy to implement it...I just wanted to make sure that I'm not missing a super obvious solution, and that this is something people would want!

ammgws:
[...]

PR #1095, opened 2021-02-24
add "material-nf" icon set, add `icons_format` config option

MaxVerevkin:
With `icons_format`
```toml
icons = "material-nf"
icons_format = " <span size='large' stretch='ultraexpanded' font_family='NotoSans Nerd Font'>{icon}</span> "
```
image

Without `icons_format`
```toml
icons = "material-nf"
```
image
Font used
```
    font pango:JetBrainsMono Nerd Font 10.6
```
### Related issues
- #981
- #1086
### TODO
~~Do not apply `icons_format` for overridden icons? (Not possible with the current implementation of icons deserialization.)~~
- [x] Rename `icons_style` to `icons_format`
- [x] Write documentation
- [x] Make use of more moon icons (for backlight block).
- [ ] Deprecate old `material` icon set?


src/icons.rs
```diff
@@ -5,280 +5,356 @@ use lazy_static::lazy_static;
...
+        "bat_quarter" => "\u{f57a}",
+        "bat_three_quarters" => "\u{f57f}",
+        "bell" => "\u{f599}",
+        "bell-slash" => "\u{f59a}",
+        "bluetooth" => "\u{f5ae}",
+        "calendar" => "\u{f5ec}",
+        "cogs" => "\u{f992}",
+        "cpu" => "\u{f9c4}",
+        "disk_drive" => "\u{f7c9}",
+        "docker" => "\u{f308}", // nf-linux-docker
```
ammgws, line 306: Can you add these comments for the other icons as well? It makes it much easier to figure out the intention.

Can get them from here: https://www.nerdfonts.com/cheat-sheet

src/icons.rs
```diff
@@ -5,280 +5,356 @@ use lazy_static::lazy_static;
...
+        "update" => "\u{e8d7}",
+        "uptime" => "\u{e192}", // Same as time symbol.
+        "volume_empty" => "\u{e04e}",
+        "volume_full" => "\u{e050}",
+        "volume_half" => "\u{e04d}",
+        "volume_muted" => "\u{e04e} \u{e04f}",
+        "xrandr" => "\u{e31e}"
+    };
+
+    // Material an in NerdFont
```
ammgws, line 284: "an in"?

ammgws, line 284: Can you add a link to https://www.nerdfonts.com/cheat-sheet here?

ammgws (changes requested):

src/icons.rs
MaxVerevkin: Sure

MaxVerevkin: I can't even remember what I was trying to write :laughing:

src/icons.rs
MaxVerevkin: Will take some time.

ammgws:
Are the Nerd Fonts mappings different than for the official Google material font? I imagine they would be the same if Nerd Fonts is just an aggregator?

MaxVerevkin:
I checked google's [codes](https://github.com/google/material-design-icons/blob/master/font/MaterialIcons-Regular.codepoints) and no, they are not the same. That's weird.

ammgws:
Should icons_format be able to be overridden on a per-block basis like with themes?

MaxVerevkin:
> Should icons_format be able to be overridden on a per-block basis like with themes?

I'm not sure. I think that disabling `icons_format` for icons that were overridden is enough, but needs some more work. Actually, allowing each block to override `icons_format` it simpler. But current implementation of icons' deserialization is messy and should be refactored anyway. I mean, it uses one specific custom macro that is used just once.

ammgws:
> Actually, allowing each block to override icons_format it simpler.

Yeah it might be the way to go. Now too sure but might be useful if for example you wanted to specify some different icon font for the weather block? If it's trivial to implement then I don't really see any downside to allowing users to override it.

> . But current implementation of icons' deserialization is messy and should be refactored anyway. I mean, it uses one specific custom macro that is used just once.

Sure

MaxVerevkin:
> Yeah it might be the way to go.

Done.

MaxVerevkin:
Github actions seem to be down.

src/icons.rs
```diff
@@ -287,9 +317,19 @@ lazy_static! {
         "" => "",
         "backlight_empty" => "\u{e38d}", // nf-weather-moon_new
         "backlight_full" => "\u{e39b}", // nf-weather-moon_full
-        "backlight_partial1" => "\u{e3d1}", // nf-weather-moon_alt_waxing_gibbous_3
-        "backlight_partial2" => "\u{e3ce}", // nf-weather-moon_alt_first_quarter
-        "backlight_partial3" => "\u{e3cb}", // nf-weather-moon_alt_waxing_crescent_4
+        "backlight_1" => "\u{e3d4}", // nf-weather-moon_alt_waxing_gibbous_6
```
ammgws, line 320: Update this list too: https://github.com/greshake/i3status-rust/blob/master/themes.md#available-icon-overrides

ammgws:
Looks OK to me. As a last step I want to check the differences between the patched material font, the official Google one and the Nerd Fonts one.

Icon | Patched | Google | NerdFonts
--- | --- | --- | --- | 
bat_charging | `\u{e1a3}` | `??` | `\u{f57d}`
etc

MaxVerevkin:
Actually, NerdFonts' README claims that they are using this material icons, not Google's - https://github.com/Templarian/MaterialDesign

Google's font has only one charging icon - `battery_charging_full e1a3`

I haven't fond any icons that would "work" in any two of those fonts yet.

ammgws (approved):

ammgws:
Google Material Design icons
https://github.com/google/material-design-icons
https://aur.archlinux.org/packages/ttf-material-design-icons/

Material Design icons by MaterialDesignIcons.com
https://github.com/Templarian/MaterialDesign
https://aur.archlinux.org/packages/ttf-material-design-icons-extended/

Nerd Fonts which includes MaterialDesignIcons.com, possibly patched but codepoints the same?
https://github.com/ryanoasis/nerd-fonts/blob/master/readme.md
https://aur.archlinux.org/packages/nerd-fonts-complete/

"Patched" icons from that gist
https://gist.github.com/draoncc/3c20d8d4262892ccd2e227eefeafa8ef/raw/3e6e12c213fba1ec28aaa26430c3606874754c30/MaterialIcons-Regular-for-inline.ttf

MaxVerevkin:
>  possibly patched but codepoints the same?

From what I can tell, no, codepoints are different.

ammgws:
Looks like it. Let's deal with it separately.

Thanks for this!

PR #1099, opened 2021-02-25
networkmanager: Avoid duplicate device with VPN connections.

### LINKED ISSUE #1016: networkmanager: When connected to VPN, wifi shows twice
Hey everyone!

Just a slight issue I've experienced with the `networkmanager` block.
When connected to my wifi it looks like this:
image

However when I connect to a VPN, it shows the wifi status twice:
image

I would rather think 1x wifi + 1x vpn should be sufficient, but maybe it's working as intended? 

Thank you for your time!

ravomavain:
When a VPN is enabled there's two connections created:
- One with the property "VPN = True" and "Devices" set to the upstream interface (for example the wireless connection over which the vpn is created)
- The other one with "VPN = False" and "Devices" set to the actual vpn device (tun0, vpn0, or whatever)

Since the first one only return a device which is already listed on another connection (the actual wireless connection for exemple), we can ignore it.

This patch ignore all connections with the property "VPN = True"

Fix #1016

### DIFF
--- a/src/blocks/networkmanager.rs
+++ b/src/blocks/networkmanager.rs
@@ -262,6 +262,21 @@ impl<'a> NmConnection<'a> {
         Ok(ActiveConnectionState::from(state.0))
     }
 
+    fn vpn(&self, c: &Connection) -> Result<bool> {
+        let m = ConnectionManager::get(
+            c,
+            self.path.clone(),
+            "org.freedesktop.NetworkManager.Connection.Active",
+            "Vpn",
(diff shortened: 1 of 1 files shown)


ammgws:
Awesome!

@xkonni Can you confirm?

xkonni:
yes, that helps! thank you very much!

ammgws:
Thanks! I'm glad you were able to spot the vpn entry in the dbus properties.

PR #1102, opened 2021-02-26
Fixed wpa_cli BSSID being displayed instead of SSID, see #1100

### LINKED ISSUE #1100: net block with wpa_cli displays BSSID instead of SSID
I've only got `wpa_cli` installed which according to the wiki is supported, however it seems that the displayed `ssid` is indeed the first byte of the `bssid` (MAC address).

Expected output: `MySsid`
Current output: `ssid=ac`

It seems like the block is matching on the substring `ssid` and because `wpa_cli status` prints the `bssid` before `ssid` that is what gets displayed. Furthermore the string `ssid=` also shouldn't be in the block output I believe.

knightshrub:
The net block now correctly displays the SSID instead of the BSSID when `wpa_cli` is used and fixes #1100


ammgws:
LGTM!

PR #1103, opened 2021-02-26
Make `maildir` an optional dependency

### LINKED ISSUE #1101: Make `maildir` an optional dependency
`maildir` is used in maildir block only. Why not make it an optional dependency like `pulseaudio` and `notmuch`?

MaxVerevkin:
Closes #1101


ammgws:
Thanks!

PR #1107, opened 2021-02-28
sound: Don't 0-pad volume

rst0git:
The current string format for volume prepends an additional '0' to single digit values. For example, when the volume is set to 0 it would display `00%`.

This patch replaces the zero-padding format with right-alignment, as shown in the example below.

Before:
```
    00%
    10%
   100%
```

After:
```
    0%
   10%
  100%
```

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -843,7 +843,7 @@ impl Sound {
         } else {
             output_name
         };
-        let values = map!("{volume}" => format!("{:02}", volume),
+        let values = map!("{volume}" => format!("{:>3}", volume),
                           "{output_name}" => mapped_output_name
         );
         let text = self.format.render_static_str(&values)?;


ammgws:
One thing that's been mentioned before is that this may make the block change in size depending on the font used, and then those users will make a claim...

I'd rather we dealt with this using something like #

MaxVerevkin:
@ammgws To prevent this we can set something like `min_width=" <icon> 00% "`

ammgws:
Doesn't i3/swaybar end up padding the bar based on the calculated pixel size of that text? So it would still change size if the text used for min_width was smaller than whatever ends up in there.

(If that makes sense. It's been a while since I looked at the i3/swaybar code)

MaxVerevkin:
> Doesn't i3/swaybar end up padding the bar based on the calculated pixel size of that text?

Yes, you are right.

> So it would still change size if the text used for min_width was smaller than whatever ends up in there.

Yes, but `"<icon>00 "` Is the maximum width of the block, and that's why block will not change it's size. That's how music block works.

GladOSkar:
Yeah. Most users use monospaced fonts anyway. And yes i think i remember this was introduced to prevent block resizing. @MaxVerevkin what would you do with a given `min_width`? Pad with spaces on the left or right of the block? That might look weird as well. I think i actually like the zero-padding better

ammgws:

Issue #829, opened 2020-09-20
disk block: compute state broken since 8074b0bbade191c0ab31158e666efed0fa392f04


ObiWahn:
Hi,

the calculation when the alert triggers changed in commit `8074b0bbade191c0ab31158e666efed0fa392f04`. It used to take a percentage. That was imo a much better behavior than the total that is taken now. And the size of the total does not seem to be taken in GB as the documentations suggest, but in the given unit of the block.

I was expecting an alert at 10/20 % of my 6TB hard drives. With the update all disks turned red and I am expected to have 20TB disk space left to get no alert.

Maybe you could reconsider that part of the change and restore the old behavior. I am myself a rust newbie and would try to have a go if you agree to revert that part of the change and provide me with reviews.

ammgws:
That does seem wrong. @jamesmcm do you get this too?

jamesmcm:
Could you please post your config @ObiWahn ?

Ideally we'd be able to use both absolute values and percentages, the tricky part is that if you set `unit = "Percent"` then _all_ the values are set to use percentages which is probably not what you want. That would fix the percentage alert though (but `{used}` etc. would show bytes).

I see two options, either assume to show in GB if using `unit = "Percent"` so at least the bytes issue is fixed, or add another field in the config for that.

ObiWahn:
My blocks look like this:
```
[[block]]
block = "disk_space"
path = "/mnt/system/backup"
alias = "backup"
info_type = "available"
unit = "TB"
interval = 20
warning = 20.0
alert = 10.0
```

`df` output:
```
/dev/sdc2                2,9T  1,3T  1,5T  46% /mnt/system/backup
```

jamesmcm:
Okay, I'll add an absolute_alerts boolean to set absolute values for the alerts and use percentages otherwise (regardless of the unit setting) for backwards compatibility.

ObiWahn:
@jamesmcm awesome! I was just about to suggest either that or appending `%` to trigger the percentage mode.

jamesmcm:
See #831 

Also my config is:

```toml
[[block]]
block = "disk_space"
path = "/"
alias = "/"
info_type = "available"
unit = "GiB"
interval = 20
warning = 20.0
alert = 10.0
format = "{icon}{used}/{total} {unit} ({available}{unit} / {percentage} free)"
```

ObiWahn:
Works as expected! Thank you!

PR #578, opened 2020-03-05
Add support for progress bars to some blocks

arlopurcell:
Implementation for https://github.com/greshake/i3status-rust/issues/577

PR #1112, opened 2021-03-03
Implement new formatting mechanism

### LINKED ISSUE #1108: discussion: add formatting parameters for format key specifiers
Issues regarding spacing come up often - most recently ones like these: #1107, #1051, #1049 regarding number padding.

For these "padding" issues, I propose we expand on the initial work done in #750 by also adding syntax for setting the padding/number format for format specifiers.

Memory block example where the ":04" would zero-pad the number.
```toml
[[block]]
block = "memory"
format_swap = "{Mug:02}"
```

Sound block example where the ": >2" would pad the number with spaces.
```toml
[[block]]
block = "sound"
format = "{volume: >2}"
```

MaxVerevkin:
An attempt to close #1108. Any thoughts on this PR are welcome.

## Overview
The syntax for the formatting string placeholders is (the order doesn't matter, but the name must always be first)
```
​{<name>[:[0]<min width>][^<max width>][;<min prefix>][*<unit>][#<bar max value>]}
```
### Examples of configs:
- `format = "{signal_strength:2} {graph_down}{speed_down;K}{speed_up;1*Bi/s}"`
- `format = "{volume:03}"`
- `format = "{volume:5#110} {volume:03}"` - [demo](https://imgur.com/a/CCNw04e)
### Examples of code (blocks' POV):
Temperature:
```rust
            let values = map!(
                "average" => Value::from_integer(avg).degrees(),
                "min" => Value::from_integer(min).degrees(),
                "max" => Value::from_integer(max).degrees()
            );
```
Speed test:
```rust
                let values = map!(
                    "ping" => Value::from_float(ping).seconds().icon(self.ping_icon.clone()),
                    "speed_down" => Value::from_float(down).bits_per_second().icon(self.down_icon.clone()),
                    "speed_up" => Value::from_float(up).bits_per_second().icon(self.up_icon.clone()),
                );
```
Sound:
```rust
        let values = map!(
            "volume" => Value::from_integer(volume as i64).percents(),
            "output_name" => Value::from_string(mapped_output_name)
        );
```
## TODO
- [x] Add more suffixes.
- [x] Add suffixes for bytes/bits (Ki, Mi, Gi, Ti)
- [x] Translate every block to a new formatting and remove `render_static_str()`
- [x] disk_space: alert_absolute
- [x] ~~Allow users to override placeholders' icon with `{var~icon}`~~ will not implement.
- [x] Add ability to draw bars for every numeric placeholder.
- [x] Add `max_width` for strings. Something like `"{var.5}"` will truncate the text to first 5 characters.
- [x] Add proper error handling.
- [x] Write genral documentation.
- [x] Document changes in every block
- [x] Add unit tests
- [x] ~~Extend `Value` to have colors (via pango)~~ (not useful because `i3` renders `pango` differently compared to `sway`)
- [ ] (A lot of) Testing.


ammgws:
I don't understand the use case for var~icon. Any example?

MaxVerevkin:
> I don't understand the use case for var~icon. Any examples? 

Net block displays speed as `<icon><speed><suffix><unit>`. "Var~icon" gives ability to override the icon. But yeah, this is not very useful, and most likely I'll leave it.

MaxVerevkin:
Also, an error will be shown if any icon was not found. I thing it's better than `unwrap_or_default()`ing and makes more sense.

MaxVerevkin:
Maybe instead of having `per_core` option in cpu block it is better to have multiple `utilization` and `frequency` placeholders? For example: `format = "Average: {utilization} {frequency} 1: {utilization1} {frequency1} ..."`. This gives much more flexibility. Moreover, this is the only way I see how to implement new formatting for cpu block.

ammgws:
> Maybe instead of having `per_core` option in cpu block it is better to have multiple `utilization` and `frequency` placeholders? For example: `format = "Average: {utilization} {frequency} 1: {utilization1} {frequency1} ..."`. This gives much more flexibility. Moreover, this is the only way I see how to implement new formatting for cpu block.

Sounds OK to me.

MaxVerevkin:
> Add ability to draw bars for every numeric placeholder.

Would something like this be OK: `"{var:5~110}"`? This will create a bar 5 characters long with minimum value of zero and maximum of 110. And I'm not sure if the minimum value should be configurable.

ammgws:
So it would pad the string? I don't get why the minimum is 0 in that example.

Is this what you mean (replaced spaces with □ to make it easier to tell):

`format = "{var:5~110}"`
var | output
--- | ---
"" | "□□□□□" <-- pads to the left?
"x" | "□□□□□x" <-- pads to the left?
"x□" | "□□□□□x□" <-- pads to the left?
"123456" | "12345" <-- trims from the right?

ammgws:
Oh sorry I thought you were talking about strings because of your comment on the other PR. Didn't realise you were talking about the barchart.

MaxVerevkin:
@ammgws No, I meant this function https://github.com/greshake/i3status-rust/blob/094cfdc13953b31edda8914f5e536e52b93deb38/src/util.rs#L348-L355

MaxVerevkin:
This feature would remove redundant placeholders like 
https://github.com/greshake/i3status-rust/blob/094cfdc13953b31edda8914f5e536e52b93deb38/src/blocks/disk_space.rs#L274-L275
and add give ability to draw bar for, lets say, sound block, as well as every other block.

ammgws:
Oh wow I totally forgot that even existed. It's not even documented for the sound block.

Anyway, right now I can't think of any cases where configuring the min would be useful, so perhaps it may be OK to hardcode it at 0. One nit: the proposed syntax kinda looks like a number range due to the "~", so perhaps another char would be better. And maybe explicitly declare that it's a barchart to make it more obvious as well. "{var:bar:5:110}" or something.

MaxVerevkin:
> It's not even documented for the sound block.

Sound block doesn't have this feature.

>  One nit: the proposed syntax kinda looks like a number range due to the "~", so perhaps another char would be better. And maybe explicitly declare that it's a barchart to make it more obvious as well. "{var:bar:5:110}" or something.

The parsing I currently have doesn't work well (at all) with repeated "keys" (`:`, `;`, `*`), so maybe `"{var:5#100}"`?

ammgws:
> Sound block doesn't have this feature.

```rs
src/blocks/sound.rs
40:use crate::util::{format_percent_bar, FormatTemplate};
856:                    format_percent_bar(volume as f32)
870:                format_percent_bar(volume as f32)
```

image

MaxVerevkin:
Ohhh, well, that's weird :laughing:

ammgws:
Memory block also has a bunch of undocumented format key specifiers that output a percentage bar :laughing: 

```rust
src/blocks/memory.rs
277:            "{MFpb}" => format_percent_bar(mem_free.percent(mem_total)),
282:            "{MUpb}" => format_percent_bar(mem_total_used.percent(mem_total)),
287:            "{Mupb}" => format_percent_bar(mem_used.percent(mem_total)),
292:            "{MApb}" => format_percent_bar(mem_avail.percent(mem_total)),
299:            "{SFpb}" => format_percent_bar(swap_free.percent(swap_total)),
304:            "{SUpb}" => format_percent_bar(swap_used.percent(swap_total)),
309:            "{Bpb}" => format_percent_bar(buffers.percent(mem_total)),
314:            "{Cpb}" => format_percent_bar(cached.percent(mem_total)));
```

MaxVerevkin:
> a bunch of undocumented format key specifiers 

Apparently, #578 hasn't wrote any documentation,

MaxVerevkin:
It looks like I have to remove `alert_absolute` option from [disk space](https://github.com/greshake/i3status-rust/blob/master/blocks.md#disk-space).

MaxVerevkin:
https://github.com/greshake/i3status-rust/blob/094cfdc13953b31edda8914f5e536e52b93deb38/src/blocks/xrandr.rs#L197-L218
Why not just allow the user to set `format` option? This looks like a total misuse of formatting.

MaxVerevkin:
Should I try to make everything as backwards compatible as possible, or is it OK to remove some options like "alias" on disk_space and rewrite xrandr options? The next version is 0.15.0 (as far as I know), and if we were to follow semantic versioning, this is not allowed. However, lots of braking changes have already made without changing the major version.

ammgws:
> Should I try to make everything as backwards compatible as possible, or is it OK to remove some options like "alias" on disk_space and rewrite xrandr options? The next version is 0.15.0 (as far as I know), and if we were to follow semantic versioning, this is not allowed. However, lots of braking changes have already been made without changing the major version.

Yeah I was prepared to allow some breaking changes (hence the deprecation warnings in the v0.14.4 release notes), which is why I bumped the next version to 0.15.

> It looks like I have to remove alert_absolute option from disk space, because the unit is not set to a concrete value.

This was added in https://github.com/greshake/i3status-rust/pull/714 with related discussion here: https://github.com/greshake/i3status-rust/issues/829. Is there no other way to accomplish the use-case?

> What is the point of diskspace alias placeholder?

It's just leftover from pre-format string days.

MaxVerevkin:
Just to make clear it - we do not follow semantic versioning, right?

MaxVerevkin:
I still don't fully understand this project's deprecation policy. For example, lets take xrandr block - if I add a formatting string option, can I remove old options?

Or I have to leave them with a deprecation note left if the documentation and remove them after a few versions? If that's the case, maybe we should display a warning regarding those deprecated options for a few seconds in the beginning so the user would be informed?

MaxVerevkin:
> Is there no other way to accomplish the use-case?

If `alert_absolute` is set, we can assume the units are MB. But then it should be renamed to `alert_unit_mb` or something like that.

ammgws:
For the last couple of years the policy was basically to avoid breaking user's configs, which meant deprecating things on paper but not actually in practice. Over time this has lead to a build up of messy / hackish code in some parts, so to ease the maintenance burden I would be fine with cleaning up the code of options that have been deprecated since a few releases back.

For the new deprecations like xrandr that you mentioned above, if possible leave a deprecation note so that they can be removed in a later release. However if there is no other way to proceed without removing them, I give this PR precedence since it's a feature that is sorely needed.

> maybe we should display a warning regarding those deprecated options for a few seconds in the beginning so the user would be informed?

Could be a good way to get the deprecation more visibility for those who don't read the release notes, but I wouldn't consider it a blocker for this PR.

ammgws:
> If alert_absolute is set, we can assume the units are MB. 

Another way could be to leave the `unit` option and rename the description:
Before: "Unit that is used to display disk space. Options are MB, MiB, GB, GiB, TB and TiB"
After: "Unit used for alert_absolute. Options are MB, MiB, GB, GiB, TB and TiB"
... or did you mention that this wasn't possible?

@jamesmcm Do you have any suggestions?

MaxVerevkin:
> can be removed in a later release

It can be any release, not just a "minor" release (0.13.x -> 0.14.0), right?

MaxVerevkin:
> ... or did you mention that this wasn't possible?

The way you have suggested it, it is possible, at least I think so.

ammgws:
> It can be any release, not just a "minor" release (0.13.x -> 0.14.0), right?

Sure. If it means better visibility for the "breaking changes" then I wouldn't mind just bumping it up from 0.15.x to 0.16.0 when the time comes.

MaxVerevkin:
I'll leave xrandr formatting option to another PR for now.

jamesmcm:
> Another way could be to leave the `unit` option and rename the description:
> Before: "Unit that is used to display disk space. Options are MB, MiB, GB, GiB, TB and TiB"
> After: "Unit used for alert_absolute. Options are MB, MiB, GB, GiB, TB and TiB"
> ... or did you mention that this wasn't possible?

I'm happy with that - how do we specify the size unit for the display (i.e. if I want to use GiB or TiB) with the new config?

The only other thing to consider is if there's a way we could direct users to some page for config deprecation issues if there is an issue in their config file - i.e. a quick summary of what has changed, a bit like some Rust crates have on their Rust docs page.

ammgws:
> The only other thing to consider is if there's a way we could direct users to some page for config deprecation issues if there is an issue in their config file - i.e. a quick summary of what has changed, a bit like some Rust crates have on their Rust docs page.

Hmm right now the info is scattered around in a few places - the release notes on GitHub / NEWS.md, blocks.md and this severely outdated wiki: https://github.com/greshake/i3status-rust/wiki/past-breaking-changes.

MaxVerevkin:
Eh, it is not possible to use dot in `​{<name>[:[0]<min width>][.<max width>][;<min suffix>][*<unit>]}` because that makes it impossible to use floats in formatting config. Any ides what to use instead?

MaxVerevkin:
> Add ability to draw bars for every numeric placeholder.

That's how it looks with this config
```toml
[[block]]
block = "sound"
on_click = "pavucontrol"
format = "{volume:5#110} {volume:03}"
```
https://imgur.com/a/CCNw04e

GladOSkar:
> Eh, it is not possible to use dot in `​{<name>[:[0]<min width>][.<max width>][;<min suffix>][*<unit>]}` because that makes it impossible to use floats in formatting config. Any ideas what to use instead?

I don't really get why? The width value can never be a float, so can't you just manually split it before arsing the number?

MaxVerevkin:
> I don't really get why? The width value can never be a float, so can't you just manually split it before arsing the number?

Not width but `bar_max_value`:
```toml
[[block]]
block = "sound"
format = "{volume:5#110.5} {volume:03}"
```

MaxVerevkin:
> How about ^ for the max width?

Yeah, that's what I ended up using :smile: :

> ​`{<name>[:[0]<min width>][^<max width>][;<min suffix>][*<unit>][#<bar max value>]}`

MaxVerevkin:
In terms of functionality I've implemented everything I wanted. Any thoughts? Maybe I missed something?

jamesmcm:
This is a great contribution!

If we continue to add to the formatting language, it might be worth using [nom](https://github.com/Geal/nom) to parse it in the future rather than dealing with it character by character directly.

There's a lot of documentation to update between the README, man page, and the blocks.md. Perhaps this could be merged first without a release (with some general documentation on the formatting language), and then it would be easier for people to contribute to the blocks documentation individually?

GladOSkar:
Issues i've noticed trying to use this:

- `disk_space` gives `Internal error in context 'util': Unknown placeholder in format string: alias` after i removed the now illegal options `alias` and `unit`.
- `memory` gives `Configuration error: incorrect `min_suffix`. Cause: suffix 'M' cannot be used to format byte value` when no formats specified. Also the placeholder `Mup` isn't recognized anymore but i assume that's intentional and will be documented?
- `battery` shows as `99%%` by default (i assume the default format string still has a `%` somewhere)
- `battery` `{time}` placeholder doesn't print anything anymore
- `speedtest` default format prints `37.ms 42.MiBi/s 47.MiBi/s` -> there should never be a decimal point at the end of the number. `format_number()` used to prevent this.

I haven't tried any of the new placeholder stuff yet

MaxVerevkin:
> there should never be a decimal point at the end of the number.

I've made it this way in order to maintain a constant amount of digits.

> memory gives ...

Memory formatting placeholders are completely rewritten.

> disk_space gives ...

Alias is removed.

> speedtest now colors the entire block based on ping instead of just 1 widget

Because I'm not sure how to use formatting placeholders and multiple widgets at the same. I have ideas, but on the time to implement it.

> Using {var:0} in a format string causes

> battery {time} placeholder doesn't print anything anymore

Will investigate.

> assume the default format string still has a % somewhere

With this PR every placeholder has a unit in it.

Thanks for the feedback!

GladOSkar:
Added some more issues to my first comment. LMK if another form of feedback works better for you

MaxVerevkin:
> and will be documented?

Sure, but basically every block has some changes so it will take some time.

GladOSkar:
Yeah no worries i can help with that if you want

MaxVerevkin:
Looks like I have to review default formatting strings again :laughing:

MaxVerevkin:
https://github.com/greshake/i3status-rust/blob/094cfdc13953b31edda8914f5e536e52b93deb38/src/blocks/docker.rs#L61-L63

Why percents?

MaxVerevkin:
@GladOSkar is it better now?

By the way, you can refer to this until there is a documentation for memory block:

https://github.com/MaxVerevkin/i3status-rust/blob/0b4bf2ac71cd33f295092394feb26d2aedf341b1/src/blocks/memory.rs#L214-L233

MaxVerevkin:
> Could be done with " 37ms" or "037ms"?

Yeah, that's definitely better.

MaxVerevkin:
> speedtest with format = "{speed_down:1*Mb/s}" causes thread 'main' panicked at 'called 

Fixed

GladOSkar:
Neat! I crossed off everything that works now.

> > speedtest with format = "{speed_down:1*Mb/s}" causes thread 'main' panicked at 'called
> 
> Fixed

Not for me, i'm now getting:
```
Configuration error: Can not parse unit.
Cause: unknown unit: 'Mb/s'
```

GladOSkar:
No i wanted it to display Megabits as `Mb` instead of `MiBi`. But i see what the issue is now, the unit parameter is not meant for adjusting how the value is displayed but only to choose from a range of given unit options. Hm. I had imagined thati could choose that using that parameter as well but that might be difficult

MaxVerevkin:
> Using {var:0} in a format string causes thread 'main' panicked at 'called ``Result::unwrap()`` on an ``Err`` value: ParseIntError { kind: Empty }', src/formatting.rs:112:52

Error handling is not completed yet, but this *is* wrong formatting string. If `min_width` starts with a zero, that means that `pad_with='0'`. Buy you haven't set the min width itself. I suspect `{var:00}` works as expected.

MaxVerevkin:
Ah, so you want to show *bits* but the unit to be `B` instead of `Bi`?

GladOSkar:
> > Using {var:0} in a format string causes thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: Empty }', src/formatting.rs:112:52
> 
> Error handling is not completed yet, but this _is_ wrong formatting string. If `min_width` starts with a zero, that means that `pad_with='0'`. Buy you haven't set the min width itself. I suspect `{var:00}` works as expected.

Yeah i assumed so. It's probably not needed anyway, but crashing the main thread is also not great.
C `printf` prints `1234` for `printf("%0d", 1234);` which seems logical to me.

GladOSkar:
> > > battery {time} placeholder doesn't print anything anymore

OK ignore this i was stupid and looked at it while charged.

MaxVerevkin:
> No, as b. And Mi as M, ideally. I now it's technically not correct but way less distracting imo. That's the way i had it before this PR

So basically, your point is that, for example, suffix (BTW, are those things called "prefixes"?) `M`, it should mean `2^20` for bytes/bits and `10^6` for everything else?

MaxVerevkin:
> speedtest uses unit MiBi/s by default but i think Mib/s makes more sense for bits.

Ah, I get it. "B" for bytes, "b" for bits?

GladOSkar:
> > No, as b. And Mi as M, ideally. I now it's technically not correct but way less distracting imo. That's the way i had it before this PR
> 
> So basically, your point is that, for example, suffix (BTW, are those things called "prefixes"?) `M`, it should mean `2^20` for bytes/bits and `10^6` for everything else?

No not that either. It's 2 things:

1. Use `b` instead of `Bi` to display `bits` because it's easier to read. No change in number value.
2. Use `M` instead of `Mi` (and so on for `G`/`Gi`...) because it's easier to read. No change in number value. Maybe optional?

GladOSkar:
> > speedtest now colors the entire block based on ping instead of just 1 widget
> 
> Because I'm not sure how to use formatting placeholders and multiple widgets at the same time. I have ideas, but on the time to implement it.

Interesting. I thought that'd not be fixable, but i also don't really need it to be. I always found it confusing that just the ping value would always be green when it's `good` while everything else stayed normal. I'd suggest just not making the block green on low pings, because then there's way too much green.

MaxVerevkin:
> No not that either. It's 2 things:
> 
> 1. Use `b` instead of `Bi` to display `bits` because it's easier to read. No change in number value.
> 2. Use `M` instead of `Mi` (and so on for `G`/`Gi`...) because it's easier to read. No change in number value. Maybe optional?

That makes sense. I'm convinced :smile:.

GladOSkar:
Sorry for my difficulties explaining. Not my native language yadda yadda. Thanks for doing all this, very neat!

MaxVerevkin:
> BTW, are those things called "prefixes"?

Well, those *are* prefixes. Rename everything? :smile:

MaxVerevkin:
Renamed just in case to prevent future confusions.

MaxVerevkin:
Does everyone agree on 0b703c9?

GladOSkar:
That broke the default `disk_space` format :D

MaxVerevkin:
I think it is OK to remove bluetooth's `label` option? Or it's better to just leave a deprecation note?

MaxVerevkin:
> Using {var:0} in a format string causes thread 'main' panicked at 'called ``Result::unwrap()`` on an ``Err`` value: ParseIntError { kind: Empty }', src/formatting.rs:112:52

Hey, you are right! Only now I got that it is not possible to set `pad_with=' '` and `min_width=0` at the same time using current implementation. So I guess, `{var:0}` should mean "pad with spaces" and '{var:00}` "pad with zeros".

formatting.md
```diff
@@ -0,0 +1,103 @@
+# Formatting
```
ammgws, line 1: Shall we move this section to blocks.md either below or above the Escaping Text section?

GladOSkar: Yeah i think that makes sense with regard to un-scattering the documentation

MaxVerevkin: OK, sure.

TilCreator:
I used your fork to implement the radeontop block in [block_add_radeontop](https://github.com/TilCreator/i3status-rust/tree/block_add_radeontop) and noticed that the Bytes are not converted to GB, but the Hz are converted to Ghz. How is that handeled?
screenshot_1615750107

MaxVerevkin:
@TilCreator That's because you define those bytes as integers. From blocks.md:

> Float values are formatted following [engineering notation](https://en.wikipedia.org/wiki/Engineering_notation).

> Also the formatting is awesome!

Thanks :smiley:

TilCreator:
Ah, thx
will use that in [block_add_lact](https://github.com/TilCreator/i3status-rust/tree/block_add_lact) (nothing jet)

blocks.md
```diff
@@ -62,17 +62,21 @@ critical_updates_regex = "(linux |linux-lts|linux-zen)"
 Key | Values | Required | Default
 ----|--------|----------|--------
 `interval` | Update interval in seconds. | No | `600`
-`format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{count}"`
-`format_singular` | Same as `format`, but for when exactly one update is available. | No | `"{count}"`
-`format_up_to_date` | Same as `format`, but for when no updates are available. | No | `"{count}"`
+`format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{count:1}"`
```
ammgws, line 65: Can you add something like "Refer to [formatting documentation](blocks.md#formatting) to customize formatting strings' placeholders."  in the description for `format`?

blocks.md
```diff
@@ -62,17 +62,21 @@ critical_updates_regex = "(linux |linux-lts|linux-zen)"
...
+`format_singular` | Same as `format`, but for when exactly one update is available. | No | `"{count:1}"`
+`format_up_to_date` | Same as `format`, but for when no updates are available. | No | `"{count:1}"`
 `warning_updates_regex` | Display block as warning if updates matching regex are available. | No | `None`
 `critical_updates_regex` | Display block as critical if updates matching regex are available. | No | `None`
 
 #### Available Format Keys
```
ammgws, line 72: Or perhaps it could be added here. For every block that has format strings, to increase visibility of the feature.

blocks.md
MaxVerevkin: Do you mean to add this reference for every block? Given the direction of moving all the blocks to formatting strings, I don't think it's a good idea, because it is just too redundant. Maybe add a note in the beginning of blocks.md?

ammgws: Yeah I meant to add it to every block like the Escaping Text reference, however the only reason is for visibility. Moving it to the beginning of blocks.md could be OK too. In the future it might be best to overhaul the docs, especially for top-level config stuff.

blocks.md
```diff
@@ -384,40 +394,35 @@ Creates a block which displays disk space information.
...
-format = "{icon}{used}/{total} {unit} ({available}{unit} free)"
+format = "{icon} {used}/{total} ({available} free)"
 ```
 
 #### Options
 
 `alert` | Available disk space critical level as a percentage or Unit. | No | `10.0`
-`alias` | Alias that is displayed for path. | No | `"/"`
```
ammgws, line 398: I thought we were keeping this around for now and deprecating in a later release?

MaxVerevkin: Ah, yeah, I'll move it back.

src/blocks/speedtest.rs
```diff
@@ -203,22 +159,16 @@ impl Block for SpeedTest {
...
+                let values = map!(
+                    "ping" => Value::from_float(ping).seconds().icon(self.ping_icon.clone()),
+                    "speed_down" => Value::from_float(down).bits_per_second().icon(self.down_icon.clone()),
+                    "speed_up" => Value::from_float(up).bits_per_second().icon(self.up_icon.clone()),
+                );
+                self.text.set_text(self.format.render(&values)?);
 
                 // ping is in seconds
-                self.text[0].set_state(match (ping * 1000.) as i32 {
+                self.text.set_state(match (ping * 1000.) as i32 {
```
ammgws, line 171: @GladOSkar You mentioned in another comment:

> speedtest now colors the entire block based on ping instead of just 1 widget. I assume this is not fixable, but should be accounted for with some saner defaults

The reason is because speedtest was changed to use a format string by combining the two text widgets into one. The previous functionality should be able to kept if we keep the two widgets, and instead have one format string for each widget (Similar to how the NetworkManager block does it: https://github.com/greshake/i3status-rust/blob/master/blocks.md#networkmanager)

@MaxVerevkin, @GladOSkar Reckon we should go with that?

One disadvantage would be that the order of the widgets would be unconfigurable (as it is now), but that could be fixed by adding another config option for the widget order. This would allow us to have the best of both worlds.

blocks.md
ammgws: Can create a `#### Deprecated Format Keys` section for it, like those other blocks

src/blocks/networkmanager.rs
```diff
@@ -736,27 +727,19 @@ impl Block for NetworkManager {
...
-                                            truncated.truncate(self.max_ssid_width);
-                                            truncated
-                                        }
-                                        Err(_) => "".to_string(),
-                                    };
-                                    let strength = match ap.strength(&self.dbus_conn) {
-                                        Ok(v) => format!("{}", v).to_string(),
-                                        Err(_) => "0".to_string(),
-                                    };
+                                    let ssid = ap.ssid(&self.dbus_conn).unwrap_or_else(|_| "N/A".to_string());
```
ammgws, line 730: So now this displays "N/A" instead of just an empty string? Is the change documented somewhere?

src/blocks/notify.rs
```diff
@@ -31,6 +33,7 @@ pub struct Notify {
 #[serde(deny_unknown_fields)]
 pub struct NotifyConfig {
     /// Format string for displaying phone information.
+    /// "phone"?
```
ammgws, line 36: Copy paste error :laughing:  Can just delete the comment since we all know what the format string is for.

src/blocks/pacman.rs
```diff
@@ -134,11 +136,7 @@ impl PacmanConfig {
             // most likely a mistake: {count}, {pacman}, {aur}, {both} not found in format string
             Err(ConfigurationError(
                 "pacman".to_string(),
-                (
-                    "No formatter ({count}|{pacman}|{aur}|{both}) found in format string"
-                        .to_string(),
-                    "invalid format".to_string(),
-                ),
+                "No formatter ({count}|{pacman}|{aur}|{both}) found in format string".to_string(),
```
ammgws, line 139: IIRC we don't do this kind of check for the other blocks, do we?

src/blocks/temperature.rs
```diff
@@ -93,7 +95,7 @@ pub struct TemperatureConfig {
 
 impl TemperatureConfig {
     fn default_format() -> String {
-        "{average}° avg, {max}° max".to_owned()
```
ammgws, line 96: Is the degrees symbol deleted for a reason?

src/blocks/xrandr.rs
```diff
@@ -194,13 +195,16 @@ impl Xrandr {
...
-                              "{brightness}" => m.brightness.to_string(),
-                              "{brightness_icon}" => self.shared_config.get_icon("backlight_full").unwrap_or_default().trim().to_string(),
-                              "{resolution}" => m.resolution.clone(),
-                              "{res_icon}" => self.shared_config.get_icon("resolution").unwrap_or_default().trim().to_string());
-
-            self.text.set_icon("xrandr");
+            let values = map!(
+                "display" => Value::from_string(m.name.clone()),
+                "brightness" => Value::from_integer(m.brightness as i64).percents(),
+                "brightness_icon" => Value::from_string(self.shared_config.get_icon("backlight_full")?), //FIXME?
```
ammgws, line 201: What's with the FIXME? Because the icon doesn't change like the backlight block?

src/blocks/pacman.rs
MaxVerevkin: No, we don't. By the way, from a closer look it doesn't even work with the new formatting.

src/config.rs
```diff
@@ -58,12 +58,15 @@ impl SharedConfig {
...
-                .replace("{icon}", self.icons.0.get(icon)?),
-        )
+    pub fn get_icon(&self, icon: &str) -> crate::errors::Result<String> {
+        use crate::errors::OptionExt;
+        Ok(self.icons_format.clone().replace(
+            "{icon}",
+            self.icons
+                .0
+                .get(icon)
+                .internal_error("get_icon()", &format!("icon '{}' not found", icon))?,
```
ammgws, line 68: Perhaps we could improve the error message here: tell the user to check their current theme's config is correct or something like that.

src/blocks/xrandr.rs
MaxVerevkin: Yes

src/formatting/value.rs
```diff
@@ -0,0 +1,253 @@
...
+
+fn format_bar(value: f64, length: usize) -> String {
+    let value = value.clamp(0., 1.);
+    let chars_to_fill = value * length as f64;
+    (0..length)
+        .map(|i| {
+            let printed_chars = i as f64;
+            let val = (chars_to_fill - printed_chars).clamp(0., 1.) * 8.;
+            match val as usize {
+                0 => ' ',
```
ammgws, line 108: Perhaps we could add these to icons.rs as bar0, bar1, etc or something like that.

src/blocks/pacman.rs
ammgws: I'm fine with removing the check here.

src/blocks/temperature.rs
MaxVerevkin: Yes. Every placeholder displays a unit on it's own.

src/blocks/xrandr.rs
ammgws: Oh OK. Can you update the comment to make it more obvious? Then we can just leave it for the future.

src/blocks/temperature.rs
ammgws: Oh right, sorry about that!

src/formatting/value.rs
MaxVerevkin: I'm not sure if there are alternative blocks symbols, in which case it's a reasonable thing to to.

src/blocks/temperature.rs
ammgws: Then the docs need updating too since it still has the degrees symbol in the example config and in the format string description.

blocks.md
MaxVerevkin: By the way, when deprecation placeholders, it is OK to change the default format not to include it?

blocks.md
```diff
@@ -1636,3 +1626,107 @@ command = "echo '<b>1 &amp;</b>'"
...
+`0.01`             | "0.0"
+`0.1`              | "0.1"
+`1.0`              | "1.0"
+`12.0`             | " 12"
+`123.0`            | "123"
+`1234.0`           | "1.23K"
+
+### `<unit>`
+
+Some placeholders have a "unit". For example, `net` block displays speed in `B/s`. This option gives abitity to convert one units into another. Ignored for strings.
```
ammgws, line 1712: Seems like the unit should be added to each block's format key table, for example for the Temperature block:

 Key | Value | Type | Unit
-----|-------|-----| -----
`{min}` | Minimum temperature among all sensors | Integer | ° 
`{average}` | Average temperature among all sensors | Integer | ° 
`{max}` | Maximum temperature among all sensors | Integer | °

src/blocks/weather.rs
```diff
@@ -274,14 +272,16 @@ impl Weather {
...
-                                  "{temp}" => format!("{:.0}", raw_temp),
-                                  "{humidity}" => format!("{:.0}", raw_humidity),
-                                  "{apparent}" => format!("{:.0}",apparent_temp),
-                                  "{wind}" => format!("{:.1}", raw_wind_speed),
-                                  "{wind_kmh}" => format!("{:.1}", kmh_wind_speed),
-                                  "{direction}" => convert_wind_direction(raw_wind_direction),
-                                  "{location}" => raw_location);
+                self.weather_keys = map!(
+                    "weather" => Value::from_string(raw_weather),
+                    "temp" => Value::from_integer(raw_temp as i64),
```
ammgws, line 277: Not .degrees() like the Temperature block?

src/formatting/value.rs
ammgws: Can just add a TODO here to be revisited in the future.

blocks.md
ammgws: So there is no way to have it display the same thing using different placeholders? What's the before/after look like in this case?

MaxVerevkin: > So there is no way to have it display the same thing using different placeholders?

There is, just asking.

ammgws: In that case the best thing is to have the output be the same :)

src/blocks/weather.rs
MaxVerevkin: Forgot to add. And by the way, should it be `°C` instead of `°`?

ammgws: I've overridden it to °C in my config, but then again there are people who use fahrenheit so maybe just leave it the same as it is now?

src/blocks/speedtest.rs
MaxVerevkin: So your idea is basically to split `format` into
``` toml
format_ping = "{ping}"
format_speed = "{speed_down} {speed_up}"
```
I think It's good enough.

MaxVerevkin: Or maybe I have a different solution: we can just use pango's color attributes. This way, `Value` can be extended to have a background and foreground color. I even prefer this solution as it is more generic.

GladOSkar: That would be very neat of course but i doubt that'll render the color across the entire bar height, and instead only color the text height. But it's worth a shot.

MaxVerevkin: @GladOSkar Can you test 9f50e09? It works for me just fine.

GladOSkar: Perrsonally, i always disabled ping coloring entirely because i found it distracting and also not very meaningful. But that's probably not viable for everyone

GladOSkar: Nope, on `i3` it's as i feared:
image

src/formatting/unit.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl TryInto<Unit> for &str {
+    type Error = crate::errors::Error;
+
+    fn try_into(self) -> Result<Unit> {
+        match self {
+            "b/s" => Ok(Unit::BitsPerSecond),
+            "B/s" => Ok(Unit::BytesPerSecond),
+            "%" => Ok(Unit::Percents),
+            "°" => Ok(Unit::Degrees),
```
MaxVerevkin, line 47: I don't think it makes much sense to use a degree symbol in configs. Should I change it to "degrees" or something?

GladOSkar: `deg` is the classig shorthand i think

src/blocks/speedtest.rs
MaxVerevkin: That's what I have on `sway`
image

MaxVerevkin: > the neighboring placeholder is also not colored.

Should it?

GladOSkar: > That's what I have on sway

Yeah we had similar issues before. Sway renders quite differently from i3, since the latter has sort of 2 border rows that aren't part of the pango render area.

GladOSkar: > Should it be?

Yes. See `master`:
image

MaxVerevkin: Oh, you mean the separator. Well, this *is* unfixable without splitting the block into two widgets.

GladOSkar: Oh yeah sorry i confused those for a sec there

GladOSkar: But yeah the "widget" paradigm and the "format string" paradigm seem to be somewhat opposing at the moment. Do you think there is a way we could specify widgets within the format string? That seems a bit overkill. So probably multiple format strings are the way to go, but that isn't nice either since it's not very flexible with regard to adding/removing widgets

MaxVerevkin: > Do you think there is a way we could specify widgets within the format string? That seems a bit overkill

Translating formatting string like `"text {var} {var2}"` into four widgets seems overkill indeed. But maybe it is not that bad?

GladOSkar: I was thinking of something like `"{{ping}} {{up} {down}}"` or `"{ping} {|} {up} {down}"`. But i have absolutely no idea about the viability of this

MaxVerevkin: I mean... 714b29f1f0244c976780a0bf9c35c8fe9bbd6149 kinda works.

GladOSkar: It does, and quite well. Just some spacing to figure out:
image

ammgws: @MaxVerevkin on phone now so hard to check but will render_widgets still allow per widget click actions to work? (For when applying the same method to other blocks such as Nvidia)

MaxVerevkin: > will render_widgets still allow per widget click actions to work?

Yes, but I haven't tried yet. That's how I see it: if I wanted to handle a click on ping separately from the rest of the block, I can set it's `instance` right after I set it's status.

blocks.md
```diff
@@ -161,33 +176,23 @@ Key | Values | Required | Default
...
-`upower` | Deprecated in favour of `device`. When `true`, use the Upower D-Bus driver. | No | `false`
-
-The `show` option is deprecated, and will be removed in future versions. In the meantime, it will override the `format` option when present.
-
 #### Available Format Keys
 
-Placeholder | Description
-------------|-------------
-`{percentage}` | Battery level, in percent
-`{bar}` | The current battery level in a bar chart
```
ammgws, line 188: Is it possible to keep {bar} around and deprecate it instead? Then we can remove it in a future release.

blocks.md
```diff
@@ -251,22 +269,16 @@ Key | Values | Required | Default
 `critical` | Minimum usage, where state is set to critical. | No | `90`
 `interval` | Update interval, in seconds. | No | `1`
 `format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{utilization}"`
-`per_core` | Display CPU frequencies and utilization per core. | No | `false`
```
ammgws, line 254: There's no way to keep this around is there?

blocks.md
```diff
@@ -384,40 +396,47 @@ Creates a block which displays disk space information.
 [[block]]
 block = "disk_space"
 path = "/"
-alias = "/"
 info_type = "used"
 unit = "GiB"
-format = "{icon}{used}/{total} {unit} ({available}{unit} free)"
+format = "{icon} {used}/{total} ({available} free)"
```
ammgws, line 401: Is the space added because the icon no longer has spaces inside of it?

blocks.md
```diff
@@ -384,40 +396,47 @@ Creates a block which displays disk space information.
...
+`alias` | Sets the value for `{alias}` placeholder | No | `"/"`
+
+
 #### Available Format Keys
 
-Key | Value
-----|-------
-`{alias}` | Alias for disk path
-`{available}` | Available disk space (free disk space minus reserved system space)
-`{bar}` | Display bar representing percentage
```
ammgws, line 413: Is it possible to keep {bar} around and deprecate it instead? Then we can remove it in a future release.

blocks.md
```diff
@@ -384,40 +396,47 @@ Creates a block which displays disk space information.
...
-`{free}` | Free disk space
-`{icon}` | Disk drive icon
-`{path}` | Path used for capacity check
-`{percentage}` | Percentage of disk used or free (depends on info_type setting)
-`{total}` | Total disk space
-`{unit}` | Unit used for disk space (see above)
```
ammgws, line 419: Is it possible to keep {unit} around and deprecate it instead? Then we can remove it in a future release. I'm guessing not due to the formatter changes, in that case we need to mention it somewhere.

blocks.md
```diff
@@ -811,44 +830,26 @@ Key | Values | Required | Default
 
 #### Available Format Keys
 
-  Key    | Value
```
ammgws, line 814: There's no way to keep these around is there? (Deprecate first and remove in future release). If not, need to figure out how to document the changes. Perhaps can re-add this table as "Removed Format Keys" and add a column showing what the new equivalent placeholder is?

blocks.md
```diff
@@ -939,11 +940,13 @@ but the 'use_bits' flag can be set to `true` to convert the units to bps (little
 
 #### Examples
 
+Displays ssid, signal_strength, ip, down speed and up speed as bits per second
```
ammgws, line 943: Maybe add "with minimum unit set to K" or similar to the end?

blocks.md
```diff
@@ -953,29 +956,25 @@ use_bits = false
 Key | Values | Required | Default
 ----|--------|----------|--------
 `device` | Network interface to monitor (name from /sys/class/net). | No | Automatically chosen from the output of `ip route show default`
-`format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{speed_up} {speed_down}"`
+`format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{speed_up;K} {speed_down;K}"`
 `format_alt` | If set, block will switch its formatting between `format` and `format_alt` on every click. | No | None
-`speed_digits` | Number of digits to use when displaying speeds. | No | `3`
-`speed_min_unit` | Smallest unit to use when displaying speeds. Possible choices: `"B"`, `"K"`, `"M"`, `"G"`, `"T"`. | No | `"K"`
```
ammgws, line 959: Need to document that this has been removed. Perhaps better if we show what the new way is as well.

blocks.md
```diff
@@ -953,29 +956,25 @@ use_bits = false
 Key | Values | Required | Default
 ----|--------|----------|--------
 `device` | Network interface to monitor (name from /sys/class/net). | No | Automatically chosen from the output of `ip route show default`
-`format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{speed_up} {speed_down}"`
+`format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{speed_up;K} {speed_down;K}"`
 `format_alt` | If set, block will switch its formatting between `format` and `format_alt` on every click. | No | None
-`speed_digits` | Number of digits to use when displaying speeds. | No | `3`
```
ammgws, line 958: Need to document that this has been removed. Perhaps better if we show what the new way is as well.

blocks.md
```diff
@@ -998,37 +997,36 @@ interface_name_include = []
 Key | Values | Required | Default
 ----|--------|----------|---------
 `primary_only` | Whether to show only the primary active connection or all active connections. | No | `false`
-`max_ssid_width` | Truncation length for SSID. | No | `21`
```
ammgws, line 1001: Need to document that this has been removed. Perhaps better if we show what the new way is as well.

blocks.md
```diff
@@ -953,29 +956,25 @@ use_bits = false
...
+`format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{speed_up;K} {speed_down;K}"`
 `format_alt` | If set, block will switch its formatting between `format` and `format_alt` on every click. | No | None
-`speed_digits` | Number of digits to use when displaying speeds. | No | `3`
-`speed_min_unit` | Smallest unit to use when displaying speeds. Possible choices: `"B"`, `"K"`, `"M"`, `"G"`, `"T"`. | No | `"K"`
-`use_bits` | Display speeds in bits instead of bytes. | No | `false`
 `interval` | Update interval, in seconds. Note: the update interval for SSID and IP address is fixed at 30 seconds, and bitrate fixed at 10 seconds. | No | `1`
 `hide_missing` | Whether to hide interfaces that don't exist on the system. | No | `false`
 `hide_inactive` | Whether to hide interfaces that are not connected (or missing). | No | `false`
-`max_ssid_width` | Truncation length for SSID. | No | `21`
```
ammgws, line 964: Need to document that this has been removed. Perhaps better if we show what the new way is as well.

blocks.md
```diff
@@ -953,29 +956,25 @@ use_bits = false
 Key | Values | Required | Default
 ----|--------|----------|--------
 `device` | Network interface to monitor (name from /sys/class/net). | No | Automatically chosen from the output of `ip route show default`
-`format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{speed_up} {speed_down}"`
+`format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{speed_up;K} {speed_down;K}"`
 `format_alt` | If set, block will switch its formatting between `format` and `format_alt` on every click. | No | None
-`speed_digits` | Number of digits to use when displaying speeds. | No | `3`
-`speed_min_unit` | Smallest unit to use when displaying speeds. Possible choices: `"B"`, `"K"`, `"M"`, `"G"`, `"T"`. | No | `"K"`
-`use_bits` | Display speeds in bits instead of bytes. | No | `false`
```
ammgws, line 960: Need to document that this has been removed. Perhaps better if we show what the new way is as well.

blocks.md
```diff
@@ -998,37 +997,36 @@ interface_name_include = []
...
 
 #### Connection format string
 
-Placeholder | Description
-------------|-------------
-`{devices}` | The list of devices, each formatted with the device format string
+ Key | Value | Type
+-----|-------|-----
+`{devices}` | The list of devices, each formatted with the device format string | String
+`{id}` | ??? | String
```
ammgws, line 1029: What's this? An undocumented one you found?

blocks.md
```diff
@@ -1319,10 +1316,17 @@ interval = 1800
 
 Key | Values | Required | Default
 ----|--------|----------|--------
-`bytes` | Whether to use bytes or bits in the display (true for bytes, false for bits). | No | `false`
```
ammgws, line 1322: Need to document that this has been removed. Perhaps better if we show what the new way is as well.

blocks.md
```diff
@@ -1319,10 +1316,17 @@ interval = 1800
...
+`format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{ping}{speed_down}{speed_up}"`
 `interval` | Update interval in seconds. | No | `1800`
-`speed_digits` | Number of digits to use when displaying speeds and latencies. | No | `3`
```
ammgws, line 1324: Need to document that this has been removed. Perhaps better if we show what the new way is as well.

blocks.md
```diff
@@ -1319,10 +1316,17 @@ interval = 1800
...
-`speed_min_unit` | Smallest unit to use when displaying speeds. Possible choices: `"B"`, `"K"`, `"M"`, `"G"`, `"T"`. | No | `"K"`
```
ammgws, line 1325: Need to document that this has been removed. Perhaps better if we show what the new way is as well.

ammgws (changes requested):

blocks.md
MaxVerevkin: I don't think there is.

blocks.md
MaxVerevkin: I'm not sure if it's possible.

blocks.md
MaxVerevkin: Yes

blocks.md
MaxVerevkin: So you want it to be an alias to `unit` config option?

blocks.md
MaxVerevkin: There is no sane way to keep them. I can imagine some hacky ways, but I really what to avoid them.

> Perhaps can re-add this table as "Removed Format Keys" and add a column showing what the new equivalent placeholder is?

Good idea.

blocks.md
MaxVerevkin: I'll add some examples.

blocks.md
MaxVerevkin: Yeah, it was undocumented and I decided to leave it as that.

blocks.md
ammgws: OK then we can document the new way to do it somewhere.

MaxVerevkin: In case of battery block I've already added an example.

ammgws: Let's document these breaking changes in NEWS.md. I started it off with a couple.

blocks.md
ammgws: Let's document these breaking changes in NEWS.md. I added this one to the file to start with.

blocks.md
ammgws: Actually, we can just remove it since leaving it around would make the output weird and users would have to look up what changed anyway.

Add it to the list of breaking changes in NEWS.md and we can resolve this that way.

blocks.md
ammgws: If not possible then document the breaking change in NEWS.md. I started it off with a couple of examples from the other blocks.

blocks.md
ammgws: Let's document these breaking change in NEWS.md. I started it off with a couple of examples from other blocks.

blocks.md
ammgws: Let's document these breaking change in NEWS.md. I started it off with a couple of examples from other blocks.

blocks.md
ammgws: Let's document these breaking change in NEWS.md. I started it off with a couple of examples from other blocks.

blocks.md
ammgws: Let's document these breaking change in NEWS.md. I started it off with a couple of examples from other blocks.

src/blocks/speedtest.rs
MaxVerevkin: I'll merge 714b29f then.

blocks.md
ammgws: Let's just leave this one undocumented for now.

ammgws:
@MaxVerevkin Apart from docs, are you happy with this to go?

@GlasOSkar Have all the points you raised previously now addressed?

MaxVerevkin:
> Apart from docs, are you happy with this to go?

Well, yes. The only thing I'm not happy with is `render_widgets()`, but I've left a note not to use it and most likely I'll rewrite it later.

GladOSkar:
> @GladOSkar Have all the points you raised previously now addressed?

Yes they have. I already ticked all of them off in my original comment. I also only see very minor issues with `render_widgets()` that can be saved for later. So green light from me.

blocks.md
```diff
@@ -1636,3 +1692,107 @@ command = "echo '<b>1 &amp;</b>'"
...
+`123.0`            | "123"
+`1234.0`           | "1.23K"
+
+### `<unit>`
+
+Some placeholders have a "unit". For example, `net` block displays speed in `B/s`. This option gives abitity to convert one units into another. Ignored for strings.
+
+#### Example
+
+`"{speed_down*b/s}"` - show the download speed in bits per second.
```
ammgws, line 1782: Does this actually work for you?

I get 

> Configuration error: Can not convert unit.
> Cause: it is not possible to convert 'BytesPerSecond' to 'BitsPerSecond'

with 

```toml
[[block]]
block = "net"
format = "{speed_down*b/s}"
```

blocks.md
```diff
@@ -1636,3 +1692,107 @@ command = "echo '<b>1 &amp;</b>'"
...
+`0.01`             | "0.0"
+`0.1`              | "0.1"
+`1.0`              | "1.0"
+`12.0`             | " 12"
+`123.0`            | "123"
+`1234.0`           | "1.23K"
+
+### `<unit>`
+
+Some placeholders have a "unit". For example, `net` block displays speed in `B/s`. This option gives abitity to convert one units into another. Ignored for strings.
```
ammgws, line 1778: Shouldn't we list the available units here?

blocks.md
```diff
@@ -1636,3 +1692,107 @@ command = "echo '<b>1 &amp;</b>'"
...
+
+#### Example
```
ammgws, line 1780: Can we also add an example for disabling the unit?

Maybe I'm being silly but I thought this should be able to do it:

```toml
[[block]]
block = "net"
format = "{speed_down*\"\"}"
```

But somehow it doesn't get caught by the `match` in unit.rs and gets flagged invalid.

ammgws:
@GladOSkar Does the below config produce the output you would expect?

```toml
[[block]]
block = "net"
format = "{speed_down;M}"
```

blocks.md
MaxVerevkin: Fixed

blocks.md
MaxVerevkin: Yes, I'll add them.

blocks.md
MaxVerevkin: Fixed. No need to use `''`, just `format = "{speed_down*}"`

blocks.md
```diff
@@ -1281,7 +1330,7 @@ format = "{output_name} {volume}"
 Key | Values | Required | Default
 ----|--------|----------|--------
 `driver` | `"auto"`, `"pulseaudio"`, `"alsa"`. | No | `"auto"` (Pulseaudio with ALSA fallback)
-`format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `{volume}%`
+`format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `{volume}`
```
ammgws, line 1333: The percentage needs to be removed from the examples above this too

blocks.md
MaxVerevkin: Will something like this be OK?

```toml
[[block]]
block = "net"
format = "{speed_down*none}"
```

ammgws: I think that's fine

MaxVerevkin:
Did I documented all the breaking changes?

GladOSkar:
> @GladOSkar Does the below config produce the output you would expect?
> 
> ```toml
> [[block]]
> block = "net"
> format = "{speed_down;M}"
> ```

I don't think so:
image

GladOSkar:
> Have I documented all the breaking changes?

Seems complete to me.

One nit for `blocks.md`: In the "Memory" section "examples", the old placeholders are used instead of the new ones. That should be changed.

MaxVerevkin:
> I don't think so:

Should be fixed.

GladOSkar:
Yep, works great now. Fun typo ^^

GladOSkar:
Not sure if it's related to this PR, but i've noticed something weird with my `brightness` icon:

brightness

I use this to reverse the direction in my config and it usually worked fine:
```toml
[icons.overrides]
backlight_empty = " 🌑 "
backlight_partial1 = " 🌘 "
backlight_partial2 = " 🌗 "
backlight_partial3 = " 🌖 "
backlight_full = " 🌕 "
```

MaxVerevkin:
@GladOSkar Does it work correctly on master? By the way, I think it's unnecessary to put spaces around icons anymore (#1095)

GladOSkar:
Oh yeah it doesn't work on `master` either. Sorry, should have checked that. Will open an issue

blocks.md
```diff
@@ -939,11 +992,13 @@ but the 'use_bits' flag can be set to `true` to convert the units to bps (little
...
+Displays ssid, signal strength, ip, down speed and up speed as bits per second. Minimal prefix is set to `K` in order to prevent the block to change it's size.
+
 ```toml
 [[block]]
 block = "net"
 device = "wlp2s0"
-format = "{ssid} {signal_strength} {ip} {speed_down} {graph_down}"
+format = "{ssid} {signal_strength} {ip} {speed_down;K*b/s} {graph_down;K*b/s}"
 interval = 5
 use_bits = false
```
ammgws, line 1003: Need to remove use_bits from the example.

ammgws:
> Will something like this be OK?
> 
> ```toml
> [[block]]
> block = "net"
> format = "{speed_down*none}"
> ```

This is working fine, however one possible nit: with `use_bits` removed, you can't have the block display the speed in bits but without the unit appended.

MaxVerevkin:
> with use_bits removed, you can't have the block display the speed in bits and at the same time have the unit omitted.

True. Any ideas? And is it even necessary?

ammgws:
What's going on here? Looks like it's applying the engineering notation to the percentage (usage is around 0.03%) /

```toml
[[block]]
block = "memory"
format_mem = "{swap_used_percents}"
```

> "full_text":" MEM  27m% "

GladOSkar:
> > with use_bits removed, you can't have the block display the speed in bits and at the same time have the unit omitted.
> 
> True. Any ideas? And is it even necessary?

Best i can come up with is some sort of modifier. Maybe:

```
"{down:B}" -> "1KB"
"{down:b}" -> "8Kb"
"{down:_B}" -> "1"
"{down:_b}" -> "8"
```

MaxVerevkin:
> What's going on here? Looks like it's applying the engineering notation to the percentage (usage is around 0.03%) 

An exception  ~~can~~ should be made for percents.

GladOSkar:
Ah yes the classic millipercent, also known as the centipermille :D

MaxVerevkin:
> Also i always thought for network speeds bits were the default unit instead of bytes?

No, bits are default only for `speedtest`.

GladOSkar:
> > Also i always thought for network speeds bits were the default unit instead of bytes?
> 
> No, bits are default only for `speedtest`.

I meant in general use, not this software. But it doesn't really matter as long as both are possible

ammgws:
> is it even necessary?

It may be, since before it would have been displayed as "1.92Kb" and now it's "1.92Kb/s", and I imagine there are people who want to keep it short by removing the "/s" part. (Actually I am one of those people but I don't have it set to bits :smiley: )

ammgws:
> Also i always thought for network speeds bits were the default unit instead of bytes?

Bits for network speed, but when actually downloading files to disk (browser, torrents, etc) I think it's mostly been bytes.

GladOSkar:
Honestly same, but i don't really know how we could completely decouple the actual unit from the displayed unit unless we made the entire unit after the prefix customizable, which would then make it hard to semantically get the desired units (b vs B) from it (but it would probably be possible)

MaxVerevkin:
A little TODO:
- [ ] Do not apply engineering notation for percents
- [ ] `_b/s` should mean "convert to bits per second but don't display it"

GladOSkar:
Is there a way to do what @ammgws wanted? A way do discard the `/s`? I'm imagining something like:

```
"{down*B/s}" -> "1KB/s"
"{down*b/s}" -> "8Kb/s"
"{down*Bps}" -> "1KBps"
"{down*bps}" -> "8Kbps"
"{down*B}" -> "1KB"
"{down*b}" -> "8Kb"
"{down*_B}" -> "1"
"{down*_b}" -> "8"
```

Or is that too difficult?

MaxVerevkin:
`"[units_overrides]"` :laughing:?

GladOSkar:
> `"[units_overrides]"` laughing?

Oof

ammgws:
How about adding bits as a unit and then just allowing conversion between BitsPerSecond/BytesPerSecond and Bits/Bytes?

MaxVerevkin:
> How about adding bits as a unit and then just allowing conversion between BitsPerSecond/BytesPerSecond and Bits/Bytes?

It still would not give enough flexibility to, for example, switch between `b/s` and `Bi/s`, so I don't think it's a good idea.

If just hiding the unit with `_unit` doesn't work good enough ether, then I suppose, `[units_overrides]` doesn't look that bad.

ammgws:
Since we only have bits per second and not per minute/hour/etc then just remove them and have Bits and Bytes only. The "/s" part can be added in the format string. Especially since there are other ways people might want to write it, such as "Kbps" instead "Kb/s"

MaxVerevkin:
> Since we only have bits/bytes per second and not per minute/hour/etc then just remove them and have Bits and Bytes only. The "/s" part can be added in the format string. Especially since there are other ways people might want to write it, such as "Kbps" instead "Kb/s"

OK, at least it would look the same way it was before.

MaxVerevkin:
By the way, would it be possible to use `rustdoc` to generate documentation? This would make it much easier to keep it up to date.

ammgws:
If it makes things easier, I don't see why not. Something to look at in the future.

src/errors.rs
```diff
@@ -6,6 +6,7 @@ pub use self::Error::{BlockError, ConfigurationError, InternalError};
 /// Result type returned from functions that can have our `Error`s.
 pub type Result<T> = ::std::result::Result<T, Error>;
 
+// Why `ResultExtBlock` and `ResultExtInternal` are splitted?
```
MaxVerevkin, line 9: So is there any particular reason to split them?

MaxVerevkin: OK, I get why.

ammgws:
OK I've done as much testing as I can.

Excellent work on this!

GladOSkar:
Yeah honestly great work :) Thanks!

PR #1114, opened 2021-03-07
Add description and max_width to sound block

TilCreator:
This adds the usage of Pulse device descriptions (more human friendly names) (without breaking mapping).
Also the `output_name` now can have a `max_width`.

### DIFF
--- a/blocks.md
+++ b/blocks.md
@@ -1264,6 +1264,12 @@ step_width = 3
 ```toml
 [[block]]
 block = "sound"
+format = "{output_description} {volume}%"
+```
+
+```toml
+[[block]]
+block = "sound"
 format = "{output_name} {volume}"
 [block.mappings]
 "alsa_output.usb-Harman_Multimedia_JBL_Pebbles_1.0.0-00.analog-stereo" = "🔈"
@@ -1285,7 +1291,7 @@ Key | Values | Required | Default
 `on_click` | Shell command to run when the sound block is clicked. | No | None
 `show_volume_when_muted` | Show the volume even if it is currently muted. | No | `false`
 `max_width` | Max width of output\_name | No | `None`
-`use_description` | Pulseaudio provides some more human friendly names called description, it will replace the value of output\_name (mapping will still use the name) | No | `true`
+
 
 #### Available Format Keys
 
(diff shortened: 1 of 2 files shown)


blocks.md
```diff
@@ -1284,6 +1284,8 @@ Key | Values | Required | Default
 `max_vol` | Max volume in percent that can be set via scrolling. Note it can still be set above this value if changed by another application. | No | `None`
 `on_click` | Shell command to run when the sound block is clicked. | No | None
 `show_volume_when_muted` | Show the volume even if it is currently muted. | No | `false`
+`max_width` | Max width of output\_name | No | `None`
+`use_description` | Pulseaudio provides some more human friendly names called description, it will replace the value of output\_name (mapping will still use the name) | No | `true`
```
ammgws, line 1288: I'd prefer it as a format specifier instead: `{output_description}`

Then you can override the default format string and use it instead of `{output_name}`

blocks.md
```diff
@@ -1284,6 +1284,8 @@ Key | Values | Required | Default
 `max_vol` | Max volume in percent that can be set via scrolling. Note it can still be set above this value if changed by another application. | No | `None`
 `on_click` | Shell command to run when the sound block is clicked. | No | None
 `show_volume_when_muted` | Show the volume even if it is currently muted. | No | `false`
+`max_width` | Max width of output\_name | No | `None`
```
ammgws, line 1287: This could perhaps be covered by #1112

MaxVerevkin: Yes, I have plans to implement this.

ammgws: Awesome!

blocks.md
TilCreator: But that would not allow using name mapping and descriptions at the same time and description could be empty.

blocks.md
TilCreator: Nice

TilCreator: Should I remove this then or leave it to be removed when #1112 gets merged?

MaxVerevkin: To be honest, it would be better for me if you remove it, because it will make things a little simpler. However, I'm not sure when I'll finish that PR.

TilCreator: I think will leave for now. Is there anything specific one can help you on #1112? (If only merging to upstream when this will be merged before #1112 ^^)

MaxVerevkin: > Is there anything specific one can help you on #1112?

Not anything in particular yet, but, testing is appreciated.

MaxVerevkin: Implemented in https://github.com/greshake/i3status-rust/pull/1112/commits/5e3eec707bd76577abf402c26df3dd18ffe7a3dd

ammgws:
Thanks!

PR #1116, opened 2021-03-08
Added a block for tracking bitcoin value in multiple currencies

ThyW:
This PR adds Bitcoin tracking in multiple currencies(list available [here](https://www.blockchain.com/ticker)). 

Configuration looks like this:

```toml
[[block]]
block = "btc"
interval = 10 # (default is 60 seconds)
currency = "EUR" # (default is USD)
currency_list = ["EUR", "USD", "RUB", "CNY"] # (default is just USD)
```
To switch between the currencies, you can just scroll down on the text.

### DIFF
--- a/src/blocks/btc.rs
+++ b/src/blocks/btc.rs
@@ -23,9 +23,31 @@ pub struct Bitcoin {
     #[allow(dead_code)]
     tx_update_request: Sender<Task>,
 
-    // for some currency
+    /// use this for a list of supported currencies: https://blockchain.info/ticker
     currency: String,
+    currency_list: Vec<String>,
+    list_index: usize,
+}
 
+impl Bitcoin {
+    fn compute_index(&mut self) {
+        for (i, cur) in self.currency_list.iter().enumerate() {
+            if cur == &self.currency {
+                self.list_index = i
+            }
+        }
+    }
+
+    fn next_currency(&mut self) {
+        let mut index = self.list_index + 1;
+        let len = self.currency_list.len();
+        if index >= len {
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
```toml
[[block]]
block = "custom"
command = ''' echo 1B = $(curl -s https://blockchain.info/ticker | jq '.EUR.last')€ '''
```

Isn't this enough?

ammgws:
Thanks for your contribution! Unfortunately we prefer Custom blocks to be used where possible in order to reduce maintainer burden. As discussed in #715 it may be a good idea to have a "custom block example repository" that people could contribute to. Perhaps in the Wiki?

MaxVerevkin:
> Perhaps in the Wiki?

I think it is still impossible to contribute to the Wiki. So maybe `examples/custom`?

MaxVerevkin:
@ammgws If we do not accept this kind of blocks, functionality of which can be easily achieved using a `custom` block, then why do we have `uptime`???

ammgws:
See https://github.com/greshake/i3status-rust/pull/141

Also something like `uptime` is not likely to break compared to this which relies on an online service.

hjco:
Thanks @MaxVerevkin I added  bitcoin price custom block.

`[[block]]
block = "custom"
command = ''' echo  $(curl -s https://api.coindesk.com/v1/bpi/currentprice/usd.json | grep -o 'rate":"[^"]*' | cut -d\" -f3) USD '''
interval = 10`

2024-11-09-21:22:16-screenshot

Issue #991, opened 2020-12-22
networkmanager: slow DHCP offer, interface stays disconnected


damirsamardzic:
I'm working in dev environments with multiple network interfaces where DHCP servers are often responding with delay (dnsmasq with ping check as an example) - sometimes I see up to 5 seconds between NetworkManager state "prepare" to state "activated" on a specific network interface. In this case the networkmanager block will show the device as disconnected (×) even though it eventually received IP and is working properly. The status will not change unless I reload i3 (or another network event happens).

As far as I understand, the networkmanager block will be triggered by dbus events generated by NetworkManager components, but I couldn't figure out in which moment the block stops processing and the "activated" state doesn't get through. Maybe it would be useful to add polling mechanism next to the existing event driven? That would also help in situations where you manually add one or more temporary static IPs to an interface besides the DHCP offered one.

Maybe the issue will not happen in most of the network environments, but additional flexibility could be of great use for sure.
In any case, thanks for the great piece of software!

damirsamardzic:
Update:

While testing further, I see that the issue happens with wifi interfaces as well, maybe even more often. Looks like we get the state too soon and leave it like that. Does anyone else experience this issue?

My current workaround is to reload i3status-rs on click:
```
[[block]]
block = "networkmanager"
on_click = '''zsh -c "kill -SIGUSR1 $(pidof i3status-rs)"'''
```

ammgws:
It might be easiest to start debugging this by adding a print statement (`eprintln!("{:#?}", event)`) or similar here:

https://github.com/greshake/i3status-rust/blob/2b63bcd6c77f8d30155c1d41cac92b6a83f4d67e/src/blocks/networkmanager.rs#L563-L564

Then we can at least see what events are coming in. If we find an event that we should be dealing with but seemingly aren't, we can then start tracking it further elsewhere in the code.

damirsamardzic:
I'm sorry for late reply, here is config and log which shows events and missing IP info on eth0 (USB ethernet adapter) interface:
https://pastebin.com/BQLgFEN6

ammgws:
Hmm, could you try `busctl monitor org.freedesktop.NetworkManager` and see what signals there are when eth0 finally gets an IP?

damirsamardzic:
Here is the log file with busctl monitor running in parallel with i3status-rs:
https://www.dropbox.com/s/7jg30za7uehi5ga/i3status-rs_busctl.log?dl=0

[...]

PR #1119, opened 2021-03-11
Add additional match rule for active connections

### LINKED ISSUE #1117: NetworkManager block not ubdating non primary connections
I have noticed that the NetworkManager block doesn't update when my wifi connection changes state while being connected to a lan cable.
After a bit of digging I found that the NM block only subscribes to `PropertiesChanged` of `/org/freedesktop/NetworkManager` and not to every single connection.
https://github.com/greshake/i3status-rust/blob/094cfdc13953b31edda8914f5e536e52b93deb38/src/blocks/networkmanager.rs#L553-L556
This detects _almost_ every change, but it doesn't detect when a connection changes that is not the primary connection, because no property of `/org/freedesktop/NetworkManager` is changed.

This is a problem when for example: The computer has two internet connections, one lan and one wifi, the lan connection is the primary. Now when the wifi connection is disabled and re-enabled it changes the `connections` property of `/org/freedesktop/NetworkManager`, but it will be marked as disconnected because the connection is not established jet. The event of the wifi connection being established fully never reaches the block, because no property of `/org/freedesktop/NetworkManager` changes as a consequence.

One fix for this would be to subscribe to `PropertiesChanged` of every connection, but that is not super easy and maybe there is a better solution anyways. Maybe someone who knows dbus better than me knows of some was to subscribe to `/org/freedesktop/NetworkManager/connections/*` with one match rule or something even easier.
If I don't find a better solution I will just implement this one (in a few days).

TilCreator:
This fixes #1117 and #991. (For explanation see #1117)
Also as a nice side effect the wifi signal strength now updates.

**But** I was debugging a bit and noticed that the block now updates 14 times when a connection is disabled and 19 times when a new wifi connection is added that seams a _bit_ excessive (most of the time nothing changes between updates, because we listen for properties we don't display). Removing the rule for `/org/freedesktop/NetworkManager` helps, but it's still not great (7-10 updates) and we could miss events (again).
I have looked a bit into the dbus crate, but didn't manage to deconstruct or filter an event (jet)...

### DIFF
--- a/src/blocks/networkmanager.rs
+++ b/src/blocks/networkmanager.rs
@@ -550,12 +550,21 @@ impl ConfigBlock for NetworkManager {
             .name("networkmanager".into())
             .spawn(move || {
                 let c = Connection::get_private(BusType::System).unwrap();
-                let rule = "type='signal',\
-                        path='/org/freedesktop/NetworkManager',\
-                        interface='org.freedesktop.NetworkManager',\
-                        member='PropertiesChanged'";
 
-                c.add_match(&rule).unwrap();
+                c.add_match(
+                    "type='signal',\
+                    path='/org/freedesktop/NetworkManager',\
+                    interface='org.freedesktop.NetworkManager',\
+                    member='PropertiesChanged'",
+                )
+                .unwrap();
+                c.add_match(
+                    "type='signal',\
+                    path_namespace='/org/freedesktop/NetworkManager/ActiveConnection',\
+                    interface='org.freedesktop.NetworkManager.Connection.Active',\
+                    member='PropertiesChanged'",
+                )
+                .unwrap();
 
                 loop {
                     let timeout = 300_000;


ammgws:
@damirsamardzic does this solve your issue?

ammgws:
> I was debugging a bit and noticed that the block now updates 14 times when a connection is disabled and 19 times when a new wifi connection is added that seams a bit excessive.


On phone so can't check the code but pretty sure it's because the thread listening for the PropertiesChanged signal forces a block update without checking the signal message contents, and then inside the block update function we're making some redundant method calls to grab the information that we could have got from the signal message. (Or at least that's how some other blocks such as music used to operate before I updated them)

If that's the case then changing the block so it's more like how the music block does it should solve the issue.

Even if NetworkManager is spammy with updates, you could always check within the signal handler whether the block needs to be updated or not. 

This would likely require a large overhaul of the block, and there may be some other nuances I'm missing that may make it hard to implement.

ammgws:
On the other hand, if the update spam only happens on connection add/remove which realistically shouldn't happen that often (?), then it could be something to live with fit the time being.

TilCreator:
It's at least better than having an apparently broken wifi connection for hours ^^

damirsamardzic:
> @damirsamardzic does this solve your issue?

Yes, it seems it's working correctly now. Both wifi and usb eth interfaces eventually get updated and show correct status.
Thanks, @TilCreator !

ammgws:
Ok let's go with this for now.

Thanks!

PR #1126, opened 2021-03-21
`impl Default` for blocks' configs

MaxVerevkin:
`impl Default` for blocks' configs instead of setting default value for each field separately.

It is just a little cleaner, behavior should not have been changed.


ammgws:
Yeah this looks OK to me. Not sure why it wasn't done in the first place.

Issue #1124, opened 2021-03-20
Music block not updating although the song has changed. (ncspot)


epixinvites:
I have ncspot running, and somehow I'll have to like manually Alt+Shift+R (i3 keybind) to refresh the whole thing and then the song only updates.

ammgws:
Can you see whether there are any signals being sent when `ncspot` changes tracks?
`busctl --user monitor org.mpris.MediaPlayer2 --match 'member=PropertiesChanged'`

epixinvites:
Yes I do, but my block isn't responding as normal.

ammgws:
OK I'll see if I can test ncspot using spotify free.

epixinvites:
ncspot doesnt work with spotify free, only spotify premium tho

ammgws:
That's a shame then :(

OK so does the block update if you pause the song from ncspot?

epixinvites:
nope

ammgws:
Can you add these lines and see what output you get in the terminal?

https://github.com/greshake/i3status-rust/blob/03d4fdfcc043e0c61b268bc295fe554f42d602b3/src/blocks/music.rs#L389
```rust
eprintln!("{:#?}", player);
```

https://github.com/greshake/i3status-rust/blob/03d4fdfcc043e0c61b268bc295fe554f42d602b3/src/blocks/music.rs#L397
```rust
eprintln!("{:#?}", raw_metadata);
```

ammgws:
OK I so started a Premium trial and the block works fine?

Test config:

```toml
[[block]]
block = "music"
player = "ncspot"
```

Opened ncspot, went to playlist tab and started playing random songs and the block updates accordingly (I'm on `master`).

epixinvites:
well I did have that in my config and sometimes it works sometimes it doesn't.

ammgws:
I'm not sure what else I can do to help you since I can't seem to reproduce it. The info needed would be to add those debug print statements and/or also log the output `busctl --user monitor org.mpris.MediaPlayer2 --match 'member=PropertiesChanged'` and see what is (or isn't) happening when it "doesn't work".

Or otherwise figure out if there is a way to reliably reproduce the issue (preferably within the next couple of weeks while I still have Premium)

ammgws:
Are you sure the whole bar isn't freezing? Perhaps try running a separate instance of i3status-rust in a terminal with *only* this as a config:

```toml
[[block]]
block = "music"
player = "ncspot"
```

[...]

PR #1128, opened 2021-03-21
music block: merge two threads into one

MaxVerevkin:
Also: initialize `Player` struct when a new player is created.

### DIFF
--- a/src/blocks/music.rs
+++ b/src/blocks/music.rs
@@ -388,25 +388,25 @@ impl ConfigBlock for Music {
                             }
                         }
                         // Add/remove player
-                        else if let Ok((name, old_owner, new_owner)) = signal.read3::<&str, &str, &str>() {
-
-                            match (old_owner, new_owner) {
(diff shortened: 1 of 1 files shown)


ammgws:
Glad we can finally combine the two threads.

> solves some issues I've faced (https://github.com/greshake/i3status-rust/issues/1124#issuecomment-803599620)

Which issues does it solve?

Some initial testing with mpv: on play/pause the block is updating twice, presumably because the "updated = true" checks were removed.

MaxVerevkin:
> Which issues does it solve?

Sometimes a new player wouldn't work properly because when the block discovers it, it doesn't query the info and just waits for the properties to be changed.

> presumably because the "updated = true" checks were removed.

Ah yes, will add it.

ammgws:
> Sometimes a new player wouldn't work properly because when the block discovers it, it doesn't query the info and just waits for the properties to be changed.

Ah OK that makes sense.

ammgws:
Thanks, much better!

Issue #7, opened 2017-05-19
[Med Prio] Change architecture to support more output formats


greshake:
Because i3status-rust is already using an abstract representation for UI elements (widgets) it shouldn't be too much effore to abstract away from the i3bar protocol, to support other bars like dzen2.

svmnotn:
Would we do that as a feature flag, or as a cli option? This should probably be easy to do by altering the `print_blocks` function. Perhaps making a special trait that knows how to print blocks to the specific bar, and then just implementing the trait on an empty struct so that we can select the right impl at runtime; or just select the right impl at compile time.

greshake:
Hmm the first thing I think we need to do is research how other bar protocols work... If it's always just printing somehting to stdout, it should be easy... But even then, there is trouble. Currently, the widgets cache a rendered version of themselves, which is obviously dependent on the output format. There is some consideration and work to come if we are going to support other bars.

ammgws:
Doesn't look like there is any demand for this.

PR #1129, opened 2021-03-22
group protocol-related code together

MaxVerevkin:
Just move protocol-related code in `src/protocol/`

I know that `i3status-rust` probably will not support any backend other that i3bar, but still.


GladOSkar:
Wouldn't `print_blocks()` also qualify for moving to `protocol`? I always found it weird that such an essential function is in `util`...

ammgws:
Fair enough I guess.

For reference there was an issue open early on regarding other output formats: https://github.com/greshake/i3status-rust/issues/7

Issue #1069, opened 2021-02-17
net block: investigate using netlink/nl80211 instead of shelling out


ammgws:
As long as dependencies don't get too out of hand, this would be a step towards simplifying the block and for us to truly be "pure Rust" (or closer anyway by using C bindings).

- [ ] ip
- [ ] ipv6
- [x] bitrate (https://github.com/greshake/i3status-rust/commit/99062389e24e9026b25874514ace70c60f03634c)
- [x] ssid (completed with #1142)
- [x] frequency (completed with #1142)
- [x] signal strength (completed with #1142)

MaxVerevkin:
The most "popular" crate I found is https://github.com/Eonm/nl80211 and it works fine to get SSID and frequency. Also it is super easy to use (but I didn't figure out how to get signal strength).

Speed can still be calculated form `/sys/class/net/.../statistics/`, there is no problem with it.

https://github.com/jbaublitz/neli can be used to get IPs.

I don't know about bitrate.

ammgws:
https://github.com/doums/baru might be a good reference

MaxVerevkin:
> https://github.com/doums/baru

It displays my signal strength as "100%", but it should be ~50%.

Also found that https://github.com/Eonm/nl80211 in fact has a signal strength in "dBm", only need to convert it.

GladOSkar:
Yeah i honestly have no idea what the `%` numbers are supposed to say anyway, like `%` of what? It's probably some arbitrary threshold

MaxVerevkin:
https://www.intuitibits.com/2016/03/23/dbm-to-percent-conversion/

GladOSkar:
Interesting. As a summary: Yes, there are arbitrary choices (interval, perception curve) but rough common values are:

Logarithmic interval: `[ -100 dBm ; 0 dBm ]`
Perception Curve: Quadratic

MaxVerevkin:
Maybe https://github.com/torvalds/linux/blob/9ff9b0d392ea08090cd1780fb196f36dbb586529/drivers/net/wireless/intel/ipw2x00/ipw2200.c#L4322-L4334 ?

MaxVerevkin:
OK, nl80211 + that formula from some driver gives the same percentage as current block. I'll assume it's correct.

MaxVerevkin:
What I can tell so far: `netlink` is hard.

MaxVerevkin:
[...]

PR #1132, opened 2021-03-25
New wifi block

MaxVerevkin:
Relevant issue: #1069

This is a simple block that relies on https://github.com/Eonm/nl80211 to get wifi info. I decided to create a new block instead of modifying `net` since it's already bloated (750 SLOC) and I don't want to break it.

This block has four format keys: SSID, signal strength, interface name and wifi frequency.

### TODO
- [ ] Write docs
- [x] Add `{interface}` format key (for example "wlp3s0")
- [ ] Add new `wifi` and `no-wifi` icons. And maybe add `wifi-(1/2/3)` icons for different signal strength.
- [ ] Figure out which runtime dependencies it requires (if any)
- [x] Maybe change color (state) when disconnected?
- [ ] Add an option to hide when disconnected?
- [ ] Deprecate wifi-related options of `net` block and remove them in a later release?


MaxVerevkin:
For some reason [nl80211](https://github.com/Eonm/nl80211)'s `interface` has a trailing "\u0000". And for some reason `sway`'s parser fails to parse it :eyes:

MaxVerevkin:
Now I wonder if it's possible to disable wifi icon only for `net` block.

MaxVerevkin:
> Now I wonder if it's possible to disable wifi icon only for net block.

I'll just add a new icon "wifi" for the new block.

src/widgets/text.rs
```diff
@@ -69,9 +69,11 @@ impl TextWidget {
...
 
     pub fn set_text(&mut self, content: String) {
-        if content.is_empty() {
-            self.spacing = Spacing::Hidden;
-        }
+        self.spacing = if content.is_empty() {
+            Spacing::Hidden
+        } else {
+            Spacing::Normal
+        };
```
MaxVerevkin, line 76: If we already do a check if text is empty, why not check if it's not empty?

It doesn't break any of the blocks I use.

ammgws: Not sure if it is something block author's would actually use, but the only thing I can think of is if a block sets the spacing to something other than `Normal` while still displaying text - this change would change the spacing behind the block's back.

ammgws:
> I decided to create a new block instead of modifying net since it's already bloated (750 SLOC) and I don't want to break it.

To be honest, I would rather this was implemented within the net block (although I understand it will need much more work). From a quick glance, the "bloat" looks like it would mostly go away since a lot of the code could be removed (e.g. the code for parsing SSIDs which is around 250 SLOC by itself.)

- [x] signal_strength: nl80211
- [ ] bitrate: should be possible using netlink
- [ ] ip: should be possible using netlink
- [ ] ipv6: should be possible using netlink
- [x] speed_up: read from /sys/class/net/
- [x] speed_down: read from /sys/class/net/

src/widgets/text.rs
MaxVerevkin: Without this change, this config
```toml
[[block]]
block = "net"
device = "..."
format = "{speed_down;K}{speed_up;K}"
format_alt = ""
```
after two clicks has incorrect spacing (no space on the right).

> change the spacing behind the block's back.

It already does it :eyes:

MaxVerevkin:
> bitrate: should be possible using netlink

Does anyone know how?

MaxVerevkin:
I think there are reasons to split `net` and `wifi`. For example, my laptop can be connected to wifi and ethernet at the same time, and I want to know the speed of my active connection (ethernet) and the signal strength (wifi) at the same time. Currently it is not possible. However I understand your point to minimize the number of blocks.

ammgws:
> I want to know the speed of my active connection (ethernet) and the signal strength (wifi) at the same time.

Can't you just have two `net` blocks but with `device` set differently for each one?

ammgws:
No, but that case seems a bit contrived to me. Then again I'm 97% a wired desktop user so perhaps I'm missing out something?

MaxVerevkin:
> but that case seems a bit contrived to me

Fair enough

ammgws:
Sorry I meant the double wifi card thing.

MaxVerevkin:
> Sorry I meant the double wifi card thing.

No, it was just a hypothetical situation which I probably will never face.

PR #1135, opened 2021-03-25
Move pre-compiled themes and icons into files

### LINKED ISSUE #1109: Discussion: Move pre-compiled themes and icon sets to `/usr/share/i3status-rust`
With #1106 merged, this config is now correct:
```toml
[icons]
file = "material-nf"
[theme]
file = "slick"
```
given that `/usr/share/i3status-rust/themes/slick.toml` and `/usr/share/i3status-rust/icons/material-fn.toml` exist.

So my idea is that instead of having hard-coded icon sets and themes, we can provide them in `.toml` files. This makes `themes.rs` and `icons.rs` cleaner by splitting data and the actual code.

Any thoughts?

MaxVerevkin:
Closes #1109

Themes and icon sets should be somewhere in:
- `XDG_CONFIG_HOME/i3status-rust`
- `~/.local/share/i3status-rust`
- `/usr/share/i3status-rust`


MaxVerevkin:
Now tests fail because obviously in test environment there are no theme files.

MaxVerevkin:
Is it worth trying to make that test work again?

install.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+mkdir ~/.local/share/i3status-rust
```
ammgws, line 2: Might need to change it to `mkdir --parents` just in case

README.md
```diff
@@ -40,18 +40,17 @@ Stable releases are packaged on some distributions:
...
-$ cargo install --git https://github.com/greshake/i3status-rust i3status-rs
+$ git clone https://github.com/greshake/i3status-rust
+$ cd i3status-rust
+$ cargo install --path .
 ```
 
 By default, this will install the binary to `~/.cargo/bin/i3status-rs`.
 
-or manually:
+For functional themes and icon sets, you also have to copy files from `files/` to `~/.local/share/i3status-rust/`:
```
ammgws, line 50: Hmm, I wonder if we should just say to run `install.sh` to install the default themes/icons, otherwise some people may get confused and put their custom themes in `~/.local/share/i3status-rust/` as well?

MaxVerevkin: OK

src/config.rs
```diff
@@ -182,53 +182,53 @@ where
...
-                    "format = \"{1m}\"",
-                )
-                .as_ref(),
-            )
-            .unwrap();
-        let config: Result<Config, _> = deserialize_file(config_file_path.path());
-        config.unwrap();
-    }
-}
+//#[cfg(test)]
```
ammgws, line 185: I guess we can just remove this test

MaxVerevkin:
@ammgws can you restart github actions? For some reason it's trying to run a test for 15 minutes.

ammgws:
> running 1 test
> {"version": 1, "click_events": true}
> [
> [{"full_text":" Configuration error: failed to parse TOML from file contents.
> Cause: Theme 'plain' not found. for key `theme` at line 4 column 1 ","border":"#ff0000","name":"0","instance":"0","separator":false,"separator_block_width":0,"markup":"pango"}],
> test run_binary::debug_build ... test run_binary::debug_build has been running for over 60 seconds

MaxVerevkin:
What did that test test anyway? Undocumented `--one-shot` option?

MaxVerevkin:
I can change the theme in that test to `native` (doesn't require a file), but what's the point of that test?

ammgws:
Yeah and I don't think we need it anymore

MaxVerevkin:
Those who are still running :eyes: https://github.com/greshake/i3status-rust/actions?query=is%3Ain_progress

ammgws:
Ok let's see how this goes.

Thanks for your work!

ammgws (approved):

PR #1140, opened 2021-03-27
Allow reloading with usr2 signal after displaying error

MaxVerevkin:
No more restarting `sway` after making a typo in the config!


ammgws:
Thanks!

PR #1152, opened 2021-04-03
Cpu boost stat

indlin:
Read the cpu turbo boost status. Add the new format key {boost} which will output "⏼" if the boost on, or "◌" if off. It's a quick and easy addition, helpful if you use a hot intel cpu and sometime turn off a turbo boost.


src/blocks/cpu.rs
```diff
@@ -206,3 +213,18 @@ impl Block for Cpu {
...
+/// Read the cpu turbo boost status from kernel sys interface
+/// or intel pstate interface
+fn boost_status() -> Result<bool> {
+    if let Ok(boost) = fs::read_to_string("/sys/devices/system/cpu/cpufreq/boost") {
+        return Ok(boost.starts_with("1"));
+    } else if let Ok(no_turbo) = fs::read_to_string("/sys/devices/system/cpu/intel_pstate/no_turbo")
+    {
+        return Ok(no_turbo.starts_with("0"));
+    }
+    Err(BlockError(
```
MaxVerevkin, line 226: Note that you can use `block_error()' function on `Resut` type instead of manually creating `BlockError'. The internals if `BlockError` will change at some point to include addition information.

indlin: Yes, but I think, may be remove error at all? Because if boost_status return Err, that will break cpu block, even if no {boost} template used... Just return bool?

MaxVerevkin: Good point. Or perhaps `Option<bool>` and default to `""`?

indlin: Done

src/blocks/cpu.rs
```diff
@@ -170,10 +170,18 @@ impl Block for Cpu {
             barchart.push(BOXCHARS[(7.5 * utilization) as usize]);
         }
 
+        let boost = match boost_status() {
+            Some(true) => "⏼",
+            Some(false) => "◌",
```
MaxVerevkin, line 175: I think it would make sense to add those icons into icon sets.

indlin: It was a little more complicated :) In all sets I add the same icons, they work in all fonts.

MaxVerevkin: > In all sets I add the same icons, they work in all fonts

Are you sure though? Maybe it would be better to use different icons? You can try searching for "power" or "toggle" in https://www.nerdfonts.com/cheat-sheet

src/blocks/cpu.rs
```diff
@@ -71,13 +73,15 @@ impl ConfigBlock for Cpu {
         Ok(Cpu {
             id,
             update_interval: block_config.interval,
-            output: TextWidget::new(id, 0, shared_config).with_icon("cpu")?,
+            output: TextWidget::new(id, 0, shared_config.clone()).with_icon("cpu")?,
```
MaxVerevkin, line 76: I think you can avoid using `clone()` by moving this line under `boost_icon_off`.

files/icons/awesome.toml
```diff
@@ -81,3 +81,5 @@ weather_snow = "\uf2dc" # fa-snowflake-o
 weather_sun = "\uf185" # fa-sun-o
 weather_thunder = "\uf0e7" # fa-bolt
 xrandr = "\uf26c" # fa-television
+cpu_boost_on = "\u23fc"
```
ammgws, line 84: So which icon set is this actually from?

src/blocks/cpu.rs
```diff
@@ -1,4 +1,4 @@
-use std::fs::File;
+use std::fs::{self, File};
```
ammgws, line 1: Just import `read_to_string` here instead of `self`

src/blocks/cpu.rs
```diff
@@ -206,3 +217,15 @@ impl Block for Cpu {
         self.id
     }
 }
+
+/// Read the cpu turbo boost status from kernel sys interface
+/// or intel pstate interface
+fn boost_status() -> Option<bool> {
+    if let Ok(boost) = fs::read_to_string("/sys/devices/system/cpu/cpufreq/boost") {
```
ammgws, line 224: and remove the "fs::" here

src/blocks/cpu.rs
```diff
@@ -206,3 +217,15 @@ impl Block for Cpu {
...
+        return Some(boost.starts_with("1"));
+    } else if let Ok(no_turbo) = fs::read_to_string("/sys/devices/system/cpu/intel_pstate/no_turbo")
```
ammgws, line 226: and also remove the "fs::" here

ammgws (changes requested):

files/icons/awesome.toml
indlin: to be honest, I don't know. I just search symbol in unicode table (https://unicode-table.com/en/23FC/)...

src/blocks/cpu.rs
indlin: Yeah... If You think so. At beginning I just use read_to_string, but for readability, I thought the fs:: prefix more helpful. You immediately see where this function from and what it do - read string from file system.

src/blocks/cpu.rs
indlin: Cool! You are right. I thought, fields must be initialized in the declaration order.

src/blocks/cpu.rs
indlin: Sorry, busy week... I will try it later

ammgws:
I'll fix the icon stuff after merging. Thanks!

PR #1157, opened 2021-04-05
Allow custom blocks with JSON not to set an icon

### LINKED ISSUE #1156: Error with custom block json with no icon
```
Internal error in context 'get_icon()': icon '' not found: please check your icon file or open a new issue on GitHub if you use a precompiled icons.
```

icons = "awesome"

The json I emitted was:
```
{"state":"Critical", "text":" Reboot Required"}
```

GladOSkar:
Resolves #1156

Regressed in #1135, feedback welcome

### DIFF
--- a/src/blocks/custom.rs
+++ b/src/blocks/custom.rs
@@ -161,7 +161,11 @@ impl Block for Custom {
             let output: Output = serde_json::from_str(&*raw_output).map_err(|e| {
                 BlockError("custom".to_string(), format!("Error parsing JSON: {}", e))
             })?;
-            self.output.set_icon(&output.icon)?;
+            if output.icon.is_empty() {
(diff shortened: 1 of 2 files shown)


MaxVerevkin:
One issue: If the custom script is dynamic, then there is no way to disable the icon once it is already shown.

GladOSkar:
Ah yes i see. I don't think there's a way to explicity unset an icon programmatically yet, right?

MaxVerevkin:
Probably have to add a new function to the `TextWidget`.

GladOSkar:
Okay this should work now. Not sure if this is the proper way to do it, maybe this could be done easier during deserialization of the json but i have no idea how

ammgws:
Thanks!

PR #1161, opened 2021-04-06
Fix apt block

### LINKED ISSUE #1160: Apt block broken
On `759602d8554`

When the apt block is enabled, it throws an error:
```
[[block]]
block = "apt"
interval = 1800
```

```
Internal error in context 'util': Unknown placeholder in format string: count
```

MaxVerevkin:
Closes #1160


ammgws:
Thanks! 

A quick grep shows that there aren't any other ones left.

Issue #1163, opened 2021-04-07
Volume block when muted has extra brace


cameronswinoga:
On 3d3f165f7bd

See how on mute the volume icon has a closing brace next to it:

![volume_muted](https://user-images.githubusercontent.com/1660815/113904352-fcb13c00-979f-11eb-8fb8-6bbf071ddf91.jpg)

Doesn't show when not muted:

![volume_not_muted](https://user-images.githubusercontent.com/1660815/113904822-7a754780-97a0-11eb-9e64-8e8b4ef1fc03.jpg)

PR #1167, opened 2021-04-08
Fix some icons (related with #1163)

szaffarano:


GladOSkar:
Oops i guess i missed another one

Issue #1097, opened 2021-02-25
weather block: add more icons


ammgws:
Currently we only support clear, rain, clouds, thunderstorm and snow: 
https://github.com/greshake/i3status-rust/blob/d39cd39977bd056d13e165ee52c010ffb5c779ad/src/blocks/weather.rs#L258-L265
https://github.com/greshake/i3status-rust/blob/d39cd39977bd056d13e165ee52c010ffb5c779ad/src/icons.rs#L148-L153

We are grabbing the weather state from the JSON `weather.main` field which has a limited set of states. If we instead grab the `weather.id` field then we have much more detail: https://openweathermap.org/weather-conditions

Meanwhile, Font Awesome has some icons we are not using yet, such as "clouds-shower-heavy", which we could then use if we made the change mentioned above: https://fontawesome.com/icons?d=gallery&p=2&c=weather&m=free

PR #1169, opened 2021-04-09
weather: support config language

halfcrazy:
localized weather in the description field

default
```toml
[[block]]
block = "weather"
format = "{weather} {weather_verbose}"
service = { name = "openweathermap", api_key = "", city_id = "1799962", units = "metric"  }
```
```json
[{"full_text":" WEATHER \"Clouds\" \"overcast clouds\" ","name":"0","instance":"0","markup":"pango"}]
```

zh
```toml
[[block]]
block = "weather"
format = "{weather} {weather_verbose}"
service = { name = "openweathermap", api_key = "", city_id = "1799962", units = "metric" , lang = "zh"}
```
```json
[{"full_text":" WEATHER \"Clouds\" \"阴，多云\" ","name":"0","instance":"0","markup":"pango"}]
```

### DIFF
--- a/doc/blocks.md
+++ b/doc/blocks.md
@@ -1639,6 +1639,7 @@ Key | Values | Required | Default
 `place` | OpenWeatherMap 'By city name' search query. See [here](https://openweathermap.org/current) | Yes* | None
 `coordinates` | GPS latitude longitude coordinates as a tuple, example: `["39.236229089090216","9.331730718685696"]`
 `units` | Either `metric` or `imperial`. | Yes | `metric`
+`lang` | Language code. See [here](https://openweathermap.org/current#multi). Currently only affects `weather_verbose` key. | No | `en`
 
 One of `city_id`, `place` or `coordinates` is required. If more than one are supplied, `city_id` takes precedence over `place` which takes place over `coordinates`.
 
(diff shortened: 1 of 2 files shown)


ammgws:
Perhaps need to document the language codes, or link to somewhere on the OMW site that has them?

halfcrazy:
> Perhaps need to document the language codes, or link to somewhere on the OMW site that has them?

Updated. Add the website link.

ammgws:
What language are you using this with?

halfcrazy:
> What language are you using this with?
zh_cn

ammgws:
Oh ok, because for english this results in the descriptions being displayed in lower case, and also the descriptions are now more verbose (which may be good or bad depending on how you look at it). Don't think there's a trivial way to capitalise the strings either, considering we could be dealing with non-ASCII characters.

halfcrazy:
Yes, the description filed is more verbose and maybe needs more width.
From [doc](https://openweathermap.org/weather-conditions) the longest description "thunderstorm with heavy drizzle" may a little bit longer.
Maybe we can set a max length for the description if exceed we prefer to use the main field.

halfcrazy:
If more icons were provided the weather field may not such useful. #1097

ammgws:
Perhaps. IIRC I only checked for extra icons in FontAwesome but still needed to see how the Material Design compared to it.

halfcrazy:
Emm, how about introducing a new template var `weather_verbose` it takes value from the description and lets `weather` remain.

ammgws:
That could work too

halfcrazy:
> That could work too

update

ammgws:
Just need to clarify in the docs that lang only applies to weather_verbose

halfcrazy:
done

ammgws:
Thanks! Glad it worked out in the end.

PR #1170, opened 2021-04-09
widgets/rotatingtext: never provide `min_width` for dynamic width

remi-dupre:
This field was only filled when not necessary (ie. when `text_width >= max_width`) with `dynamic_width` set to `true` but would potentially add a padding when not using a fixed sized font.

I personally encountered this issue while using the music block.


ammgws:
Long time no see ;)

Thanks!

PR #1171, opened 2021-04-09
music: set state to idle when the music is not playing

remi-dupre:
I think it makes more sense like this, the block should be as little intrusive as it can be while no music plays.

"Not playing" states are "Paused", "Stopped" and "Unknown".

### DIFF
--- a/src/blocks/music.rs
+++ b/src/blocks/music.rs
@@ -569,6 +569,18 @@ impl Block for Music {
             }
         }
 
+        let state = match metadata.playback_status {
+            PlaybackStatus::Playing => State::Info,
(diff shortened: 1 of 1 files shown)


GladOSkar:
I absolutely agree, the `info` coloring is unnecessarily distracting

ammgws:
That makes sense. 

Thanks!

MaxVerevkin:
I think it would also make sense to set the state it idle when no player is available:

https://github.com/greshake/i3status-rust/blob/a31158f9bc191c48c16ffd23fe15b90500190f0f/src/blocks/music.rs#L530-L537

PR #1174, opened 2021-04-09
Introduce `FormatTemplate::contains()`; fix net block

### LINKED ISSUE #1173: Net Block: ip gets not updated if not in `format` string
```sh
$ i3status-rs --version
i3status-rs 0.20.0 (commit 4804ebe 2021-04-09)
```

I would like to use the `net`-block with the following configuration:

```toml
[[block]]
block = "net"
format_alt = "{ip}"
```

So normally showing the default but on click can show the current IP.

I looked at the source and pinpointed the error in this line (I think):
https://github.com/greshake/i3status-rust/blob/4804ebe6d464fa860897d153553ff66eb3aa188b/src/blocks/net.rs#L446-L460

While I thought this should be easy enough even without rust-knowledge, it turns out that it takes more time than I have at the moment and therefore I was unable to create a PR. 

tl;dr checks on `ip` are disabled if they are not in the `format`. If they are only in the alternative `format_alt` the check gets entirely deactivated.

The same happens with `bitrate` and `ipv6`.

MaxVerevkin:
Resolves #1173

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -419,6 +419,7 @@ impl ConfigBlock for Net {
         let wireless = device.is_wireless();
         let vpn = device.is_vpn();
 
+        let format = FormatTemplate::from_string(&block_config.format)?;
         let format_alt = if let Some(f) = block_config.format_alt {
             Some(FormatTemplate::from_string(&f)?)
         } else {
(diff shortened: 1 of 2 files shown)


ammgws:
Thanks for the quick fix!

PR #1177, opened 2021-04-10
Replace FontAwesome5 disk_drive icon to be free

farwayer:
The icon currently provided available in Pro paid version only
https://fontawesome.com/icons?d=gallery&p=2&q=uf8b5


ammgws:
Huh, never noticed that. Thanks!

PR #1178, opened 2021-04-10
Net block: do not throw an error if frequency/signal is unavailable

MaxVerevkin:
Relevant comment https://github.com/greshake/i3status-rust/issues/1069#issuecomment-817090540


ammgws:
Thanks!

PR #1179, opened 2021-04-10
music block: escape pango

MaxVerevkin:
Some titles contain `&` and other symbols that should be escaped.


ammgws:
Cheers, surprised I haven't come across it yet

PR #1186, opened 2021-04-16
formatting: better unicode support

MaxVerevkin:
Count chars, not bytes.


ammgws:
Did you run into some unexpected behaviour?

MaxVerevkin:
> Did you run into some unexpected behaviour?

Yes, then placeholder contains non ascii text (russian in my case), min/max width works incorrectly.

PR #1187, opened 2021-04-17
disk_space: fix alert_absolute

### LINKED ISSUE #1182: alert_absolute in disk space is not working correctly, when true
I build latest master and use this config
```
[[block]]
block = "disk_space"
alert = 10.0
alert_absolute = true
format = "{path^2}{available;G}"
info_type = "available"
interval = 300
path = "/data"
unit = "GB"
warning = 50.0
```
Currently I have 34 GB available, but block stay the same color as it was before. If I change `alert_absolute` to false and use correct percentage for warning, the block styled as it should be styled for warning level. In my case I use dark-gruvbox theme and warning color is yellow

MaxVerevkin:
Closes #1182

### DIFF
--- a/src/blocks/disk_space.rs
+++ b/src/blocks/disk_space.rs
@@ -178,20 +178,20 @@ impl Block for DiskSpace {
         let alert_type;
         match self.info_type {
             InfoType::Available => {
-                result = available;
+                result = available as f64;
                 alert_type = AlertType::Below;
             }
             InfoType::Free => {
-                result = free;
+                result = free as f64;
                 alert_type = AlertType::Below;
             }
             InfoType::Used => {
-                result = used;
+                result = used as f64;
                 alert_type = AlertType::Above;
             }
         }
 
-        let percentage = (result as f64) / (total as f64) * 100.;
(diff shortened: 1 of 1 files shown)


src/blocks/disk_space.rs
```diff
@@ -208,10 +208,10 @@ impl Block for DiskSpace {
         // Send percentage to alert check if we don't want absolute alerts
         let alert_val = if self.alert_absolute {
             (match self.unit {
-                Prefix::Tera => result << 40,
-                Prefix::Giga => result << 30,
-                Prefix::Mega => result << 20,
-                Prefix::Kilo => result << 10,
+                Prefix::Tera => result >> 40,
```
ammgws, line 211: Maybe it's better to leave `result` as bytes and shift the warning/alert values to bytes instead. For example, with this way a `result` of 4.6GB ends up becoming just `4`

MaxVerevkin: You are right. Or just convert to float and divide.

ammgws: Yeah either way

ammgws:
Thanks!

Issue #1036, opened 2021-02-10
Add signal support for all blocks


ammgws:
Right now only the Custom block supports setting an update signal, however other blocks should be able to use it too as mentioned in https://github.com/greshake/i3status-rust/pull/822 (and there are requests for that e.g. https://github.com/greshake/i3status-rust/issues/1015).

GladOSkar:
IIUC #1015 is click related and not about signals?

ammgws:
Related in this way: adding both on_click_right support and signal support would achieve their use case

MaxVerevkin:
But what are the use cases for sending signals to blocks (except for custom)?

ammgws:
To force updates early. For example the pacman block could have an interval of 1 hour but if you send a signal you could trigger an update without waiting.

MaxVerevkin:
Well, I still don't get it. Is writing `pkill -SIGRTMIN+x i3status-rs` simpler that just `checkupdates`? I know, I can set a keybinding to send a signal, but still.

It is already possible to update every block with `USR1` signal, right?

ammgws:
I originally thought of it as a way to solve https://github.com/greshake/i3status-rust/issues/1015 without introducing any block-specific stuff. In that case, the `on_click` could be set to run `pkill -SIGRTMIN+x i3status-rs` and the `signal` set appropriately to catch it. Another way without using `signal` might be to overhaul `on_click` and give it some special syntax so you can choose to force update the block instead of running a custom command, but the `signal` way seemed the easier way to go (yet could be seen as a bit hackish)

So if it can be generically implemented for every block like `on_click` then I don't see the harm in `signal` also having the same treatment. Depending on the config, reloading every block in the bar can take some time so I can see the point of being able to pinpoint force update a block. So basically the argument is, if it can be implemented fairly trivially then it seems like a good way to allow power users finer control over their i3status-rs bar.

MaxVerevkin:
> and give it some special syntax so you can choose to force update the block instead of running a custom command

IMO, every block should always update when clicked, even if `on_click` is set.

ammgws:
Interesting idea, though I could imagine misclicking on a long blocking block might be annoying (though not so bad if it was async?)

Also, would that just be for left clicks, or any button click? Ignore scrolling?

MaxVerevkin:
> Also, would that just be for left clicks, or any button click? Ignore scrolling?

I'm not sure.

[...]

PR #1190, opened 2021-04-21
Block update changed to right click

ram02z:
Initially, I just wanted to change the apt block to use right click to update instead of left because left click is used for the `on_click` property. However, I realised there are a few blocks that also have some sort of update function when left click is clicked, so I updated them accordingly.

In a nutshell, I just wanted to be able to have an extra bind (left click), whilst retaining the block's existing update functionality. This is particularly useful for blocks that tend to have long intervals like the apt block.

### DIFF
--- a/src/blocks/apt.rs
+++ b/src/blocks/apt.rs
@@ -223,7 +223,7 @@ impl Block for Apt {
     }
 
     fn click(&mut self, event: &I3BarEvent) -> Result<()> {
-        if event.button == MouseButton::Left {
+        if event.button == MouseButton::Right {
             self.update()?;
         }
         Ok(())
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -634,7 +634,7 @@ impl Block for Net {
     }
 
     fn click(&mut self, event: &I3BarEvent) -> Result<()> {
-        if event.button == MouseButton::Left {
+        if event.button == MouseButton::Right {
             if let Some(ref mut format) = self.format_alt {
                 std::mem::swap(format, &mut self.format);
             }
--- a/src/blocks/notify.rs
+++ b/src/blocks/notify.rs
@@ -144,7 +144,7 @@ impl Block for Notify {
     }
 
     fn click(&mut self, e: &I3BarEvent) -> Result<()> {
-        if let MouseButton::Left = e.button {
+        if let MouseButton::Right = e.button {
             let c = Connection::get_private(BusType::Session).block_error(
                 "notify",
                 &"Failed to establish D-Bus connection".to_string(),
--- a/src/blocks/notmuch.rs
+++ b/src/blocks/notmuch.rs
@@ -140,7 +140,7 @@ impl Block for Notmuch {
     }
 
     fn click(&mut self, event: &I3BarEvent) -> Result<()> {
-        if event.button == MouseButton::Left {
+        if event.button == MouseButton::Right {
             self.update()?;
         }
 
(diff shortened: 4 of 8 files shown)


MaxVerevkin:
I don't believe that's the right thing to do.

> > and give it some special syntax so you can choose to force update the block instead of running a custom command
> 
> IMO, every block should always update when clicked, even if `on_click` is set.

Would something like that work for you (from #1036)?

ram02z:
> I don't believe that's the right thing to do.
> 
> > > and give it some special syntax so you can choose to force update the block instead of running a custom command
> > 
> > IMO, every block should always update when clicked, even if `on_click` is set.
> 
> Would something like that work for you (from #1036)?

Yeah, that's probably a better method because I just want to retain the update functionality of some blocks.

MaxVerevkin:
@ammgws what do you think?

ammgws:
> @MaxVerevkin: IMO, every block should always update when clicked, even if on_click is set.

That is what [py3status](https://py3status.readthedocs.io/en/latest/configuration.html#custom-click-events) (and possibly other ones) does so it may be the right way to go: "As an added feature and in order to get your i3bar more responsive, every on_click command will also trigger a module refresh". However, do you think we would need to allow that behaviour to be disabled as well? For example if you have set a custom `on_click` command for the Pacman block that pops up the list of pending updates, but you don't want it to trigger the block update function when you click on it.

ram02z:
> For example if you have set a custom `on_click` command for the Pacman block that pops up the list of pending updates, but you don't want it to trigger the block update function when you click on it.

Speaking of list of pending updates, how would you go about getting the list of pending updates from apt's temporary package database? Currently I have to run `sudo apt update && apt list --upgradable` which takes a while to update.

MaxVerevkin:
> However, do you think we would need to allow that behaviour to be disabled as well?

```toml
[[block]]
block = "pacman"
on_click = "..."
update_on_click = false # defaults to true
```

ammgws:
> @ram02z: Speaking of list of pending updates, how would you go about getting the list of pending updates from apt's temporary package database? Currently I have to run `sudo apt update && apt list --upgradable` which takes a while to update.

I will see if i I can look into it (not my daily driver so will need to grab another machine)

ram02z:
> @ram02z, this should do it: `APT_CONFIG=/tmp/i3rs-apt/apt.conf apt list --upgradable`

Thank you!

ammgws:
I'll add it to the block documentation. If you come up with an `on_click` command that pops up a menu with the list of updates or something, please open a PR to add the example to the docs :+1:

ram02z:
> I'll add it to the block documentation. If you come up with an `on_click` command that pops up a menu with the list of updates or something, please open a PR to add the example to the docs +1

Sure! I currently use [i3-quickterm](https://github.com/lbonn/i3-quickterm) in order to show a popup terminal with the updates.

MaxVerevkin:
> > However, do you think we would need to allow that behaviour to be disabled as well?
> 
> ```toml
> [[block]]
> block = "pacman"
> on_click = "..."
> update_on_click = false # defaults to true
> ```

Or perhaps something like that? (not sure if this is a valid toml)

```toml
[[block]]
block = "pacman"
[[block.click]]
button = "left"
cmd = "..."
update = false # defaults to true
[[block.click]]
button = "right"
cmd = "..."
```

ram02z:
> I'll add it to the block documentation. If you come up with an `on_click` command that pops up a menu with the list of updates or something, please open a PR to add the example to the docs +1

Created the example for apt PR.

Also I think a separate PR should be created about the ideas discussed here, since this PR is really just a quickfix.

PR #1195, opened 2021-04-25
Remove quotes from weather block

### LINKED ISSUE #1191: QUESTION: Should the Weather block be removing quotes?
The wiki well documents escaping in custom blocks; is it possible to perform (custom) replacements in non-custom blocks? 

The weather block (openweather) returns the description as a quoted string for me, and the quotes are retained by the block s.t. the output is something like:

```
{"full_text":"  \"Clouds\" 42° ","color":"#1ebce8","background":"#000000","name":"1","instance":"0","separator":false,"separator_block_width":0,"markup":"pango"}
```

I'd like to remove the quotes around "Clouds".  I'm currently doing it in a trivial script that just does `i3status-rs | sed 's/\\"//g'`, but it'd be nice to be able to do a regexp replacement, or just have the Weather block remove quotes.

Because it seems odd that I'd be the only person to notice this, I wonder if I have something set incorrectly, or if I'm missing a way of transforming the string in a non-custom widget.

i3status-rust is from 2021-03-26 (d0caebc2d0a7b2608746bcfaf87a6bfc1edd6a32).

grvxs:
Fixes #1191

### DIFF
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -231,6 +231,7 @@ impl Weather {
 
                 let raw_weather = json
                     .pointer("/weather/0/main")
+                    .and_then(|v| v.as_str())
                     .ok_or_else(malformed_json_error)?
                     .to_string();
 


ammgws:
Thanks!

Issue #1205, opened 2021-05-01
Hide block when there's no place on a bar


ivankovnatsky:
The use case: on an external big screen I can easily fit all my blocks, though when I'm my laptop it surely hides those blocks on the left.

Would be cool to be able to flag blocks that are less important and hide them when there's no free space on a bar.

Please point me, if I missed a way to accomplish this or similar behavior. 

```
[[block]]
block = "memory"
format_mem = "{mem_used}/{mem_total}({mem_used_percents})"
format_swap = "{swap_used}/{swap_total}({swap_used_percents})"
display_type = "memory"
hide_on_bar_full = true
```

MaxVerevkin:
Relevant issue: #676 
Relevant PR: #703

GladOSkar:
AFAIK this is not reasonably possible for us to do until `i3/i3#4113` is implemented, which, knowing i3's development cycle will still take a while.

Until then i think we can only switch all blocks to short mode at once which might allow for some workarounds (make the short mode for all the "important" blocks the same as their long mode to only shorten non-important ones?)

GladOSkar:
Actually nvm that feature request is suggesting something a bit different from what you're asking. But still, this definitely requires some interaction with the DE/WM not currently possible with i3 without workarounds

GladOSkar:
Also, with basically all blocks having been reworked to use format strings and a unified click interface being in the works (or done, not sure?) it's probably worth revisiting #676 (and #703) since this might have become easier to do

GladOSkar:
Okay i just took a quick look at trying to implement this and the main problem i ran into was the following: `i3status-rust` has no information about whether a block will be short or not and we have to pass a single big list of widgets to `i3`, some of which should now get a `short_text` attribute.

But `format_short` would be specified on a block level. Since a block contains a series of widgets built from the format string, it's really hard to build one widget list from 2 format strings, since we can't really correlate widgets of the format strings together between the 2.

So another idea would be integrating this into the format string syntax as a modifier for the placeholders (like `{speed_down?}` means can be hidden). But that would probably limit the functionality quite a bit because it can't be used to, say, shorten the decimal precision or hide text when becoming short.

[...]

Issue #1038, opened 2021-02-10
Implement on_click-like options for all other mouse buttons (applicable to all blocks)


ammgws:
#1006 implemented this for `on_click`, but it would be nice to allow the same for other mouse buttons.

Issues:
- what to name the options... `on_click_right`, `on_click_middle`, `on_scroll_up`, `on_scroll_down`? or some other way is better?

- can we do anything about blocks with multiple widgets? Otherwise just override for the whole block

MaxVerevkin:
Maybe delay this feature for another release?

Implementing this feature would require some rewrites of `base_block`, but in the future I hope we will be able to move to `async`, with different click handling mechanism anyway.

Doing this work twice feels like a waste of time.

ammgws:
I figured there may be some breaking changes due to this so thought 0.2 would be good timing since all the breaking changes could be bundled into the same release.

Are @remi-dupre and yourself continuing work/discussion on the async stuff?

MaxVerevkin:
> breaking changes

There shouldn't be any breaking changes in introducing new options.

> Are @remi-dupre and yourself continuing work/discussion on the async stuff?

Yes. All the blocks that are already ported work pretty well.

ammgws:
Oh, I meant in the case were the `on_click` functionality was enhanced to allow it to force an update of the block instead of running a command (in order to cover use cases such as https://github.com/greshake/i3status-rust/issues/1015). I imagined that might need some changes to the current config syntax. Unless it can be done another way without breaking configs then it could be left for later.

ammgws:
I'll remove it from the v0.2 milestone since it looks like there won't be any breaking changes

MaxVerevkin:
> what to name the options... on_click_right, on_click_middle, on_scroll_up, on_scroll_down? or some other way is better?

I think that the syntax below is the most flexible solution:
```toml
[[block]]
block = "pacman"
[[block.click]]
button = "left"
cmd = "..."
update = false # defaults to true
[[block.click]]
button = "right"
cmd = "..."
```
(I've implemented it in `swaystaus`, so someone can port is while I'm working on the async stuff)

ammgws:
Looks good!

asdf8dfafjk:
Speaking of breaking changes, perhaps consider dropping TOML? https://lobste.rs/s/dn91bz/why_broot_is_switching_from_toml_hjson_for 

[...]

PR #1207, opened 2021-05-04
Add short_text support

GladOSkar:
This is mostly a port from `MaxVerevkin/swaystatus@b26224d` and based on the discussions from #1205.

It works quite well for me.

Things that could be improved are:

- [x] ~~The `[block.format]` section *requires* the `full` field to be set and i couldn't figure out how to make `serde` fall back to the default if it's not given.~~
- [x] ~~The `pacman` block doesn't support the `[block.format]` section yet because it has some weird internal logic that requires direct access to the format string (which can't be accessed from the `FormatTemplate`). Possible fixes include either adding the format string to the `FormatTemplate` or reworking the way `pacman` figures out which package managers to monitor.~~
- `RotatingTextWidget`s can't be shortened from the format string, the interface doesn't really facilitate that. This only impacts the `music` block afaik. Something for a future PR imo.

Feedback + testing appreciated :)

### DIFF
--- a/doc/blocks.md
+++ b/doc/blocks.md
@@ -1724,6 +1724,18 @@ command = "echo '<b>1 &amp;</b>'"
 
 # Formatting
 
+All blocks that have a `format` field can be reformatted by changing their format strings.
+
+The field can be set as a string (`format = "{my_format}"`) or as a section:
+```toml
+[[block]]
+block = "my_block"
+[block.format]
+full = "{my_full_format}"
+short = "{my_short_format}"
+```
+Your `i3` or `sway` will switch all blocks over to the `short` variant whenever there isn't enough space on your screen for the `full` status bar.
+
 ## Syntax
 
 The syntax for placeholders is
--- a/src/blocks/bluetooth.rs
+++ b/src/blocks/bluetooth.rs
@@ -231,26 +231,12 @@ pub struct BluetoothConfig {
     //DEPRECATED
     //TODO remove
     pub label: Option<String>,
-    #[serde(default = "BluetoothConfig::default_hide_disconnected")]
+    #[serde(default)]
     pub hide_disconnected: bool,
-    #[serde(default = "BluetoothConfig::default_format")]
-    pub format: String,
-    #[serde(default = "BluetoothConfig::default_format_unavailable")]
-    pub format_unavailable: String,
-}
-
-impl BluetoothConfig {
-    fn default_hide_disconnected() -> bool {
-        false
-    }
-
-    fn default_format() -> String {
-        "{label} {percentage}".into()
-    }
-
-    fn default_format_unavailable() -> String {
-        "{label} x".into()
-    }
+    #[serde(default)]
+    pub format: FormatTemplate,
+    #[serde(default)]
+    pub format_unavailable: FormatTemplate,
 }
 
 impl ConfigBlock for Bluetooth {
@@ -276,8 +262,8 @@ impl ConfigBlock for Bluetooth {
             })?,
             device,
             hide_disconnected: block_config.hide_disconnected,
-            format: FormatTemplate::from_string(&block_config.format)?,
-            format_unavailable: FormatTemplate::from_string(&block_config.format_unavailable)?,
+            format: block_config.format.with_default("{label} {percentage}")?,
+            format_unavailable: block_config.format_unavailable.with_default("{label} x")?,
         })
     }
 }
@@ -316,7 +302,7 @@ impl Block for Bluetooth {
                     _ => State::Warning,
                 });
             }
-            self.output.set_text(self.format.render(&values)?);
+            self.output.set_texts(self.format.render(&values)?);
         } else {
             let values = map!(
                 "label" => Value::from_string(self.device.label.clone()),
@@ -324,7 +310,7 @@ impl Block for Bluetooth {
             );
             self.output.set_state(State::Idle);
             self.output
-                .set_text(self.format_unavailable.render(&values)?);
+                .set_texts(self.format_unavailable.render(&values)?);
         }
 
         Ok(None)
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -54,13 +54,13 @@ pub struct PacmanConfig {
     pub interval: Duration,
 
     /// Format override
-    pub format: String,
+    pub format: FormatTemplate,
 
     /// Alternative format override for when exactly 1 update is available
-    pub format_singular: String,
+    pub format_singular: FormatTemplate,
 
     /// Alternative format override for when no updates are available
-    pub format_up_to_date: String,
+    pub format_up_to_date: FormatTemplate,
 
     /// Indicate a `warning` state for the block if any pending update match the
     /// following regex. Default behaviour is that no package updates are deemed
@@ -81,9 +81,9 @@ impl Default for PacmanConfig {
     fn default() -> Self {
         Self {
             interval: Duration::from_secs(600),
-            format: "{pacman}".to_string(),
-            format_singular: "{pacman}".to_string(),
-            format_up_to_date: "{pacman}".to_string(),
+            format: FormatTemplate::default(),
+            format_singular: FormatTemplate::default(),
+            format_up_to_date: FormatTemplate::default(),
             warning_updates_regex: None,
             critical_updates_regex: None,
             aur_command: None,
@@ -94,20 +94,21 @@ impl Default for PacmanConfig {
 
 impl PacmanConfig {
     fn watched(
-        format: &str,
-        format_singular: &str,
-        format_up_to_date: &str,
+        format: &FormatTemplate,
+        format_singular: &FormatTemplate,
+        format_up_to_date: &FormatTemplate,
         aur_command: Option<String>,
     ) -> Result<Watched> {
-        let aur_format = "{aur";
-        let pacman_format = "{pacman";
-        let both_format = "{both";
-        let pacman_deprecated_format = "{count";
-        let concatenated_format_str = format!("{}{}{}", format, format_singular, format_up_to_date);
-        let aur = concatenated_format_str.contains(aur_format);
-        let pacman = concatenated_format_str.contains(pacman_format)
-            || concatenated_format_str.contains(pacman_deprecated_format);
-        let both = concatenated_format_str.contains(both_format);
+        macro_rules! any_format_contains {
+            ($name:expr) => {
+                format.contains($name)
+                    || format_singular.contains($name)
+                    || format_up_to_date.contains($name)
+            };
+        }
+        let aur = any_format_contains!("aur");
+        let pacman = any_format_contains!("pacman") || any_format_contains!("count");
+        let both = any_format_contains!("both");
         if both || (pacman && aur) {
             let aur_command = aur_command.block_error(
                 "pacman",
@@ -142,18 +143,6 @@ impl ConfigBlock for Pacman {
         Ok(Pacman {
             id,
             update_interval: block_config.interval,
-            format: FormatTemplate::from_string(&block_config.format)
-                .block_error("pacman", "Invalid format specified for pacman::format")?,
-            format_singular: FormatTemplate::from_string(&block_config.format_singular)
-                .block_error(
-                    "pacman",
-                    "Invalid format specified for pacman::format_singular",
-                )?,
-            format_up_to_date: FormatTemplate::from_string(&block_config.format_up_to_date)
-                .block_error(
-                    "pacman",
-                    "Invalid format specified for pacman::format_up_to_date",
-                )?,
             output,
             warning_updates_regex: match block_config.warning_updates_regex {
                 None => None, // no regex configured
@@ -187,6 +176,9 @@ impl ConfigBlock for Pacman {
             )?,
             uptodate: false,
             hide_when_uptodate: block_config.hide_when_uptodate,
+            format: block_config.format.with_default("{pacman}")?,
+            format_singular: block_config.format_singular.with_default("{pacman}")?,
+            format_up_to_date: block_config.format_up_to_date.with_default("{pacman}")?,
         })
     }
 }
@@ -380,7 +372,7 @@ impl Block for Pacman {
             }
             Watched::None => (std::collections::HashMap::new(), false, false, 0),
         };
-        self.output.set_text(match cum_count {
+        self.output.set_texts(match cum_count {
             0 => self.format_up_to_date.render(&formatting_map)?,
             1 => self.format_singular.render(&formatting_map)?,
             _ => self.format.render(&formatting_map)?,
@@ -415,6 +407,7 @@ mod tests {
     use crate::blocks::pacman::{
         get_aur_available_updates, get_update_count, PacmanConfig, Watched,
     };
+    use crate::formatting::FormatTemplate;
 
     #[test]
     fn test_get_update_count() {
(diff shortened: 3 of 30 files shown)


src/blocks/bluetooth.rs
```diff
@@ -224,32 +224,27 @@ pub struct Bluetooth {
...
-    fn default_format() -> String {
-        "{label} {percentage}".into()
-    }
-
-    fn default_format_unavailable() -> String {
-        "{label} x".into()
+impl Default for BluetoothConfig {
+    fn default() -> Self {
+        Self {
+            mac: String::new(),
```
MaxVerevkin, line 242: Why? This makes `mac` optional.

src/formatting.rs
```diff
@@ -52,7 +58,21 @@ impl FormatTemplate {
         false
     }
 
-    pub fn from_string(s: &str) -> Result<Self> {
+    // TODO: Remove?
```
MaxVerevkin, line 61: Only `xrandr` uses it, right?

GladOSkar: And `pacman` but that might change. I can replave both of them

src/blocks/bluetooth.rs
GladOSkar: Oops yes that was stupid

src/blocks/bluetooth.rs
```diff
@@ -224,27 +224,32 @@ pub struct Bluetooth {
...
 pub struct BluetoothConfig {
     pub mac: String,
     //DEPRECATED
     //TODO remove
     pub label: Option<String>,
+    #[serde(default = "BluetoothConfig::default_hide_disconnected")]
     pub hide_disconnected: bool,
+    #[serde(default = "BluetoothConfig::default_format")]
     pub format: Option<FormatTemplate>,
+    #[serde(default = "BluetoothConfig::default_format_unavailable")]
```
MaxVerevkin, line 238: By the way, `Option<T>` implements Default that returns `None`, so no need to to create an explicit function for that. `#[serde(default)]` should work.

MaxVerevkin: Same for `bool`

GladOSkar: Oh neat, thanks!

MaxVerevkin:
> The pacman block doesn't support the [block.format] section yet because it has some weird internal logic that requires direct access to the format string

To be honest, `pacman` block seems over-complicated for some reason. I think it's better to simplify it somehow not to rely on text of format string.

MaxVerevkin:
> The [block.format] section requires the full field to be set and i couldn't figure out how to make serde fall back to the default if it's not given.

Perhaps turn`FormatTemplate`'s `full` field into an `Option` and modify the `default_format!` macro to take this into account? And let `FormatTemplate::render` return `""` if `full` is None?

src/blocks/pacman.rs
```diff
@@ -142,14 +142,13 @@ impl ConfigBlock for Pacman {
...
-            format_singular: FormatTemplate::from_string(&block_config.format_singular)
-                .block_error(
-                    "pacman",
-                    "Invalid format specified for pacman::format_singular",
-                )?,
-            format_up_to_date: FormatTemplate::from_string(&block_config.format_up_to_date)
+            format_singular: FormatTemplate::new(&block_config.format_singular, None).block_error(
+                "pacman",
+                "Invalid format specified for pacman::format_singular",
+            )?,
```
MaxVerevkin, line 150: Off topic but you can remove those `.block_error()`s to be consistent with other blocks.

GladOSkar:
> Perhaps turnFormatTemplate's full field into an Option and modify the default_format! macro to take this into account? And let FormatTemplate::render return "" if full is None?

Good idea, i did basically that but then decided to change a bunch of other things because doing that within the macro was quite cumbersome and also the `FormatTemplate` in the BlockConfigs don't need to be Options anymore then. Plus it made more sense to me this way. But it works :)

I also updated the `pacman` block now in a way that i think should work, but since i don't have arch I'd appreciate some more thorough testing, as with the rest of this PR.

I'd consider this PR feature-complete now, feel free to let me know if you'd recommend any other changes

MaxVerevkin:
> Edit: (Oh and GH actions seems to be complaining about clippy lints but i can only find some unrelated clippy warnings so i wouldn't include those changes here)

`formatting` module's test are not compiling.

src/formatting.rs
```diff
@@ -39,20 +30,51 @@ fn unexpected_token<T>(token: char) -> Result<T> {
...
+    short: Option<Vec<Token>>,
+}
+
 impl FormatTemplate {
     /// Whether the format string contains given placeholder
     pub fn contains(&self, var: &str) -> bool {
-        for token in &self.tokens {
-            if let Token::Var(ref placeholder) = token {
-                if placeholder.name == var {
-                    return true;
```
MaxVerevkin, line 48: I think `FormatTemplate::contains()` should take `short` into account too.

MaxVerevkin:
> Edit: (Oh and GH actions seems to be complaining about clippy lints but i can only find some unrelated clippy warnings so i wouldn't include those changes here)

Build fails because of formatting. Try `cargo fmt`.

ammgws:
@matclab Any comments? (since you did something similar in https://github.com/greshake/i3status-rust/pull/703, though I believe it stalled because we were waiting on someone to implement an enhancement on the i3 side)

MaxVerevkin:
So what about `time`?

src/formatting.rs
GladOSkar: As in check both? I figured everything in `short` would also be in `full` but i guess that's reasonable

GladOSkar:
> So what about time?

Not sure what exactly you mean but `time` uses a very different `format` system (directly passing it to the unix `date` binary afaik), hence this system is not suited to that. But of course it would be neat for it to be able to be shortened as well

GladOSkar:
> Build fails because of formatting. Try cargo fmt.

Didn't help apparently. I already ran `rustfmt` and `clippy` as well. I reakky can't figure out what it's complaining about

ammgws:
Doesn't look like any of the strftime modifiers conflict with our format string stuff (curly brackets), so don't see why time's `format` option couldn't be handled as a FormatTemplate internally?

https://man7.org/linux/man-pages/man3/strftime.3.html

GladOSkar:
Hm i guess you're right it'd just be a static format without any placeholders then

MaxVerevkin:
> Didn't help apparently. I already ran rustfmt and clippy as well. I really can't figure out what it's complaining about

Right now test don't compile.

matclab:
> @matclab Any comments? (since you did something similar in #703, though I believe it stalled because we were waiting on someone to implement an enhancement on the i3 side)

i3 side support would be the butter. 
On #703, I added some user UI which allows to switch between three states (short enforced, long enforced, choosen by i3) by midlle clicking. It was somewhat usefull to me, but I'm not fond of the UX (poor discoverability and was difficult to know in which state the widget was).
Also my solution was to add specific code to each widget. Current solution seems much better.

So I think I'll give a try to this PR this evening to give you some feedbac, but it looks nice to me.

GladOSkar:
> Right now test don't compile.

Oops

GladOSkar:
@matclab i agree. The `i3` feature would make the functionality of this PR nicer as well but isn't required.

Being able to switch the format by clicking might be neat but also a bit overkill for now. We can maybe do that along with/after the unification of the click handlers discussed in #1038.

src/formatting.rs
```diff
@@ -119,40 +157,86 @@ impl FormatTemplate {
...
-                                    "util",
-                                    &format!("Unknown placeholder in format string: {}", var.name),
-                                )?
-                                .format(&var)?,
-                        ),
-                )),
+impl Default for FormatTemplate {
+    fn default() -> Self {
+        Self {
+            full: None,
```
MaxVerevkin, line 165: A simple `derive` is sufficient.

MaxVerevkin:
> Doesn't look like any of the strftime modifiers conflict with our format string stuff (curly brackets), so don't see why time's format option couldn't be handled as a FormatTemplate internally?

Perhaps just add an explicit `format_short`? Or add `fn FormatTemplate::raw()` that would return the string it was constructed with?

src/formatting.rs
GladOSkar: Ah yes those are the parts of Rust that i don't have an intuition for yet

ammgws:
What's the status on this?

matclab:
I'm using the version built 2 days ago and it works fine to my taste.

GladOSkar:
> What's the status on this?

Well it's complete imo and I implemented all of @MaxVerevkin's feedback. But it could probably need a bunch more testing since a lot was changed. Especially `pacman` with the various placeholders since I don't run Arch.

MaxVerevkin:
>What's the status on this?

At the moment I'm running `swaystatus` which has exact same `short_text` management and it works fine. 

However, I can note that on the `swaybar` side, the support is kinda raw. In addition to lack of progressive collapsing, if *one* of my outputs switches to `short_text`, all the outer outputs switch too, which is sad (I have a small laptop screen and an external monitor).

> Especially `pacman` with the various placeholders since I don't run Arch.

I'll test it today (even though I don't like the idea of always seeing the number of updates (and if did, I would just use `checkupdates | wc -l`))

ammgws:
Pacman block is displaying the respective arch and AUR update counts correctly here.

MaxVerevkin:
@ivankovnatsky do you want to test this?

doc/blocks.md
```diff
@@ -1724,6 +1724,18 @@ command = "echo '<b>1 &amp;</b>'"
...
+
+The field can be set as a string (`format = "{my_format}"`) or as a section:
+```toml
+[[block]]
+block = "my_block"
+[block.format]
+full = "{my_full_format}"
+short = "{my_short_format}"
+```
+Your `i3` or `sway` will switch all blocks over to the `short` variant whenever there isn't nough space on your screen for the `full` status bar.
```
ivankovnatsky, line 1737:
```suggestion
 Your `i3` or `sway` will switch all blocks over to the `short` variant whenever there isn't enough space on your screen for the `full` status bar.
```

GladOSkar:
@ammgws if `pacman` works and nobody has any issues anymore i think this is good to go

ivankovnatsky:
guys, thanks a lot for taking time implementing this!

laptop/monitor bars running at the same time, i like it. though running in a plain theme i guess:

laptop
monitor

running my config as is gives me:

```
May 10 15:24:24 thinkpad sway[34355]: Configuration error: failed to parse TOML from file contents.
May 10 15:24:24 thinkpad sway[34355]: Cause: Icon set 'awesome5' not found for key `icons` at line 107 column 1
```

using plain theme gives the same error:

```
May 10 15:34:10 thinkpad sway[46970]: Configuration error: failed to parse TOML from file contents.
May 10 15:34:10 thinkpad sway[46970]: Cause: Theme 'plain' not found. for key `theme` at line 2 column 9
```

i was able to run i3status-rust only without `theme` and `icons` definitions.

i guess i compiled it in a weird way, i used nix overlay to built it:

```nix
{ lib, rustPlatform, fetchFromGitHub, pkg-config, makeWrapper, dbus
, libpulseaudio, notmuch, openssl, ethtool }:

rustPlatform.buildRustPackage rec {
  pname = "i3status-rust";
  version = "4bf8495dc5a8292acacc39f27114ce838a6c4183";

  src = fetchFromGitHub {
    owner = "GladOSkar";
    repo = pname;
    rev = "${version}";
    sha256 = "sha256-sgbT9vD/d/2obW+f3DcaSdkEXTlPyUFTk5wYsPASL3o=";
  };

  cargoSha256 = "sha256-+frd9sLelFLfv99nQl5bIQlVFXhkC7AP2ZD8hmPj+u0=";

  nativeBuildInputs = [ pkg-config makeWrapper ];

  buildInputs = [ dbus libpulseaudio notmuch openssl ];

  cargoBuildFlags = [ "--features=notmuch" "--features=maildir" ];

  postFixup = ''
    wrapProgram $out/bin/i3status-rs --prefix PATH : "${ethtool}/bin"
  '';

  # Currently no tests are implemented, so we avoid building the package twice
  doCheck = false;

  meta = with lib; {
    description =
      "Very resource-friendly and feature-rich replacement for i3status";
    homepage = "https://github.com/greshake/i3status-rust";
    license = licenses.gpl3;
    maintainers = with maintainers; [ backuitist globin ma27 ];
    platforms = platforms.linux;
  };
}
```

Anyway, the hiding is works as expected.

ivankovnatsky:
@GladOSkar yeah, i guess, i rebuilt your current change and restarted sway all the way, and the icons appeared as expected.

ivankovnatsky:
bigger screen:
image

smaller screen:
image

ignore mem widgets, didn't formatted yet.

```toml
theme = "space-villain"
icons = "awesome5"

[[block]]
block = "custom"
command = "echo  $(uname -r)"
interval = "once"

[[block]]
block = "custom"
command = "[ $(nixos-version --revision) != $(curl -s https://api.github.com/repos/NixOS/nixpkgs/git/refs/heads/nixos-unstable | jq '.object.sha' -r ) ] && echo '{\"state\":\"Info\", \"text\": \"Update\"}' || echo '{\"state\":\"Idle\", \"text\": \"No Update\"}'"
interval = 1800
json = true

[[block]]
block = "cpu"
interval = 1

[block.format]
full = "{utilization} {frequency}"
short = ""

[[block]]
block = "load"
interval = 1

[block.format]
full = "{1m} {5m} {15m}"
short = ""

[[block]]
block = "temperature"
chip = "*-isa-*"
collapsed = false
inputs = ["temp1"]
interval = 1

[block.format]
full = "{max}"
short = ""

[[block]]
block = "custom"
command = "echo  $(cat /sys/class/hwmon/hwmon4/fan1_input) RPM"
interval = 1

[[block]]
block = "memory"
display_type = "memory"

[[block]]
block = "memory"
display_type = "swap"

[[block]]
alert = 200
alias = "/"
block = "disk_space"
info_type = "used"
path = "/"
warning = 150

[block.format]
full = " {used}/{total}"
short = ""

[[block]]
block = "net"
device = "wlp2s0"
hide_inactive = true
interval = 1

[block.format]
full = "{speed_down} {speed_up}"
short = ""

[[block]]
block = "battery"
driver = "upower"

[block.format]
full = "{percentage}"
short = ""

[[block]]
block = "keyboard_layout"
driver = "sway"
format = " {layout}"

[[block]]
block = "sound"
on_click = "pavucontrol"

[[block]]
block = "weather"

[block.format]
full = "{temp} {apparent}  {humidity} 煮 {wind_kmh} km/h"
short = ""

[block.service]
api_key = ""
city_id = ""
name = "openweathermap"
units = "metric"

[[block]]
block = "time"
format = "%a %b %d %H:%M"
```

GladOSkar:
Hm. I'll have another look at that

ivankovnatsky:
i think keeping the widget icon is good idea (in short version), since it could be highlighted with color, we might not see the details, but the color will tell in general.

going back to what i imagined this implementation is much-much better!

MaxVerevkin:
> bigger screen:

>smaller screen:

Do you use `i3` or `sway`? for me on `sway`, all outputs have the same, either short or full text.

GladOSkar:
On my `i3` it is able to keep them separate, `full` on a bar with enough space and `short` on one without.

ivankovnatsky:
> Do you use i3 or sway? for me on sway, all outputs have the same, either short or full text.

sway, i meant that on a bigger screen all my widgets get it's space, on a laptop it will fallback to short versions, where applicable and configured.

if i deliberately clone some widgets to fill the width it will fallback to short versions on both screens:

screen with longer width: 
image

shorter width: 
image

MaxVerevkin:
OK, so it's `swaybar`'s bug.

GladOSkar:
@ammgws any blockers left?

PR #1213, opened 2021-05-07
Escape html entities in focused_window block

brookst:
Block would fail to render text containing `&` or `<` with pango emitting errors e.g:

    pango_layout_set_markup_with_accel: Error on line 1:
    Entity did not end with a semicolon; etc.

I've used the [html-escape crate](https://crates.io/crates/html-escape) which may be overkill for this but was easiest. I can write a smaller implementation if you don't want to take the dependency.

### DIFF
--- a/src/blocks/focused_window.rs
+++ b/src/blocks/focused_window.rs
@@ -11,6 +11,7 @@ use crate::blocks::{Block, ConfigBlock, Update};
 use crate::config::SharedConfig;
 use crate::errors::*;
 use crate::scheduler::Task;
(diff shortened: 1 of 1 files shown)


ammgws:
Did you try this approach: https://github.com/greshake/i3status-rust/pull/1179?

It should be sufficient.

brookst:
Yeah, that's all that was needed. I'm not sure how I missed that, I did search through issues to see if this had been reported yet.

So no dependency - just pass though that escape function.

PR #1220, opened 2021-05-09
fixed pacman printing to stdout/err

### LINKED ISSUE #1214: pacman block prevents use
```toml
[[block]]
block = "pacman"
interval = 300
format_up_to_date = "0"
critical_updates_regex = "(linux|linux-lts|linux-zen|linux-ck|firefox|sudo|librewolf|linux-amd|firefox)"
```
causes invalid i3bar-json because the output of the pacman DB update is not piped away. Causing output similar to this:
```json
:: Paketdatenbanken werden synchronisiert …
Fehler: failed to synchronize all databases (Kann Datenbank nicht sperren)
[[{"background":null,"color":"#424242","full_text…
```

If I read the source correct then it should put the database in `/tmp`, since it is tmpfs potential issues should be resolved after a restart, but they are not.

poldi1405:
This should fix #1214

### DIFF
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -195,7 +195,6 @@ fn run_command(var: &str) -> Result<()> {
     Command::new("fakeroot")
         .args(&["--", var])
         .stdout(Stdio::null())
-        .stderr(Stdio::null())
         .spawn()
         .block_error("pacman", &format!("Failed to run command '{}'", var))?
         .wait()
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -255,7 +255,7 @@ fn get_pacman_available_updates() -> Result<String> {
 
     // Update database
     run_command(&format!(
-        "pacman -Sy --dbpath \"{}\"",
+        "pacman -Sy --dbpath \"{}\" --logfile /dev/null",
         updates_db
     ))?;
 
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -263,10 +263,7 @@ fn get_pacman_available_updates() -> Result<String> {
     String::from_utf8(
         Command::new("sh")
             .env("LC_ALL", "C")
-            .args(&[
-                "-c",
-                &format!("pacman -Qu --dbpath \"{}\"", updates_db),
-            ])
+            .args(&["-c", &format!("pacman -Qu --dbpath \"{}\"", updates_db)])
             .output()
             .block_error("pacman", "There was a problem running the pacman commands")?
             .stdout,
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -3,7 +3,7 @@ use std::ffi::OsString;
 use std::fs;
 use std::os::unix::fs::symlink;
 use std::path::Path;
-use std::process::Command;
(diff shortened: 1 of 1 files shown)


src/blocks/pacman.rs
```diff
@@ -264,7 +266,7 @@ fn get_pacman_available_updates() -> Result<String> {
             .env("LC_ALL", "C")
             .args(&[
                 "-c",
-                &format!("fakeroot pacman -Qu --dbpath \"{}\"", updates_db),
+                &format!("pacman -Qu --dbpath \"{}\"", updates_db),
```
ammgws, line 269: Why remove fakeroot here?

src/blocks/pacman.rs
```diff
@@ -254,7 +256,7 @@ fn get_pacman_available_updates() -> Result<String> {
 
     // Update database
     run_command(&format!(
-        "fakeroot -- pacman -Sy --dbpath \"{}\" --logfile /dev/null &> /dev/null",
+        "pacman -Sy --dbpath \"{}\"",
```
ammgws, line 259: I'm not sure why we had `--logfile /dev/null` in the first place, but do you really need to remove it to fix your issue? If not then might be better to leave it for now, and perhaps remove it later once it's proven that it is really not needed.

ammgws (changes requested):

src/blocks/pacman.rs
poldi1405: Instead of calling it through a shell which brings no advantage to my knowledge

src/blocks/pacman.rs
poldi1405: all output is discarded anyways and this just keeps the command simpler

src/blocks/pacman.rs
ammgws: Oh sorry I missed that fakeroot was moved to the other function. That function is kinda useless since it's only used in one place but that's another issue.

src/blocks/pacman.rs
poldi1405: > do you really need to remove it to fix your issue

technically I don't but since I was there I just did since it had no effect.

ammgws: OK so it looks like we're just doing exactly what the `checkupdates` script does, so I would prefer to keep it for now even though it doesn't appear to be needed. There might be some weird issue that it solves.

src/blocks/pacman.rs
```diff
@@ -192,8 +192,10 @@ impl ConfigBlock for Pacman {
 }
 
 fn run_command(var: &str) -> Result<()> {
-    Command::new("sh")
-        .args(&["-c", var])
+    Command::new("fakeroot")
+        .args(&["--", var])
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
```
MaxVerevkin, line 198: Maybe leave `stderr` piped? If something is printed to `stderr` it doesn't break the bar, but might (?) be useful for debugging.

poldi1405: sounds logical. Will do.

ammgws:
Just need to run rustfmt on it.

ammgws:
Thanks!

Issue #1201, opened 2021-04-28
Loading icons and themes relative to binary


liclac:
As of #1135, `/usr/share/i3status-rust` is hardcoded as as the last path to check for icons and themes. I'm using NixOS, and I'd like the themes to live alongside the binary in the nix store: `/nix/store/<hash>-i3status-rust-0.20.0/share/i3status-rust`.

One way to do this, which also covers `/usr/loca/share`, etc. is to simply make the path `../share/i3status-rust` relative to the binary. If that sounds good, I'd be happy to send a PR that implements this.

GladOSkar:
Wouldn't that break it for users run `i3status-rs` from `<repo_dir>/target/release/i3status-rs` or `~/.local/bin/i3status-rs` or similar?

liclac:
> Wouldn't that break it for users who run i3status-rs from `<repo_dir>/target/release/i3status-rs` or `~/.local/bin/i3status-rs` or similar?

Just to be clear, I'm proposing changing the current logic from:

1. `$XDG_CONFIG_HOME/i3status-rust`
1. `~/.local/share/i3status-rust`
1. `/usr/share/i3status-rust`

To:

1. `$XDG_CONFIG_HOME/i3status-rust`
1. `~/.local/share/i3status-rust`
1. `</path/to/i3status-rs>/../share/i3status-rust`

...or, if there's a use case for keeping `/usr/share` hardcoded in there, appending `3.` to the list.

My assumption would be that `/usr/bin/i3status-rs` and `/usr/share/i3status-rust` would be installed as a pair by a package, and that mixing and matching versions isn't normally desirable - if a new icon is introduced, for example, the default icon sets would also be updated to provide it, but a user's custom icon sets would still live in their home directory.

(That said, there probably should be a path in there that `cargo run`, `cargo build`, etc. builds will find.)

> Also afaik a relative path would be relative to the cwd of the calling process (i3bar/swaybar), not relative to the binary, right?
> I wonder how other packages do this on NixOS?

Right, this is basically working around the fact that there's no easy way to set a traditional `PREFIX=` with cargo.

A common workaround is to use paths relative to `argv[0]`, which is normally the (occasionally relative) path to the executable, eg. `./i3status-rs` or `/usr/bin/i3status-rs`.

Go programs tend to have a similar problem, but as they're statically linked by default, they tend to just embed their resources into the executable itself.

[...]

PR #1221, opened 2021-05-09
Find icons/themes relative to binary path

Mephistophiles:
Fixes: #1201

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -45,6 +45,7 @@ fn lookup_file(path: Option<PathBuf>, subdir: Option<&str>, file: &Path) -> Opti
 /// - Then try XDG_CONFIG_HOME
 /// - Then try `~/.local/share/`
 /// - Then try `/usr/share/`
+/// - Then try `<i3status-rs>/../share/`
 ///
 /// Automaticaly append an extension if not presented.
 pub fn find_file(file: &str, subdir: Option<&str>, extension: Option<&str>) -> Option<PathBuf> {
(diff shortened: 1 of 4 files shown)


MaxVerevkin:
Can you please update the comments for `fn find_file()` and write a short summary of what `fn lookup_file()` does?

Mephistophiles:
> Can you please update the comments for `fn find_file()` and write a short summary of what `fn lookup_file()` does?

Fixed! Thanks for the review!

src/util.rs
```diff
@@ -10,18 +10,57 @@ use serde::de::DeserializeOwned;
...
+    file: &Path,
+) -> Option<PathBuf>
+where
+    P: Into<PathBuf>,
+    F: FnOnce(&mut PathBuf),
+{
+    let mut path: PathBuf = match path {
+        Some(path) => path.into(),
+        None => return None,
+    };
```
MaxVerevkin, line 40: You can use `?` operator here
```rust
let mut path: PathBuf = path?.into();
```

src/util.rs
```diff
@@ -10,18 +10,57 @@ use serde::de::DeserializeOwned;
...
+
+    transform(&mut path);
```
MaxVerevkin, line 42: Is there a reason to pass a `transofrm` function to `lookup_file()` instead of directly calling it? I mean something like
```rust
if let Some(local_share_path) = lookup_file(
        std::env::var("HOME")
            .ok()
            .map(|path| PathBuf::from(path).join(".local/share")),
        subdir,
        &file,
    ) {
        return Some(local_share_path);
    }
```
instead of
```rust
if let Some(local_share_path) = lookup_file(
        std::env::var("HOME").ok(),
        |path| path.push(".local/share/"),
        subdir,
        &file,
    ) {
        return Some(local_share_path);
    }
```

src/util.rs
Mephistophiles: Fixed, thanks

src/util.rs
Mephistophiles: Hm. It looks better, but I would like to remove unnecessary allocation in join function and replace it with push. I will think about replacing the callback with map.

MaxVerevkin: Both are fine, but current `lookup_file`'s API is just a little weird (to me).

Mephistophiles: Now `lookup_file` accepts only 3 parameters (without callback and generic).

ammgws:
I'm not against merging this if someone can clarify this

For reference are there any other projects that have implemented your requested changes?

_Originally posted by @ammgws in https://github.com/greshake/i3status-rust/issues/1201#issuecomment-834051022_

PR #1223, opened 2021-05-10
Custom block examples

### LINKED ISSUE #1215: discussion: Custom block script/example repository
Copying comment from https://github.com/greshake/i3status-rust/issues/715#issue-627886430:

> Note that there are a lot of simple blocks here that could be handled by the custom block, especially those that just call external programs, so I don't see the advantage of adding those into the Rust codebase since there's no performance gains etc to be made. However we could look into the idea of having contributed configs for the custom block either in the repo or wiki or somewhere. 

Would like to hear everyone's opinion on the best way forward.

GladOSkar:
Resolves #1215

Feedback welcome


ammgws:
Good stuff! Just needs a link leading to the README somewhere in blocks.md

Issue #1222, opened 2021-05-10
[v0.20.0] invalid json generated for any block


jthomaschewski:
### Config

```
[[block]]
block = "load"
interval = 120
format = "{5m}"
```
(Simple example config, can be reproduced with any config)

### Version

i3status-rust v0.20.0 and latest master.
Tried `i3status-rust v0.20.0-2` package from arch `community-testing` and `i3status-git` from AUR.

i3-gaps v4.19.1

### Expected behavior

Should write valid json to stdout and show bar in i3

```console
$ i3status-rs 
{"version": 1, "click_events": true}
[{"full_text":"","color":"#002b36","separator":false,"separator_block_width":0,"markup":"pango"},{"full_text":"  1.6 ","color":"#93a1a1","background":"#002b36","name":"0","instance":"0","separator":false,"separator_block_width":0,"markup":"pango"}],
```

### Actual behavor

Writes invalid open bracket in stdout.

```console
$ i3status-rs    // v0.20.0
{"version": 1, "click_events": true}
[
[{"full_text":"","color":"#002b36","separator":false,"separator_block_width":0,"markup":"pango"},{"full_text":"  1.6 ","color":"#93a1a1","background":"#002b36","name":"0","instance":"0","separator":false,"separator_block_width":0,"markup":"pango"}],
```

When downgrading to v0.14.7 and starting with the same config output is the following:

```console
$ i3status-rs   // v0.14.7
{"version": 1, "click_events": true}
[[{"background":null,"color":"#002b36","full_text":"","markup":"pango","separator":false,"separator_block_width":0},{"background":"#002b36","color":"#93a1a1","full_text":"  2.07 ","name":"0","separator":false,"separator_block_width":0}],
```

i3 reports:
`Could not parse JSON (lexical error: invalid character inside string.)`

Tried multiple theme/icon combinations and different block types.
Always the same result.

jthomaschewski:
Uh... sorry I failed. I've set a different config file in my i3 config.
The additional opening bracket seems still weird to me and is new in v0.20.0
But i3 seems to ignore it.

Additional info if interested:

The bracket seems to be added during init. When ran with `--no-init` flag, it's not written to stdout:
```
i3status-rs --no-init                          
[...]

PR #1224, opened 2021-05-11
Fix json output of configuration error

jthomaschewski:
JSON of `ConfigurationError` which is passed to i3bar is separated by newline.
Because of that, a JSON parsing error is displayed by i3 instead of the proper i3status-rs error message.

before:
```
[{"full_text":" Configuration error: failed to parse TOML from file contents.
Cause: expected an equals, found a newline at line 1 column 13 ","color":"#dc322f","background":"#000000","name":"0","instance":"0","separator":false,"separator_block_width":0,"markup":"pango"}],
```
image

after:
```
[{"full_text":" Configuration error: failed to parse TOML from file contents. Cause: expected an equals, found a newline at line 1 column 13 ","color":"#dc322f","background":"#000000","name":"0","instance":"0","separator":false,"separator_block_width":0,"markup":"pango"}],
```
image

Discussed in #1222 

Note: Escaping `\n` is not a solution as it's then displayed as literal. Not sure if it's possible or even desirable to have the line break interpreted by i3bar.
Alternatively a simplified error message could be written to i3bar json and the full only to stderr

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -88,11 +88,11 @@ impl fmt::Debug for Error {
                 f.write_str(&format!("Error in block '{}': {}", block, message))
             }
             ConfigurationError(ref message, ref cause) => f.write_str(&format!(
-                "Configuration error: {}.\nCause: {}",
+                "Configuration error: {}. Cause: {}",
                 message, cause
             )),
             InternalError(ref context, ref message, Some((ref cause, _))) => f.write_str(&format!(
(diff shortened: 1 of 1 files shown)


ammgws:
Yeah that explains why I haven't seen this until now, since newlines are handled differently by sway (it just renders the newline character and doesn't error out - noticed something similar in the past: https://github.com/i3/i3/issues/3976)

Anyway thanks for the fix!

PR #1226, opened 2021-05-11
Adds example for custom pi-hole status block

EugeneKorshenko:
Sharing my custom block for the local ads blocking server [https://pi-hole.net/](https://pi-hole.net/). I'd be happy to get a feedback :slightly_smiling_face: 
It partially replaces previous "native" implementation: #1007
In response to: https://github.com/greshake/i3status-rust/pull/1007#issuecomment-838414515


ammgws:
I don't use pihole but it looks like it should work for others too.

Thanks!

PR #1229, opened 2021-05-12
Custom block example: liquid cooling system status

EugeneKorshenko:
Here I'm adding an example block for displaying information from `liquidctl` tool. 
And here how it looks like:
screenshot_2021-05-12-220426


GladOSkar:
The usual block style has a space before and after each icon. Otherwise looks great!

ammgws:
Thanks!

PR #1233, opened 2021-05-14
Better error message on custom blocks

### LINKED ISSUE #1184: custom block: improve error messages
> Not related to your issue but some better error messages would be nice too. Like if you had set `shell = nonexistent` then it results in "Error in block 'custom': Error parsing JSON: expected value at line 1 column 1", where as the actual error prior to it trying to parse the JSON is "No such file or directory (os error 2)".

_Originally posted by @ammgws in https://github.com/greshake/i3status-rust/issues/1156#issuecomment-814145953_

> Yeah, https://github.com/greshake/i3status-rust/blob/b298e267fbd00f4ce32cc4e89ece826dc4175bc5/src/blocks/custom.rs#L158 doesn't make much sense.

_Originally posted by @MaxVerevkin in https://github.com/greshake/i3status-rust/issues/1156#issuecomment-814149456_

jespino:
Fixes #1184

### DIFF
--- a/src/blocks/custom.rs
+++ b/src/blocks/custom.rs
@@ -151,11 +151,14 @@ impl Block for Custom {
             .or_else(|| self.command.clone())
             .unwrap_or_else(|| "".to_owned());
 
-        let raw_output = Command::new(&self.shell)
+        let raw_output = match Command::new(&self.shell)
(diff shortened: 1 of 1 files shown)


ammgws:
That seems about right.

Thanks!

PR #1234, opened 2021-05-14
Adding get_crypto block (based on getcrypto from bumblebee-status)

jespino:


ammgws:
I'd prefer this added as a Custom block example instead. What do you think?

PR #1240, opened 2021-05-16
pacman: fix default template not being used for `PacmanConfig::watched()`

### LINKED ISSUE #1238: pacman default string no longer valid
My pacman block looks like this:

```
[[block]]
block = "pacman"
```

Now the following happens:

```
i3status-rs
[
[{"full_text":" Internal error in context 'util': Unknown placeholder in format string: pacman ","color":"#dc322f","background":"#000000","name":"0","instance":"0","separator":false,"separator_block_width":0,"markup":"pango"}],


Internal error in context 'util': Unknown placeholder in format string: pacman
```

There seems to be an issue with the default string of the block. 

---

Version: i3status-rs 0.20.1 (commit 67238c5 2021-05-13)

GladOSkar:
Fixes #1238

`PacmanConfig::watched()` was using a version of the `FormatTemplate` that didn't include the default format.

Please test as i don't have `pacman`, although according to my debugging it should work now

### DIFF
--- a/src/blocks/pacman.rs
+++ b/src/blocks/pacman.rs
@@ -140,6 +140,10 @@ impl ConfigBlock for Pacman {
     ) -> Result<Self> {
         let output = TextWidget::new(id, 0, shared_config).with_icon("update")?;
 
+        let fmt_normal = block_config.format.with_default("{pacman}")?;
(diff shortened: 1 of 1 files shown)


ammgws:
It works

ammgws:
Thanks!

PR #1241, opened 2021-05-16
pacman: fix invocation of fakeroot/pacman command

ammgws:
Need to revert back to calling `fakeroot` via `sh` since there seems to be some quoting issues otherwise.


ammgws:
@mpldr Can you check whether this regresses your issue?

PR #1245, opened 2021-05-18
format strings: fix parsing of "unit" part

MaxVerevkin:
Now format strings like `"{speed_down*_}"` work correctly - it hides the
unit but does not override the default setting.


ammgws:
Thanks!

PR #1248, opened 2021-05-22
formatting: refactor, implement prefix hiding and spacing

### LINKED ISSUE #1228: How to hide units in memory block?
I used to have the following configuration:

```toml
[[block]]
block = "memory"
display_type = "memory"
format_mem = "{Mug}/{MTg} GB"
format_swap = "{SUg}/{STg} GB"
```

This displayed, for example, `1.8/32 GB`, displaying the unit only once, which is the perfect balance between readability and space usage. After the update, I tried the replacement

```toml
[[block]]
block = "memory"
display_type = "memory"
format_mem = "{mem_used;G}/{mem_total;G} GB"
format_swap = "{swap_used;G}/{swap_total;G} GB"
```

However, this renders `1.8GB/ 31GB GB`. Even if I removed the manual unit from the format string, this still shows the unit twice, which is unneccessary and wastes space. Also, there is a random space after the slash that I didn't ask for.

How can I configure i3status-rust to show the memory string it used to show?

### LINKED ISSUE #1242: Spaces between numbers and units as per NIST standard
Hello, I've noticed that since the update with the new formatting, many of the default outputs lack a space between the number and the units. I'm wondering on the reasons behind this choice?

It is a [global standard](https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-7-rules-and-style-conventions-expressing-values) to leave a space between a value and its units, plus it's easier (for me, and I presume most others) to read.

I appreciate that it's possible to manually accomplish this by removing the units and then adding a custom unit through Formatting, but this doesn't seem to work with values that have their own units that can change, e.g. those in the net block, and since there's a standard, I'd suggest that leaving a space would be a better default.

### LINKED ISSUE #1243: Formatting net block with consistent units and width
Hello. 

Perhaps I'm missing something, or I'm dense, or just having a bad day but I'm genuinely finding it difficult to understand how to use the new Formatting documentation, particularly in application to the net block. Here are my ideal requirements:
- Fix the block's width to avoid shifting the rest of my bar around every second as the download/upload speed length strings change.
- Have a space between the number and the units (see #1242) and ensure the units are "b", "kb", "Mb", "Gb".
- Scale between b, kb, Mb, Gb 

Here's where I'm at now: 
`format = "{speed_down:2;n*_b} b {speed_up:2;n*_b} b"`

Which leads to this:
image

I appreciate that not all my requirements might be possible, but I'd really appreciate any help that could be provided further to the Formatting documentation. I'm not demanding any changes, I just feel like I might be missing something which could get me most of the way. 

Many thanks,

MaxVerevkin:
Closes #1228
Closes #1242
Possibly closes #1243

- Rewrite formatting strings' parsing
- Remove a test that checks the internal representation of `FormatTemplate` (why did I write it in the first place?)
- Allow to hide the prefix (`"{key;_K}"` sets the min prefix to "K" and hides it)
- Allow a space between the value and prefix/unit (`"{key; K*b}"` results in `"val Kb"`, `"{key; _K*b}"` results in `"val b"`)
- TODO: update the docs

### DIFF
--- a/src/formatting.rs
+++ b/src/formatting.rs
@@ -4,13 +4,13 @@ pub mod unit;
 pub mod value;
 
 use std::collections::HashMap;
-use std::convert::TryInto;
 use std::fmt;
 
 use serde::de::{MapAccess, Visitor};
 use serde::{de, Deserialize, Deserializer};
 
 use crate::errors::*;
+use placeholder::unexpected_token;
 use placeholder::Placeholder;
 use value::Value;
 
@@ -20,24 +20,38 @@ enum Token {
     Var(Placeholder),
 }
 
-fn unexpected_token<T>(token: char) -> Result<T> {
-    Err(ConfigurationError(
-        format!(
-            "failed to parse formatting string: unexpected token '{}'",
-            token
-        ),
-        String::new(),
-    ))
-}
-
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Default, Clone)]
 pub struct FormatTemplate {
     full: Option<Vec<Token>>,
     short: Option<Vec<Token>>,
 }
 
 impl FormatTemplate {
-    /// Whether the format string contains given placeholder
+    pub fn new(full: &str, short: Option<&str>) -> Result<Self> {
+        Self::new_opt(Some(full), short)
+    }
+
+    pub fn new_opt(full: Option<&str>, short: Option<&str>) -> Result<Self> {
+        let full = match full {
+            Some(full) => Some(Self::tokens_from_string(full)?),
+            None => None,
+        };
+        let short = match short {
+            Some(short) => Some(Self::tokens_from_string(short)?),
+            None => None,
+        };
+        Ok(Self { full, short })
+    }
+
+    /// Initialize `full` field if it is `None`
+    pub fn with_default(mut self, default_full: &str) -> Result<Self> {
+        if self.full.is_none() {
+            self.full = Some(Self::tokens_from_string(default_full)?);
+        }
+        Ok(self)
+    }
+
+    /// Whether the format string contains a given placeholder
     pub fn contains(&self, var: &str) -> bool {
         Self::format_contains(&self.full, var) || Self::format_contains(&self.short, var)
     }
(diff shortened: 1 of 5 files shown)


MaxVerevkin:
Future idea - we can add another modifier to "min prefix" config to force the prefix (i.e. actually set it to a specific value). This make a ton of sense when using `_` modifier. For example, `"{key;+_G}"` may be used to _force_ the prefix to "G" and hide it.

MaxVerevkin:
By the way, is it a reasonable idea to package formatting "subsystem" as a separate crate?

GladOSkar:
Fails to build on stable `rustc 1.51.0`:
```
error[E0658]: use of unstable library feature 'str_split_once': newly added
```

GladOSkar:
Works great for me, i really like it! Needs up-to-date `rustc` though, build with `1.51.0` fails.

MaxVerevkin:
@GladOSkar interesting, can you post the error message?

GladOSkar:
```
error[E0658]: use of unstable library feature 'str_split_once': newly added
```

MaxVerevkin:
Oh, that's sad. All my parsing code relies on `split_once()` xD

GladOSkar:
It's in the latest stable so it's fine if you ask me

src/formatting.rs
```diff
@@ -71,64 +43,91 @@ impl FormatTemplate {
...
-        }
 
         Ok(tokens)
     }
 
     pub fn render(&self, vars: &HashMap<&str, Value>) -> Result<(String, Option<String>)> {
         let full = match &self.full {
             Some(tokens) => Self::render_tokens(tokens, vars)?,
-            None => String::new(),
+            None => String::new(), // TODO: throw an error that says that it's a bug?
```
ammgws, line 130: Perhaps something to do in the future

ammgws:
> By the way, is it a reasonable idea to package formatting "subsystem" as a separate crate?

Hmm I don't really see the need to split it out, I'm guessing it's just to make it easier to work between swaystatus and here, and/or a learning experience?

ammgws:
> Works great for me, i really like it! Needs up-to-date `rustc` though, build with `1.51.0` fails.

Stablised in Version 1.52.0 (2021-05-06), literally only a couple of weeks ago. We already recommend rustup and I think there was a similar situation a couple months ago with clamp or another feature so I'd say it's OK.

ammgws:
@MaxVerevkin  I know it's not related to this PR but could you please fix the formatting as well?

MaxVerevkin:
> > By the way, is it a reasonable idea to package formatting "subsystem" as a separate crate?
> 
> Hmm I don't really see the need to split it out, I'm guessing it's just to make it easier to work between swaystatus and here, and/or a learning experience?

Yeah, exactly. I'm interested in the process of publishing crates and so on. However, I'm not sure if this crate would be useful for other projects.

MaxVerevkin:
>  I know it's not related to this PR but could you please fix the formatting as well?

It's weird, but my `cargo fmt` doesn't detect it. Will fix it manually later.

ammgws:
Alright I'll do it on this side then.

Thanks!

PR #1249, opened 2021-05-23
Add net support for material-icons

### LINKED ISSUE #1244: Net block does not have icons defined for material
Title; if you use material icons and the net block, you'll get an error unless you "override" `net_wireless`, `net_up`, and `net_down`.

K4rakara:
Fixes #1244


ammgws:
Thanks!

I wonder why this wasn't caught until now. Perhaps it was silently failing (blank icon) before.

PR #1258, opened 2021-06-07
blocks: Add function to find default battery

### LINKED ISSUE #1237: Better defaults for battery block
Perhaps it could be improved by defaulting to the first file found in `/sys/class/power_supply/` instead of being hardcoded to BAT0.

_Originally posted by @ammgws in https://github.com/greshake/i3status-rust/issues/1236#issuecomment-841609021_

orvij:
Resolves #1237

### DIFF
--- a/doc/blocks.md
+++ b/doc/blocks.md
@@ -179,7 +179,7 @@ format = "{percentage} {time}"
 
 Key | Values | Required | Default
 ----|--------|----------|--------
-`device` | The device in `/sys/class/power_supply/` to read from. When using UPower, this can also be `"DisplayDevice"`. | No | `"BAT0"`
+`device` | The device in `/sys/class/power_supply/` to read from. When using UPower, this can also be `"DisplayDevice"`. Defaults to finding the first `"BAT"` file in `/sys/class/power_supply`. | No | `"BAT0"`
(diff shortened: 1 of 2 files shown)


ammgws:
Can you update the docs too?

ammgws:
Final thing: the order is not guaranteed, is it? Should we sort it by name? Or by time created?

orvij:
> Final thing: the order is not guaranteed, is it? Should we sort it by name? Or by time created?

Sorted by name. Still default to `BAT0` if no directory is found in `/sys/class/power_supply`. Replaces with first found dir, then with the first ordered dir past that, e.g. `BAT1` replaces `BAT2`.

ammgws:
Alright, thanks!

matthiaskrgr:
This broke i3status-rs for me:
````
[
thread 'main' panicked at 'byte index 3 is out of bounds of `AC`', src/blocks/battery.rs:561:25
`````

`/sys/class/power_supply/` looks like this:
````
AC  BAT0  BAT1  ucsi-source-psy-USBC000:001
````

matthiaskrgr:
The problem seems to be that `AC` only has 2 letters so the array index `&f[..3] == "BAT"` is out of bounds

PR #1260, opened 2021-06-13
battery: fix array out of bounds access when directories with only 2 letters are encountered in /sys/class/power_supply

matthiaskrgr:
cc https://github.com/greshake/i3status-rust/pull/1258
cc @orvij


ammgws:
Yeah that looks good to me.

Thanks!

orvij:
@matthiaskrgr thanks for catching this!

Issue #1254, opened 2021-05-31
Improvements for Battery-Block


SebTM:
I'm currently sed-patching `sed -i "s/{}:{:02}/ | {}:{:02}h/" src/blocks/battery.rs` while building with nix-overlays so I can have a separator and hour-unit after time which is only shown when there is something to show.

Besides this I'm experiencing issues with empty time on 99% and never turning 100% (on some not original ac-adapter | Thinpad T14s - AMD).

So my Ideas are:
- Adding support for seconds/minutes/hours in https://github.com/greshake/i3status-rust/blob/master/doc/blocks.md#formatting
- Maybe another formatting-option to be able to prepend a string only shown while the variable is not empty?
- Option for battery-block which makes an offset between full calculated and full shown possible?

ammgws:
> I'm experiencing issues with empty time on 99% and never turning 100% (on some not original ac-adapter | Thinpad T14s - AMD).

Perhaps create another issue for this if you are implying it's an issue with our calculations. Or is it purely a hardware limitation that causes it to only show up to 99%?

>     * Option for battery-block which makes an offset between full calculated and full shown possible?

What do you mean by this? To set the full threshold to something less than 100%, e.g. I guess in your case you would set it to 99%?

>     * Adding support for seconds/minutes/hours in https://github.com/greshake/i3status-rust/blob/master/doc/blocks.md#formatting
>     * Maybe another formatting-option to be able to prepend a string only shown while the variable is not empty?

@MaxVerevkin What do you think about these?

MaxVerevkin:
> Adding support for seconds/minutes/hours

How exactly do you expect it to work? Can you give some details?

GladOSkar:
> To set the full threshold to something less than 100%, e.g. I guess in your case you would set it to 99%?

"Full" is currently not determined by percentage, but by `BAT0/status == 'Full'`. I recently also had to swap my battery for an aftermarket one and it now also doesn't set 'Full', but switches the `status` to `'Unknown'` once it stops charging, so i'm currently patching this as well:

```diff
diff --git a/src/blocks/battery.rs b/src/blocks/battery.rs
index 8b702f3..846e6c5 100644
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -658,7 +658,7 @@ impl Block for Battery {
             },
         );
 
-        if status == "Full" || status == "Not charging" {
+        if status == "Full" || status == "Not charging" || status == "Unknown" {
[...]

PR #1261, opened 2021-06-15
battery: Add `full_threshold` option for setups that don't reach 100%

### LINKED ISSUE #875: Battery block with TLP
Using [TLP](https://linrunner.de/tlp/) allows you to set charging thresholds for laptop batteries. I'm currently using this on a ThinkPad t14s.

I currently have the settings below:

```
START_CHARGE_THRESH_BAT0=85
STOP_CHARGE_THRESH_BAT0=90
```

Meaning that when the battery reaches 90% it stops charging and is considered full.

The values for the actual filesystem BAT0 status is `Unknown` - and looking at the code it doesn't seem we care about handling this case at the moment.

Currently when using i3status-rust my block just gets "stuck" at 89/90%, but using py3status or i3status, it's possible to set either `status_unk` or `threshold_full` in order to work alongside the charge stop threshold setting.

I'm curious if it's something you'd want to consider supporting? or if there is currently a way around this at the moment? (had a look at the code but can't see any).

Thanks!

GladOSkar:
Resolves #875
Partially addresses #1254

Usecases:

- TLP (or other) setups that intentionally only charge to a certain percentage
- Aftermarket batteries that don't achieve the full charge necessary to show up as full in the sysfs

### DIFF
--- a/doc/blocks.md
+++ b/doc/blocks.md
@@ -187,8 +187,9 @@ Key | Values | Required | Default
 `missing_format` | Same as `format` but for when the specified battery is missing. | No | `"{percentage}"`
 `allow_missing` | Don't display errors when the battery cannot be found. Only works with the `sysfs` driver. | No | `false`
 `hide_missing` | Completely hide this block if the battery cannot be found. Only works in combination with `allow_missing`. | No | `false`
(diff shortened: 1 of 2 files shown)


ammgws:
LGTM, thanks!

PR #1264, opened 2021-06-20
pomodoro: add notify-send

ammgws:
Also remove hardcoded icons

Also run `cargo +nightly clippy --fix -Z unstable-options`


MaxVerevkin:
Perhaps a better approach - https://github.com/MaxVerevkin/swaystatus/commit/aaef3215aed9be5c24b315404bc37950ac821f43

PR #1142, opened 2021-03-31
Net block: use nl80211 to get WiFi information

MaxVerevkin:
Net block: use https://github.com/Eonm/nl80211 to get WiFi information.

PR #1274, opened 2021-07-11
Fix SSID escape code decoding

### LINKED ISSUE #1271: Crash when adding net block for wireless device
i3status-rust crashes when I add a net block for the wireless device on my laptop. I'm on Archlinux. `ethtool`, `iw`, and `ip` are installed.
I'm running i3status-rust 0.20.2. This problem occurs since 0.20.0 in May.

```
[[block]]
block = "net"
device = "wlan0"
format = "{signal_strength} {speed_down;K} {speed_up;K}"
interval = 5
hide_inactive = true
```

```
$ RUST_BACKTRACE=full i3status-rs config.toml
[
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }', src/blocks/net.rs:668:60
stack backtrace:
   0:     0x559f38cf96c0 - <unknown>
   1:     0x559f38d2156f - <unknown>
   2:     0x559f38cf5955 - <unknown>
   3:     0x559f38cfba3b - <unknown>
   4:     0x559f38cfb50d - <unknown>
   5:     0x559f38cfc05d - <unknown>
   6:     0x559f38cfbbe7 - <unknown>
   7:     0x559f38cf9b5c - <unknown>
   8:     0x559f38cfbb49 - <unknown>
   9:     0x559f38924b51 - <unknown>
  10:     0x559f38924c43 - <unknown>
  11:     0x559f389981f7 - <unknown>
  12:     0x559f38ae61bd - <unknown>
  13:     0x559f389e832a - <unknown>
  14:     0x559f38a322ba - <unknown>
  15:     0x559f389ed103 - <unknown>
  16:     0x559f38ad65f9 - <unknown>
  17:     0x559f38cfc55a - <unknown>
  18:     0x559f38a36a92 - main
  19:     0x7f58bff8bb25 - __libc_start_main
  20:     0x559f3892531e - <unknown>
  21:                0x0 - <unknown>
```

```
$ ip addr
...
4: wlan0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether xx:xx:xx:xx:xx:xx brd ff:ff:ff:ff:ff:ff
    inet xx.xx.xx.xx/xx brd xx.xx.xx.xx scope global dynamic noprefixroute wlan0
       valid_lft 84293sec preferred_lft 84293sec
    inet6 xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx/xxx scope global dynamic noprefixroute
       valid_lft 49sec preferred_lft 19sec
    inet6 xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx/xxx scope global dynamic noprefixroute
       valid_lft 172792sec preferred_lft 86392sec
    inet6 xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx/xxx scope link noprefixroute
...
```

GladOSkar:
Resolves #1271

Use a regex to find only valid escape sequences, not just backslashes.

Everything else is kept literal, this should not be able to fail ever.

Feedback welcome

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -7,7 +7,7 @@ use std::time::{Duration, Instant};
 
 use crossbeam_channel::Sender;
 use lazy_static::lazy_static;
-use regex::bytes::Regex;
+use regex::bytes::{Captures, Regex};
 use serde_derive::Deserialize;
 
 use crate::blocks::{Block, ConfigBlock, Update};
@@ -657,21 +657,14 @@ struct IpAddrInfo {
 }
 
 fn decode_escaped_unicode(raw: &[u8]) -> String {
-    let mut result: Vec<u8> = Vec::new();
+    // Match escape sequences like \x2a or \x0D
+    let re = Regex::new(r"\\x([0-9A-Fa-f]{2})").unwrap();
 
-    let mut idx = 0;
-    while idx < raw.len() {
-        if raw[idx] == b'\\' {
-            idx += 2; // skip "\x"
-
-            let hex = std::str::from_utf8(&raw[idx..idx + 2]).unwrap();
-            result.extend(Some(u8::from_str_radix(hex, 16).unwrap()));
-            idx += 2;
-        } else {
-            result.extend(Some(&raw[idx]));
-            idx += 1;
-        }
-    }
+    let result = re.replace_all(raw, |caps: &Captures| {
+        let hex = std::str::from_utf8(&caps[1]).unwrap();
+        let byte = u8::from_str_radix(hex, 16).unwrap();
+        [byte; 1]
+    });
 
     String::from_utf8_lossy(&result).to_string()
 }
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
Isn't regrx somewhat overkill here? Anyway, if it works it works :laughing:

GladOSkar:
Sure, maybe. I thought about it and figured it made sense because:

- Regex is very fast in rust afaik
- Complexity shouldn't increase (still only O(n) i think?)
- It allows me to check for a fully valid escape sequence without any further checks
- Keeping the previous loop algorithm with all the checks would have been a lot longer and harder to read

MaxVerevkin:
By the way, can someone remind me why we need this in the first place? I believe that with https://github.com/greshake/i3status-rust/pull/1142 we actually don't need it. I just changed my SSID to contain an emoji and it turns out I don't actually need that block of code :eyes:

GladOSkar:
Interesting. I have no idea since i never used SSID display anyway. Maybe it's a legacy function? or is it because some APs escape their unicode SSIDs and some don't?

MaxVerevkin:
@GladOSkar 
> or is it because some APs escape their unicode SSIDs and some don't?

I don't think any AP does this. However,`iw dev <dev> link` does.

Since we don't use `iw` as a source of SSID anymore, I think it should be safe to ignore escape sequences. But I'm not sure.

GladOSkar:
That makes sense i guess. I'd be fine with it but i'll leave that decision to those of you that actually worked on that stuff :)

ammgws:
Thanks! I'll merge this for now and if someone wants to remove it later then go ahead!

PR #1277, opened 2021-07-15
networkmanager: Update DBus interface

### LINKED ISSUE #1269: Networkmanager block no longer updating with NetworkManager v1.32.0
After upgrading to NetworkManager 1.32.0 from 1.30.4, the NetworkManager block no longer updates on connection changes. Downgrading to 1.30.4 fixes the issue so there must be some breaking change in the new version.

mailhost:
Resolves #1269

### DIFF
--- a/src/blocks/networkmanager.rs
+++ b/src/blocks/networkmanager.rs
@@ -525,14 +525,14 @@ impl ConfigBlock for NetworkManager {
                 c.add_match(
                     "type='signal',\
                     path='/org/freedesktop/NetworkManager',\
-                    interface='org.freedesktop.NetworkManager',\
+                    interface='org.freedesktop.DBus.Properties',\
                     member='PropertiesChanged'",
                 )
                 .unwrap();
                 c.add_match(
                     "type='signal',\
                     path_namespace='/org/freedesktop/NetworkManager/ActiveConnection',\
-                    interface='org.freedesktop.NetworkManager.Connection.Active',\
+                    interface='org.freedesktop.DBus.Properties',\
                     member='PropertiesChanged'",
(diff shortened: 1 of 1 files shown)


ammgws:
What version of NetworkManager are distros like Debian on? This may fix it for some and break it for others. Perhaps it could be checked?

mailhost:
NetworkManager 1.2.0, released April 20, 2016
Debian Stretch (NetworkManager 1.6.2), released June 17, 2017
Ubuntu 16.04 (NetworkManager 1.2.0), released June 2, 2016
Slackware 14.2 (NetworkManager 1.2.2), released June 30 2016
Fedora 24 (NetworkManager 1.2.2), released June 21, 2016
CentOS 7.3 (NetworkManager 1.4.0), released November 4, 2016

ammgws:
So pretty much all of those would be broken by this change, right? Since they are all below version 1.32.

Can you do something similar to https://github.com/greshake/i3status-rust/commit/c9e75f2c22e0e7d2a81dc99de975898494a91918#diff-cd7ed854d8e55aa1b7f0392ca8e466d531fab1a143d5379d6973377294d167e4?
Using the output of nmcli --version to decide which dbus address to use.

MaxVerevkin:
But does it actually break stuff on older versions?
> * API CHANGE: D-Bus: remove long deprecated "PropertiesChanged" signal from D-Bus
>  API. They are replaced by standard "PropertiesChanged" signal on the
 > "org.freedesktop.DBus.Properties" interface. It is not expected that anybody
 > is still using this API.

ammgws:
That's true actually!

PR #1280, opened 2021-07-21
replace glyphs in material-nf which are overwritten arabic chars

### LINKED ISSUE #1206: `material-nf`: `net` block is printed in reverse order when loopback device is used
Output in the terminal seems OK.

## Config

```toml
icons = "material-nf"

[[block]]
block = "net"
format = "{signal_strength} {graph_down}{speed_down;K}{speed_up;K}"
interval = 1
```

## Video

https://user-images.githubusercontent.com/34583604/116897308-c9968700-ac3d-11eb-946d-ec0465214474.mp4

freswa:
NerdFonts has one big issue. They overwrite code points which are
reserved for chinese and arabic letters (range 0xf500-0xfd46).
While overwriting chinese letters is bad, they are LTR glyphs and print
"normally".
`net_loopback` and `update` used to be symbols that usually map
to arabic letters which make pango print those strings RTL. I
tried to fix this with the LTR marker 0x200e in #1206 but it failed.
This PR replaces the symbols with similar ones that don't map to
the arabic charset.
nerd-font issue: https://github.com/ryanoasis/nerd-fonts/issues/365
fixes #1206


ammgws:
Thanks! This is the best approach so far since it has been two years and #365 still has not been addressed.

MaxVerevkin:
> Thanks! This is the best approach so far since it has been two years and #365 still has not been addressed.

@ammgws How does #365 relate it this?

GladOSkar:
I assume they mean ryanoasis/nerd-fonts#365

PR #1281, opened 2021-07-21
fix `temperature` block fallback

### LINKED ISSUE #1273: `temperature` block fallback doesn't work
the `temperature` block uses `has_command` to check if the installed `sensors` command supports `-j`:

https://github.com/greshake/i3status-rust/blob/38dfb061dcd37655c273110fd70f6b4e3ffb7a01/src/blocks/temperature.rs#L157

but that isn't how `has_command` works – it only checks if the command is installed on the system:

https://github.com/greshake/i3status-rust/blob/952f06847ad01ef9d336a4431c9d7c6c599fc2b8/src/util.rs#L134-L146

so any system with lm-sensors installed will always pass the test, and the fallback case will never be used.

freswa:
The temperature block used the `has_command` to check for `-j` support.
But `has_command` only works if the binary can be executed. The commit
searches for json output support in the `--help` ouput to determine the
fallback property.

fixes #1273

### DIFF
--- a/src/blocks/temperature.rs
+++ b/src/blocks/temperature.rs
@@ -13,7 +13,6 @@ use crate::formatting::value::Value;
 use crate::formatting::FormatTemplate;
 use crate::protocol::i3bar_event::{I3BarEvent, MouseButton};
 use crate::scheduler::Task;
-use crate::util::has_command;
 use crate::widgets::{text::TextWidget, I3BarWidget, Spacing, State};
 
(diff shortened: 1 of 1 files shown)


ammgws:
Thanks! Did you come across this through actual usage or was it just a happenstance discovery?

freswa:
Due to another problem, I searched the issues and came across #1273. Couldn't resist to try a fix :)

Issue #1104, opened 2021-02-27
discussion: add more battery icons?


ammgws:
MaterialDesignIcons.com / Nerd Fonts have battery icons in 10% increments:

e.g. `nf-mdi-battery_10`, `nf-mdi-battery_20`, etc
https://pictogrammers.github.io/@mdi/font/5.4.55/
https://www.nerdfonts.com/cheat-sheet

Since extra icons for the backlight were added in #1095, I figure why not do this as well?

PR #1282, opened 2021-07-21
add more battery icons

freswa:
As nerd fonts has battery icons for every 10 % increment, this
commit makes the battery charge display more granular.

implements #1104


GladOSkar:
Why not do something like
```rust
let icon = &format!("bat_{}", (level / 10 + 1) * 10)
```
instead of that big map?

freswa:
I thought about it, but we return `'static &str` here, which doesn't allow things like
```
    match charge_level {
        Ok(level) if level <=90 && level >= 0 => &format!("bat_{}", (level / 10 + 1) * 10),
        _ => "bat_full",
    }
```
We can change the signature to `String` but that probably would involve changing the signature of `set_icon()` too...
In the end the map was the smallest change, but I'm down to change that :)

GladOSkar:
Ah that makes sense. Was just wondering

src/util.rs
```diff
@@ -85,10 +85,15 @@ pub fn escape_pango_text(text: String) -> String {
 
 pub fn battery_level_to_icon(charge_level: Result<u64>) -> &'static str {
     match charge_level {
-        Ok(0..=5) => "bat_empty",
-        Ok(6..=25) => "bat_quarter",
-        Ok(26..=50) => "bat_half",
-        Ok(51..=75) => "bat_three_quarters",
+        Ok(0..=10) => "bat_10",
```
ammgws, line 88: This is essentially a breaking change, isn't it? Anyway around it?

freswa: Unless we can detect an older icon mapping scheme, there seems to be no way of backward compatibility.
I thought about doing something like
```
        Ok(0..=10) => "bat_empty",
        Ok11..=20) => "bat_20",
        Ok(21..=30) => "bat_quarter",
```
But that would lead to unknown icons too. If you can think of any solution to this, I'm happy to implement it :)

ammgws: Perhaps you could check whether the icon exists, and if not revert to the previous icon.

freswa: I ran down that path. To check for icons, we need a copy of `SharedConfig` (at least if we don't want to parse the config again). The instance is created in `main.rs` and then passed to every block. So either we pass it from the block to `utils.rs` or we create another config, which would be a lot of code duplication.
But we could update the `get_icon` error, which seems to be the instance that handled upgrades like these in the past.
Atm it's this line:
`internal_error("get_icon()", &format!("icon '{}' not found in your icons file. If you recently upgraded to v0.2 please check NEWS.md.", icon))?`
We could change it accordingly. What do you think?

ammgws: I was thinking to just check it in the block where it's called, which right now I think is only battery and kdeconnect.

MaxVerevkin: I suppose @ammgws meant that you can insert the following code into `new()` functions of battery and kdeconnect blocks
```rust
let fallback = shared_config.get_icon("bat_empty").is_some() || shared_config.get_icon("bat_half").is_some(); // and so on;
```
Also, if `fallback` is `true`, you can print a warning. Or even run `notify-send <warning message>`, because I doubt anyone reads stderr.

freswa: We've got config section for the notify daemon used in the `pomodoro` block. If we want to issue notifications globally we should introduce a global config for this imho.
I added the requested code in `battery.rs` and `kdeconnect.rs`. Hope this is what you intended :)

ammgws (approved):
This should be OK

ammgws:
Thanks!

Issue #469, opened 2019-11-21
temperature: read values from `/sys` instead of parsing `sensors -u`


The-King-of-Toasters:
I thought it was weird that the program was parsing `sensors -u` instead of reading values from `/sys` like `sensors` does. [A `strace`](https://www.sgregoratto.me/paste/sensors.log.txt) shows that it reads values from `/sys/class/hwmon/hwmon4` on my machine at least. We can confirm this with the following script:

```sh
while true; do
	pr -mtw11 <(sensors -u coretemp-isa-0000 | awk '/temp._input/{print $2 * 1000}') \
               <(cat /sys/class/hwmon/hwmon4/temp?_input)
	echo
	sleep 1
done
```

The-King-of-Toasters:
To test this, I wrote a PoC Go that outputs something similar to `sensors -u`

```go
package main

import (
	"fmt"
	"path/filepath"
	"io/ioutil"
	"bytes"
	"strings"
	"strconv"
	"sort"
)

type CoreTemp struct {
	input, max, crit, crit_alarm int
	label string
}

func (c *CoreTemp) Print() {
	fmt.Printf(`%s:
  input: %d.000
  max: %d.000
  crit: %d.000
  crit_alarm: %d.000
`, c.label, c.input, c.max, c.crit, c.crit_alarm)
}

func readTemps(path string) {
	matches, err := filepath.Glob(path + "/temp*")
	if err != nil {
		panic(err)
	}
	sort.Strings(matches)

	ct := CoreTemp{}
	for i, f := range(matches) {
		if (i + 1) % 5 == 0 {
			ct.Print()
		}

		buf, err := ioutil.ReadFile(f)
		if err != nil {
			panic(err)
		}
		// remove trailing newline
		buf = buf[0:len(buf) - 1]

		var field *int
		switch {
		case strings.HasSuffix(f, "input"):
			field = &ct.input
		case strings.HasSuffix(f, "max"):
			field = &ct.max
		case strings.HasSuffix(f, "crit"):
			field = &ct.crit
		case strings.HasSuffix(f, "crit_alarm"):
			field = &ct.crit_alarm
		case strings.HasSuffix(f, "label"):
			ct.label = string(buf)
			fallthrough
		default:
			continue
		}

		num, err := strconv.Atoi(string(buf))
		if err != nil {
			panic(err)
		}
		*field = num / 1000
	}
}

func main() {
[...]

PR #1286, opened 2021-08-02
Github, Nvidia_gpu, Temperature block additions and changes

ZachCook:
Changed Nvidia_gpu block to use the interval argument of nvidia-smi, and added sysfs backend for Temperature block, as I personally dislike constantly spawning lots of processes if it can be avoided. Some relevant discussion in #469.

Added colour theme support to Github block to make it much easier to tell when it changes. Allows setting different colours depending on notification type.

Change recommended pacman hook to use SIGUSR1 to merely force an update rather than restarting.


ammgws:
Looks reasonable to me.

Thanks!

MaxVerevkin:
Is there a reason to keep "sensors" driver?

ZachCook:
> Is there a reason to keep "sensors" driver?

I didn't remove the sensors driver since it has different naming format so every user would need to change their config

PR #1287, opened 2021-08-05
http: Escape URLs before they are used

rbuch:
Previously, URLs with spaces, etc. (e.g. a city name with spaces when `autolocate=true` for the Weather block) would fail inside curl.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ serde_json = "1.0"
 signal-hook = "0.3"
 swayipc = "2.7"
 toml = "0.5"
+url = "2.2"
 
 # Optional features/blocks
 libpulse-binding = { optional = true, version = "2.0", default-features = false }
--- a/src/http.rs
+++ b/src/http.rs
@@ -1,6 +1,7 @@
 use curl::easy::Easy;
 use serde_json::value::Value;
 use std::time::Duration;
+use url::Url;
 
 use crate::errors;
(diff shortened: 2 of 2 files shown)


ammgws:
Can we implement the encoding function ourselves to save adding a new crate?

rbuch:
> Can we implement the encoding function ourselves to save adding a new crate?
> 
> Or perhaps use something more lightweight: https://crates.io/crates/urlencoding

Yes, we could implement it directly here, in fact,  `url.replace(" ", "%20");` is enough to fix the issue I'm seeing. I chose to use `url` instead for the sake of robustness, since it should handle other characters that should be encoded, like diacritics.

Something like the urlencoding crate would also work, but it's a little less clean to integrate it; since it'll encode everything except alphanumerics and `-_.~`, running an entire URL through it will give you results like `https%3A%2F%2Fapi.openweathermap.org`, which obviously causes failures, so it would instead need to be applied on the individual pieces of the URL before they're assembled, so the calls would then have to go inside `blocks/weather.rs`, which seemed to be crossing the abstraction boundary a bit too much.

In any case, if you think using the simpler approach is better, then I can move this stuff into `blocks/weather.rs`.

ammgws:
Yeah let's go with the trivial fix for now and if the need arises we can revisit your initial approach.

ammgws:
I'll close this in anticipation of a new PR for the simple fix.

PR #1289, opened 2021-08-09
http: Replace spaces in URL with "%20"

rbuch:
Fixes failures in Weather block when place names have a space in them. Simplified version of #1287 based on feedback received there.


ammgws:
Thanks!

PR #1290, opened 2021-08-09
Weather: Remove duplicate degree symbol from default format

rbuch:
A degree symbol is added by the Unit formatter already, so there should
not be an additional one added in the default format string.


ammgws:
Good catch!

Issue #1081, opened 2021-02-19
brightness block: icon order is backwards


hacmorgan:
On full brightness, the icon shown is of a black moon, while on min brightness the icon is a full moon. I'm not sure if this is intentional, but it seems backwards to me. 

icon set: awesome
OS: Arch Linux 
i3status-rust version: i3status-rs 0.14.5 (commit 1e40e8f 2021-02-15)

The screenshots below are after I reversed the order of the icons, so this is what I think it **should** look like:
![image](https://user-images.githubusercontent.com/19177940/108568932-1d901000-735f-11eb-92df-832eae16ec36.png)
![image](https://user-images.githubusercontent.com/19177940/108569013-46b0a080-735f-11eb-884e-5fb183ae4e91.png)

ammgws:
See #845

ammgws:
Though perhaps since this keeps coming up, maybe an option to flip out would be good?

hacmorgan:
Oh sorry, I didn't check past issues, that was dumb. I will try to implement an option to flip them, I think that would be cool.

Issue #845, opened 2020-10-05
Backlight block - awesome icons flipped


karthink:
The font-awesome icons used to indicate backlight brightness are flipped, with the darkest "moon phase" corresponding to the brightest setting:

![2020-10-05-124636_122x26_scrot](https://user-images.githubusercontent.com/8607532/95113946-cfa8a180-0708-11eb-9a04-d9849e1cd02d.png)
![2020-10-05-124704_107x27_scrot](https://user-images.githubusercontent.com/8607532/95113997-e0f1ae00-0708-11eb-8b16-8f13fa7eb928.png)
![2020-10-05-124730_89x26_scrot](https://user-images.githubusercontent.com/8607532/95114034-ef3fca00-0708-11eb-9754-5708833952cd.png)
![2020-10-05-124746_84x26_scrot](https://user-images.githubusercontent.com/8607532/95114060-f961c880-0708-11eb-9f3a-502997c141f9.png)

I am not sure if this is intentional. Seems to me the progression should be flipped. I am using `ttf-font-awesome` version `5.14.0-1` from the Arch repos.

ammgws:
See #573

PR #1291, opened 2021-08-10
Backligth Icons: Correct backlight order

alt-art:
In the old backligth icons the sun is 5% and the moon is 100%, i just replace the order to make sun 100% and moon 5%


ammgws:
Just to be sure, does the `invert_icons` option help you?
See https://github.com/greshake/i3status-rust/blob/master/doc/blocks.md#options-1

Past issues: 
- #183
- #452
- #573
- #845
- #1081

alt-art:
Was bad I didn't know I had an option for that, You know that having an option for this is kind of meaningless, I mean you know that everywhere the sun gets hot right?

alt-art:
And by default it should be like I said, it only happens on awesome and awesome5

alt-art:
You ended up making people believe they were wrong

ammgws:
Let's keep it civil.

@GladOSkar Looking back at this, what do you reckon now? (I don't use this block daily)

alt-art:
The fact that they have an option to invert is super cool but that won't change the fact that the order is wrong, sorry if it felt rude

MaxVerevkin:
@alt-art Do you have colorful icons? Can you post a short GIF in with you change your backlight through all the values? I have monochrome font and it both `awesome` and `awesome5` work for me.

GladOSkar:
Hi @alt-art - The reason these icons are the way they are is because they stem from a time when colored emoji weren't a thing on most systems. When using these icons with a monospaced font with a light text color on a dark background, it actually looks right that way.

Therefore, when color emoji started to become more prevalent, the option was introduced to allow those to look right as well while still maintaining backwards compatibility for existing setups. This is why just changing the icons is not an option - it would break the setups of everyone who uses light monochrome fonts on a dark background (which is actually still most people AFAIK).

MaxVerevkin:
Btw, dark text on light background works too.

GladOSkar:
Yeah in that case it's kinda ambiguous last time I checked it, it's really down to interpretation what side of the icons is "bright" and which is "dark" then

alt-art:
Some icons are small, not all

MaxVerevkin:
@alt-art
> Would it be possible to increase an icon without increasing the font?

Yes, you can try the following config option

```toml
icons_format = " <span size='x-large'>{icon}</span> "
```

For more info, refer to [pango docs](https://docs.gtk.org/Pango/pango_markup.html)

PR #1296, opened 2021-08-27
Fix nvidia block falling behind on lines from nvidia-smi

ZachCook:
ab2f7bbe2e changed to running nvidia-smi once and reading lines one at a time
but any delay in running update would lead to multiple lines being buffered
and never catching up, if any other blocks have a long running time or with
a non integer interval it would fall further behind every update

Read over all lines currently in buffer instead and use the most recent


ammgws:
Thanks!

Issue #1268, opened 2021-06-28
Missing `libssl-dev` in requirements


MiniGod:
I got the following error on a fresh Ubuntu 20.04 install:

```
error: failed to run custom build command for `openssl-sys v0.9.65`

Caused by:
  process didn't exit successfully: `/home/kristjan/git/i3status-rust/target/release/build/openssl-sys-a14e45a2013c726c/build-script-main` (exit status: 101)
  --- stdout
  cargo:rustc-cfg=const_fn
  cargo:rerun-if-env-changed=X86_64_UNKNOWN_LINUX_GNU_OPENSSL_LIB_DIR
  X86_64_UNKNOWN_LINUX_GNU_OPENSSL_LIB_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_LIB_DIR
  OPENSSL_LIB_DIR unset
  cargo:rerun-if-env-changed=X86_64_UNKNOWN_LINUX_GNU_OPENSSL_INCLUDE_DIR
  X86_64_UNKNOWN_LINUX_GNU_OPENSSL_INCLUDE_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_INCLUDE_DIR
  OPENSSL_INCLUDE_DIR unset
  cargo:rerun-if-env-changed=X86_64_UNKNOWN_LINUX_GNU_OPENSSL_DIR
  X86_64_UNKNOWN_LINUX_GNU_OPENSSL_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_DIR
  OPENSSL_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=OPENSSL_STATIC
  cargo:rerun-if-env-changed=OPENSSL_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH
[...]

PR #1300, opened 2021-08-31
Update compilation requirements according to #1268

Henriquelay:
This has been tested by using a Docker image to replicate a fresh Ubuntu install. I chose to do it on Ubuntu since it's the distro mentioned in #1268, and is also a very common distro.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ For a list of available blocks, see the [block documentation](https://github.com
 
 ## Requirements
 
-The Rust language and the `cargo` package manager are required to build the binary.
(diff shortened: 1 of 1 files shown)


ammgws:
Thanks!

Henriquelay:
Should the library name `openssl-sys` be swapped for the package name, `libssl-dev`? Knowing this is package-manager-specific, I chose to use the library name itself instead of package name, even tough they tend to be similar.

ammgws:
@Henriquelay Right below that it mentions `libdbus-1-dev` which is package manager specific, so I guess we could either match that or change the dbus one to be package manager agnostic.

PR #1301, opened 2021-09-01
Display "available" memory instead of "free" memory in examples

buckley310:
The example configurations come with a memory block which displays total memory used, including buffers, cache, and ZFS ARC. I would imagine the majority of people do not want this, and would rather see the amount of memory available for applications.

It appears the example used to work this way, but was changed in https://github.com/greshake/i3status-rust/commit/04105c7535e1fc6100ce76bb4bb2bfc4f69d8656 (perhaps accidentally?) to include buffers/cache.

`Mup == mem_used_percents`
`MUp == mem_total_used_percents`
https://github.com/greshake/i3status-rust/blob/v0.20.2/NEWS.md

### DIFF
--- a/examples/config.toml
+++ b/examples/config.toml
@@ -14,7 +14,7 @@ alert = 10.0
 [[block]]
 block = "memory"
 display_type = "memory"
-format_mem = "{mem_total_used_percents}"
+format_mem = "{mem_used_percents}"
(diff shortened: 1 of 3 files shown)


ammgws:
Fair enough

Thanks!

PR #1302, opened 2021-09-05
refactor: FormatTemplate::render(): accept generic hashmap

MaxVerevkin:
https://github.com/greshake/i3status-rust/blob/db144a00a29c8fabaee7cf22dbd7984401ed0619/src/blocks/cpu.rs#L188

https://github.com/greshake/i3status-rust/blob/db144a00a29c8fabaee7cf22dbd7984401ed0619/src/blocks/cpu.rs#L195

Yes, there is.

### DIFF
--- a/src/blocks/cpu.rs
+++ b/src/blocks/cpu.rs
@@ -179,26 +179,21 @@ impl Block for Cpu {
             _ => String::new(),
         };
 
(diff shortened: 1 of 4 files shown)


ammgws:
Thanks as always!

PR #1310, opened 2021-09-25
allow non-default docker setups

### LINKED ISSUE #1309: docker block doesn't work for rootless-mode or other, non-standard socket locations
Not really a bug. The socket path is hardcoded [here](../blob/master/src/blocks/docker.rs#L85) so anyone using non-standard paths can't really use the block.

I think it'd be reasonable (and reasonably straightforward?) to expose that path as part of the config (with `/var/run/docker.sock` being the default). I'd be happy to take a pass at it if there's interest.

JTarasovic:
Allows configuration of the docker socket path so that non-default installations such as rootless-mode can use to block.

fixes #1309 

```
[[block]]
block = "docker"
format = "T:{total} R:{running} S:{stopped} P:{paused} I:{images}"
interval = 60

[[block]]
block = "docker"
format = "T:{total} R:{running} S:{stopped} P:{paused} I:{images}"
interval = 60
socket_path = "/run/user/1001/docker.sock"
```

### DIFF
--- a/doc/blocks.md
+++ b/doc/blocks.md
@@ -534,6 +534,7 @@ Key | Values | Required | Default
 ----|--------|----------|--------
 `interval` | Update interval, in seconds. | No | `5`
 `format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{running}"`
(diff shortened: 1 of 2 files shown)


ammgws:
Thanks!

PR #1311, opened 2021-09-25
Block for DNF updates (Fedora packages)

sigvei:
A block for DNF. Much of the code is copied from the src/blocks/apt.rs.

Some of it is a bit simpler, though, as dnf allows unprivileged users to update the database and includes its own caching logic.

I guess these modules, and pacman, could be refactored into some shared logic - it's a lot of repeated code in these three modules now. Not sure if the abstraction is worth it, though.

### DIFF
--- a/doc/blocks.md
+++ b/doc/blocks.md
@@ -515,6 +515,53 @@ Key | Value | Type
 
 ###### [↥ back to top](#list-of-available-blocks)
 
+## Dnf
+
+Creates a block which displays the pending updates available for your Fedora system.
+
+#### Examples
+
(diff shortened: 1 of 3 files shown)


sigvei:
I realise there is probably some way to implement most of this with `custom`, but the warning/critical filtering would require quite a hacky script and jq or something as a dependency.

sigvei:
Re: #715

ammgws:
Thanks!

I actually wouldn't mind consolidating everything in one block at some stage due to all the shared code and the fact that it can easily be extended to include other package managers without polluting the list of blocks. Perhaps someone would like to give it a go.

PR #1313, opened 2021-10-05
Add optional headphones indicator to sound block

codicodi:
This allows to dynamically change sound block icon when headphones are plugged in/out. I've made it off-by-default since it may not work for people with weird audio setups.


ammgws:
Seems reasonable, thanks!

I had been doing it this way since also had some other speaker types:

https://github.com/ammgws/dotfiles/blob/c5921e407bc0caea41dac20347b8fe62288e722b/i3status-rs/.config/i3status-rust/config.toml#L82-L89

Issue #1322, opened 2021-10-26
Support inotify in Custom block


BrendanBall:
Hi

I currently have a Custom block that has an interval of 1s  to update a status. 
If the Custom block supports inotify, then I could remove the interval which would be pretty great since this status doesn't update that often.
This is actually pretty similar to the Backlight block that already uses inotify.

Would you accept this addition to the Custom block?
Since signals have been added to this block already, it makes sense to support inotify for even more use cases.

PR #1325, opened 2021-10-30
[custom block] Add watch files feature with inotify

BrendanBall:
See [#1322](https://github.com/greshake/i3status-rust/issues/1322).
This basically copies and extends the inotify implementation for the backlight block to the custom block.
Currently it doesn't normalize paths, so only absolute paths are supported.
This means you can't do e.g. `~/file_to_watch` or `$HOME/file_to_watch`.
I'm not sure what's the best way to support this, since it seems we'd have to add a dependency.


src/blocks/custom.rs
```diff
@@ -101,6 +108,42 @@ impl ConfigBlock for Custom {
...
+            let paths: Vec<PathBuf> = paths.iter().map(|p| PathBuf::from(p)).collect();
+            let tx_inotify = custom.tx_update_request.clone();
+            thread::Builder::new()
+                .name("custom".into())
+                .spawn(move || {
+                    let mut notify = Inotify::init().expect("Failed to start inotify");
+                    for p in paths {
+                        notify
+                            .add_watch(p, WatchMask::MODIFY)
+                            .expect("Failed to watch file");
```
MaxVerevkin, line 122: I think the initialization code should be moved to the main thread so that it'll be possible to handle this error. `Inotify` is `Send` and `Sync` so this shouldn't be a problem.

src/blocks/custom.rs
```diff
@@ -101,6 +108,42 @@ impl ConfigBlock for Custom {
             custom.signal = Some(convert_to_valid_signal(signal)?);
         };
 
+        if let Some(paths) = block_config.watch_files {
+            // If the signal is not in the valid range we return an error
```
MaxVerevkin, line 112: Which signal :smile: ?

src/blocks/custom.rs
```diff
@@ -101,6 +108,42 @@ impl ConfigBlock for Custom {
             custom.signal = Some(convert_to_valid_signal(signal)?);
         };
```
ammgws, line 109: Wrong comment?

ammgws (commented):
What dependency would be needed for resolving paths?

src/blocks/custom.rs
BrendanBall: done, it now returns an error:
```
Configuration error: custom. Cause: Failed to watch file $HOME/test_i3status: No such file or directory (os error 2)
```

BrendanBall:
It seems like [shellexpand](https://docs.rs/shellexpand/2.1.0/shellexpand/) is the right dependency to use

BrendanBall:
I've added `shellexpand`. It's at least a very minimal dependency so hopefully you're fine with that.

ammgws:
Nice timing, I was literally just typing a comment asking if you could implement it in a separate commit.

ammgws:
Should the default update interval be ignored when watching files?

BrendanBall:
That's a good question. I would say yes, but I'm not sure how easy it would be to implement currently? I think the block implementation can't distinguish between default and set config?
I don't think this is a big deal since the user can easily set the interval to `once` or something else.

ammgws:
Maybe update the example in the docs to use 'once' then

BrendanBall:
done

ammgws:
> I think the block implementation can't distinguish between default and set config? 

In update() can't you just do something like
```rs
if self.watch_mode (add to pub struct custom)
    Ok(None)
else
    Ok(Some(self.update_interval.clone()))
```
?

BrendanBall:
If we're going to do that then we should probably have signal and watch_files behave the same regarding update interval.

ammgws:
That's a good point. In my mind `signal` was for users to arbitrarily force a block update (for example their interval might be set to 1 hour but they want to trigger an update right now), but I guess people might use it in a script to update the block periodically. Where as in my mind watch_files would be the primary update mechanism so figured interval should be turned off.

Anyway I will leave that for another time since it shouldn't affect the implementation of the filewatch feature.

ammgws:
Thanks!

BrendanBall:
ye, I guess it's possible for people to use it in different ways, so it's probably better to just leave it like this. 
Thanks for merging :D

Issue #534, opened 2020-01-29
Panics when "focused_window" is the only block configured


ammgws:
As per title.

<details close>
<summary>config.toml contents</summary>
<br>

```toml
theme = "solarized-dark"
icons = "awesome"

[[block]]
block = "focused_window"
```
</details>

<details close>
<summary>stdout at time of panic:</summary>
<br>

```
{"version": 1, "click_events": true}
[[],
thread 'main' panicked at 'overflow when adding duration to instant', src/libcore/option.rs:1185:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```
</details>

<details close>
<summary>backtrace:</summary>
<br>

```
{"version": 1, "click_events": true}
[[],
thread 'main' panicked at 'overflow when adding duration to instant', src/libcore/option.rs:1185:5
stack backtrace:
   0:     0x557fb6fcace4 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h4c5e39dd6bf38933
   1:     0x557fb6fedc8c - core::fmt::write::h94b22f3cbda6422c
   2:     0x557fb6fc79a7 - std::io::Write::write_fmt::he7e02466cb3dc860
   3:     0x557fb6fcd38e - std::panicking::default_hook::{{closure}}::h077f2d0f0f4550db
   4:     0x557fb6fcd081 - std::panicking::default_hook::hcc1ceebfaf9d0a58
   5:     0x557fb6fcda8b - std::panicking::rust_panic_with_hook::h328d8e7ed959397b
   6:     0x557fb6fcd62e - std::panicking::continue_panic_fmt::h745b12464b662c8f
   7:     0x557fb6fcd516 - rust_begin_unwind
   8:     0x557fb6fe9ace - core::panicking::panic_fmt::hdb91b8826feb6fc9
   9:     0x557fb6fe9b37 - core::option::expect_failed::h02ca44b00cc41c7f
  10:     0x557fb6fca879 - <std::time::SystemTime as core::ops::arith::Add<core::time::Duration>>::add::h93fc4eab81208764
  11:     0x557fb6dad0b3 - crossbeam_channel::flavors::after::Channel::new::h7cf3b0b56e89f831
                               at /home/ammgws/.cargo/registry/src/github.com-1ecc6299db9ec823/crossbeam-channel-0.4.0/src/flavors/after.rs:31
[...]

PR #1327, opened 2021-10-31
Fix music marquee updates after closing player

jamesmcm:
Previously closing the player would allow no further updates to be scheduled, since song change and right-click events would only be processed directly and not add to the scheduled queue (they could not schedule further events - necessary for the marquee).

This is fixed by making these events add to the scheduler queue _if_ the block does not currently have any update scheduled. So simply playing a new song or right-clicking the block will make the marquee start scheduling events again :)

See issue #1124 for more details.

This has been tested with the music block and seems to work well, it may be worth quickly checking other blocks to make sure that none of them rely upon the `update_requests` channel never scheduling further events.

Note that writing to the scheduler directly like this is okay for now since it runs on the same thread.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -219,13 +219,13 @@ fn run(matches: &ArgMatches) -> Result<()> {
                 blocks.get_mut(req.id)
                     .internal_error("scheduler", "could not get required block")?
                     .update()?;
-                protocol::print_blocks(&blocks, &shared_config)?;
                 } else {
                 // Otherwise add to scheduler tasks and trigger update
                 // In case this needs to schedule further updates e.g. marquee
                 scheduler.schedule.push(req);
                 scheduler.do_scheduled_updates(&mut blocks)?;
                 }
+                protocol::print_blocks(&blocks, &shared_config)?;
             },
             // Receive update timer events
             recv(ttnu) -> _ => {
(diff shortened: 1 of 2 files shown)


ammgws:
Does the Custom DBus block still work for you after this?

Try a config of just:
```toml
[[block]]
block = "custom_dbus"
name = "test"
```

and run `busctl --user call i3.status.rs /test i3.status.rs SetStatus s 123` in another terminal.

ammgws:
Actually I believe it will occur with any block that does not schedule updates in its `update` function. I can reproduce with only the music block as well.

jamesmcm:
@ammgws it seems to work fine for the custom dbus block with the patch applied still.

ammgws:
With just that one block in the config?

jamesmcm:
Oh damn, in that case it doesn't. Is the scheduler not used in that case?

ammgws:
Actually I remember something related from a while back:
https://github.com/greshake/i3status-rust/issues/534 -> https://github.com/greshake/i3status-rust/pull/535 & https://github.com/greshake/i3status-rust/pull/582 -> https://github.com/greshake/i3status-rust/pull/685/

jamesmcm:
Actually it works fine, it was just missing the `print_blocks` call in that case.

ammgws:
Indeed it was!

Thanks!

Issue #979, opened 2020-12-14
block idea: Fifo/Local Messages


BourgondAries:
A block that shows any line sent to a fifo.

Use-case: When changing i3's pane mode (tabbed/horizontal/vertical/stacked), I normally send a message using `notify-send` just as a little feedback mechanism, however, this overlaps completely with the (top) i3bar and is rather ugly. It would be better if this kind of feedback were integrated into i3status-rust.

Solution: Instead just echo a line into a local fifo and show it integrated into `i3status-rust` for a user-defined amount of time, or keep it around (depending on the block config).

GladOSkar:
Wouldn't this be very easy with a custom block? Just write your message to a temp file and cat it in the custom block, optionally send a signal to reload immediately. Or use a custom dbus block which is probably even easier

BourgondAries:
@GladOSkar it's a bit of bookkeeping to use a custom block, for instance, clearing the block 3 seconds after the last update becomes a bit of work. You'd need some kind of scheduler such that if 2 notifies happen in short succession, that the first doesn't clear the second. Also requires a script to notify and send a signal for updating.

By having a dedicated block, this is all made easier for the user, it's as simple as `echo Hey > /tmp/myfifo`.

GladOSkar:
Sounds like the custom dbus block would be perfect for that

BourgondAries:
> Sounds like the [custom dbus](https://github.com/greshake/i3status-rust/blob/master/blocks.md#custom-dbus) block would be perfect for that:
> 
> ```
> [[block]]
> block = "custom_dbus"
> name = "MyDBus"
> ```
> 
> `qdbus i3.status.rs /MyDBus i3.status.rs.SetStatus Hey`

That looks useful, is there any easy way to clear this based on time, without potentially clearing right after a new message has been set?

GladOSkar:
Nah i can only think of

```
qdbus i3.status.rs /MyDBus i3.status.rs.SetStatus Hey
sleep 3
qdbus i3.status.rs /MyDBus i3.status.rs.SetStatus There
```

ammgws:
Hmm would a new config for the custom blocks be useful?

i.e. "clear_timeout" or something which will clear the output of the block after x seconds.

Combined with the hide when empty option it could allow the block to act as an adhoc notification

ammgws:
Or maybe the timeout could be specified via the existing JSON option. That way you could customise it on the fly.

BourgondAries:
[...]

PR #1329, opened 2021-11-06
custom_dbus: add option to clear contents after timeout expires

ammgws:
For #979

### DIFF
--- a/src/blocks/custom_dbus.rs
+++ b/src/blocks/custom_dbus.rs
@@ -11,6 +11,7 @@ use serde_derive::Deserialize;
 
 use crate::blocks::{Block, ConfigBlock, Update};
 use crate::config::SharedConfig;
+use crate::de::deserialize_opt_duration;
 use crate::errors::*;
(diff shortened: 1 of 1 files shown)


src/blocks/custom_dbus.rs
```diff
@@ -27,12 +28,18 @@ pub struct CustomDBus {
...
 }
 
 #[derive(Deserialize, Debug, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct CustomDBusConfig {
     pub name: String,
+
+    /// Timeout for clearing the block output after an update (in seconds)
+    #[serde(default, deserialize_with = "deserialize_opt_duration")]
+    pub timeout: Option<Duration>,
```
MaxVerevkin, line 42: Perhaps it could be added as an optional parameter to the DBus API?

ammgws: Sure. Though I think the current API needs some reworking so you can set it without having to supply icon/state as well.

Will leave it for later

PR #1330, opened 2021-11-06
custom_dbus: allow initial text to be customised

ammgws:


src/blocks/custom_dbus.rs
```diff
@@ -37,11 +37,24 @@ pub struct CustomDBus {
...
+
     /// Timeout for clearing the block output after an update (in seconds)
     #[serde(default, deserialize_with = "deserialize_opt_duration")]
     pub timeout: Option<Duration>,
 }
 
+impl Default for CustomDBusConfig {
+    fn default() -> Self {
+        Self {
+            name: Default::default(),
```
MaxVerevkin, line 51: Shouldn't `name` be a required parameter? (Like `mac` in `bluetooth`)

ammgws: Indeed

PR #1335, opened 2021-11-07
update to latest release of openssl-sys which supports OpenSSL 3.0.0

smcavoy:


ammgws:
Why?

smcavoy:
> Why?

Specifically Alpine Linux edge.  The upcoming stable release of Alpine Linux will also use openssl 3.0.0 by default. 
The upstream crate supports it.

ammgws:
If we really need to update the lock file then I think the correct way would be to run `cargo update` instead. You can send a PR for that, but if we didn't do anything will it mean the package will be broken/fail to build on Alpine?

smcavoy:
> If we really need to update the lock file then I think the correct way would be to run `cargo update` instead. You can send a PR for that, but if we didn't do anything will it mean the package will be broken/fail to build on Alpine?

The PR was the result of 
` cargo update -p openssl-sys:0.9.63 --precise 0.9.70`


Alpine edge has the package at version [0.20.5](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/i3status-rust). This is with the [patch](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/27229/diffs) to update `openssl-sys`.  


Generally where a patch is needed to get the package into the repo, upstream will be notified. When the next release of `i3status-rust` is out someone will see if the patch is still needed to build against the default OpenSSL. 

If you can integrate it that would be great. I run `i3status-rust` with the patch on Alpine Linux (aarch64)

ammgws:
Done in 4b5140fdc77630eaaaa83356e75a5d7c0bd937b6. There should hopefully be another release this month.

Again to clarify, what happens on Alpine Linux edge (using OpenSSL v3 by default) when you try to build and/or use i3status-rust without updating `openssl-sys`?

smcavoy:
> Done in [4b5140f](https://github.com/greshake/i3status-rust/commit/4b5140fdc77630eaaaa83356e75a5d7c0bd937b6). There should hopefully be another release this month.

Thanks!

> 
> Again to clarify, what happens on Alpine Linux edge (using OpenSSL v3 by default) when you try to build and/or use i3status-rust without updating `openssl-sys`?

The error the package building systems ran into was: 
`Failed to find OpenSSL development headers` [line 438](https://gitlab.alpinelinux.org/smcavoy/aports/-/jobs/532503#L438).

Since my PR Alpine edge has reverted to `1.1.1l`  for `openssl-dev`. 

This means `i3status-rust 0.20.5` will build on Alpine edge with the existing version `0.9.63` of `openssl-sys`.

So the change was not necessary except of course when the default version of `openssl`  in Alpine edge is switched to 3 at some point in the future. 

Thanks!

PR #1337, opened 2021-11-07
hueshift block: Add support for wlsunset

### LINKED ISSUE #1316: Add support for wlsunset to hueshift
[wlsunset](https://git.sr.ht/~kennylevinsen/wlsunset) is a simple gamma shifter (like redshift, et al.) for wlroots based Wayland compositors. It would be nice to add support for it in the hueshift block. `wlsunset -T` sets the high color temperature, `wlsunset -t` sets the low color temperature, and `wlsunset -g` sets the gamma.

DerVerruckteFuchs:
This commit adds support for wlsunset and should close  #1316.

As I was reading how redshift/gammastep and sct do things so that I could make the behavior for wlsunset as simiar as possible, there were a couple of things to note.
1. While wlsunset has a gamma option, there does not seem to be an option to reset the gamma ramps like redshift/gammastep. I skimmed through wlsunset's code, and it seems like it might handle that automatically. I am going to leave the gamma CLI argument alone for now.
2. redshift/gammastep have a color reset option, sct and wlsunset do not. On the [github README for sct](https://github.com/faf0/sct), it states "_If `xsct` is called without parameters, **the current display temperature is estimated**._" (emphasis mine) I do not know what this means at a glance. I suppose if I used xsct, then I may figure that out. Calling wlsunset without arguments, like sct, does work. For wlsunset, the default color temperature and gamma values in its `config` struct work out fine. However, both `latitude` and `longitude` in its `config` struct are set to `NaN`. Running wlsunset without the latitude and longitude arguments works, but `sun_condition` gets set to `POLAR_NIGHT`. The value of `latitude`/`longitude` eventually is interpreted as a float, so the default `NaN` value is getting interpreted as a float. Until the hueshift block takes options for latitude and longitude, the solution that seems to work right now is just killing the wlsunset process when resetting the color temperature.

### DIFF
--- a/src/blocks/hueshift.rs
+++ b/src/blocks/hueshift.rs
@@ -101,6 +101,37 @@ impl HueShiftDriver for Gammastep {
         Ok(())
     }
 }
+struct Wlsunset();
+impl HueShiftDriver for Wlsunset {
+    fn update(&self, temp: u16) -> Result<()> {
+        Command::new("sh")
+            // wlsunset does not have a oneshot option, so set both day and
+            // night temperature. wlsunset dose not allow for day and night
+            // temperatures to be the same, so increment the day temperature.
+            .args(&[
+                "-c",
+                &format!("killall wlsunset; wlsunset -T {} -t {} &", temp + 1, temp),
+            ])
+            .spawn()
+            .block_error(
+                "hueshift",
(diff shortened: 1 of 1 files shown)


ammgws:
Thanks!

I will go ahead and merge this since it looks to be usable even with the caveats you mentioned.

PR #1348, opened 2021-11-16
sound: detect more portable speaker types

xentec:
Add detection for headset, earpiece and handsfree devices.
Taken from https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/blob/0ce3008605e5f644fac4bb5edbb1443110201ec1/src/pulse/def.h#L1073-1100

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -955,7 +955,7 @@ impl Block for Sound {
         let headphones = self
             .device
             .active_port()
-            .map(|p| p.contains("headphones"))
+            .map(|p| p.starts_with("head") || p.starts_with("hand") || p.starts_with("ear"))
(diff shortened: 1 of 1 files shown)


ammgws:
Seems reasonable. Thanks!

codicodi:
Hmm this breaks headphones detection for me. 
The actual port names for pulseaudio are defined here for pulseaudio (in the left column):
https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/blob/0ce3008605e5f644fac4bb5edbb1443110201ec1/src/modules/alsa/alsa-mixer.c#L2709-L2731
None of them starts with "head", "hand" or "ear".

On my system specifically port names are
```
analog-output-lineout
analog-output-headphones
```

PR #1349, opened 2021-11-16
`backlight` block: adapt configuration options to precedent set by `sound` block

Vanille-N:
## Current state
The `backlight` block has a few configuration differences compared to `sound`, the last of which may be considered a bug.
- there is no `format` option (`volume` has one)
- `on_click` does nothing by default  (`volume` toggles on click)
- `on_click` override _erases the preset scroll action_ of changing brightness by small increments (scrolling on `volume` is still available when `on_click` is set)

## Proposed fix
This PR fixes these three differences by adding behavior to the `backlight` block:
- `format` is by default set to the only placeholder available, `"{brightness}"`
- by default, both left and right clicks "toggle" brightness (same as `volume`)
- if `on_click` is set, left click becomes `on_click` and right click remains the default toggle (same (I think) as `volume`)

"toggle" brightness: if less than 50% set to 100%, otherwise set to 0%

### Other (somewhat) unrelated fix
```rust
if brightness < 100 {
    set_brightness(brightness + step_width);
}
if brightness > step_width {
    set_brightness(brightness - step_width)?;
}
```
this code behaves inconsistently with regards to edge cases:
- `5 - 10 -> 5`
- `10 - 10 -> 10`
- `11 - 10 -> 1`
- `99 + 10 -> 109`
- `100 + 10 -> 100`
etc

The following is much more consistent:
```rust
(brightness + sign * step_width).clamp(0, 100)
```

### DIFF
--- a/src/blocks/backlight.rs
+++ b/src/blocks/backlight.rs
@@ -141,12 +141,6 @@ impl BacklitDevice {
             .block_error("backlight", "Failed to write into brightness file")
     }
 
-    fn toggle(&mut self) -> Result<()> {
-        let current = self.brightness()?;
-        let target = if current > 50 { 0 } else { 100 };
-        self.set_brightness(target)
-    }
-
     fn set_brightness_via_dbus(&self, raw_value: u64) -> Result<()> {
         let device_name = self
             .device_path
@@ -189,6 +183,9 @@ pub struct Backlight {
     output: TextWidget,
     device: BacklitDevice,
     step_width: u64,
+    min_bright: u64,
+    max_bright: u64,
+    curr_bright: u64,
     scrolling: Scrolling,
     invert_icons: bool,
     on_click: Option<String>,
@@ -204,6 +201,11 @@ pub struct BacklightConfig {
 
     /// The steps brightness is in/decreased for the selected screen (When greater than 50 it gets limited to 50)
     pub step_width: u64,
+
+    /// the min and max brightness limit the range over which the brightness can be in/decreased
+    /// as well as determine the values between which toggle alternates
+    pub min_bright: u64,
+    pub max_bright: u64,
 
     /// Format string for displaying backlight information.
     /// placeholders: {brightness}
@@ -232,7 +234,40 @@ impl Default for BacklightConfig {
             invert_icons: false,
             on_click: None,
             format: FormatTemplate::default(),
+            min_bright: 5,
+            max_bright: 100,
         }
+    }
+}
+
+impl Backlight {
+    fn toggle_device(&mut self) -> Result<()> {
+        let target = {
+            let brightness = self.device.brightness()?;
+            if self.curr_bright == self.min_bright || self.curr_bright == self.max_bright {
+                // special case: if curr is an extremity, alternate between the two extremities
+                if brightness == self.min_bright {
+                    self.max_bright
+                } else {
+                    self.min_bright
+                }
+            } else {
+                // normal case: cycle through min -> max -> curr
+                if brightness == self.min_bright {
+                    self.max_bright
+                } else if brightness == self.max_bright {
+                    self.curr_bright
+                } else {
+                    self.min_bright
+                }
+            }
+        };
+        self.device.set_brightness(target)
+    }
+
+    fn register_brightness(&mut self) -> Result<()> {
+        self.curr_bright = self.device.brightness()?;
+        Ok(())
     }
 }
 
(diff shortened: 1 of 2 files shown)


MaxVerevkin:
> "toggle" brightness: if less than 50% set to 100%, otherwise set to 0%

To be honest, I don't think it's a good idea, in particular considering that some drivers disable the backlight when set to zero.

Vanille-N:
5% then ?

GladOSkar:
Disabling backlight might be a desirable feature, but I wouldn't enable it by default because it might be difficult for users to re-enable it with a completely dark screen

Vanille-N:
For now I have set the behavior on my personal machine to be `set_backlight(if backlight < 50 { 100 } else { 1 })`.
My backlight drivers had no issue going down to 0%, so I didn't imagine it could be a problem.
1% should be OK for all backlight drivers though, right ?

If needed I can do any of the following:
- nothing
- commit as is set on my local instance
- change 1 to `step_width`
- make it so that it behaves more like the volume: click -> 1%, click again -> back to initial
- remove the backlight toggle feature from my pull request since it can be implemented anyway with the `on_click` feature

GladOSkar:
>  1% should be OK for all backlight drivers though, right ?

Yes i think so. Only `0` means "off" on some devices (like mine)

GladOSkar:
> make it so that it behaves more like the volume: click -> 1%, click again -> back to initial

How about a mix:

Click 1 -> Max brightness
Click 2 -> Min brightness
Click 3 -> Back to original
repeat.

Also, one could make `min_brightness` (and `max_brightness`?) configurable with a default of `1`, because some users might want to turn their display off like this and others might want a smaller range

ammgws:
Related note: if someone ports the click handling code from swaystatus to i3status-rust then you could just set your right command in the config

Vanille-N:
> Click 1 -> Max brightness Click 2 -> Min brightness Click 3 -> Back to original repeat.
> 
> Also, one could make `min_brightness` (and `max_brightness`?) configurable with a default of `1`, because some users might want to turn their display off like this and others might want a smaller range

I do like both of these points. I'll implement them locally right away and see how it fares.

Vanille-N:
Current behavior:
- scroll sets `curr`
- if `curr` is `max` or `min` then click toggles between `max` and `min`
- otherwise click cycles between `min`, `max` and `curr`

It must still be decided how `curr` interacts with external modifications:
currently the sequence
- scroll to 50%
- manually reduce brightness to 25%
- repeat click
will end cycling between 1%, 100% and 50%, not 25%


Unrelated: should `max_bright` and `min_bright` rather be named `maximum` and `minimum` ? In the context of the `backlight` block should be obvious that it refers to brightness.

Vanille-N:
Some notes about `min` and `max`:
- typing ensures that both are nonnegative
- there is no way to enforce that both be `<= 100`, this is clamped manually
- although they are named `min` and `max`, they will actually be the reverse if the configuration sets `max < min`
- if the current brightness is outside of `min <= x <= max`, scrolling in any direction will immediately bright back the brightness in the right interval

MaxVerevkin:
One more idea: maybe instead of `min`/`max` provide `cycle` or something which would accepts a list of values, and clicking on the block would cycle through those values? But `min`/`max` could be used just to clamp as in your last point?

I just don't like that if I want to cycle between `min`/`max`/`cur` I cannot set a value less than `min` and greater then `max`.

Vanille-N:
> I just don't like that if I want to cycle between min/max/cur I cannot set a value less than min and greater then max.

Technically you can, in the current implementation.
- manually set to 100%
- `min = 1`, `max = 50`
- repeat click

will cycle through 1%, 50%, 100%

A major limitation for now is that the brightness outside of `min`/`max` needs to be set before starting the bar, but this will fix itself along with the issue I already mentioned of `curr` not being set by external modifications.

Vanille-N:
How do you think `cycle = [b1, b2, ..., bn]` should interract with `curr`?
Cycle through `b1` -> `b2` -> ... -> `bn` -> `curr` -> `b1` ?

MaxVerevkin:
I don't know. I wouldn't mind if the block didn't even remember `curr`, but that might by just me.

Vanille-N:
I just experimented with a `cycle` option, and I quite like it.

It feels less hacky than the previous `min`/`max`/`current` cycle, and it has the pleasing side-effect that `cycle = []` disables cycling and `cycle = [n]` provides a way to reset brightness to a fixed value on each click.
To minimize surprise I made sure that the cycle continues at the point closest to the current brightness.
This also makes one issue I mentioned above irrelevant.

src/blocks/backlight.rs
```diff
@@ -234,16 +292,27 @@ impl ConfigBlock for Backlight {
...
+            (block_config.minimum, block_config.maximum)
+        } else {
+            (block_config.maximum, block_config.minimum)
+        };
+
+        let backlight = Self {
             id,
             device,
-            step_width: block_config.step_width,
+            step_width: block_config.step_width.max(1),
```
MaxVerevkin, line 306: Why `.max(1)`? Shouldn't `step_width = 0` disable scrolling?

Vanille-N: That's a good question. There was probably a reason I though of, but I can't remember. I'll check.

Vanille-N: Nope, looks like a mistake on my part.
Possibly left over from a first attempt at guaranteeing `brightness > 0` when clamping brightness between `step_width` and `100`. Definitely no use now.

ammgws:
Looks like this reached a good state

Thanks!

Issue #1340, opened 2021-11-09
Code cleanup: remove profiling and --one-shot


ammgws:
I don't think these are useful anymore.

jgoday:
Hi, I've started using i3status-rust with i3 (a pleasure to use), and  looking for 'good first issues', so ... here it is a PR #1350 for this.

colemickens:
I think this broke the build, it looks like Cargo.lock didn't get properly updated.

PR #1350, opened 2021-11-16
#1340: Code cleanup: remove profiling and --one-shot

jgoday:
See #1340 
1) Remove one-shot command line argument
2) Remove profile feature


Cargo.toml
```diff
@@ -37,8 +35,6 @@ toml = "0.5"
 libpulse-binding = { optional = true, version = "2.0", default-features = false }
 notmuch = { optional = true, version = "0.7.0" }
 maildir = { optional = true, version = "0.5" }
-# Used only in debug build for profiling blocks
-cpuprofiler = { version = "0.0.4", optional = true }
 progress = { version = "0.2", optional = true }
```
MaxVerevkin, line 38: Is think `progress` is safe to remove too.

ammgws:
Thanks!

colemickens:
I think this broke the build, it looks like Cargo.lock didn't get properly updated.

ammgws:
Should be OK now

PR #1356, opened 2021-11-23
Create rofication block

cfsmp3:
Rofication is the notification system used by regolith linux (and others I assume, but this is my use case).

It's a simple system - there's a daemon that receives the notications on d-bus and that exports them via a UNIX socket. The block itself just needs to display a counter and a color (based on any notification being critical or not) and spawn the GUI when clicked.

### DIFF
--- a/src/blocks/rofication.rs
+++ b/src/blocks/rofication.rs
@@ -90,7 +90,7 @@ impl Block for Rofication {
         match rofication_status(&self.socket_path) {
             Ok(status) => {
                 self.text.set_icon("bell")?;
-                self.text.set_text(format!("{}", status.num).to_string());
+                self.text.set_text(status.num.to_string());
                 if status.crit > 0 {
                     self.text.set_state(State::Critical)
                 } else {
--- a/src/blocks/rofication.rs
+++ b/src/blocks/rofication.rs
@@ -13,6 +13,8 @@ use crate::blocks::{Block, ConfigBlock, Update};
 use crate::config::SharedConfig;
 use crate::de::deserialize_duration;
 use crate::errors::*;
+use crate::formatting::value::Value;
+use crate::formatting::FormatTemplate;
 use crate::protocol::i3bar_event::I3BarEvent;
 use crate::protocol::i3bar_event::MouseButton;
 use crate::scheduler::Task;
@@ -20,8 +22,6 @@ use crate::subprocess::spawn_child_async;
 use crate::widgets::text::TextWidget;
 use crate::widgets::I3BarWidget;
 use crate::widgets::State;
-use crate::formatting::value::Value;
-use crate::formatting::FormatTemplate;
 
 #[derive(Debug)]
 struct RotificationStatus {
--- a/src/blocks/rofication.rs
+++ b/src/blocks/rofication.rs
@@ -1,0 +1,168 @@
+use std::str::FromStr;
+use std::time::Duration;
+
+use crossbeam_channel::Sender;
+use serde_derive::Deserialize;
+
+use std::os::unix::net::UnixStream;
+use std::io::Write;
+use std::io::Read;
+use std::path::Path;
+
+use crate::blocks::{Block, ConfigBlock, Update};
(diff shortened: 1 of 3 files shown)


cfsmp3:
@greshake Could you review?

ammgws:
Does this support both the [original](https://github.com/DaveDavenport/Rofication) rofication as well as the regolith fork? If not can you mention that in the docs?

src/blocks/rofication.rs
```diff
@@ -0,0 +1,195 @@
...
+    text: TextWidget,
+    update_interval: Duration,
+
+    //useful, but optional
+    #[allow(dead_code)]
+    shared_config: SharedConfig,
+    #[allow(dead_code)]
+    tx_update_request: Sender<Task>,
+    // UNIX socket to read from
+    pub socket_path: String,
```
ammgws, line 41: Can you add a `format` string too? It seems inevitable so we may as well do it from the start.

src/blocks/rofication.rs
```diff
@@ -0,0 +1,195 @@
...
+    pub interval: Duration,
+    // UNIX socket to read from
+    pub socket_path: String,
+}
+
+impl Default for RoficationConfig {
+    fn default() -> Self {
+        Self {
+            interval: Duration::from_secs(1),
+            socket_path: String::from_str("/tmp/rofi_notification_daemon").unwrap(),
```
ammgws, line 58: Why not just `"/tmp/rofi_notification_daemon".to_string()` or `String::from("/tmp/rofi_notification_daemon".to_string()")`?

src/blocks/rofication.rs
```diff
@@ -0,0 +1,195 @@
...
+    type Config = RoficationConfig;
+
+    fn new(
+        id: usize,
+        block_config: Self::Config,
+        shared_config: SharedConfig,
+        tx_update_request: Sender<Task>,
+    ) -> Result<Self> {
+        let text = TextWidget::new(id, 0, shared_config.clone())
+            .with_text("0")
```
ammgws, line 73: Should this just be a blank string `""` or `"?"` instead since we don't have the actual count yet?

ammgws (changes requested):

cfsmp3:
> Does this support both the [original](https://github.com/DaveDavenport/Rofication) rofication as well as the regolith fork? If not can you mention that in the docs?

Yes, it only uses one command to get the number of notifications and that came from upstream.

I personally use, and contribute to, the fork but it is a superset of the original.

src/blocks/rofication.rs
cfsmp3: > Can you add a format string too? It seems inevitable so we may as well do it from the start.

Done.

src/blocks/rofication.rs
cfsmp3: Changed.

src/blocks/rofication.rs
cfsmp3: Changed to ?.

ammgws (approved):

ammgws:
Thanks!

MaxVerevkin:
> > Does this support both the [original](https://github.com/DaveDavenport/Rofication) rofication as well as the regolith fork? If not can you mention that in the docs?
> 
> Yes, it only uses one command to get the number of notifications and that came from upstream.
> 
> I personally use, and contribute to, the fork but it is a superset of the original.

Actually, it doesn't work with the original.

- It works only with if the command is `b"num"`, not `b"num\n"`.
- The original separates `num` and `critical` with a new line, not a comma.

MaxVerevkin:
So, perhaps rename it?

MaxVerevkin:
Or, if the fork forks with `"b"num"` too, then the block can be adapted to work with both versions.

cfsmp3:
Nah, let's do things right. I'm away but I'll fix this properly asap.

cfsmp3:
@MaxVerevkin https://github.com/greshake/i3status-rust/pull/1390

Issue #431, opened 2019-08-22
Debian packaging


atheriel:
At the moment we basically encourage users to compile from source, although I'm aware of packages in the AUR, [in Void Linux](https://github.com/void-linux/void-packages/tree/master/srcpkgs/i3status-rust), and [in NixOS](https://github.com/NixOS/nixpkgs/blob/b8572d322fa3ec8f21205249821a684dc91586af/pkgs/applications/window-managers/i3/status-rust.nix).

One way to broaden the reach of this project considerably would be to make a Debian package (`.deb`) available, since many distributions can use them either directly or as the basis of their own packages. A Debian package could also provide a partial solution to #401, and there appears to be [a popular `cargo deb` subcommand](https://github.com/mmstick/cargo-deb) to make this process easier, too.

I'm also willing to attempt to get i3status-rust included in Debian itself, if that's appealing to others (although I am not a Debian user myself).

atheriel:
For anyone interested in this, I build a `.deb` file for release 0.11.0 that can be downloaded via GitHub. I'd love to hear feedback.

atheriel:
Also, for future reference, you can build `.deb` files by using the `cargo-deb` branch of this repository. Rebase the branch against master and run `cargo deb` in the project root.

If you don't have a Debian environment around to check if things are working, you can use Docker, e.g.

```
$ docker run -v `pwd`/target/debian/:/deb -it debian
root@13068fa4ee15:/# ls /deb/
ext  i3status-rust_0.11.0_amd64.deb
```

To poke around with the `.deb` file try `dpkg-deb -I <deb>` and `dpkg-deb -c <deb>`. To install it, use `dpkg -i <deb>` -- you will then be prompted to install the dependencies.

ammgws:
I guess the next step could be providing a ppa? Don't know how much of a hassle it is though

ammgws:
If someone wants to submit a RFP or ITP refer to the guide here: https://www.debian.org/devel/wnpp/

cfsmp3:
I'm working on this. But if the branch cargo-deb is not being used - it should just be removed to avoid confusion.

Packaging this is being painful because there's plenty of dependencies that just don't exist yet at Debian. So in order to successfully package i3status-rust I need to first package all of this:

```
[...]

PR #1357, opened 2021-11-25
Debian: Catch up with current master.

cfsmp3:
Sync cargo-deb with master.


ammgws:
Thanks but not currently using that branch. Would be nice if we could get packaged by Debian/Ubuntu instead: https://github.com/greshake/i3status-rust/issues/431

PR #1358, opened 2021-11-25
Focused window format

cfsmp3:
This PR adds format support to the block focused_window.

The reason I need this is that I really want all my blocks to have a fixed width (not just maximum, also minimum) because I find it annoying when a block changing width moves the location of all the rest.

So I have this now in my config:
```
[[block]]
block = "focused_window"
max_width = 100
show_marks = "visible"
format = "{composed:50^50}"
```

### DIFF
--- a/src/blocks/focused_window.rs
+++ b/src/blocks/focused_window.rs
@@ -190,7 +190,7 @@ impl ConfigBlock for FocusedWindow {
             text,
             max_width: block_config.max_width,
             show_marks: block_config.show_marks,
-            format: block_config.format.with_default("{composed}")?,
+            format: block_config.format.with_default("{combo}")?,
             title,
             marks,
         })
@@ -222,7 +222,7 @@ impl Block for FocusedWindow {
             }
         };
         let values = map!(
-            "composed" => Value::from_string(escape_pango_text(&out_str)),
+            "combo" => Value::from_string(escape_pango_text(&out_str)),
             "marks" => Value::from_string(escape_pango_text(&marks_string)),
             "title" => Value::from_string(escape_pango_text(&title_string))
         );
--- a/src/blocks/focused_window.rs
+++ b/src/blocks/focused_window.rs
@@ -11,9 +11,11 @@ use crate::blocks::{Block, ConfigBlock, Update};
 use crate::config::SharedConfig;
 use crate::errors::*;
 use crate::scheduler::Task;
-use crate::util::escape_pango_text;
+// use crate::util::escape_pango_text;
 use crate::widgets::text::TextWidget;
 use crate::widgets::I3BarWidget;
+use crate::formatting::value::Value;
+use crate::formatting::FormatTemplate;
 
 #[derive(Copy, Clone, Debug, Deserialize)]
 #[serde(rename_all = "lowercase")]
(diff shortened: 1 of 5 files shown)


ammgws:
Thanks, I think every block should have format support. It's not the case for this block but for others it's not so trivial to implement so not every block has been done yet.

Can you remove the unrelated commits?

cfsmp3:
> Can you remove the unrelated commits?

Oops. Done.

src/blocks/focused_window.rs
```diff
@@ -11,7 +11,9 @@ use crate::blocks::{Block, ConfigBlock, Update};
 use crate::config::SharedConfig;
 use crate::errors::*;
 use crate::scheduler::Task;
-use crate::util::escape_pango_text;
+// use crate::util::escape_pango_text;
```
ammgws, line 14: What's going on here?

cfsmp3: Sorry, that's no longer being used and I commented it out rather that remove it. Fixed.

ammgws: We still need to escape the text, don't we? The title/marks might have pango characters in them.

cfsmp3: Fixed

src/blocks/focused_window.rs
```diff
@@ -177,12 +184,13 @@ impl ConfigBlock for FocusedWindow {
             })
             .expect("failed to start watching thread for `window` block");
 
-        let text = TextWidget::new(id, 0, shared_config);
+        let text = TextWidget::new(id, 0, shared_config).with_text("x");
```
ammgws, line 187: Why do we need this fallback?

src/blocks/focused_window.rs
```diff
@@ -213,7 +221,11 @@ impl Block for FocusedWindow {
                 }
             }
         };
-        self.text.set_text(escape_pango_text(out_str));
+        let values = map!(
+            "composed" => Value::from_string(escape_pango_text(out_str))
```
ammgws, line 225: Should we also add "title" and "marks"?

src/blocks/focused_window.rs
cfsmp3: I just took it from whatever block I was looking at as a reference to add format support. Removed.

src/blocks/focused_window.rs
cfsmp3: I'm not convinced they're useful (since "composed" comes from some logic that makes sense) but they don't hurt so added.

ammgws: Yes the existence of show_marks is probably why no one tried to add a format string previously, since it's a bit tricky to make it fit in with the furthest string.

It makes as much sense as the music block having "artist", "title" and "combo". Also it will allow showing both marks and the title, which isn't possible right now (although not sure if anyone would do that)

cfsmp3: > Also it will allow showing both marks and the title, which isn't possible right now (although not sure if anyone would do that)

Well, done. Anything else? 
Don't love having too many pending PRs, my fork is deviating fast :-)

ammgws:
Thanks!

PR #1360, opened 2021-11-28
docker block: Mark errors as critical

cfsmp3:
Trivial change - if we can't connect to the docker daemon, change the color.

A very typical reason for not being able to connect (on top of course of the daemon being down, but I don't know if that's typical on a properly configured system) is that the user doesn't have permissions to access the socket file. 

Anyway, I prefer to make it obvious,  but feel free to close the PR.


ammgws:
Thanks!

PR #1361, opened 2021-12-01
Add missing bat_not_available icon

ram02z:
Can't use `material-nf` without getting an error if I disconnect and reconnect a device that has a battery block like my mouse for example. 

(Seperate issue) It flags the mouse as unavailable since it increments on re-connection, i.e. `mouse_hidpp_battery_1` to `mouse_hidpp_battery_2`, not sure how to workaround that problem, since I frequently disconnect my mouse.


ammgws:
These keep popping up from time to time - wonder if we need some kind of script to check so it doesn't happen again.

Thanks!@

PR #1365, opened 2021-12-03
rofication-block: Solve clippy messages

cfsmp3:
No functional changes, just take care of clippy's suggestions.

### DIFF
--- a/src/blocks/rofication.rs
+++ b/src/blocks/rofication.rs
@@ -145,21 +145,11 @@ fn rofication_status(socket_path: &str) -> Result<RotificationStatus> {
     };
 
     // Request count
-    if stream.write(b"num\n").is_err() {
-        return Err(BlockError(
-            "rofication".to_string(),
-            "Failed to write to socket".to_string(),
-        ));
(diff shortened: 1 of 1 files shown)


src/blocks/rofication.rs
```diff
@@ -148,26 +145,20 @@ fn rofication_status(socket_path: &str) -> Result<RotificationStatus> {
...
     // Request count
-    match stream.write(b"num\n") {
-        Err(_) => {
-            return Err(BlockError(
-                "rofication".to_string(),
-                "Failed to write to socket".to_string(),
-            ))
-        }
-        Ok(_) => {}
+    if stream.write(b"num\n").is_err() {
```
MaxVerevkin, line 148: You can also use `stream.write(b"num\n").error("rofication", "Failed to write to socket")?;`

cfsmp3: Thanks @MaxVerevkin  :-) Done, applied in other places as well.

ammgws:
Thanks!

PR #1366, opened 2021-12-05
Create block to display external IP address information

cfsmp3:
Display our external IP address and information about it (country, region, location, ISP, etc). 

- To test

```
[[block]]
block = "external_ip"
format = "{address} {country} {country_code} {region} {region_name} {city} {zip} {latitude} {longitude} {timezone} {isp} {org} {autonomous_system} {country_flag}"
```

- Possibly some non-rustic code there, feel free to school me.
- serde_derive and service don't get really well (both export Serialize). Used an alias for one of them. Proper solution welcome.
- Using a free API to get all the info. Only limitation this API has that we might sort of care about is that it's http, not https (on the free tier).
- Includes country flag representation but it requires a font that includes them. See blocks doc for details.
-Tested with IPv4 only.

### DIFF
--- a/src/blocks/external_ip.rs
+++ b/src/blocks/external_ip.rs
@@ -1,0 +1,196 @@
+use std::thread;
+use std::time::Instant;
+
+use crossbeam_channel::Sender;
+use dbus::ffidisp::{BusType, Connection, ConnectionItem};
+use serde::{Deserialize as des, Serialize as ser};
+use serde_derive::Deserialize;
+
+use crate::blocks::{Block, ConfigBlock, Update};
+use crate::config::SharedConfig;
+use crate::errors::*;
+use crate::formatting::value::Value;
+use crate::formatting::FormatTemplate;
+use crate::http;
+use crate::scheduler::Task;
+use crate::util::country_flag_from_iso_code;
+use crate::widgets::text::TextWidget;
+use crate::widgets::{I3BarWidget, State};
+use crate::Duration;
+
+const API_ENDPOINT: &str = "http://ip-api.com/json";
+
+#[derive(ser, des, Default)]
+struct IPAddressInfo {
+    #[serde(rename = "query")]
+    address: String,
+    status: String,
+    country: String,
+    #[serde(rename = "countryCode")]
+    country_code: String,
+    region: String,
+    #[serde(rename = "regionName")]
+    region_name: String,
+    city: String,
+    zip: String,
+    lat: f64,
+    lon: f64,
+    timezone: String,
+    isp: String,
+    org: String,
+    #[serde(rename = "as")]
+    autonomous_system: String,
+}
+
+pub struct ExternalIP {
+    id: usize,
+    output: TextWidget,
+    format: FormatTemplate,
+    refresh_interval_success: u64,
+    refresh_interval_failure: u64,
+}
+
+#[derive(Deserialize, Debug, Clone)]
+#[serde(deny_unknown_fields, default)]
+pub struct ExternalIPConfig {
+    /// External IP formatter.
+    pub format: FormatTemplate,
+    pub refresh_interval_success: u64,
+    pub refresh_interval_failure: u64,
+}
+
+impl Default for ExternalIPConfig {
+    fn default() -> Self {
+        Self {
+            format: FormatTemplate::default(),
+            refresh_interval_success: 300,
+            refresh_interval_failure: 15,
+        }
+    }
+}
+
+impl ConfigBlock for ExternalIP {
+    type Config = ExternalIPConfig;
+
+    fn new(
+        id: usize,
+        block_config: Self::Config,
+        shared_config: SharedConfig,
+        send: Sender<Task>,
+    ) -> Result<Self> {
+        thread::Builder::new()
+            .name("externalip".into())
+            .spawn(move || {
+                let c = Connection::get_private(BusType::System).unwrap();
+                c.add_match(
+                    "type='signal',\
+                    path='/org/freedesktop/NetworkManager',\
+                    interface='org.freedesktop.DBus.Properties',\
+                    member='PropertiesChanged'",
+                )
+                .unwrap();
+                c.add_match(
+                    "type='signal',\
+                    path_namespace='/org/freedesktop/NetworkManager/ActiveConnection',\
+                    interface='org.freedesktop.DBus.Properties',\
+                    member='PropertiesChanged'",
+                )
+                .unwrap();
+                c.add_match(
+                    "type='signal',\
+                    path_namespace='/org/freedesktop/NetworkManager/IP4Config',\
+                    interface='org.freedesktop.DBus',\
+                    member='PropertiesChanged'",
+                )
+                .unwrap();
+
+                loop {
+                    let timeout = 300_000;
+
+                    for event in c.iter(timeout) {
+                        match event {
+                            ConnectionItem::Nothing => (),
+                            _ => {
+                                send.send(Task {
+                                    id,
+                                    update_time: Instant::now(),
+                                })
+                                .unwrap();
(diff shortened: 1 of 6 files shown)


cfsmp3:
Self note: 
1) Use the included http stuff  instead of that random thing from eip
2) Use this API instead : https://api.myip.com/  - Free and it gives us a country code

MaxVerevkin:
> Using a free API to get all the info. Only limitation this API has that we might sort of care about is that it's http, not https (on the free tier).

This should probably be mentioned in the docs.

cfsmp3:
> > Using a free API to get all the info. Only limitation this API has that we might sort of care about is that it's http, not https (on the free tier).
> 
> This should probably be mentioned in the docs.

Added.

src/formatting.rs
```diff
@@ -61,6 +61,11 @@ impl FormatTemplate {
         Self::format_contains(&self.full, var) || Self::format_contains(&self.short, var)
     }
 
+    pub fn has_tokens(&self) -> bool {
+        (self.full.is_some() && !self.full.as_ref().unwrap().is_empty())
+            || (self.short.is_some() && !self.short.as_ref().unwrap().is_empty())
```
MaxVerevkin, line 66: I think
```rust
!self.full.as_ref().map(Vec::is_empty).unwrap_or(true)
```
is more "rustic".

cfsmp3: > is more "rustic".

Yup. Changed.

MaxVerevkin: By the way, it's a bit weird that a commenter cannot mark their comments as "resolved"

src/blocks/networkmanager.rs
```diff
@@ -784,7 +826,53 @@ impl Block for NetworkManager {
...
+                        "timezone" => Value::from_string (ip_info.timezone),
+                        "isp" => Value::from_string (ip_info.isp),
+                        "org" => Value::from_string (ip_info.org),
+                        "autonomous_system" => Value::from_string (ip_info.autonomous_system),
+                        "country_flag" => Value::from_string(flag),
+                    );
+                    if let Ok(s) = self.external_ip_format.render(&values) {
+                        s.0
+                    } else {
+                        "[invalid external_ip format string]".to_string()
```
MaxVerevkin, line 861: It might be better to propagate the error (and thus telling the user what's the problem).

MaxVerevkin:
This addition doesn't look NetworkManager-specific, so perhaps it could be added to `net` block instead? And just to point out, this functionality can be accomplished with `custom` block:
```toml
# County flag
[[block]]
block = "custom"
command = """curl -s http://ip-api.com/json | jq -r '.countryCode' | python -c 'print("".join(map(lambda x: chr(ord(x)+127397),input())))'"""
interval = 60

# County code
[[block]]
block = "custom"
command = """curl -s http://ip-api.com/json | jq -r '.countryCode'"""
interval = 60

# Address
[[block]]
block = "custom"
command = """curl -s http://ip-api.com/json | jq -r '.query'"""
interval = 60
```

cfsmp3:
> This addition doesn't look NetworkManager-specific, so perhaps it could be added to `net` block instead?

The net block doesn't have the very useful DBus listener stuff to detect changes as they happen. I actually started adding it there but it was clearly going in the wrong direction.

I could create a separate block just for this (which would be a copy of this one minus part of the preexisting stuff), but I'm not sure that's also a good idea.

>And just to point out, this functionality can be accomplished with custom block:

First - I personally refuse to be spawning processing every X seconds to populate the bar, if I wanted that I would be using the old bar which has a trillion of those scripts :-) Second, with a custom block you won't be able to immediately react to changes anyway.

src/blocks/networkmanager.rs
cfsmp3: How? That check is inside a block that returns an Option, not an Error. I guess that's what the other blocks do the same thing.

cfsmp3:
I've moved all the new stuff stuff to a new block with blackjack and hookers.

MaxVerevkin:
> Only limitation this API has that we might sort of care about is that it's http, not https (on the free tier).

Why not use `https://ipapi.co/json/` which is already [used by `weather` block](https://github.com/greshake/i3status-rust/blob/f3cda7be0c4573e5e6c0dcbb00b22da4df8fa07b/src/blocks/weather.rs#L77)?

cfsmp3:
> Why not use https://ipapi.co/json/ which is already used by weather block?

Because I wasn't aware of it :-)

cfsmp3:
@MaxVerevkin I've switched to that one. It's missing the ISP though, which is somewhat valuable, but I prefer having https over that specific field.

MaxVerevkin:
Also I think NetworkManager thread should be optional (perhaps on by default?) because this block can be useful for non NetworkManager users too.

cfsmp3:
> Also I think NetworkManager thread should be optional (perhaps on by default?) because this block can be useful for non NetworkManager users too.

I've added that.
However, I do have a problem with the periodic updates: They aren't happening anywhere close to the configured interval. Can you take a look at the code, where am I going wrong? It seems quite straightforward, but with this config:

```
[[block]]
block = "external_ip"
format = "{ip} {country_flag}"
refresh_interval_success = 60
```

I'd expect one poll every minute, and that's not happening (I'm monitoring traffic with wireshark).

MaxVerevkin:
It appears that `postal` can be `null`. Besides that, I don't know what could be wrong with your code - the `update` function runs correctly every N seconds. I don't have experience with linux network monitoring, but it seems to work just fine.

cfsmp3:
>- the `update` function runs correctly every N seconds.

How do you know for sure?

MaxVerevkin:
> How do you know for sure?

I've put an `eprintln!` in there :)

cfsmp3:
> > How do you know for sure?
> 
> I've put an `eprintln!` in there :)

OK, can't argue with that :-)

src/blocks/external_ip.rs
```diff
@@ -0,0 +1,233 @@
...
+    country: String,
+    country_name: String,
+    country_code: String,
+    country_code_iso3: String,
+    country_capital: String,
+    country_tld: String,
+    continent_code: String,
+    in_eu: bool,
+    #[serde(default)]
+    postal: String,
```
MaxVerevkin, line 42: `#[serde(default)]` doesn't help when postal in `null`.

src/blocks/external_ip.rs
```diff
@@ -0,0 +1,233 @@
...
+use crate::widgets::{I3BarWidget, State};
+use crate::Duration;
+
+const API_ENDPOINT: &str = "https://ipapi.co/json/";
+const BLOCK_NAME: &str = "external_ip";
+
+#[derive(ser, des, Default)]
+struct IPAddressInfo {
+    #[serde(default = "bool::default")]
+    error: bool,
```
MaxVerevkin, line 27: When `error` is `true`, most of the other fields are missing, so I suppose you have to derive `Default` and add `#[serde(default)]` for `IPAddressInfo`

src/blocks/external_ip.rs
cfsmp3: Fixed.

src/blocks/external_ip.rs
cfsmp3: Fixed.

MaxVerevkin:
There are clippy warnings, but apart from that, looks good to me.

cfsmp3:
> There are clippy warnings, but apart from that, looks good to me.

Fixed.

MaxVerevkin:
Thanks! By the way, if you want to tinker with some async rust code, you can port this block overt to [async version](https://github.com/MaxVerevkin/swaystatus) which will hopefully be merged at some point (#1369) :)

PR #1373, opened 2021-12-19
:bug: Escape SSID in NetworkManager

nzig:
This fixes error messages from sway when the SSID contains unescaped characters like `&`.


MaxVerevkin:
Thanks!

PR #1374, opened 2021-12-24
taskwarrior: Use inotify to get instant changes

cfsmp3:
Pretty much copy/pasted the Thread block from backlight, not a lot to it. Watch for changes on ~/.task and update the taskwarrior block if needed.

I haven't removed the polling code.

To test, just create a task:
```
task add "Fake task"
```
And you'll see the block updating instantly.

### DIFF
--- a/src/blocks/taskwarrior.rs
+++ b/src/blocks/taskwarrior.rs
@@ -91,7 +91,6 @@ pub struct TaskwarriorConfig {
 
 impl Default for TaskwarriorConfig {
     fn default() -> Self {
-
         let home_dir = match dirs::home_dir() {
             Some(path) => path.into_os_string().into_string().unwrap(),
             None => "".to_owned(),
@@ -110,7 +109,7 @@ impl Default for TaskwarriorConfig {
             format: FormatTemplate::default(),
             format_singular: FormatTemplate::default(),
             format_everything_done: FormatTemplate::default(),
-            data_location: task_dir
+            data_location: task_dir,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


ammgws:
We don't actually need to use the dirs crate, do we? IIRC we already have a helper function, or it was done trivially per block. Can you have a look?

Also why is interval still around?

On mobile so only had a brief look

MaxVerevkin:
Should the path be configurable?

cfsmp3:
> We don't actually need to use the dirs crate, do we? IIRC we already have a helper function, or it was done trivially per block. Can you have a look?

What we have in notmuch.rs is this:

```
#[allow(deprecated)]
let home_dir = match env::home_dir() {
```
And:

 Deprecated since 1.29.0:

This function’s behavior is unexpected and probably not what you want. Consider using a crate from crates.io instead.

dirs seems to have the sanest implementation (I did look).

> Also why is interval still around?

From inotify's documentation:

https://man7.org/linux/man-pages/man7/inotify.7.html

```
 Limitations and caveats
[...]
 Inotify reports only events that a user-space program triggers
       through the filesystem API.  As a result, it does not catch
       remote events that occur on network filesystems.  
[...]
       The inotify API does not report file accesses and modifications
       that may occur because of mmap(2), msync(2), and munmap(2).
[...]
 Note that the event queue can overflow.  In this case, events are
       lost.  Robust applications should handle the possibility of lost
       events gracefully.  For example, it may be necessary to rebuild
       part or all of the application cache. 
```
I believe when you put all of that together it's sane to keep some polling (maybe we can increase the default interval though).

cfsmp3:
> Should the path be configurable?

Yes - because it is in taskwarrior itself.
Done.

MaxVerevkin:
> From inotify's documentation:

I'm surprised it works for backlight block, even though It uses inotify to watch sysfs, which, AFAIK, is not supported at all and is considered to be a misuse.

src/blocks/taskwarrior.rs
```diff
@@ -80,10 +83,20 @@ pub struct TaskwarriorConfig {
...
+    pub data_location: String,
 }
 
 impl Default for TaskwarriorConfig {
     fn default() -> Self {
+        let home_dir = match dirs::home_dir() {
+            Some(path) => path.into_os_string().into_string().unwrap(),
+            None => "".to_owned(),
+        };
+        let task_dir = format!("{}/.task", home_dir);
```
MaxVerevkin, line 98: `shellexpand` will expand `~` as a home directory anyway, so I suppose you can just do `let task_dir = "~/.task"`.

cfsmp3: > `shellexpand` will expand `~` as a home directory anyway, so I suppose you can just do `let task_dir = "~/.task"`.

That's better. Done.

MaxVerevkin:
The interval was already high, so I think everything is good. Thanks!

PR #1375, opened 2021-12-25
temperature: use libsensors bindings instead of sensors binary

MaxVerevkin:
One step closer to be "written in pure Rust".

`sensors` binary is basically a wrapper around `libsensors`, so this change shouldn't break anything.

## Open questions:
- Now, are there any reasons to keep `sysfs` driver?


MaxVerevkin:
It... segfaults if built in release mode?

ammgws:
> * Now, are there any reasons to keep `sysfs` driver?

I don't think so, but should try avoid breaking configs. Perhaps by marking it deprecated for the next release, and in the release after that you could remove the option.

MaxVerevkin:
> It... segfaults if built in release mode?

It works perfectly fine in debug mode.

It doesn't detect any chips in release mode (on stable rustc).

It segfaults in release mode on nightly with `codegen-units = 1`.

So, probably the issue is with rust compiler?

MaxVerevkin:
https://github.com/nyantec/sensors/pull/6 fixes it!

MaxVerevkin:
@ammgws Is it OK to use git dependencies?

ammgws:
Better not to, especially since it would make it unpackagable for even Arch Linux.

MaxVerevkin:
Why? I as far as I understand Arch's PKGBUILD literally clones the repo and runs cargo build. Same with Void.

But it would be unpackageble on Debian, apparently.

So, would you prefer to just copy the file (the bindings consist of just lib.rs)?

ammgws:
Oops, you're right.

We could just vendor it as you mentioned.

ammgws:
The backend change seems fine to me, only worry is making sure not to break configs, e.g. this config (might be silly but it's allowed) seems to not work anymore:
```toml
[[block]]
block = "temperature"
```

```
[
[{"full_text":" Error in block 'temperature': Failed to get input value ","color":"#DC322FFF","background":"#000000FF","name":"0","instance":"0","separator":false,"separator_block_width":0,"markup":"pango"}],


Error in block 'temperature': Failed to get input value
```

ishitatsuyuki:
Upstream has released a new `sensors` version with the bugfix.

MaxVerevkin:
@ammgws It works for me...

ammgws:
It's this sensor subfeature that is tripping it up, which when we call libsensors is returning an "Unknown" error:

```rs
subfeat: Subfeature {
    inner: 0x000055639e468310,
    chip_ptr: 0x000055639e45ff90,
    name: "temp1_input",
    number: 0,
    subfeature_type: SENSORS_SUBFEATURE_TEMP_INPUT,
    mapping: 0,
    flags: 5,
}
```

Where as `sensors` returns "N/A" instead of erroring out, which explains why it works:
```sh
>sensors
...
iwlwifi_1-virtual-0
Adapter: Virtual device
temp1:            N/A
```
https://github.com/lm-sensors/lm-sensors/blob/42f240d2a457834bcbdf4dc8b57237f97b5f5854/prog/sensors/chips.c#L373-L381

So perhaps we should just skip trying to read values from ones like these?

MaxVerevkin:
> So perhaps we should just skip trying to read values from ones like these?

Sounds reasonable.

> It's this sensor subfeature that is tripping it up, which when we call get_value(), libsensors is returning an "Unknown" error:

I'm just curious: do you have the corresponding `temp1_input` file in your sysfs? If so, what does is contains?

ammgws:
It's empty:

```sh
>ll /sys/class/hwmon/hwmon2/
total 0
lrwxrwxrwx 1 root root    0 Jan 17 19:35 device -> ../../thermal_zone2
-r--r--r-- 1 root root 4.0K Jan 17 19:35 name
drwxr-xr-x 2 root root    0 Jan 18 14:01 power
lrwxrwxrwx 1 root root    0 Jan 17 19:35 subsystem -> ../../../../../class/hwmon
-r--r--r-- 1 root root 4.0K Jan 17 19:35 temp1_input
-rw-r--r-- 1 root root 4.0K Jan 18 14:01 uevent

>cat name
iwlwifi_1

>cat temp1_input
cat: temp1_input: No data available
```

MaxVerevkin:
Should work as expected now.

PR #1378, opened 2021-12-30
Enable allow_missing for the UPower driver

bim9262:
Added additonal match filters in the monitor function to let us know
when a upower devices is added or removed.

Implimented the is_availible function for UpowerDevice.

Deduplicated logic in Battery's update function regarding the return
values.

Updated blocks.md to remove note about allow_missing not working for
drivers other than the sysfs driver.

### DIFF
--- a/doc/blocks.md
+++ b/doc/blocks.md
@@ -247,7 +247,7 @@ Key | Values | Required | Default
 `format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{percentage}"`
 `full_format` | Same as `format` but for when the battery is full. | No | `"{percentage}"`
 `missing_format` | Same as `format` but for when the specified battery is missing. | No | `"{percentage}"`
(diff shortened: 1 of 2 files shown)


bim9262:
Updated comment in battery.rs regarding allow_missing only working on sysfs devices.

bim9262:
Rebased as there would otherwise be a merge conflict.

ammgws:
In theory this should be ok

ammgws:
Can you resolve the conflicts?

bim9262:
Yes, will do.

ammgws:
Thanks!

ammgws:
This still fails on startup if the device isn't available at that time, so we will need some additional logic for that case..

MaxVerevkin:
I don't have any removable batteries, so I can't check myself, but what's wrong with the `IsPresent` property? Does it even work?

bim9262:
@MaxVerevkin , I don't have any device that will show `IsPresent` when it's present and false when it's not. I have a computer with two different removable batteries (separate bays). When the batteries are removed they simply aren't listed by UPower. Maybe we need to find somebody with an "expensive UPS" (see: https://upower.freedesktop.org/docs/Device.html#Device:IsPresent)

Issue #1200, opened 2021-04-28
Networkmanager block: Use connection name for VPN connections


jsmnbom:
Current I use a networkmanager block like the one below to show VPN connections, but they only show up as "tun0" or "tun1".

```
[[block]]
block = "networkmanager"
on_click = "nm-dmenu"
interface_name_exclude = []
interface_name_include = ["tun"]
device_format = "{icon}{name}"
icons_format = "{icon} "
```

I was wondering if it would be somehow possible to instead show the connection name? So that {name} instead of outputting the interface name, actually outputs the connection id.

ammgws:
I'm guessing this is what you meant to say:

> I was wondering if it would be somehow possible to instead show the connection ~name~ id? So that {name} instead of outputting the interface name, actually outputs the connection id.

If it's an actual D-Bus field I don't see why it couldn't be done.

jsmnbom:
Ah yes, that is what i meant, I got confused since the connection manager actually does call that field for "Connection Name" while in d-bus it looks to be "Id" in the dbus.

ammgws:
PRs welcome

jsmnbom:
I've tried my best to figure out how this would be implemented but I'm gonna have to give up. The way NetworkManager handles VPN connections makes absolutely no sense as far as I can tell:

If i have a WiFi connection it will be ActiveConnection/1, under devices it has Devices/3 which is wlo1.
Then when I connect to a VPN using OpenSSH, two new ActiveConnections get added:
ActiveConnection/2 which has VPN = true, and the VPN connection name as its id, under devices it has Devices/3 which is wlo1,
and ActiveConnection/3 which has VPN = false and the interface as its id (tun0), under devices it has Devices/5 which is tun0.

This means that when using `interface_name_include = ["tun"]` the current code only finds the ActiveConnection/3 since that's the only one connected to Devices/5 with that interface name. To me it doesn't seem like there's any way of figuring out what ActiveConnection (with VPN = true and the wrong interface name) corresponds to an ActiveConnnection (with VPN = false and the proper interface name), so I'm unsure if this is even possible now actually.

ammgws:
Copied the screenshots from https://github.com/greshake/i3status-rust/issues/1016#issuecomment-775125507 since it's easier to grok.

Wifi connection:
![image](https://user-images.githubusercontent.com/20397027/119359219-5df78500-bce4-11eb-9246-8db1226163fe.png)

VPN connection:
[...]

PR #1381, opened 2021-12-31
networkmanager: allow connection types to be specified

ammgws:
Possibly fixes https://github.com/greshake/i3status-rust/issues/1200


ammgws:
Closing since this block has been removed in v0.30.0

PR #1383, opened 2022-01-01
Add new apcupsd battery driver

bim9262:
Added new apcupsd battery driver called apcaccess (like the command
line tool wich is used to query apcupsd for status and events).

Added information about new driver to blocks.md


ammgws:
Seems ok to me

ammgws:
Thanks!

PR #1389, opened 2022-01-04
Fix battery block's spacing

bim9262:
Currently the full format for the battery block uses the hidden spacing.
However the hidden spacing should only be used when there's no tokens
in the full format.

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -22,7 +22,7 @@ use crate::formatting::FormatTemplate;
 use crate::scheduler::Task;
 use crate::util::{battery_level_to_icon, read_file};
 use crate::widgets::text::TextWidget;
-use crate::widgets::{I3BarWidget, Spacing, State};
+use crate::widgets::{I3BarWidget, State};
 
 /// A battery device can be queried for a few properties relevant to the user.
 pub trait BatteryDevice {
@@ -700,7 +700,6 @@ impl Block for Battery {
             self.output.set_icon("bat_full")?;
             self.output.set_texts(self.full_format.render(&values)?);
             self.output.set_state(State::Good);
-            self.output.set_spacing(Spacing::Hidden);
         } else {
             self.output.set_texts(self.format.render(&values)?);
 
@@ -735,7 +734,6 @@ impl Block for Battery {
                 "Charging" => "bat_charging",
                 _ => battery_level_to_icon(capacity, self.fallback_icons),
             })?;
-            self.output.set_spacing(Spacing::Normal);
         }
 
         match self.driver {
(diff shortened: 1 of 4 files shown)


bim9262:
Forgot to run `cargo fmt` the first time :(

MaxVerevkin:
I think you can remove this check completelys since `set_texts()` does it anyway.

MaxVerevkin:
> Forgot to run `cargo fmt` the first time :(

What editor do you use? I'm pretty sure you can automate this in every editor :)

bim9262:
@MaxVerevkin , I use vs code. Adding `"editor.formatOnSave": true` to my config so I should be good format-wise from here on out...

I didn't realise that `set_texts` sets the spacing too. That got me interested in `set_spacing`/`set_texts` pairs in the code base. 

For `sound.rs` could be cleaned up
```
diff --git a/src/blocks/sound.rs b/src/blocks/sound.rs
index 39a9155..54b07b0 100644
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -964,7 +964,6 @@ impl Block for Sound {
         if self.device.muted() {
             self.text.set_icon(&self.icon(0, headphones))?;
             if self.show_volume_when_muted {
-                self.text.set_spacing(Spacing::Normal);
                 self.text.set_texts(texts);
             } else {
                 self.text.set_text(String::new());
@@ -972,7 +971,6 @@ impl Block for Sound {
             self.text.set_state(State::Warning);
         } else {
             self.text.set_icon(&self.icon(volume, headphones))?;
-            self.text.set_spacing(Spacing::Normal);
             self.text.set_state(State::Idle);
             self.text.set_texts(texts);
         }```

Should we clean up all of the blocks that follow the same pattern (`temperature.rs`, nvidia_gpu.rs`, `sound.rs`, and `battery.rs`)

bim9262:
@MaxVerevkin, let me know what you think about the other blocks. I can update the branch name or open a new PR for the other blocks if you want. Happy to squash changes and/or rebase onto master too.

MaxVerevkin:
I remember there were some issues related to spacing before, but maybe some of those checks can be removed. I can't check it now, but I'll take a look when I can.

Don't worry about branch name.

MaxVerevkin (approved):

MaxVerevkin:
> Should we clean up all of the blocks that follow the same pattern (`temperature.rs`, nvidia_gpu.rs`, `sound.rs`, and `battery.rs`)

Yes, those checks are either redundant or incorrect as in battery block.

bim9262:
@MaxVerevkin in the same PR or should a new one be opened? The battery block has an actual bug whereas the other blocks are just clean up.

MaxVerevkin:
I'm fine with just a separate commit.

MaxVerevkin (approved):

MaxVerevkin:
Thanks!

bim9262:
No problem :)

PR #1393, opened 2022-01-11
Fix for batteries allow_missing

bim9262:
@ammgws pointed out that UPower devices were throwing an error if the
device was not present when the bar started. If the battery has
`allow_missing=true` this is not a case that we want to throw an
error. There was a similar issue present in the new apcaccess
battery driver that is fixed as well. Both drivers are able to
pull information from a device is that not present when the status
bar is launched, when `allow_missing=true`.

### DIFF
--- a/src/apcaccess.rs
+++ b/src/apcaccess.rs
@@ -27,8 +27,8 @@ impl ApcAccess {
         Ok(stream)
     }
 
-    pub fn is_available(&self) -> bool {
-        if let Ok(status_data) = self.get_status() {
+    pub fn is_available(&self, status_result: &Result<HashMap<String, String>>) -> bool {
+        if let Ok(status_data) = status_result {
             if let Some(status) = status_data.get("STATUS") {
                 return !status.contains("COMMLOST");
             }
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -537,6 +537,25 @@ impl UpowerDevice {
             })
             .unwrap();
     }
+
+    fn get_upower_value<T: for<'b> dbus::arg::Get<'b>>(
+        &self,
+        key: &str,
+        fallback_value: T,
+    ) -> Result<T> {
+        self.con
+            .with_path("org.freedesktop.UPower", &self.device_path, 1000)
+            .get::<T>("org.freedesktop.UPower.Device", key)
+            .or_else(|_| {
+                if self.allow_missing {
+                    return Ok(fallback_value);
+                }
+                Err(BlockError(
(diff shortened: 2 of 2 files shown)


ammgws:
Just tried with the only battery device I actually use and ran into a panic when disconnecting the device (it's connected by bluetooth). When disconnected it completely disappears off the UPower bus, which is why it's failing to query. Does that not happen with your device(s) too? Maybe we just need to tweak the logic in the update() function..

```toml
[[block]]
block = "battery"
driver = "upower"
device = "gaming_input_ps_controller_battery_d0obcoc1oe6o57o5c"
format = "PS5: {percentage}"
full_format = "PS5: {percentage}"
missing_format = "PS5: N/A"
allow_missing = true
```

```
[
Icon bat_10 not found in your icons file. Please check NEWS.md
[{"full_text":"|","color":"#A9A9A9FF","background":"#000000FF","separator":false,"separator_block_width":0,"markup":"pango"},{"full_text":" FULL PS5: 85% ","color":"#859900FF","background":"#000000FF","name":"0","instance":"0","separator":false,"separator_block_width":0,"markup":"pango"}],
[{"full_text":" Error in block 'battery': Failed to read UPower State property. ","color":"#DC322FFF","background":"#000000FF","name":"0","instance":"0","separator":false,"separator_block_width":0,"markup":"pango"}],


Error in block 'battery': Failed to read UPower State property.
thread 'battery' panicked at 'called `Result::unwrap()` on an `Err` value: "SendError(..)"', src/blocks/battery.rs:529:38
```

bim9262:
@ammgws when I disconnect my UPS usb cable (that is used to feed status info back to the computer) I see this:

```
$ upower --enumerate              
/org/freedesktop/UPower/devices/ups_hiddev0
/org/freedesktop/UPower/devices/mouse_hidpp_battery_2

#Disconecting usb from ups
Broadcast message from root@xxx (somewhere) (Wed Jan 12 17:54:54 2022)
                                                                               
Warning communications lost with UPS xxx

$ upower --enumerate
/org/freedesktop/UPower/devices/mouse_hidpp_battery_2
```

Does it crash constantly? It seems that your device must have disconnected between these two lines:
```
self.device.refresh_device_info()?;
let status = self.device.status()?;
```

I'll take a look at the functions in `impl BatteryDevice for UpowerDevice` so that instead of checking to see if the device is available and then trying to run the dbus get command and then either return an error or a default value if allow_missing is set. This should eliminate the race condition.

bim9262:
You're welcome!

PR #1394, opened 2022-01-12
Named PA sound devices don't know current volume

bim9262:
When adding a sound block using the auto/pulseaudo driver and specifying
a device name:

For example:
```
[[block]]
block = "sound"
name = "alsa_output.pci-0000_41_00.1.hdmi-stereo-extra2"
```
The device volume is unknown. This results in an error being thrown when
 when you try to modify the volume (for example by scrolling over the
 block to change it by step).

The workaround I'd used was to also set a custom block that would modify
the volume (slightly) so that the mainloop would pick up the change,
like this:

```
[[block]]
block = "custom"
command = "pactl set-sink-volume
alsa_output.pci-0000_41_00.1.hdmi-stereo-extra2 +1"
interval = "once"
hide_when_empty= true
```

Instead of creating a `PulseAudioSoundDevice` pointing at the default
device and then mapping the device using `with_name` we can just pass in
the `block_config.name` along with `block_config.device_kind` to
`PulseAudioSoundDevice::new`. By doing this we make sure that the name
is set before `PulseAudioClientRequest::GetInfoByName` is invoked.

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -408,8 +408,11 @@ impl PulseAudioClient {
 
                 loop {
                     // make sure mainloop dispatched everything
-                    for _ in 0..10 {
+                    loop {
                         connection.iterate(false).unwrap();
+                        if connection.context.borrow().get_state() == PulseState::Ready {
+                            break;
+                        }
                     }
 
                     match recv_req.recv() {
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
Thanks!

src/blocks/sound.rs
```diff
@@ -408,8 +408,11 @@ impl PulseAudioClient {
 
                 loop {
                     // make sure mainloop dispatched everything
-                    for _ in 0..10 {
+                    loop {
```
ammgws, line 411: Did you run into an issue with this, or was this just something you noticed in the code? (No idea where the magic number came from in the original)

MaxVerevkin: For me the loop always runs just once. Block works as expected. I suppose this is just the "correct" way to do this.

bim9262: No, I didn't run into an issue, but I changed it while debugging. It didn't make sense to run the loop an arbitrary number of times instead of just making sure that the context state is ready.

Issue #1397, opened 2022-01-16
hueshift: killall command is unreliable


stelcodes:
Hi there, love this project. I'm having an issue with the Hueshift block. The `killall` command used to end previous hueshifting processes does not work reliably. I'm using `gammastep` but this could be recreated with any provider. When I scroll on the block my displays glitch out, go black, and I have to restart. This is because `i3status-rs` fails to kill previous `gammastep` processes and they keep piling up. This is the error log I keep for `i3status-rs`:
```
❯ cat /tmp/i3status-rs-errors.log 
gammastep: no process found
gammastep: no process found
gammastep: no process found
gammastep: no process found
gammastep: no process found
gammastep: no process found
gammastep: no process found
gammastep: no process found
gammastep: no process found
gammastep: no process found
gammastep: no process found
```
This is because `killall` uses the output `/proc/<pid>/stat` to match the name provided [1], and that name is not always going to be exactly "gammastep", "wlsunet", etc. For example, I'm using gammastep installed from the Nix package manager which wraps `gammastep` in a script, a common technique in Nix packaging.
```
❯ pgrep gammastep
66269
66272
66301
66304
66308
66336
66339
66342
66345
66348
66376

❯ cat /proc/66269/stat
66269 (.gammastep-wrap) S 1 1261 1261 1025 1261 4194304 388 0 4 0 0 0 0 0 20 0 1 0 173688 11653120 1207 18446744073709551615 4210688 4245901 140726936919616 0 0 0 0 6 0 0 0 0 17 2 0 0 0 0 0 4273072 4276304 11685888 140726936927432 140726936927481 140726936927481 140726936932261 0
```
A more reliable strategy would be to use `killall -r` or `pkill`:
```
❯ gammastep &
[1] 44924

✦ ❯ pgrep gammastep
44924

✦ ❯ ps aux | grep gammastep
stel       44924  0.3  0.0  11380  5120 pts/0    S    11:09   0:00 /home/stel/.nix-profile/bin/gammastep
stel       45053  0.0  0.0   6224   908 pts/0    S+   11:09   0:00 grep --color=auto gammastep

✦ ❯ killall gammastep && echo 'Success!'
gammastep: no process found

✦ ❯ killall -r gammastep && echo 'Success!'
Success!

❯ gammastep &
[1] 46944

✦ ❯ pkill gammastep && echo 'Success!'
Success!
```
[1] https://unix.stackexchange.com/a/14482

heitorPB:
[...]

PR #1398, opened 2022-01-16
hueshift: Replace killall with pkill

stelcodes:
The usage of killall is unreliable when using wrapped executables such
as gammastep installed from the Nix package manager. When killall
doesn't kill the process, many hueshifting processes build up which
result in monitor flicker, total blackout, etc.

Addresses #1397


MaxVerevkin:
Thanks for pointing that out and providing a fix!

PR #1407, opened 2022-01-19
Add dracula theme

welcoMattic:
Based on https://draculatheme.com/contribute#color-palette

Render

dracula


MaxVerevkin:
I think it's better to set `separator = "\ue0b2"` to be more consistent with the other themes. Also, can you please add an image to https://github.com/greshake/i3status-rust/blob/master/doc/themes.md?

welcoMattic:
I didn't tested with Powerline separator. Then I'll try to make a screenshot similar to other themes

welcoMattic:
It miss the yellow color on the screenshot, is it important or not?

PR #1409, opened 2022-01-19
Issue-1405: UPower property type mismatch

### LINKED ISSUE #1405: Error in block 'battery': Failed to read UPower Type property
Just happened upon upgrading to 0.21.1 from 0.20.7. I am running arch linux. All packages are up-to-date on my system.

bim9262:
Fixes #1405 

State and Type need to be read in as u32 not i32.
This was missed when testing with `allow_missing = true`

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -458,11 +458,8 @@ pub struct UpowerDevice {
 
 impl UpowerDevice {
     /// Create the UPower device from the `device` string, which is converted to
-    /// the path `"/org/freedesktop/UPower/devices/battery_<device>"`, except if
-    /// `device` equals `"DisplayDevice"`, in which case it is converted to the
-    /// path `"/org/freedesktop/UPower/devices/DisplayDevice"`. Raises an error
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
Thanks!

FlorianFranzen:
This is still not completely fixed in '0.21.2', most likely due to the new default value and some broken error messages.

It use to default to `battery_BAT0` but now it just takes the first available value, which if the first device is not a battery, will still trigger `Error in block 'battery': Failed to read UPower Type property.`

I am just using a very basic block:
```toml
[[block]]
block = "battery"
driver = "upower"
```

But my first device is the power adapter:
```
$ upower --enumerate
/org/freedesktop/UPower/devices/line_power_ADP0
/org/freedesktop/UPower/devices/battery_BAT0
```

I would at least expect a `UPower device is not a battery` as the error message or even better, that i3bar-rust selects the first battery device in the enumeration.

FlorianFranzen:
Ok, this is much more messed up than I thought. Now also `DisplayDevice` does no longer work with `allow_missing`, while with`battery_BAT0` will produce a working block.

Issue #988, opened 2020-12-21
Processes spawned by xrandr block on brightness updates are not cleaned up properly


CyanoKobalamyne:
After I change the brightness by scrolling on the block, many new `xrandr` processes are spawned. They seem to be zombies ("Z" notation in `ps`/`top`/`htop`).

ammgws:
Should be fixed if you make a similar change:

https://github.com/greshake/i3status-rust/commit/75204ed5fdeee85170dc0fed35e7851716753285

Need to go through and check if there are any other places left that need fixing.

ammgws:
Also you wouldn't happen to also experience https://github.com/greshake/i3status-rust/issues/668?

ammgws:
Can you try #990? I don't have an X system to test with.

CyanoKobalamyne:
I haven't noticed any performance degradation like those mentioned in #668. (I'm also on Arch btw.)

I just checked out #990 and it fixed the issue!

PR #1411, opened 2022-01-22
hueshift: do not leave zombies

Naarakah:
The hueshift block spawn zombie processes when interacted with, similar to what is described in #988.
I fixed it the same way it was in #990, with `spawn_child_async`.

### DIFF
--- a/src/blocks/hueshift.rs
+++ b/src/blocks/hueshift.rs
@@ -1,4 +1,3 @@
-use std::process::Command;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
Since each driver at each point of time has at most one command running, maybe it would be better to store the `Child` and kill it before running the next command? This will also remove the need to use `pkill` in some drivers.

MaxVerevkin:
Anyway, this is already much better. And in the future `spawn_child_async()` probably will not spawn new threads (#1413).

PR #1412, opened 2022-01-22
Add srcery color theme

barskern:
https://srcery-colors.github.io/


GladOSkar:
Please also document this in doc/themes.md

barskern:
> Please also document this in doc/themes.md

Ah of course, will do.

barskern:
This fell through for me and I forgot about it, but now I spotted it again and uploaded an image. I used the example setup config that comes with i3status, but it is not the same as the other example bars. If you have the config that was used to generate the other images of the bars I could reupload an image.

files/themes/srcery.toml
```diff
@@ -0,0 +1,13 @@
...
+idle_fg = "#FCE8C3" # Bright White
+info_bg = "#1C1B19" # Black
+info_fg = "#FCE8C3" # Bright White
+good_bg = "#1C1B19" # Black
+good_fg = "#98BC37" # Bright Green
+warning_bg = "#1C1B19" # Black
+warning_fg = "#FED06E" # Bright Yellow
+critical_bg = "#1C1B19" # Black
+critical_fg = "#F75341" # Bright Red
+separator = ""
```
MaxVerevkin, line 11: Hmm, is this intentionally empty? Maybe at least just a space would be a better default?

MaxVerevkin:
An you please extend the example image to also demonstrate `info` and `warning` states?

files/themes/srcery.toml
barskern: Yes, this works really great for me actually. I enjoy the minimalism of it.

MaxVerevkin:
You can use `music` for info and muted `sound` for warning.

ammgws:
I wonder if we could script this to get consistent example images for each theme..

PR #1415, opened 2022-01-29
add if_command field to block config

### LINKED ISSUE #1326: Add test command for all blocks?
Just used my laptop for the first time in ages and remembered the below discussion from a while back. I feel like it would be a pretty handy feature when you are sharing the same config between multiple machines:

> I'm also sharing the configuration between multiple computers and was bitten by this issue.
> 
> I have another suggestion, what do you think about adding a tag for all blocks for a test command? E.g.:
> 
> ```toml
> [[block]]
> block = "battery"
> if_command = "upower -e"
> ```
> 
> If the command `upower -e` exits with an error code, the block will not be rendered.
> 

_Originally posted by @jlesquembre in https://github.com/greshake/i3status-rust/issues/50#issuecomment-362104891_

LordMZTE:
closes #1326

### DIFF
--- a/examples/config.toml
+++ b/examples/config.toml
@@ -33,3 +33,7 @@ block = "sound"
 block = "time"
 interval = 5
 format = "%a %d/%m %R"
+
+[[block]]
+block = "battery"
+if_command = "test -f /sys/class/power_supply/BAT0"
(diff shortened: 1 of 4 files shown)


MaxVerevkin:
Thanks!

LordMZTE:
Oh sorry! I should've thought of that. Well at least I got an example of it in the example config.

PR #1417, opened 2022-01-30
Issue 1405

### LINKED ISSUE #1405: Error in block 'battery': Failed to read UPower Type property

bim9262:
Fixes #1405

@FlorianFranzen pointed out two issues.

1. I changed how we matched the `device` to the `device_path`.
2. `DisplayDevice` does no longer work with `allow_missing`

First we can restore the original logic that was in place for how a
`device` to the `device_path`.
Second the `DisplayDevice` is not in the `EnumerateDevices` so we need
to account for that in is_available.

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -451,7 +451,8 @@ impl BatteryDevice for ApcUpsDevice {
 }
 
 pub struct UpowerDevice {
-    device_path: String,
+    device: String,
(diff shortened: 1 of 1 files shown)


ammgws:
Thanks!

PR #1421, opened 2022-02-02
Add wl-gammarelay driver for hueshift

bim9262:
Add support for wl-gammarelay.
While reviewing hueshift.rs It was noted that `Hueshift::update`
does not need to be called on the `update_interval`.

If drivers have a way of polling for the current temperature then it
makes sense to have an update interval otherwise it has no effect.
None of the drivers besides WlGammarelay has a mechanism to get the
current temperature if they are changed outside of the statusbar.
Although WlGammarelay can get the current temperature it doesn't need
to run update on an update interval as it is listening to dbus events.
Something like this:
```
Ok(match self.hue_shifter {
    HueShifter::X | HueShifter::Y => Some(self.update_interval.into()),
    _ => None,
})
```

Updated man pages (it looks like it's been some time since this has been
done). Perhaps it would be better if there was a step of the release
process that generated the updated man pages at that time.

### DIFF
--- a/src/blocks/hueshift.rs
+++ b/src/blocks/hueshift.rs
@@ -1,6 +1,11 @@
-use std::time::Duration;
+use std::sync::atomic::{AtomicU16, Ordering};
+use std::sync::Arc;
+use std::thread;
+use std::time::{Duration, Instant};
 
 use crossbeam_channel::Sender;
+use dbus::arg::RefArg;
+use dbus::ffidisp::stdintf::OrgFreedesktopDBusProperties;
 use serde_derive::Deserialize;
 
 use crate::blocks::{Block, ConfigBlock, Update};
@@ -18,11 +23,12 @@ use crate::widgets::I3BarWidget;
 pub struct Hueshift {
     id: usize,
     text: TextWidget,
-    update_interval: Duration,
+    // update_interval: Duration,
     step: u16,
     current_temp: u16,
     max_temp: u16,
     min_temp: u16,
+    hue_shifter: HueShifter,
     hue_shift_driver: Box<dyn HueShiftDriver>,
     click_temp: u16,
     scrolling: Scrolling,
@@ -31,6 +37,9 @@ pub struct Hueshift {
 trait HueShiftDriver {
     fn update(&self, temp: u16) -> Result<()>;
     fn reset(&self) -> Result<()>;
+    fn get_current_temperature(&mut self) -> Result<Option<u16>> {
+        Ok(None)
+    }
 }
 struct Redshift();
 impl HueShiftDriver for Redshift {
@@ -132,6 +141,123 @@ impl HueShiftDriver for Wlsunset {
     }
 }
 
+struct WlGammarelay {
+    con: dbus::ffidisp::Connection,
+    current_temperature: Arc<AtomicU16>,
+}
+
+impl WlGammarelay {
+    fn attempt_to_get_current_temperature(
+        con: &dbus::ffidisp::Connection,
+        delay: u64,
+        max_attempts: usize,
+    ) -> Result<u16> {
+        for attempt in 1..=max_attempts {
+            match con
+                .with_path("rs.wl-gammarelay", "/", 1000)
+                .get::<u16>("rs.wl.gammarelay", "Temperature")
+            {
+                Ok(temperature) => {
+                    return Ok(temperature);
+                }
+                Err(_) => {
+                    if attempt == max_attempts {
+                        return Err(BlockError(
+                            "hueshift".to_string(),
+                            "Unable to get current temperature for rs.wl.gammarelay".to_string(),
+                        ));
+                    } else {
+                        thread::sleep(Duration::from_millis(delay));
+                    }
+                }
+            }
+        }
+        Ok(0)
+    }
+
+    fn new(command: &str, id: usize, update_request: Sender<Task>) -> Result<Self> {
+        spawn_child_async(
+            "sh",
+            &["-c", format!("{} >/dev/null 2>&1", command).as_str()],
+        )
+        .block_error("hueshift", format!("Failed to start {}.", command).as_str())?;
+        let con = dbus::ffidisp::Connection::new_session()
+            .block_error("hueshift", "Failed to establish D-Bus connection.")?;
+
+        let current_temperature: Arc<AtomicU16> = Arc::new(AtomicU16::new(
+            WlGammarelay::attempt_to_get_current_temperature(&con, 100, 5)?,
+        ));
+
+        {
+            let current_temperature = current_temperature.clone();
+            thread::Builder::new()
+                .name("hueshift".into())
+                .spawn(move || {
+                    let con = dbus::ffidisp::Connection::new_session()
+                        .expect("Failed to establish D-Bus connection.");
+
+                    con.add_match(
+                        "type='signal',\
+                            interface='org.freedesktop.DBus.Properties',\
+                            member='PropertiesChanged',\
+                            arg0namespace='rs.wl.gammarelay'",
+                    )
+                    .expect("Failed to add D-Bus match rule.");
+
+                    // First we're going to get an (irrelevant) NameAcquired event.
+                    con.incoming(10_000).next();
+
+                    loop {
+                        if let Some(message) = con.incoming(10_000).next() {
+                            if let (_, Some(changed_properties)) =
+                                message.get2::<String, dbus::arg::PropMap>()
+                            {
+                                if let Some(temperature_variant) =
+                                    changed_properties.get("Temperature")
+                                {
+                                    if let Some(temperature) = temperature_variant.as_u64() {
+                                        let temperature = temperature as u16;
+                                        current_temperature.store(temperature, Ordering::SeqCst);
+                                        update_request
+                                            .send(Task {
+                                                id,
+                                                update_time: Instant::now(),
+                                            })
+                                            .unwrap();
+                                    }
+                                }
+                            }
+                        }
+                    }
+                })
+                .unwrap();
+        }
+
+        Ok(WlGammarelay {
+            con,
+            current_temperature,
+        })
+    }
+}
+
+impl HueShiftDriver for WlGammarelay {
+    fn update(&self, temp: u16) -> Result<()> {
+        self.con
+            .with_path("rs.wl-gammarelay", "/", 1000)
+            .set("rs.wl.gammarelay", "Temperature", temp)
+            .map_err(|e| BlockError("hueshift".to_string(), e.to_string()))?;
+        Ok(())
+    }
+    fn reset(&self) -> Result<()> {
+        // wl-gammarelay does not have a reset option just set the temp back to 6500
+        self.update(6500)
+    }
+
+    fn get_current_temperature(&mut self) -> Result<Option<u16>> {
+        Ok(Some(self.current_temperature.load(Ordering::SeqCst)))
+    }
+}
+
 #[derive(Deserialize, Debug, Clone)]
 #[serde(rename_all = "lowercase")]
 pub enum HueShifter {
@@ -139,6 +265,8 @@ pub enum HueShifter {
     Sct,
     Gammastep,
     Wlsunset,
+    WlGammarelay,
+    WlGammarelayRs,
 }
 
 #[derive(Deserialize, Debug, Clone)]
@@ -178,6 +306,10 @@ impl Default for HueshiftConfig {
                 Some(HueShifter::Gammastep)
             } else if has_command("hueshift", "wlsunset").unwrap_or(false) {
                 Some(HueShifter::Wlsunset)
+            } else if has_command("hueshift", "wl-gammarelay-rs").unwrap_or(false) {
+                Some(HueShifter::WlGammarelayRs)
+            } else if has_command("hueshift", "wl-gammarelay").unwrap_or(false) {
+                Some(HueShifter::WlGammarelay)
             } else {
                 None
             },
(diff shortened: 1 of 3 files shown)


src/blocks/hueshift.rs
```diff
@@ -178,6 +259,8 @@ impl Default for HueshiftConfig {
                 Some(HueShifter::Gammastep)
             } else if has_command("hueshift", "wlsunset").unwrap_or(false) {
                 Some(HueShifter::Wlsunset)
+            } else if has_command("hueshift", "wl-gammarelay").unwrap_or(false) {
```
MaxVerevkin, line 262: Please check for either `wl-gammarelay` or `wl-gammarelay-rs`, since they have compatible DBus interfase.

MaxVerevkin:
Please move the mangapge update into a separate commit.

src/blocks/hueshift.rs
```diff
@@ -132,13 +138,88 @@ impl HueShiftDriver for Wlsunset {
...
+                .expect("Failed to add D-Bus match rule.");
+
+                loop {
+                    if let Some(msg) = con.incoming(10_000).next() {
+                        if let Some(interface) =
+                            msg.interface().map(|interface| interface.to_string())
+                        {
+                            if interface == "org.freedesktop.DBus.Properties"
+                                && msg.get1::<String>().expect("Unable to get string argument")
+                                    == "rs.wl.gammarelay"
```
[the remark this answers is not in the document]
bim9262: Thanks for letting me know! That would have made the upower `monitor` a bit cleaner too.

bim9262:
> Please move the mangapge update into a separate commit.

Will do

bim9262:
@MaxVerevkin , I think I covered all your suggestions/requests. Let me know if there's anything else I should change.

src/blocks/hueshift.rs
```diff
@@ -132,13 +138,81 @@ impl HueShiftDriver for Wlsunset {
...
+                        update_request
+                            .send(Task {
+                                id,
+                                update_time: Instant::now(),
+                            })
+                            .unwrap();
+
+                        // Avoid update spam.
+                        // TODO: Is this necessary?
+                        thread::sleep(Duration::from_millis(1000))
```
MaxVerevkin, line 175: Not sure if this is necessary, but if it is, one second is too long.

MaxVerevkin:
I think it might also be a good idea to start the daemon if it isn't started already (see the implementation in `async` branch).

bim9262:
> I think it might also be a good idea to start the daemon if it isn't started already (see the implementation in `async` branch).

I had trouble starting a second wl-gammarelay if one is already running. 
First instance:
```
$ wl-gammarelay; echo $?
failed to set gamma table
failed to set gamma table
```
Second instance:
```
$ wl-gammarelay; echo $?
panic: failed to connect to dbus: failed to register dbus: name already taken

goroutine 1 [running]:
main.main()
        ~/.cache/paru/clone/wl-gammarelay/src/wl-gammarelay-0.0.8/main.go:57 +0x5a
2
```

I know we could look at ps to see if the process already running, is there a cleaner way to do that in rust that you'd suggest?

MaxVerevkin:
> is there a cleaner way to do that in rust that you'd suggest?

Just ignore the error?

bim9262:
> Just ignore the error?

Although the first instance didn't die it stops setting the gamma value (the screen is returned to 6500k)

MaxVerevkin:
> > Just ignore the error?
> 
> Although the first instance didn't die it stops setting the gamma value (the screen is returned to 6500k)

Please report this bug.

bim9262:
Thanks for filing the bug for wl-gammarelay. I'd already gone to bed last night.
So I was thinking about the loop where we wait for a dbus signal. It seems like adding in the delay wouldn't actually reduce the number of times we perform the update, only slows the rate of updates.

This can be see with this small demo program:

```
use std::io::Error;
use std::result::Result;
use std::{thread, time};

fn main() -> Result<(), Error> {
    let con = dbus::ffidisp::Connection::get_private(dbus::ffidisp::BusType::Session)
        .expect("Failed to establish D-Bus connection.");

    con.add_match(
        "type='signal',\
                    interface='org.freedesktop.DBus.Properties',\
                    member='PropertiesChanged',\
                    arg0namespace='rs.wl.gammarelay'",
    )
    .expect("Failed to add D-Bus match rule.");

    // First we're going to get an (irrelevant) NameAcquired event.
    con.incoming(10_000).next();

    let sleep_for = time::Duration::from_millis(500);

    loop {
        match con.incoming(10_000).next() {
            Some(message) => {
                println!("{:?}", message);
                thread::sleep(sleep_for);
            }
            _ => println!("Nothing happened"),
        }
    }
}

```

 It seems like once we know there we should consume all of the messages we can at the desired max update rate instead. Something like this:
 
 ```
use std::io::Error;
use std::result::Result;

fn main() -> Result<(), Error> {
    let con = dbus::ffidisp::Connection::get_private(dbus::ffidisp::BusType::Session)
        .expect("Failed to establish D-Bus connection.");

    // First we're going to get an (irrelevant) NameAcquired event.
    con.incoming(10_000).next();

    con.add_match(
        "type='signal',\
                    interface='org.freedesktop.DBus.Properties',\
                    member='PropertiesChanged',\
                    arg0namespace='rs.wl.gammarelay'",
    )
    .expect("Failed to add D-Bus match rule.");

    loop {
        let mut was_message = false;
        // Read as many messages as possible in the following interval.
        for message in con.incoming(500) {
            println!("{:?}", message);
            was_message = true;
        }
        if was_message {
            println!("Time to update!");
        } else {
            println!("Nothing happened")
        }
    }
}
```

**Comment by MaxVerevkin** [comment]
> It seems like once we should consume all of the messages we can at the desired max update rate instead. Something like this:

Or, a better approach would be to filter the events and request an update only when temperature actually changes. (`zbus` would be nice to use here.)

**Comment by bim9262** [comment]
> Or, a better approach would be to filter the events and request an update only when temperature actually changes. (`zbus` would be nice to use here.)

Is there a way to achieve that with sync rust? (I know you're working on a migration to an async code base)

**Comment by MaxVerevkin** [comment]
> Is there a way to achieve that with sync rust?

Sure, its a bit more complicated, but you have to create ["property changed"](https://docs.rs/dbus/latest/dbus/ffidisp/stdintf/org_freedesktop_dbus/struct.PropertiesPropertiesChanged.html) event from the message and then check which property changed.

**Comment by bim9262** [comment]
Perhaps something like this?

```
use std::result::Result;
use std::time::Duration;

fn main() -> Result<(), dbus::Error> {
    let con =
        dbus::blocking::Connection::new_session().expect("Failed to establish D-Bus connection.");

    con.add_match(
        dbus::message::MatchRule::new_signal(
            "org.freedesktop.DBus.Properties",
            "PropertiesChanged",
        ),
        |signal: dbus::blocking::stdintf::org_freedesktop_dbus::PropertiesPropertiesChanged,
         _: &dbus::blocking::Connection,
         _: &dbus::message::Message| {
            if signal.interface_name == "rs.wl.gammarelay" {
                println!("Time to update!");
            }
            true
        },
    )?;

    // Listen to incoming signals forever.
    loop {
        con.process(Duration::from_millis(1000))?;
    }
}
```

**Comment by MaxVerevkin** [comment]
This doesn't really change anything. The issue here is that the block will request an update even if the brightness changes. But we should care only about temperature.

Also, each time you receive a signal containing the new temperature, you request the block to read the new value from dbus, even though you already have it. You can solve this by sharing an `Arc<AtomicU16>` between the driver and the spawned thread.

**Comment by MaxVerevkin** [comment]
By the way, you can specify the language in markdown to enable [syntax highlighting](https://www.markdownguide.org/extended-syntax/#syntax-highlighting).

**Comment by bim9262** [comment]
Here's the demo implementation I came up with the `Arc<AtomicU16>`, so that we don't need to look up the already known value. The bit with the variant as_u64 -> u16 seems messy, but I haven't been able to find a cleaner way to do the conversion.

```rs
use dbus::arg::RefArg;
use dbus::ffidisp::stdintf::OrgFreedesktopDBusProperties;
use std::io::Error;
use std::result::Result;
use std::sync::atomic::{AtomicU16, Ordering};
use std::sync::Arc;
use std::thread;
use std::time::Duration;

fn main() -> Result<(), Error> {
    let con = dbus::ffidisp::Connection::get_private(dbus::ffidisp::BusType::Session)
        .expect("Failed to establish D-Bus connection.");

    let current_temperature: Arc<AtomicU16> = Arc::new(AtomicU16::new(
        con.with_path("rs.wl-gammarelay", "/", 1000)
            .get("rs.wl.gammarelay", "Temperature")
            .unwrap(),
    ));

    {
        let current_temperature = current_temperature.clone();
        thread::Builder::new()
            .name("hueshift".into())
            .spawn(move || {
                let con = dbus::blocking::Connection::new_session()
                .expect("Failed to establish D-Bus connection.");

                con.add_match(
                    dbus::message::MatchRule::new_signal(
                        "org.freedesktop.DBus.Properties",
                        "PropertiesChanged",
                    ),
                    move |signal: dbus::blocking::stdintf::org_freedesktop_dbus::PropertiesPropertiesChanged,
                _: &dbus::blocking::Connection,
                _: &dbus::message::Message| {
                    if signal.interface_name == "rs.wl.gammarelay" {
                        if let Some(temperature_variant) = signal.changed_properties.get("Temperature") {
                            if let Some(temperature) = temperature_variant.as_u64()  {
                                let temperature = temperature as u16;
                                println!("Time to update the current temperature to : {}", temperature);
                                current_temperature.store(temperature, Ordering::SeqCst)
                            }

                        }
                    }
                    true
                }
            ).unwrap();

            // Listen to incoming signals forever.
            loop {
                con.process(Duration::from_millis(1000)).unwrap();
            }
        })
        .unwrap();
    }

    let sleep_for = Duration::from_millis(1000);

    loop {
        println!(
            "Current Temperature: {}",
            current_temperature.load(Ordering::SeqCst)
        );
        thread::sleep(sleep_for);
    }
}
```

MaxVerevkin:
Looks about right, however I'd prefer a text match rule, since in your last example the callback will be run when _any_ property on the bus changes, which seems inefficient.

bim9262:
@MaxVerevkin, in order to use a callback you need to use a MatchRule instead of a text match rule:

```rs
pub fn add_match<S: ReadAll, F>(
    &self,
    match_rule: MatchRule<'static>,
    f: F
) -> Result<Token, Error>
where
    F: FnMut(S, &Self, &Message) -> bool + Send + 'static, 
```
Adds a new match to the connection, and sets up a callback when this message arrives.

The returned value can be used to remove the match. The match is also removed if the callback returns "false".

```rs
pub fn add_match_no_cb(&self, match_str: &str) -> Result<(), Error>
```

Adds a new match to the connection, without setting up a callback when this message arrives.

(From see https://docs.rs/dbus/0.9.0/dbus/blocking/struct.Connection.html)

MaxVerevkin:
Yeah, I know, but I don't see why using a callback is important.

bim9262:
I thought that you were suggesting that we "create ["property changed"](https://docs.rs/dbus/latest/dbus/ffidisp/stdintf/org_freedesktop_dbus/struct.PropertiesPropertiesChanged.html) event from the message and then check which property have changed" by using that callback mechanism. I can blend the `dbus::ffidisp::Connection` solution with the `Arc` though.

```rs
use dbus::arg::RefArg;
use dbus::ffidisp::stdintf::OrgFreedesktopDBusProperties;

use std::io::Error;
use std::result::Result;
use std::sync::atomic::{AtomicU16, Ordering};
use std::sync::Arc;
use std::thread;
use std::time::Duration;

fn main() -> Result<(), Error> {
    let con = dbus::ffidisp::Connection::get_private(dbus::ffidisp::BusType::Session)
        .expect("Failed to establish D-Bus connection.");

    let current_temperature: Arc<AtomicU16> = Arc::new(AtomicU16::new(
        con.with_path("rs.wl-gammarelay", "/", 1000)
            .get("rs.wl.gammarelay", "Temperature")
            .unwrap(),
    ));

    {
        let current_temperature = current_temperature.clone();
        thread::Builder::new()
            .name("hueshift".into())
            .spawn(move || -> ! {
                let con = dbus::ffidisp::Connection::get_private(dbus::ffidisp::BusType::Session)
                    .expect("Failed to establish D-Bus connection.");

                con.add_match(
                    "type='signal',\
                            interface='org.freedesktop.DBus.Properties',\
                            member='PropertiesChanged',\
                            arg0namespace='rs.wl.gammarelay'",
                )
                .expect("Failed to add D-Bus match rule.");

                // First we're going to get an (irrelevant) NameAcquired event.
                con.incoming(10_000).next();

                loop {
                    if let Some(message) = con.incoming(10_000).next() {
                        let (_, changed_properties_option) =
                            message.get2::<String, dbus::arg::PropMap>();

                        if let Some(changed_properties) = changed_properties_option {
                            if let Some(temperature_variant) = changed_properties.get("Temperature")
                            {
                                if let Some(temperature) = temperature_variant.as_u64() {
                                    let temperature = temperature as u16;
                                    current_temperature.store(temperature, Ordering::SeqCst);
                                    println!(
                                        "Time to update the current temperature to : {}",
                                        temperature
                                    );
                                }
                            }
                        }
                    } else {
                        println!("Nothing happened")
                    }
                }
            })
            .unwrap();
    }

    let sleep_for = Duration::from_millis(1000);

    loop {
        println!(
            "Current Temperature: {}",
            current_temperature.load(Ordering::SeqCst)
        );
        thread::sleep(sleep_for);
    }
}
```

MaxVerevkin:
> > Just ignore the error?
> 
> Although the first instance didn't die it stops setting the gamma value (the screen is returned to 6500k)

It's fixed now.

MaxVerevkin:
LGTM. Thanks!

bim9262:
You're welcome :)

PR #1423, opened 2022-02-03
Fix a typo in blocks.md

E011011101001:
#externalip is the correct anchor as there is no space in title `ExternalIP`.


ammgws:
Thanks!

Issue #401, opened 2019-07-14
Produce Static Binaries for Releases


greshake:
I tried setting up static builds using Docker containers based on either Alpine and Ubuntu with musl versions of Rust. Unfortunately the builds fail because proc_macro used by serde_derive can't be used in static builds. I'm not quite sure if there will be a workaround for this.

You can try this yourself with this Dockerfile (build with `docker build -t i3status-rust .`):
```
FROM alpine:3.10

RUN echo 'http://dl-cdn.alpinelinux.org/alpine/edge/testing' >> /etc/apk/repositories

RUN apk update && \
    apk add dbus-dev pulseaudio rustup

RUN /usr/bin/rustup-init -y \
    --default-host x86_64-unknown-linux-musl \
    --default-toolchain stable
```

Execute this from the main repo folder to get a build:
```
docker run --rm -it -v "$(pwd)":/src i3status-rust sh -c 'cd src && cargo build --release'
```

ammgws:
@atheriel I saw [this](https://github.com/alexcrichton/proc-macro2/issues/139#issuecomment-621558749) comment today and gave it a go - it built successfully.

Dockerfile:
```
FROM alpine:3.11

RUN apk update && \
    apk add dbus-dev pulseaudio build-base

RUN apk add --update-cache \
--repository https://alpine.global.ssl.fastly.net/alpine/edge/community \
rustup

RUN /usr/bin/rustup-init -y \
    --default-host x86_64-unknown-linux-musl \
    --default-toolchain stable

RUN source $HOME/.cargo/env

ENV PATH /root/.cargo/bin:$PATH
```

`docker run --rm -it -v "$(pwd)":/src i3status-rust sh -c 'cd src && RUSTFLAGS="-C target-feature=-crt-static" cargo build --release'`

ammgws:
Last time I gave it a go, I was able to compile, but the resulting binary wasn't executable. Maybe someone else can try?

GladOSkar:
I actually get this compile error:

```
error: failed to run custom build command for `openssl-sys v0.9.60`

Caused by:
  process didn't exit successfully: `/src/target/release/build/openssl-sys-1f4459c902b3ad26/build-script-main` (exit code: 101)
  --- stdout
  cargo:rustc-cfg=const_fn
  cargo:rerun-if-env-changed=X86_64_UNKNOWN_LINUX_MUSL_OPENSSL_LIB_DIR
[...]

PR #1436, opened 2022-02-17
Cargo feature `static` to build on static binary-based systems

Nurrl:
Hello,

I've been wanting to try your project instead of `i3status` in my Alpine Linux v3.15 installation, but using either `cargo install` or the _clone_, _build_ & _install_ workflow didn't end up working.
I've seen #401, and I didn't try this one out, but I came up with this fix, so we could use `cargo install --features static` directly in Alpine.
Also I never did Alpine packages, but I though this would be a great first project to start as a package maintainer, so I might propose an `apk` package for your project at some point.

Thanks :)


Nurrl:
It worked with the fix from #401, closing to repropose a better fix.

PR #1437, opened 2022-02-19
Don't enumerate UPower devices every update

### LINKED ISSUE #1434: Frequent calls of EnumerateDevices in battery block with upower backend
I noticed that since version `0.21.1` the status bar calls roughly every second the `EnumerateDevices` method on the `dbus` to check if the battery device has changed. I think this should not be the desired behavior because there is already a monitoring function that - if I understand it correctly - updates the block only when a matching signal occurs on the bus. But there must be a bug because the update function gets called much more frequently. 
### Reproduction
config:
```toml
[[block]]
block = "battery"
driver = "upower"
allow_missing = true
device = "DisplayDevice"
format = "{percentage} {time}"
```
monitor the dbus with `dbus-monitor --system`:
```log
method call time=1645091812.673867 sender=:1.24 -> destination=org.freedesktop.UPower serial=75 path=/org/freedesktop/UPower; interface=org.freedesktop.UPower; member=EnumerateDevices
method return time=1645091812.674802 sender=:1.22 -> destination=:1.24 serial=246 reply_serial=75
   array [
      object path "/org/freedesktop/UPower/devices/line_power_AC"
      object path "/org/freedesktop/UPower/devices/battery_BAT0"
   ]
method call time=1645091812.675423 sender=:1.24 -> destination=org.freedesktop.UPower serial=76 path=/org/freedesktop/UPower; interface=org.freedesktop.UPower; member=EnumerateDevices
method return time=1645091812.676267 sender=:1.22 -> destination=:1.24 serial=247 reply_serial=76
   array [
      object path "/org/freedesktop/UPower/devices/line_power_AC"
      object path "/org/freedesktop/UPower/devices/battery_BAT0"
   ]
method call time=1645091813.670912 sender=:1.24 -> destination=org.freedesktop.UPower serial=77 path=/org/freedesktop/UPower; interface=org.freedesktop.UPower; member=EnumerateDevices
method return time=1645091813.671905 sender=:1.22 -> destination=:1.24 serial=248 reply_serial=77
   array [
      object path "/org/freedesktop/UPower/devices/line_power_AC"
      object path "/org/freedesktop/UPower/devices/battery_BAT0"
   ]
method call time=1645091813.672633 sender=:1.24 -> destination=org.freedesktop.UPower serial=78 path=/org/freedesktop/UPower; interface=org.freedesktop.UPower; member=EnumerateDevices
method return time=1645091813.673474 sender=:1.22 -> destination=:1.24 serial=249 reply_serial=78
   array [
      object path "/org/freedesktop/UPower/devices/line_power_AC"
      object path "/org/freedesktop/UPower/devices/battery_BAT0"
   ]
```

bim9262:
Fixes #1434

Instead of enumerating the devices each time the block is updated
we can ask for the device enumeration only when a device is added
or removed if allow_missing is true.


bim9262:
@MaxVerevkin this could be further improved by not querying for the various properties and update based on the emitted/changed values (like how we did in hueshift), but at least this fixes the device enumerate spam..

MaxVerevkin:
I wonder if we should just switch to zbus...

SilverlightningY:
Well done fix - but I'm afraid that this only fixes the dbus spamming if the `device` is set to `DisplayDevice`. If someone sets it to e.g. `battery_BAT0` in my case, the update function gets called again every second. The `is_available` function checks only for `self.device == "DisplayDevice"` so the call to dbus is prevented in that case, but in any other, it is made. I'm not quite sure but can it be that the scheduler updates this block regardless the update function returns `None`?

MaxVerevkin:
Thanks for fixing this :)

bim9262:
You're welcome :smile:

PR #1438, opened 2022-02-19
sound: check DEVICE_FORM_FACTOR property

kevinmost:
DEVICE_FORM_FACTOR, if present, will be used to determine whether the
headphones icon is used. If it is not present, the existing active_port
method is used to determine if the headphones icon is used.

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -48,8 +48,8 @@ trait SoundDevice {
     fn muted(&self) -> bool;
     fn output_name(&self) -> String;
     fn output_description(&self) -> Option<String>;
-    fn active_port(&self) -> Option<String>;
-    fn form_factor(&self) -> Option<String>;
+    fn active_port(&self) -> Option<&str>;
+    fn form_factor(&self) -> Option<&str>;
 
     fn get_info(&mut self) -> Result<()>;
     fn set_volume(&mut self, step: i32, max_vol: Option<u32>) -> Result<()>;
@@ -94,10 +94,10 @@ impl SoundDevice for AlsaSoundDevice {
         // TODO Does Alsa has something similar like descripitons in Pulse?
         None
     }
-    fn active_port(&self) -> Option<String> {
+    fn active_port(&self) -> Option<&str> {
         None
     }
-    fn form_factor(&self) -> Option<String> {
+    fn form_factor(&self) -> Option<&str> {
         None
     }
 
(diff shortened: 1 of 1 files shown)


codicodi:
Looks alright. FWIW [Waybar](https://github.com/Alexays/Waybar/blob/ad5ea7ad2b698c7a60c6bb7e81f06a662cdba819/src/modules/pulseaudio.cpp#L214-L237) folks seem to also check both form factor and port name

kevinmost:
Great, thanks for looking

src/blocks/sound.rs
```diff
@@ -653,6 +663,10 @@ impl SoundDevice for PulseAudioSoundDevice {
         self.active_port.clone()
     }
 
+    fn form_factor(&self) -> Option<String> {
+        self.form_factor.clone()
```
MaxVerevkin, line 667: This function can return `&str` in order to avoid `clone()`

src/blocks/sound.rs
```diff
@@ -653,6 +663,10 @@ impl SoundDevice for PulseAudioSoundDevice {
         self.active_port.clone()
```
MaxVerevkin, line 663: Not part of your PR, but can you please remove this `clone()` too?

src/blocks/sound.rs
kevinmost: How's this?

src/blocks/sound.rs
kevinmost: Ditto

src/blocks/sound.rs
```diff
@@ -825,9 +840,27 @@ impl Default for SoundConfig {
 }
 
 impl Sound {
-    fn icon(&self, volume: u32, headphones: bool) -> String {
-        if self.headphones_indicator && self.device_kind == DeviceKind::Sink && headphones {
-            return String::from("headphones");
+    fn icon(&self, volume: u32) -> String {
+        if self.headphones_indicator && self.device_kind == DeviceKind::Sink {
+            let headphones = match self.device.form_factor().as_deref() {
```
MaxVerevkin, line 845: Now you can remove this `as_deref()` as it doesn't do anything.

src/blocks/sound.rs
```diff
@@ -825,9 +840,27 @@ impl Default for SoundConfig {
...
+                Some("headset") | Some("headphone") | Some("hands-free") | Some("portable") => true,
+                // Per discussion at
+                // https://github.com/greshake/i3status-rust/pull/1363#issuecomment-1046095869,
+                // some sinks may not have the form_factor property, so we should fall back to the
+                // active_port if that property is not present.
+                None => self
+                    .device
+                    .active_port()
+                    .map(|p| p.contains("headphones"))
+                    .unwrap_or(false),
```
MaxVerevkin, line 857: Here you can use `.map_or(false, |p| p.contains("headphones"))`

kevinmost:
Done, thanks.

MaxVerevkin (approved):

MaxVerevkin:
Thanks!

Issue #1435, opened 2022-02-17
Continuous output for `custom` block


kraftwerk28:
Can i3status-rust handle a script, that hangs and continuously outputs a lines of text / JSON's, where for each line a block should be updated? It is, for example, the default behavior for custom blocks in waybar, if no `interval` is provided.
For example:
```bash
while sleep 1; do
  echo '{"text": "'"$(date)"'"}'
done
```

ammgws:
You can do this with a `CustomDbus` block - instead of `echo` in your script, use the appropriate dbus call. Then i3status-rs picks it up instantaneously and updates the bar. Can you give that a go first?

kraftwerk28:
That seems like a decent workaround, thank you

PR #1441, opened 2022-03-10
Implement "persist" command feature from i3blocks

an-empty-string:
In i3blocks, a block can be specified with interval "persist", which means its associated command runs as a child of i3blocks. Each line on that process's stdout is used to update in the block in realtime. I used this functionality to display MQTT messages in my statusline (with `mosquitto_sub`).

This PR adds similar functionality -- set a `custom` block's `persistent` config key to `true`, and its associated command will be run persistently; each line will update the block's contents.

I've tested to make sure that this doesn't break regular `command` and `cycle` functionality but haven't gone beyond that. I'm using this patch right now to run [a pomodoro tool](https://gist.github.com/an-empty-string/a4706537cd9ffaf5f67d3f49a8540fea) that talks MQTT and it seems to be working reliably so far!

This is my first time writing a nontrivial amount of Rust, so please let me know if I can improve anything! If I need to open an issue to get feedback on whether or not this is a broadly useful feature I can do that too, I just figured that since I'd already written the patch going straight to PR couldn't hurt :)

### DIFF
--- a/src/blocks/custom.rs
+++ b/src/blocks/custom.rs
@@ -13,6 +13,7 @@
 //! Key | Values | Required | Default
 //! ----|--------|----------|--------
 //! `command` | Shell command to execute & display | No | None
+//! `persistent` | Run command in the background; update display for each output line of the command | No | false
 //! `cycle` | Commands to execute and change when the button is clicked | No | None
 //! `interval` | Update interval in seconds (or "once" to update only once) | No | `10`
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
Looks good to me!

> going straight to PR couldn't hurt :)

Sure :)

> whether or not this is a broadly useful feature

It was requested at least once: #1435, so big thanks for implementing it!

Issue #1396, opened 2022-01-16
Enable [Music] block to only show specified players


epixinvites:
We're able to specify an array of players or regex of players we want to ignore, why not let us have an array of players we only want to show show up? 
Please correct me if I'm wrong, but as far as I've tried, I can't do something like `player = ["ncspot", "cmus"]`. It tells me the `player` variable only takes a string.
 
I know that maybe in some rare cases someone has two or more players on at the same time, so maybe letting us change the players with right-click would do the job?

ammgws:
IIRC the only reason you cannot do it right now is because we didn't want to break configs by changing the config from taking a string to taking an array. Perhaps a breaking change such as this could be bundled along with the other changes that are upcoming in the v0.3 async release...

epixinvites:
Can't you accept a string *or* an array? That way the current configs won't break, will it?

epixinvites:
I don't have much experience in Rust, sorry

PR #1452, opened 2022-03-30
Music: Add `players` field to the config

meryacine:
Addresses #1396 

The `players` field holds an array of players, which, along with the `player`, will be considered as the preferred players.
I couldn't build due to some linking errors, but `cargo check`ed successfully.

### DIFF
--- a/src/blocks/music.rs
+++ b/src/blocks/music.rs
@@ -186,6 +186,11 @@ pub struct MusicConfig {
     /// the block will track all players found.
     pub player: Option<String>,
 
+    /// Name of the music players. Same as the `player` field above but
+    /// for an array of players. The `player` field above is kept for
+    /// backward compatibility with older configurations.
+    pub players: Option<Vec<String>>,
+
     /// Max width of the block in characters, not including the buttons.
     pub max_width: usize,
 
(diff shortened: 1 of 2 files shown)


MaxVerevkin:
As discussed earlier, it should be possible to extend the `player` option to accept either a string or a list. Basically, you can create an "untagged" enum with two options: a string and a vector. See https://serde.rs/enum-representations.html for more info.

src/blocks/music.rs
```diff
@@ -287,13 +295,19 @@ impl ConfigBlock for Music {
                 500,
             )
             .unwrap();
+
+        let mut preferred_players = Vec::new();
+
+        if let Some(player) = block_config.player {
+            match player {
+                PlayerName::PlayerName(player) => preferred_players.push(player.clone()),
+                PlayerName::PlayerNames(players) => preferred_players.extend(players.clone()),
```
MaxVerevkin, line 304: You can avoid extra allocations here: `preferred_players = players` should be sufficient (I haven't tested this).

MaxVerevkin:
Please fix a clippy lint

MaxVerevkin:
Thanks, sorry for the delay.

Issue #1246, opened 2021-05-18
Discussion: home expansion for block paths


mplanchard:
Hello!

I was wondering how y'all would feel about a PR to automatically expand paths in the various blocks to allow specification of paths like `~/mail/gmail/Inbox` or similar.

The use case for me personally is so that I can easily check my `i3status-rust/config.coml` into my git-tracked dotfiles and have it work on systems with a variety of usernames.

We could use something like [shellexpand](https://docs.rs/shellexpand/2.1.0/shellexpand/) to do the hard work, and we could potentially limit expansion to _just_ the tilde, rather than also interpolating in env vars, to keep things simple. Alternatively we could manually look for the `~` at the beginning of the path and use the existing `xdg_config_home()` utility function to replace it.

The implementation could either do the expansion at the point of the `ConfigBlock::new()` or the point of config deserialization using `deserialize_with`. The type could remain `String` for simplicity of integrating with any existing code, or we could turn it into a `Path` or some custom struct (with appropriate `From` impls) if desired to provide a semantic hint that it's not a normal string.

Let me know if there's interest here and if so what approach y'all would like to see, and I'd be happy to work on it.

ammgws:
Which blocks would benefit from this apart from maildir?

ammgws:
@mplanchard We now will soon include `shellexpand` once #1325 is merged, so you can start on this if you want.

mplanchard:
Thanks a ton! Work's gotten pretty busy recently, so I can't guarantee
speedy progress, but I will work on it in fits and starts.

ammgws:
Now that we depend on shellexpand and already have some blocks using it, we just need to go through the remaining blocks and add the functionality. (and add docs where necessary)

MaxVerevkin:
Implemented in #1453

PR #1453, opened 2022-04-05
Shellexpand on blocks

Henriquelay:
As discussed in #1246, this runs `shellexpand::full` on paths the user might enter, thus expanding shell syntax like `~/file` and `${HOME}/file` to `/home/user/file`.
This is a created as a draft because I couldn't properly teste some of the changes, as I don't use these mail clients like a lot of users do, I take quite a beating from these cli clients, so I can't properly test maildir, notmuch and watson, all I could do is to try the config appropriately, and the error it spat out was not about _i3status-rust_'s config, or none at all, but I couldn't validate that I properly set up the program and whatnot. Watson requires extra attention because it uses `Pathbuf` on the config instead of `String`, like the others, since it needs it to spawn a watcher thread over it. I need some help here.
Since it's my first PR, I'd like some insight on idiomacies, since I've encountered some hard time with watson, but for the most time, followed what @ammgws did on custom block.

My method was to skim through the documentation, checking for anywhere a path was required, jumping in and change that to use shellexpand, then update the `.md` appropriately.

### DIFF
--- a/src/blocks/maildir.rs
+++ b/src/blocks/maildir.rs
@@ -77,6 +77,21 @@ impl ConfigBlock for Maildir {
         shared_config: SharedConfig,
         _tx_update_request: Sender<Task>,
     ) -> Result<Self> {
+        let expanded_inboxes = block_config
+            .inboxes
+            .iter()
+            .map(|inbox| {
+                shellexpand::full(inbox)
+                    .map_err(|e| {
+                        ConfigurationError(
+                            "maildir".to_string(),
+                            format!("Failed to expand inbox path {}: {}", inbox, e),
+                        )
+                    })
+                    .unwrap()
+                    .to_string()
+            })
+            .collect();
         let widget = TextWidget::new(id, 0, shared_config).with_text("");
         Ok(Maildir {
             id,
@@ -86,7 +101,7 @@ impl ConfigBlock for Maildir {
             } else {
                 widget
             },
-            inboxes: block_config.inboxes,
+            inboxes: expanded_inboxes,
             threshold_warning: block_config.threshold_warning,
             threshold_critical: block_config.threshold_critical,
             display_type: block_config.display_type,
--- a/src/blocks/maildir.rs
+++ b/src/blocks/maildir.rs
@@ -73,25 +73,20 @@ impl ConfigBlock for Maildir {
 
     fn new(
         id: usize,
-        block_config: Self::Config,
+        mut block_config: Self::Config,
         shared_config: SharedConfig,
         _tx_update_request: Sender<Task>,
     ) -> Result<Self> {
-        let expanded_inboxes = block_config
-            .inboxes
-            .iter()
-            .map(|inbox| {
-                shellexpand::full(inbox)
-                    .map_err(|e| {
-                        ConfigurationError(
-                            "maildir".to_string(),
-                            format!("Failed to expand inbox path {}: {}", inbox, e),
-                        )
-                    })
-                    .unwrap()
-                    .to_string()
-            })
-            .collect();
+        for inbox in &mut block_config.inboxes {
+            *inbox = shellexpand::full(inbox)
+                .map_err(|e| {
+                    ConfigurationError(
+                        "maildir".to_string(),
+                        format!("Failed to expand inbox path {}: {}", inbox, e),
+                    )
+                })?
+                .to_string();
+        }
         let widget = TextWidget::new(id, 0, shared_config).with_text("");
         Ok(Maildir {
             id,
@@ -101,7 +96,7 @@ impl ConfigBlock for Maildir {
             } else {
                 widget
             },
-            inboxes: expanded_inboxes,
+            inboxes: block_config.inboxes,
             threshold_warning: block_config.threshold_warning,
             threshold_critical: block_config.threshold_critical,
             display_type: block_config.display_type,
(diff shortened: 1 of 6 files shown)


src/blocks/maildir.rs
```diff
@@ -77,6 +77,21 @@ impl ConfigBlock for Maildir {
...
+            .iter()
+            .map(|inbox| {
+                shellexpand::full(inbox)
+                    .map_err(|e| {
+                        ConfigurationError(
+                            "maildir".to_string(),
+                            format!("Failed to expand inbox path {}: {}", inbox, e),
+                        )
+                    })
+                    .unwrap()
```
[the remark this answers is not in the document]
MaxVerevkin: Or, alternatively (perhaps preferably), this can be done in-place:
```rust
for inbox in &mut block_config.inboxes {
    *inbox = shellexpand::full(inbox).map_err(...)?;
}
```

src/blocks/watson.rs
```diff
@@ -64,10 +64,25 @@ impl ConfigBlock for Watson {
         shared_config: SharedConfig,
         tx_update_request: Sender<Task>,
     ) -> Result<Self> {
+        let state_path_string: String = block_config
+            .state_path
+            .clone()
+            .into_os_string()
+            .into_string()
+            .unwrap();
```
MaxVerevkin, line 72: No need to clone here: just use `state_path.to_str().unwrap()`.

P.S. not sure if panicking on non UTF8 path is the right thing to do, perhaps  `state_path.to_string_loosy()`?

MaxVerevkin (changes requested):
Thanks for this PR.

It might also be a good idea to create a newtype for expandable strings, to make the code a bit cleaner. See [async branch' `ShellString`](https://github.com/greshake/i3status-rust/blob/c338b15a3f42c3605a4a2af2d10bef12d558b965/src/wrappers.rs#L123). But it's probably a topic for a new PR.

src/blocks/watson.rs
Henriquelay: I saw the `state_path.to_string_loosy()`, and wasn't sure on what to do, that's why I cloned. I am leaning towards panicking.
The idea is that if the user uses a unicode path it's fine, as expected, but when it isn't unicode:
* `state_path.to_str()`: panics, throwing a hopefully helpful message, causing the user to rename some folder if it wants to continue using the block;
* `state_path.to_string_lossy()`: replaces sequences with an arbitrary unicode character, making Watson most certainly not work, and not generating errors at first. Error handling goes to Watson but the error was generated on i3status-rust (string encoding conversion error).

MaxVerevkin: > state_path.to_str(): panics, throwing a hopefully helpful message

Well, the message in this particular case will not be that helpful to the user, as `to_str()` returns an `Option`. 

> replaces sequences with an arbitrary unicode character

Fair point, but I would prefer `.block_error(...)?`ing in this case (in general panics should be avoided everywhere).

Henriquelay: I am experimenting with
```
let state_path_string = match block_config.state_path.to_str() {
    Some(str) => str,
    None => {
        return Err(BlockError(
            "watson".to_string(),
            format!(
                        "Error converting state_path to unicode string: '{}'. Can you try renaming one of the folders?",
                        block_config.state_path.to_string_lossy()
                    ),
        ))
    }
};
```
Or that isn't helpful enough to justify it? That we still can detect if the string is not properly decoded.

Henriquelay: Also I don't really have much of a clue on how to test this on my machine... I've never dealt with unusual encodings, at least not as far as I know.

MaxVerevkin: > Or that isn't helpful enough to justify it

You mean verbose error message? IDK. But this is definitely better than panicking.

MaxVerevkin: > I don't really have much of a clue on how to test this on my machine

Here you go: `touch $(printf 'non_utf_\xaa')` :laughing:

Henriquelay: This is probably a non-problem. Trying to load while pointing the non-UTF named file creates 
```
[
[{"full_text":" Internal error in context 'util': failed to read file. Cause: stream did not contain valid UTF-8 ","color":"#DC322FFF","background":"#000000FF","name":"0","instance":"0","separator":false,"separator_block_width":0,"markup":"pango"}],


Internal error in context 'util': failed to read file. Cause: stream did not contain valid UTF-8
```
So I suppose we can still use the match I made just for being safe?

Henriquelay: It creates an error before we get to block-level logic. 
```
[
thread '<unnamed>' panicked at 'failed to watch watson state file: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/blocks/watson.rs:121:18
[{"full_text":" Error in block 'watson': unable to open state file ","color":"#DC322FFF","background":"#000000FF","name":"0","instance":"0","separator":false,"separator_block_width":0,"markup":"pango"}],


Error in block 'watson': unable to open state file
```
I saw some of my i3 config there and tought maybe I screwed up, but no, it works fine when I delete just the non-utf char, watson block then says there is no such file (as expected). That's why I deleted last comment.

MaxVerevkin (approved):

MaxVerevkin:
Thanks!

PR #1455, opened 2022-04-14
Change battery 'device' configuration to be a regex and automatically rescan devices when not available

ferdinandschober:
This allows specifying a wildcard pattern for devices under
/sys/class/power_supply/

It is useful for e.g. wireless mice, that change their device
identifier upon unplugging.

E.g. my logitech mouse starts with
/sys/class/power_supply/hidpp_battery_0/
and when it's being unplugged it changes its name to
/sys/class/power_supply/hidpp_battery_1/
and then upon replugging it's now called
/sys/class/power_supply/hidpp_battery_2/
and so on.

This allows me to now just specify
hidpp_battery_* in my config file to use the first (and only)
device that matches this pattern.

It still requires a restart of i3status-rs uppon unplugging but I will
try to fix that in a later PR.

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -127,20 +127,18 @@ impl BatteryDevice for PowerSupplyDevice {
             "battery",
             format!("failed to read {} directory", sysfs_dir).as_str(),
         )?;
-        for entry in entries {
-            if let Ok(dir) = entry {
-                if read_to_string(dir.path().join("type"))
-                    .map(|t| t.trim() != "Battery")
-                    .unwrap_or(false)
-                {
-                    continue;
-                }
-                let dirname = dir.file_name();
-                let dirname = dirname.to_str().unwrap();
-                if self.device_regex.is_match(dirname) {
-                    self.device_path = Some(PathBuf::from(sysfs_dir).join(dirname));
-                    return Ok(());
-                }
+        for dir in entries.flatten() {
+            if read_to_string(dir.path().join("type"))
+                .map(|t| t.trim() != "Battery")
+                .unwrap_or(false)
+            {
+                continue;
+            }
+            let dirname = dir.file_name();
+            let dirname = dirname.to_str().unwrap();
+            if self.device_regex.is_match(dirname) {
+                self.device_path = Some(PathBuf::from(sysfs_dir).join(dirname));
+                return Ok(());
             }
         }
         Err(BlockError(
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -120,11 +120,20 @@ impl PowerSupplyDevice {
 
 impl BatteryDevice for PowerSupplyDevice {
     fn is_available(&self) -> bool {
-        let path = self.device_path.join("present");
-        if !path.exists() {
-            return false;
+        // in case of human interface devices (scope=="Device")
+        // we don't check for present==1 because the device subdirectory
+        // being present already implies that the device is available
+        let path = self.device_path.join("scope");
+        if path.exists() && read_file("battery", path).map_or(false, |x| x == "Device") {
+            // device is a human interface device
+            return true;
         }
-        read_file("battery", path).map_or(false, |x| x == "1")
+        // normal battery device -> check for present==1
+        let path = self.device_path.join("present");
+        if path.exists() {
+            return read_file("battery", path).map_or(false, |x| x == "1");
+        }
+        return false;
     }
 
     fn refresh_device_info(&mut self) -> Result<()> {
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -27,6 +27,7 @@ use crate::scheduler::Task;
 use crate::util::{battery_level_to_icon, read_file};
 use crate::widgets::text::TextWidget;
 use crate::widgets::{I3BarWidget, State};
+use glob::glob;
 
 /// A battery device can be queried for a few properties relevant to the user.
 pub trait BatteryDevice {
@@ -71,16 +72,58 @@ impl PowerSupplyDevice {
     /// `/sys/class/power_supply` directory. Raises an error if the directory for
     /// that device cannot be found and `allow_missing` is `false`.
     pub fn from_device(device: &str, allow_missing: bool) -> Result<Self> {
-        let device_path = Path::new("/sys/class/power_supply").join(device);
-
-        let device = PowerSupplyDevice {
-            device_path,
-            allow_missing,
-            charge_full: None,
-            energy_full: None,
-        };
-
-        Ok(device)
+        match Path::new("/sys/class/power_supply").join(device).to_str() {
+            Some(device_path) => {
+                match glob(device_path) {
+                    Ok(device_paths) => {
+                        for entry in device_paths {
+                            match entry {
+                                Ok(device_path) => {
+                                    let device = PowerSupplyDevice {
+                                        device_path,
+                                        allow_missing,
+                                        charge_full: None,
+                                        energy_full: None,
+                                    };
+                                    return Ok(device);
+                                }
+                                // we could technically also ignore GlobErrors
+                                // and continue with the next avaiale device
+                                // GlobErrors usually occur in case of I/O errors
+                                // (see https://docs.rs/glob/latest/glob/struct.Paths.html)
+                                Err(glob_error) => {
+                                    return Err(BlockError(
+                                        "battery".into(),
+                                        format!("{}", glob_error)
+                                    ));
+                                },
+                            }
+                        }
+                        Err(BlockError(
+                            "battery".into(),
+                            format!(
+                                "Pattern \"{}\" didn't match any devices!",
+                                device,
+                            )
+                        ))
+                    }
+                    Err(pattern_error) => Err(BlockError(
+                        "battery".into(),
+                        format!(
+                            "Invalid pattern: \"{}\": {}",
+                            device, pattern_error,
+                        )
+                    )),
+                }
+            }
+            None => Err(BlockError(
+                "battery".into(),
+                format!(
+                    "Invalid device path '{}'!",
+                    device
+                ),
+            )),
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


MaxVerevkin:
> It still requires a restart of i3status-rs uppon unplugging but I will
try to fix that in a later PR.

Let's fix this in this PR as it should be easy.

By the way, maybe using regex is a better option since it gives more control? Also `glob` crate is a bit overkill - it provides recursive directory search while we already have the code that iterates over devices' names.

ferdinandschober:
> > It still requires a restart of i3status-rs uppon unplugging but I will
> > try to fix that in a later PR.
> 
> Let's fix this in this PR as it should be easy.
> 
> By the way, maybe using regex is a better option since it gives more control? Also `glob` crate is a bit overkill - it provides recursive directory search while we already have the code that iterates over devices' names.

Seems reasonable. I could also implement this for upower batteries then.

MaxVerevkin:
> It is useful for e.g. wireless mice, that change their device
identifier upon unplugging.

This applies to UPower too, right?

ferdinandschober:
> > It is useful for e.g. wireless mice, that change their device
> identifier upon unplugging.
> 
> This applies to UPower too, right?

Yes

MaxVerevkin:
Actually, this is a bit more involved that that I though. Here is my attempt to implement it on `async` branch: https://github.com/greshake/i3status-rust/commit/56e42a1f3a618d57f60542dacdf21aeec01aedd8. When used with `sysfs` driver, the block will re-scan the batteries every time battery becomes unavailable. Upower, however, doesn't support `allow_missing` at all (on that branch).

ferdinandschober:
Do we really need a 'device_regex' option? As long as a path / upower device doesn't have any meta characters it should still be a valid regex, right?

MaxVerevkin:
`Regex::is_match` returns true if a string _contains_ any matches. We could hack this by prepending `^` and appending `$` to the provided device name, but... IDK

ferdinandschober:
> `Regex::is_match` returns true if a string _contains_ any matches. We could hack this by prepending `^` and appending `$` to the provided device name, but... IDK

```rust
use regex::Regex;

fn main() {
    let r = Regex::new("test").unwrap();
    println!("{}", r.is_match("test"));
}
```
prints true

It would also match asdftestasdf but I don't see that as a problem.

ferdinandschober:
Question: Is there any situation, where it would make sense for "allow_missing" to be false?

ferdinandschober:
> Question: Is there any situation, where it would make sense for "allow_missing" to be false?

Defaulting it to true / removing it completely would make configurations more portable imo  because you could just leave a battery block in your config for all devices without it crashing the bar on e.g. desktop pcs that don't have one.

GladOSkar:
I'm sure there was a reason for it once but i can't remember. I think it should definitely be an option, but the alternative shouldn't be crashing the bar but rather showing it as `critical` (red) with an `x`. Because for people with removable batteries it can definitely be of value to (explicitly) know when the battery is disconnected (i.e. to make sure not to unplug the AC adapter)

GladOSkar:
Also it's probably not wise to hide errors by default, since faulty batteries may sometimes just disappear instead of reporting their failure state, which is probably also something users want to be alerted about

MaxVerevkin:
> but the alternative shouldn't be crashing the bar but rather showing it as critical (red) with an x (like kdeconnect and others).

That's what happens with `allow_missing=true` & `hide_missing=false`, isn't it? I too don't see a reason to crash the whole bar because of a missing battery (`allow_missing=false`).

ferdinandschober:
> > but the alternative shouldn't be crashing the bar but rather showing it as critical (red) with an x (like kdeconnect and others).
> 
> That's what happens with `allow_missing=true` & `hide_missing=false`, isn't it? I too don't see a reason to crash the whole bar because of a missing battery (`allow_missing=false`).

Thats what I was thinking

ferdinandschober:
And the behaviour is basically exactly like you described @GladOSkar  with allow_missing = true and hide_missing = false

GladOSkar:
> That's what happens with allow_missing=true & hide_missing=false, isn't it? I too don't see a reason to crash the whole bar because of a missing battery (allow_missing=false).

Oh right. I forgor we had both

ferdinandschober:
I added the rescan_device() function to the BatteryDevice trait so far it's only implemented for PSU Devices

ferdinandschober:
The device rescan is now working as intended for me.
We still probably have a race condition (see #1459).

ferdinandschober:
Rescan apparently was already implemented for upower so that was pretty simple

ferdinandschober:
I will rebase this as soon as it's merged.

src/blocks/battery.rs
```diff
@@ -606,8 +684,8 @@ impl UpowerDevice {
 
     // Get device path. Raises exception if dbus communication fails.
     // If the device doesn't exist an exception raised unless allow_missing == true
-    fn get_device_path(device: &str, con: &dbus::ffidisp::Connection) -> Result<Option<String>> {
-        if device == "DisplayDevice" {
+    fn get_device_path(device: &Regex, con: &dbus::ffidisp::Connection) -> Result<Option<String>> {
+        if format!("{}", device) == "DisplayDevice" {
```
MaxVerevkin, line 688: `Regex` has `as_str()` method - no need for allocations here.

ferdinandschober:
Should I update the docs as well?

MaxVerevkin:
> Should I update the docs as well?

Yes

ferdinandschober:
One thing I changed for simplicity's sake is to _only_ match BAT.* or CMB.* devices when no device configuration is given.

This doesn't really change anything because other devices didn't work before afaik but I should probably change the documentation.

MaxVerevkin:
> I will rebase this as soon as it's merged.

Just merged.

> One thing I changed for simplicity's sake is to only match BAT.* or CMB.* devices when no device configuration is given.

Why not just _any_ battery?

---

Perhaps `BatteryDevice` API should be reduced to just 
```rust
trait BatteryDevice {
    fn get_info(&mut self) -> Result<Option<BatteryInfo>>;
}
```
?

ferdinandschober:
> > I will rebase this as soon as it's merged.
> 
> Just merged.
> 
> > One thing I changed for simplicity's sake is to only match BAT.* or CMB.* devices when no device configuration is given.
> 
> Why not just _any_ battery?
> 
If I matched any battery it could happen that hid devices get selected before the actual laptop battery which should probably be prefered. 
> ---
> 
> Perhaps `BatteryDevice` API should be reduced to just 
> ```rust
> trait BatteryDevice {
>     fn get_info(&mut self) -> Result<Option<BatteryInfo>>;
> }
> ```
> ?

I guess that would make sense. But what do we do with errors? Mostly they mean that a device is not available so we could even go as far as returning just an Option<BatteryInfo>?

MaxVerevkin:
> But what do we do with errors? Mostly they mean that a device is not available so we could even go as far as returning just an Option?

Hmm, I don't think silencing _all_ errors is the right decision. For example, what if dbus connections fails?

ferdinandschober:
> > But what do we do with errors? Mostly they mean that a device is not available so we could even go as far as returning just an Option?
> 
> Hmm, I don't think silencing _all_ errors is the right decision. For example, what if dbus connections fails?

Yeah you're right some errors should be reported.

So we add a BatteryInfo struct to the Battery struct that stores the current state (either None or Some{capacity, ...}) and then we update that struct everytime in the update function.

And view returns either nothing in case of BatteryInfo == None or vec![self.output] in case of Some(...)

I think that would simply a lot of things but I'd need to rewrite quite a lot of stuff.

src/blocks/battery.rs
```diff
@@ -70,274 +75,335 @@ impl PowerSupplyDevice {
...
+            if read_to_string(dir.path().join("type"))
+                .map(|t| t.trim() != "Battery")
+                .unwrap_or(false)
+            {
+                continue;
+            }
+            let dirname = dir.file_name();
+            let dirname = dirname.to_str().unwrap();
+            if self.device_regex.is_match(dirname) {
+                self.device_path = Some(PathBuf::from(sysfs_dir).join(dirname));
```
MaxVerevkin, line 140: Before assigning the path, it's availability should be checked. (I've implemented rescanning [like this](https://github.com/greshake/i3status-rust/blob/5f67435ef03167122385b39e6a822e096e214c4e/src/blocks/battery/sysfs.rs#L70-L96)).

ferdinandschober: Does someone with `device = "BAT"` in his config want to match any of his batteries or just available ones?

For the default config I think it definitely makes sense to match the first _available_ device though.

With your implementation BATx / CMBx devices currently don't take precedence over e.g. hid devices anymore (as you mentioned above).

Should we change that back?
In the docs it states
```
sysfs: the first battery device found in /sys/class/power_supply, with "BATx" or "CMBx" entries taking precedence.
apcaccess: "localhost:3551"
upower: `DisplayDevice`
```

I'm not sure what makes more sense really, but not checking availability would be the faster implementation. 
(except for the default configuration, where this would be a breaking change)

MaxVerevkin: > Does someone with device = "BAT" in his config want to match any of his batteries or just available ones?

I think only available. What's the point of matching unavailable batteries? If there are two batteries: `BAT0` and `BAT1`, and only `BAT1` is available, and I have `device = "BAT."`, then I don't see why `BAT0` should be matched.

Furthermore, how is a battery with `present=0` different from a non-existing battery? I mean, some drivers just don't show removed batteries in the sysfs while some drivers indicate that it's not present.

> With your implementation BATx / CMBx devices currently don't take precedence over e.g. hid devices anymore

That's a just matter of default `device` value and is easy to change.

MaxVerevkin:
Closing this since rescanning is already implemented for `sysfs` and implementing it for `upower` is a complementary different task.

ram02z:
> Closing this since rescanning is already implemented for `sysfs` and implementing it for `upower` is a complementary different task.

Does master have support for regex patterns for batteries?

MaxVerevkin:
> Does master have support for regex patterns for batteries?

Yes

Issue #1445, opened 2022-03-14
Error in block 'battery': failed to load file /sys/class/power_supply/apple_mfi_fastcharge/status


tarkh:
As topic says, after last update on `Arch Linux` I'm getting this error instead of i3status-rust panel:
```
Error in block 'battery': failed to load file /sys/class/power_supply/apple_mfi_fastcharge/status
```
My config block looks like this:
```
[[block]]
driver = "sysfs"
block = "battery"
allow_missing = true
hide_missing = true
interval = 30
format = "{percentage} {time}"
```
Also tried without `driver = "sysfs"` - same effect.
**This is desktop setup, so machine does not have any battery**, but for this reason my config contains `allow_missing = true` and `hide_missing = true`, which worked OK before.
If I set `driver = "upower"`, panel starts to work, but shows RED battery icon and 0%, which is not a valid behavior in this case.

MaxVerevkin:
@tarkh Can you please check if the last commit actually fixed this for you?

tarkh:
Yes, installed `i3status-rust-git ` package with latest upstream version and now error is gone.

PR #1456, opened 2022-04-14
Fix availability check for some battery devices

ferdinandschober:
Some battery devices like my wireless mouse for example have
an "online" sysfs file rather than the "present" file which
is being checked for.

This PR allows these devices to be used in the status bar as well.

### DIFF
--- a/doc/blocks.md
+++ b/doc/blocks.md
@@ -250,6 +250,7 @@ Key | Values | Required | Default
 `missing_format` | Same as `format` but for when the specified battery is missing. | No | `"{percentage}"`
 `allow_missing` | Don't display errors when the battery cannot be found. | No | `false`
 `hide_missing` | Completely hide this block if the battery cannot be found. Only works in combination with `allow_missing`. | No | `false`
+`ignore_present` | Some power supplies under /sys/class/power_supply/ don't have a "present" sysfs file. This option disables the check for this file to make battery devices without that file work (e.g. logitech wireless mice).
 `full_threshold` | Percentage at which the battery is considered full (`full_format` shown) | No | `100`
 `good` | Minimum battery level, where state is set to good. | No | `60`
 `info` | Minimum battery level, where state is set to info. | No | `60`
--- a/doc/blocks.md
+++ b/doc/blocks.md
@@ -250,7 +250,6 @@ Key | Values | Required | Default
 `missing_format` | Same as `format` but for when the specified battery is missing. | No | `"{percentage}"`
 `allow_missing` | Don't display errors when the battery cannot be found. | No | `false`
 `hide_missing` | Completely hide this block if the battery cannot be found. Only works in combination with `allow_missing`. | No | `false`
-`ignore_present` | Some power supplies under /sys/class/power_supply/ don't have a "present" sysfs file. This option disables the check for this file to make battery devices without that file work (e.g. logitech wireless mice).
 `full_threshold` | Percentage at which the battery is considered full (`full_format` shown) | No | `100`
 `good` | Minimum battery level, where state is set to good. | No | `60`
 `info` | Minimum battery level, where state is set to info. | No | `60`
--- a/man/i3status-rs.1
+++ b/man/i3status-rs.1
@@ -511,7 +511,7 @@ entries taking precedence.\f[V]apcaccess\f[R]:
 \[lq]localhost:3551\[rq]\f[V]upower\f[R]:
 \f[V]DisplayDevice\[ga]\[ga]\f[R]driver\f[V]| One of\f[R]\[lq]sysfs\[rq]\f[V],\f[R]\[lq]apcaccess\[rq]\f[V], or\f[R]\[lq]upower\[rq]\f[V]. | No |\f[R]\[lq]sysfs\[rq]\f[V]\f[R]interval\f[V]| Update interval, in seconds. Only relevant for\f[R]driver
 = \[lq]sysfs\[rq] ||
-\[lq]apcaccess\[rq]\f[V]. | No |\f[R]10\f[V]\f[R]format\f[V]| A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No |\f[R]\[lq]{percentage}\[rq]\f[V]\f[R]full_format\f[V]| Same as\f[R]format\f[V]but for when the battery is full. | No |\f[R]\[lq]\[lq]\f[V]\f[R]missing_format\f[V]| Same as\f[R]format\f[V]but for when the specified battery is missing. | No |\f[R]\[rq]{percentage}\[lq]\f[V]\f[R]allow_missing\f[V]| Don\[aq]t display errors when the battery cannot be found. | No |\f[R]false\f[V]\f[R]hide_missing\f[V]| Completely hide this block if the battery cannot be found. Only works in combination with\f[R]allow_missing\f[V]. | No |\f[R]false\f[V]\f[R]full_threshold\f[V]| Percentage at which the battery is considered full (\f[R]full_format\f[V]shown) | No |\f[R]100\f[V]\f[R]good\f[V]| Minimum battery level, where state is set to good. | No |\f[R]60\f[V]\f[R]info\f[V]| Minimum battery level, where state is set to info. | No |\f[R]60\f[V]\f[R]warning\f[V]| Minimum battery level, where state is set to warning. | No |\f[R]30\f[V]\f[R]critical\f[V]| Minimum battery level, where state is set to critical. | No |\f[R]15\[ga]
+\[lq]apcaccess\[rq]\f[V]. | No |\f[R]10\f[V]\f[R]format\f[V]| A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No |\f[R]\[lq]{percentage}\[rq]\f[V]\f[R]full_format\f[V]| Same as\f[R]format\f[V]but for when the battery is full. | No |\f[R]\[lq]\[lq]\f[V]\f[R]missing_format\f[V]| Same as\f[R]format\f[V]but for when the specified battery is missing. | No |\f[R]\[rq]{percentage}\[lq]\f[V]\f[R]allow_missing\f[V]| Don\[aq]t display errors when the battery cannot be found. | No |\f[R]false\f[V]\f[R]hide_missing\f[V]| Completely hide this block if the battery cannot be found. Only works in combination with\f[R]allow_missing\f[V]. | No |\f[R]false\f[V]\f[R]ignore_present\f[V]| Some power supplies under /sys/class/power_supply/ don\[aq]t have a \[dq]present\[dq] sysfs file. This option disables the check for this file to make battery devices without that file work (e.g. logitech wireless mice).\f[R]full_threshold\f[V]| Percentage at which the battery is considered full (\f[R]full_format\f[V]shown) | No |\f[R]100\f[V]\f[R]good\f[V]| Minimum battery level, where state is set to good. | No |\f[R]60\f[V]\f[R]info\f[V]| Minimum battery level, where state is set to info. | No |\f[R]60\f[V]\f[R]warning\f[V]| Minimum battery level, where state is set to warning. | No |\f[R]30\f[V]\f[R]critical\f[V]| Minimum battery level, where state is set to critical. | No |\f[R]15\[ga]
 T}
 .TE
 .SS Available Format Keys
@@ -1233,6 +1233,7 @@ T{
 \f[V]path\f[R]
 T}@T{
 Path to collect information from.
+Supports path expansions e.g.\ \f[V]\[ti]\f[R].
 T}@T{
 No
 T}@T{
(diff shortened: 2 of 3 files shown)


src/blocks/battery.rs
```diff
@@ -87,10 +87,14 @@ impl PowerSupplyDevice {
 impl BatteryDevice for PowerSupplyDevice {
     fn is_available(&self) -> bool {
         let path = self.device_path.join("present");
-        if !path.exists() {
-            return false;
+        if path.exists() {
```
ammgws, line 90: Can you refactor both checks into one block since the only different part is the path it's checking ("present" vs "online")?

MaxVerevkin: Something like

```rust
for file in ["present", "online"] {
    // ...
}
```

ferdinandschober: I double checked this and apparently online is not quite the same thing:
https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-power

I guess it would be better to just check if the "present" file exists and assume the device is available if it doesn't?

MaxVerevkin:
@ferdinandschober do you mean `present` not set to `1` or you don't have this file at all?

ferdinandschober:
l /sys/class/power_supply/hidpp_battery_3    
Permissions Size User Date Modified Name
.r--r--r--  4.1k root 14 Apr 14:44   capacity
lrwxrwxrwx     0 root 14 Apr 13:41   device -> ../../../0003:046D:4079.0018
drwxr-xr-x     - root 14 Apr 14:44   hwmon11
.r--r--r--  4.1k root 14 Apr 14:44   manufacturer
.r--r--r--  4.1k root 14 Apr 14:44   model_name
.r--r--r--  4.1k root 14 Apr 14:44   online
drwxr-xr-x     - root 14 Apr 14:44   power
lrwxrwxrwx     0 root 14 Apr 14:44   powers -> ../../../0003:046D:4079.0018
.r--r--r--  4.1k root 14 Apr 14:44   scope
.r--r--r--  4.1k root 14 Apr 14:44   serial_number
.r--r--r--  4.1k root 14 Apr 14:44   status
lrwxrwxrwx     0 root 14 Apr 13:41   subsystem -> ../../../../../../../../../../class/power_supply
.r--r--r--  4.1k root 14 Apr 14:44   type
.rw-r--r--  4.1k root 14 Apr 14:44   uevent
.r--r--r--  4.1k root 14 Apr 14:44   voltage_now
drwxr-xr-x     - root 14 Apr 14:44   wakeup47

@MaxVerevkin  it's not there at all. Might also just be a bug in the kernel with logitech devices.
Maybe we should just assume the device is present if the file doesn't exist?

MaxVerevkin:
> Might also just be a bug in the kernel with logitech devices

I haven't actually read the docs, but it this file is supposed to be a reliable source, then it's worth reporting.

> Maybe we should just assume the device is present if the file doesn't exist?

No :( (https://github.com/greshake/i3status-rust/issues/1445)

ferdinandschober:
ll /sys/class/power_supply/ucsi-source-psy-0-00081/      
Permissions Size User Date Modified Name
.r--r--r--  4.1k root 14 Apr 14:40   current_max
.r--r--r--  4.1k root 14 Apr 14:40   current_now
lrwxrwxrwx     0 root 14 Apr 13:03   device -> ../../../0-0008
drwxr-xr-x     - root 14 Apr 14:40   hwmon5
.r--r--r--  4.1k root 14 Apr 14:40   online
drwxr-xr-x     - root 14 Apr 14:40   power
lrwxrwxrwx     0 root 14 Apr 13:03   subsystem -> ../../../../../../../../../../class/power_supply
.r--r--r--  4.1k root 14 Apr 14:40   type
.rw-r--r--  4.1k root 14 Apr 14:40   uevent
.r--r--r--  4.1k root 14 Apr 14:40   usb_type
.r--r--r--  4.1k root 14 Apr 14:40   voltage_max
.r--r--r--  4.1k root 14 Apr 14:40   voltage_min
.r--r--r--  4.1k root 14 Apr 14:40   voltage_now
drwxr-xr-x     - root 14 Apr 14:40   wakeup44

Yeah I realized this might not be the optimal solution as well ... (this is my powersupply).

MaxVerevkin:
By the way, my battery doesn't have `online` property

ferdinandschober:
https://github.com/Alexays/Waybar/issues/972
waybar seems to have similar issues

ferdinandschober:
Any ideas for a workaround? Maybe a "ignore_present" configuration option?

MaxVerevkin:
> Maybe a "ignore_present" configuration option?

As a temporary workaround. Ideally, this should be fixed in the kernel. Btw, which kernel do you use?

ferdinandschober:
I'm currently on 5.17.2-arch3-1

MaxVerevkin:
Can you please try LTS version?

ferdinandschober:
```
[ferdinand@iridium hidpp_battery_0]$ uname -r
5.15.33-1-lts
[ferdinand@iridium hidpp_battery_0]$ pwd
/sys/class/power_supply/hidpp_battery_0
[ferdinand@iridium hidpp_battery_0]$ ls -lah
total 0
drwxr-xr-x 5 root root    0 Apr 14 15:24 .
drwxr-xr-x 3 root root    0 Apr 14 15:24 ..
-r--r--r-- 1 root root 4.0K Apr 14 15:25 capacity
lrwxrwxrwx 1 root root    0 Apr 14 15:25 device -> ../../../0003:046D:4079.000C
drwxr-xr-x 3 root root    0 Apr 14 15:25 hwmon11
-r--r--r-- 1 root root 4.0K Apr 14 15:25 manufacturer
-r--r--r-- 1 root root 4.0K Apr 14 15:25 model_name
-r--r--r-- 1 root root 4.0K Apr 14 15:25 online
drwxr-xr-x 2 root root    0 Apr 14 15:25 power
lrwxrwxrwx 1 root root    0 Apr 14 15:25 powers -> ../../../0003:046D:4079.000C
-r--r--r-- 1 root root 4.0K Apr 14 15:25 scope
-r--r--r-- 1 root root 4.0K Apr 14 15:25 serial_number
-r--r--r-- 1 root root 4.0K Apr 14 15:25 status
lrwxrwxrwx 1 root root    0 Apr 14 15:24 subsystem -> ../../../../../../../../../../class/power_supply
-r--r--r-- 1 root root 4.0K Apr 14 15:25 type
-rw-r--r-- 1 root root 4.0K Apr 14 15:25 uevent
-r--r--r-- 1 root root 4.0K Apr 14 15:25 voltage_now
drwxr-xr-x 2 root root    0 Apr 14 15:25 wakeup47
[ferdinand@iridium hidpp_battery_0]$ 
```
same thing on LTS

man/i3status-rs.1
```diff
@@ -511,7 +511,7 @@ entries taking precedence.\f[V]apcaccess\f[R]:
...
...s\f[R]format\f[V]but for when the battery is full. | No |\f[R]\[lq]\[lq]\f[V]\f[R]missing_format\f[V]| Same as\f[R]format\f[V]but for when the specified battery is missing. | No |\f[R]\[rq]{percentage}\[lq]\f[V]\f[R]allow_missing\f[V]| Don\[aq]t display errors when the battery cannot be found. | No |\f[R]false\f[V]\f[R]hide_missing\f[V]| Completely hide this block if the battery cannot be found. Only works in combination with\f[R]allow_missing\f[V]. | No |\f[R]false\f[V]\f[R]ignore_present\f[V]| Some power supplies under /sys/class/power_supply/ don\[aq]t have a \[dq]present\[dq] sysfs file. This option disables the check for this file to make battery devices without that file work (e.g. logitech wireless mice).\f[R]full_threshold\f[V]| Percentage at which the battery is considered full (\f[R]full_format\f[V]shown) | No |\f[R]100\f[V]\f[R]good\f[V]| Minimum battery level, where state is set to good. | No |\f[R]60\f[V]\f[R]info\f[V]| Minimum battery level, where state is set to info. | No |\f[R]60\f[V]\f[R]warning\f[V]| Minimum battery level, where state is set to warning. | No |\f[R]30\f[V]\f[R]critical\f[V]| Minimum battery level, where state is set to critical. | No |\f[R]15\[ga]
```
MaxVerevkin, line 514: No deed to generate man pages now, it will be done as part of the release procedure

MaxVerevkin:
@ferdinandschober What's the output of `readlink /sys/class/power_supply/hidpp_battery_3/device/driver`?

ferdinandschober:
> @ferdinandschober What's the output of `readlink /sys/class/power_supply/hidpp_battery_3/device/driver`?

../../../../../../../../bus/hid/drivers/logitech-hidpp-device

ferdinandschober:
AC adapters also generally don't seem to have a present file so I think we should not assume that this file exists.

I just read some more documentation and what I found is, that batteries, that are hid devices have a file named scope that reads "Device" in their power_supply subdirectory.
The reason behind this is that batteries with their scope as "Device" are not actually supplying power to the laptop / pc they are attached to.

So what could be done is check if "scope" exists, in case it reads "Device" we don't check for a present file and report the battery as normal, and in case the scope file doesn't exists we proceed as before.

ferdinandschober:
https://lore.kernel.org/lkml/alpine.LNX.2.00.1112191124190.8727@pobox.suse.cz/T/

ferdinandschober:
"[1/2] HID: logitech-hidpp: add battery support for HID++ 2.0 devices - Patchwork" https://patchwork.kernel.org/project/linux-input/patch/1467988545.2317.4.camel@hadess.net/

doc/blocks.md
```diff
@@ -250,6 +250,7 @@ Key | Values | Required | Default
 `missing_format` | Same as `format` but for when the specified battery is missing. | No | `"{percentage}"`
 `allow_missing` | Don't display errors when the battery cannot be found. | No | `false`
 `hide_missing` | Completely hide this block if the battery cannot be found. Only works in combination with `allow_missing`. | No | `false`
+`ignore_present` | Some power supplies under /sys/class/power_supply/ don't have a "present" sysfs file. This option disables the check for this file to make battery devices without that file work (e.g. logitech wireless mice).
```
MaxVerevkin, line 253:
```suggestion
`ignore_present` | Some power supplies under /sys/class/power_supply/ don't have a "present" sysfs file. This option disables the check for this file to make battery devices without that file work (e.g. logitech wireless mice). | No | `false`
```

MaxVerevkin:
> So what could be done is check if "scope" exists, in case it reads "Device" we don't check for a present file and report the battery as normal, and in case the scope file doesn't exists we proceed as before.

Seems reasonable :+1:

src/blocks/battery.rs
```diff
@@ -86,11 +89,27 @@ impl PowerSupplyDevice {
 
 impl BatteryDevice for PowerSupplyDevice {
     fn is_available(&self) -> bool {
+        // the problem with hid devices is, that in case they are 
+        // not plugged in the whole power_supply subdirectory doesn't exist
+        if self.device.starts_with("hid")
+        && !self.device_path.exists() {
```
MaxVerevkin, line 95: I think this check is redundant?

ferdinandschober:
I've tested this now with two differnt logitech mice and it seems to be working as intended.

src/blocks/battery.rs
ferdinandschober: device is just the device string from the config file and we don't want to say "not available" for devices that are normal batteries. Otherwise those would be assumed to be not available when they are actually configured wrong

MaxVerevkin: I'm not sure I follow. If the base directory does not exist, it means the battery is not available, doesn't matter whether it's hid or not. And actually, that was the only indicator before https://github.com/greshake/i3status-rust/commit/24f432fb67e5ba3cadddf5084b60c15e392f5e44.

ferdinandschober: Well but normal Batteries like BAT0 never change their sysfs path. So it would be better to show the user that they entered the wrong path instead of saying "device is missing". I guess either version makes sense but my reasoning was that in this case throwing an error is the right thing because a BATASDF that doesn't exists is not _really_ missing rather than misconfigured.

ferdinandschober: With HID devices it's a bit different because they remove the whole device path when they are unplugged (which is kind of the reason for PR #1455)

MaxVerevkin: > Well but normal Batteries like BAT0 never change their sysfs path

They can also be plugged/unplugged (I think so, I don't have the necessary hardware to test myself), so it's not true. And I don't see how this check affects anything - if `device_path` doesn't exist, the function will return `false` in _any_ case.

ferdinandschober: You're right that makes sense.

src/blocks/battery.rs
```diff
@@ -60,6 +60,7 @@ pub trait BatteryDevice {
 
 /// Represents a physical power supply device, as known to sysfs.
 pub struct PowerSupplyDevice {
+    device: String,
```
MaxVerevkin, line 63: Now this is unnecessary

ferdinandschober: yeah I just realized this

MaxVerevkin:
Thanks! A port to `async` branch would be appreciated.

ferdinandschober:
> Thanks! A port to `async` branch would be appreciated.

I will look into it.

ammgws:
> > Thanks! A port to `async` branch would be appreciated.
> 
> I will look into it.

Did this ever happen?

MaxVerevkin:
> > > Thanks! A port to `async` branch would be appreciated.
> > 
> > I will look into it.
> 
> Did this ever happen?

I implemented it.

PR #1457, opened 2022-04-15
Net graph formatting

veprolet:
The documentation gives the following example in for the `net` block:

> Displays ssid, signal strength, ip, down speed and up speed as bits per second. Minimal prefix is set to `K` in order to prevent the block to change its size.
> 
> ```toml
> [[block]]
> block = "net"
> device = "wlp2s0"
> format = "{ssid} {signal_strength} {ip} {speed_down;K*b} {graph_down;K*b}"
> interval = 5
> ```

After seeing the `{graph_down;K*b}` part, I was quite confused, because I've assumed that the format specifier would be interpreted similarly to text based blocks (since the SI prefix part was being used).

This PR implements interpretation of the format specifier for the `net` block. This is useful if there is a desire to change the width of the graph (the previously hard-coded 10 characters are kept as a default value), or for the values to be displayed on an unchanging absolute scale. Consider for example the following placeholder:
`{graph_down:8^12;_M*_b#50}`
`min_width` is 8, so the graph will be represented by 8 characters.
`max_width` is ignored by the code in this PR.
`min_prefix` is mega- and is hidden. This only affects how the `bar_max_val` value is interpreted.
`unit` is bits. This only affects how the `bar_max_val` value is interpreted.
`bar_max_val` is 50 Mb/s. The unit of this value is determined from the `min_prefix` and `unit` fields.
Thus, download at the rate of 25 Mb/s would result in half a bar being drawn even if all the previous values were 0.

If there's any interest in merging this PR, I'm not sure how the documentation needs to be updated, because this is exactly how I've assumed it would work, when I've read it in the current state.

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -421,10 +421,10 @@ impl ConfigBlock for Net {
             .with_default("{speed_down;K}{speed_up;K}")?;
         let format_alt = block_config.format_alt;
 
-        let tx_format = format.get("graph_up");
-        let rx_format = format.get("graph_down");
-        let tx_graph_width = Self::max_min_width(tx_format).unwrap_or(Self::DEFAULT_GRAPH_WIDTH);
-        let rx_graph_width = Self::max_min_width(rx_format).unwrap_or(Self::DEFAULT_GRAPH_WIDTH);
+        let tx_graph_width = Self::max_format_width(&format, &format_alt, "graph_up")
+            .unwrap_or(Self::DEFAULT_GRAPH_WIDTH);
+        let rx_graph_width = Self::max_format_width(&format, &format_alt, "graph_down")
+            .unwrap_or(Self::DEFAULT_GRAPH_WIDTH);
 
         Ok(Net {
             id,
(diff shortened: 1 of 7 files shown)


MaxVerevkin:
> The documentation gives the following example in for the net block

Copy-paste error :)

src/blocks/net.rs
```diff
@@ -423,6 +425,35 @@ impl ConfigBlock for Net {
             .with_default("{speed_down;K}{speed_up;K}")?;
         let format_alt = block_config.format_alt;
 
+        let tx_format = format.get("graph_up");
+        let rx_format = format.get("graph_down");
+        // prefers the full option
```
MaxVerevkin, line 430: That's not ideal. Right now I don't have a recommendation on how to properly implement this, but short config should not be ignored.

src/blocks/net.rs
```diff
@@ -543,12 +545,25 @@ impl Net {
...
 
         Ok(())
     }
+
+    fn max_min_width(format: (Option<&Placeholder>, Option<&Placeholder>)) -> Option<usize> {
+        let format = match format {
+            (Some(full), Some(short)) => vec![full, short],
+            (Some(full), None) => vec![full],
+            (None, Some(short)) => vec![short],
+            (None, None) => vec![],
```
MaxVerevkin, line 559: I haven't checked if `rustc` optimizes those `vec`s out, but maybe something like this would be better?

```rust
    fn max_min_width(format: (Option<&Placeholder>, Option<&Placeholder>)) -> Option<usize> {
        match format {
            (Some(full), Some(short)) => {
                let (a, b) = (full.min_width.value, short.min_width.value);
                Some(a.or(b)?.max(b.or(a)?))
            }
            (Some(f), None) | (None, Some(f)) => f.min_width.value,
            (None, None) => None,
        }
    }
```

src/formatting/value.rs
```diff
@@ -261,6 +320,10 @@ impl Value {
                 true => String::from("T"),
                 false => String::from("F"),
             },
+            InternalValue::Graph(value) => {
+                let max_val = graph_max_from_placeholder(var)?;
+                format_bar_graph(value, max_val.and(Some(0_f64)), max_val)
```
MaxVerevkin, line 325: Hmm, looks like `min_width`/`max_width` is ignored here. I think it should be
```rust
format_bar_graph(value.iter().skip(value.len().saturating_sub(width)), max_val.and(Some(0_f64)), max_val)
```

src/blocks/net.rs
```diff
@@ -423,6 +421,11 @@ impl ConfigBlock for Net {
             .with_default("{speed_down;K}{speed_up;K}")?;
         let format_alt = block_config.format_alt;
 
+        let tx_format = format.get("graph_up");
+        let rx_format = format.get("graph_down");
+        let tx_graph_width = Self::max_min_width(tx_format).unwrap_or(Self::DEFAULT_GRAPH_WIDTH);
+        let rx_graph_width = Self::max_min_width(rx_format).unwrap_or(Self::DEFAULT_GRAPH_WIDTH);
```
MaxVerevkin, line 427: This doesn't take into account `format_alt` config.

MaxVerevkin (approved):
LGTM :+1:

Just need to mention this in the docs (e.g. that "min_width" sets the width of the graph and that "min_prefix" affects how "max_value" is treated).

veprolet:
Great! Thanks for the guidance.

I'll look into updating the docs, but just to be sure: Is the proper place to document this the `net` section of `/doc/blocks.md`? Does any other documentation need updating?

MaxVerevkin:
> Is the proper place to document this the net section of /doc/blocks.md?

Well, that's the only block that uses this feature for now, so yes, I think so.

MaxVerevkin (approved):

PR #1458, opened 2022-04-16
Fallback to `charge_level` if `capacity` cannot be calculated

ferdinandschober:
Some devices report a charge level rather than an accurate percentage.
This let's you specify the {charge_level} in the format string for
those devices.

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -50,6 +50,9 @@ pub trait BatteryDevice {
     /// Query the device's current capacity, as a percent.
     fn capacity(&self) -> Result<u64>;
 
+    /// Query the device's current capacity level (for devices that don't report percentages)
+    fn capacity_level(&self) -> Result<String>;
+
     /// Query the estimated time remaining, in minutes, before (dis)charging is
     /// complete.
     fn time_remaining(&self) -> Result<u64>;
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
Perhaps it's better to just fallback to these approximations instead of introducing another placeholder?

ferdinandschober:
> Perhaps it's better to just fallback to these approximations instead of introducing another placeholder?

Good idea I thought of that too
It's working for me with ed45be99d9462fe37c8547ea12f57185c8169ead

src/blocks/battery.rs
```diff
@@ -183,6 +186,18 @@ impl BatteryDevice for PowerSupplyDevice {
         }
     }
 
+    fn capacity_level(&self) -> Result<String> {
```
MaxVerevkin, line 189: Let's move this and the logic from lines 944-961 to the `capacity` function to avoid unnecessary impls for other backends.

ferdinandschober: Would you prefer displaying the Strings from capacity_level or rather display them converted to some percentages? (5/25/50/75/100)?

MaxVerevkin: > display them converted to some percentages?

I think that's what Android does, and this make sense for `percentage` placeholder

ferdinandschober: Okay makes it easier to implement as  well.

MaxVerevkin:
Thanks!

PR #1459, opened 2022-04-17
Fix a race condition with kernel driver

ferdinandschober:
Since #1456 it is possible for status() and refresh_device_info() to
fail in case the device is not available.

If the device gets removed just after checking if it is available but
before checking for it's status this could crash i3status-rs.

So in order to prevent this I swapped the order so that we first check for the
device status and refresh the device info and then check if the device is available.

Now it's guaranteed that the device was available when we checked status() or
called refresh_device_info() before throwing an error.

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -887,6 +887,9 @@ impl Block for Battery {
     fn update(&mut self) -> Result<Option<Update>> {
         // TODO: Maybe use dbus to immediately signal when the battery state changes.
 
+        let refresh = self.device.refresh_device_info();
+        let status = self.device.status();
+
         // Exit early, if the battery device went missing, but the user
         // allows this device to go missing.
         if !self.device.is_available() && self.allow_missing {
@@ -902,11 +905,12 @@ impl Block for Battery {
             self.output.set_texts(self.missing_format.render(&values)?);
             self.output.set_state(State::Warning);
         } else {
+            if let Err(e) = refresh {
+                return Err(e);
+            }
+            let status = status?;
             // The device may have gone missing
             // It may be a different battery now, thereby refresh the device specs.
-            self.device.refresh_device_info()?;
-
-            let status = self.device.status()?;
             let capacity = self.device.capacity();
             let values = map!(
                 "percentage" => match capacity {
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -887,10 +887,22 @@ impl Block for Battery {
     fn update(&mut self) -> Result<Option<Update>> {
         // TODO: Maybe use dbus to immediately signal when the battery state changes.
 
-        // The device may have gone missing
-        // It may be a different battery now, thereby refresh the device specs.
-        let refresh = self.device.refresh_device_info();
-        let status = self.device.status();
+        let mut available_before = self.device.is_available();
+        let mut refresh;
+        let mut status;
+        // loop until available status before and after is the same
+        // to prevent race conditions between is_available() and status()
+        loop {
+            // The device may have gone missing
+            // It may be a different battery now, thereby refresh the device specs.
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
> Now it's guaranteed that the device was available when we checked status() or
called refresh_device_info() before throwing an error.

Hmm, but what if battery becomes available just before `is_available()`? Should the availability be checked twice (before and after, in case of an error)?

ferdinandschober:
> > Now it's guaranteed that the device was available when we checked status() or
> > called refresh_device_info() before throwing an error.
> 
> Hmm, but what if battery becomes available just before `is_available()`? Should the availability be checked twice (before and after, in case of an error)?

I guess that doesn't matter we will just have to wait 10 seconds for the next update?

ferdinandschober:
It only makes sense with allow_missing anyway.

ferdinandschober:
Oh wait yeah now that I think about it that might be an issue.

MaxVerevkin:
> > > Now it's guaranteed that the device was available when we checked status() or
> > > called refresh_device_info() before throwing an error.
> > 
> > 
> > Hmm, but what if battery becomes available just before `is_available()`? Should the availability be checked twice (before and after, in case of an error)?
> 
> I guess that doesn't matter we will just have to wait 10 seconds for the next update?

No, I mean, what if status was fetched while battery was unavailable, error was encountered, but then `is_available()` returns `true` and we (incorrectly) throw an error?

MaxVerevkin:
> So in order to prevent this

Did this happen to you?

ferdinandschober:
Yeah I didn't quite think that through.

> > So in order to prevent this
> 
> Did this happen to you?

It didn't happen so far and I'd imagine it would be _very_ rare.
Imo it should still be fixed because it's going to happen at some point to someone.

Maybe even a loop while (available_before != available_after) would be a simple solution?

ferdinandschober:
It could still possibly fail if the device is unplugged after the first available check then status is read and then the device is plugged back in before the second is_available() check but I'd Imagine that that is pretty much impossible to pull of.

MaxVerevkin:
If someone can plug/unplug batteries so fast, we cannot rely on anything :laughing:

MaxVerevkin:
@bim9262 what do you think? How does this relate to #1393?

ferdinandschober:
> @bim9262 what do you think? How does this relate to #1393?

https://github.com/greshake/i3status-rust/pull/1393#issuecomment-1011569349
If this is actually whats happening than this PR should fix it.

bim9262:
I think that is_avalible can actually be generalized for all of the battery drivers and can just check if we have a device path (is_some).

refresh_device_info should attempt to load the data from the data source. If while reading in the data from that source, it turns out that is available then you can raise an Error (see ApcUpsDevice's  refresh_device_info as an example), otherwise save all of the values that you got into temporary values that are stored for that driver. Then the driver's functions like capacity and time_remaining can basically just read/convert those saved values to the appropriate value.

ferdinandschober:
> I think that is_avalible can actually be generalized for all of the battery drivers and can just check if we have a device path (is_some).
> 
> refresh_device_info should attempt to load the data from the data source. If while reading in the data from that source, it turns out that is available then you can raise an Error (see ApcUpsDevice's  refresh_device_info as an example), otherwise save all of the values that you got into temporary values that are stored for that driver. Then the driver's functions like capacity and time_remaining can basically just read/convert those saved values to the appropriate value.
> 

I dont know if we should change the is_available function because some battery devices keep their device path when they get removed and only change the "present" file if they get removed. But moving the value parsing to the rescan function would actually be a good Idea and simplify the Error handling

MaxVerevkin:
Already implemented.

MaxVerevkin:
And thanks for pointing out this issue!

PR #1460, opened 2022-04-20
Fix argument in README.md

bznein:
The correct argument is `never-pause` https://github.com/greshake/i3status-rust/blob/0fa0d0a4ca0992f947ffe567c8b1425349b1aa3b/src/main.rs#L71-L74 not `never-stop`


MaxVerevkin:
:+1:

PR #1461, opened 2022-04-21
battery: remove allow_missing

MaxVerevkin:
As discussed in #1455, setting this option to `false` doesn't have any real use-cases. In addition, removing this option simplifies the code slightly. This is a breaking change.

### DIFF
--- a/doc/blocks.md
+++ b/doc/blocks.md
@@ -248,8 +248,7 @@ Key | Values | Required | Default
 `format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | No | `"{percentage}"`
 `full_format` | Same as `format` but for when the battery is full. | No | `""`
 `missing_format` | Same as `format` but for when the specified battery is missing. | No | `"{percentage}"`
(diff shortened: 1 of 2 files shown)


ammgws:
Sorry I didn't get around to this in time

AkechiShiro:
Hey, I just noticed this breaking change, I think the documentation of the battery should be updated along with this PR, I only rebuilt i3status-rust and was greeted with an error, looked into the documentation and got even more confused, then narrowed it down to this commit by looking at the commit history.

GladOSkar:
Alternatively it might be wise to just ignore the option for the foreseeable future

MaxVerevkin:
@AkechiShiro I've added a notice in `NEWS.md`. By the way, if you have any thoughts/ideas on how to make breaking changes less painful, please share them in #1462.

MaxVerevkin:
> Alternatively it might be wise to just ignore the option for the foreseeable future

I think it would be nice if there was a way to display a warning that a deprecated option is used. Would using `notify-send` be justifiable in this case?

GladOSkar:
I think yes, that's not a bad idea

MaxVerevkin:
Something like fb292c0e6563edb73f691ed1bbd76784befc5e6e

PR #1468, opened 2022-05-01
Add wttr.in text block

JayGray:
#### Why
I wanted a dead simple to use weather block to know if I need to wear a hat when going outside.
Although the weather block can provide me with the necessary information, I did not want to go through
the setup of an API key and formatting steps.
[wttr.in](https://github.com/chubin/wttr.in) does provide all this out of the box.

#### What
Add a simple text widget with as few configuration options as possible actually, none is required.
All display configuration can be done via the format query parameter of wttr.in.

#### TODO
- [ ] error handling
- [ ] cap minimum update interval to 900sec (IMHO weather data doesn't need to be updated every other second)

I just realized I should have run this through a feature request issue before, sry for that.

Samples:

image

image

### DIFF
--- a/doc/blocks.md
+++ b/doc/blocks.md
@@ -56,6 +56,7 @@ You may find that the block you desire is not in the list below. In that case, f
 - [Uptime](#uptime)
 - [Watson](#watson)
 - [Weather](#weather)
+- [Wttr](#wttr)
 - [Xrandr](#xrandr)
 
 ## Apt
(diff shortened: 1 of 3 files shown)


MaxVerevkin (changes requested):
Thanks for contributing! I've left a few comments regarding the code.

@ammgws what's your opinion on this (like, should this be a `custom` block instead?)?

ammgws:
> @ammgws what's your opinion on this (like, should this be a `custom` block instead?)?

Could the weather block be refactored in a way that `wttr.in` could be used as an alternative driver to the existing one? Otherwise not really a fan of having two blocks that do weather...

JayGray:
@MaxVerevkin appreciate your time to have a look at this. I'm not very fluent in RUST so I welcome any hint to keep the code clean and within specs.

As for the proposal to use `wttr.in` as a drop-in alternative.
I had the chance to test and observe for about a week now and it is somewhat holding up to my expectations.
Error handling and sensitive help messages are crucial, either because of human errors, eg. city name can't be resolved because of a typo, or quota exceeded within the service.

MaxVerevkin:
Added an example of how to replicate this block using `custom`: https://github.com/greshake/i3status-rust/tree/master/examples#wttrin

Issue #1371, opened 2021-12-17
Failable blocks / Improve error handling


MaxVerevkin:
The use of `refresh_interval_failure` in #1366 made me wonder if it can be applied to other blocks in a generic way.

Here is a rough idea: when an error in a block occurs, the block's state will be set to `error_state` and the text will be set according to `error_format`. Then the block will be restarted after `error_interval`. Clicking on the block will show the full error message. Clicking again will hide the full error message.

There is a major issue though: some blocks spawn threads. The situation is better for async implementation, as those threads are at least abortable (edit: `spawn_blocking()` is not really abortable).

MaxVerevkin:
Actually, it might be easier than I thought. There's no need to fully restart the blocks. Instead, each block can be composed of failable but recoverable parts.

As an example, here's a part of `github`'s block from my prototype:
```rust
let total = api.recoverable(|| get_total(&request), "Oh no!").await?;
```

That's it. Nothing else needed on the block's side. That's how it looks:

https://user-images.githubusercontent.com/34583604/146654531-c1364a64-0592-439c-bb10-f2ffb278d313.mp4

ammgws:
Closing this as it's implemented in master now.

PR #1471, opened 2022-05-05
Add block for libvirtd/qemu

dijit:
*Why*

I find it sometimes useful on my desktop to understand how many VMs I have running or that are paused/stopped at a glance.

*What*

a simple libvirt integration to count the number of running instances; optionally you can use format to display.

* "paused" (suspended)
* "stopped" (destroyed) domains 
* **and** a count of the images in the local disk store (such as virtual harddisks and ISOs, will count all activated storage).

Due to the crate [virt](https://crates.io/crates/virt) using a lot of unsafe, I figure it's best to keep it as an optional include.

I chose the icon 'copy' for the bar, as that closely resembles vmware's iconography.

### DIFF
--- a/src/blocks/libvirt.rs
+++ b/src/blocks/libvirt.rs
@@ -25,29 +25,6 @@ pub struct Libvirt {
 }
 
 #[derive(Deserialize, Debug, Clone)]
-struct Status {
-    #[allow(dead_code)]
-    #[serde(rename = "VMs")]
-    total: i64,
-
-    #[allow(dead_code)]
-    #[serde(rename = "VMsRunning")]
-    running: i64,
-
-    #[allow(dead_code)]
-    #[serde(rename = "VMsStopped")]
-    stopped: i64,
-
-    #[allow(dead_code)]
-    #[serde(rename = "VMsPaused")]
-    paused: i64,
-
-    #[allow(dead_code)]
-    #[serde(rename = "Images")]
-    images: i64,
-}
-
-#[derive(Deserialize, Debug, Clone)]
 #[serde(deny_unknown_fields, default)]
 pub struct LibvirtConfig {
     /// Update interval in seconds
--- a/src/blocks/libvirt.rs
+++ b/src/blocks/libvirt.rs
@@ -26,7 +26,7 @@ pub struct Libvirt {
 }
 
 impl Libvirt {
-    fn check_and_reconnect(&mut self) -> Result<&mut Self> {
+    fn check_and_reconnect(&mut self) -> Result<()> {
         if !self
             .qemu_conn
             .is_alive()
@@ -35,7 +35,7 @@ impl Libvirt {
             self.qemu_conn = Connect::open_read_only(&self.qemu_uri)
                 .block_error("virt", "error in reconnecting to libvirt socket")?;
         };
-        Ok(self)
+        Ok(())
     }
 }
 
--- a/src/blocks/libvirt.rs
+++ b/src/blocks/libvirt.rs
@@ -73,51 +73,95 @@ impl ConfigBlock for Libvirt {
 
 impl Block for Libvirt {
     fn update(&mut self) -> Result<Option<Update>> {
-        if !self
+        // Check if the connection is still active and re-instatiate it if it's not
+        match self
             .qemu_conn
             .is_alive()
-            .expect("unrecoverable error with qemu object")
+            .block_error("virt", "error when getting connection status to libvirt")
         {
-            self.qemu_conn = Connect::open_read_only(
-                &self
-                    .qemu_conn
-                    .get_uri()
-                    .expect("could not get URI from the currently connected QEMU object"),
-            )
-            .expect("could not re-connect to libvirtd")
+            Ok(true) => {}
+            Ok(false) => {
+                self.qemu_conn = match Connect::open_read_only(
+                    &self
+                        .qemu_conn
+                        .get_uri()
+                        .block_error("virt", "error in retrieving URI information from libvirt")
+                        .unwrap(), /* Only happens if memory corruption; then we should panic anyway */
+                )
+                .block_error("virt", "error in reconnecting to libvirt socket")
+                {
+                    Ok(c) => c,
+                    Err(e) => return Err(e),
+                };
+            }
+            Err(e) => {
+                self.text.set_state(State::Critical);
+                return Err(e);
+            }
         };
 
-        let mut paused: i64 = 0;
-        match self.qemu_conn.list_all_domains(1 << 5) {
-            Ok(d) => paused = d.len() as i64,
-            Err(e) => eprintln!("{}", e),
-        }
-
-        let mut stopped: i64 = 0;
-        match self.qemu_conn.num_of_defined_domains() {
-            Ok(d) => stopped = d as i64,
-            Err(e) => eprintln!("{}", e),
+        let paused: i64 = match self
+            .qemu_conn
+            .list_all_domains(1 << 5)
+            .block_error("virt", "unable to get paused domains")
+        {
+            Ok(d) => d.len() as i64,
+            Err(e) => {
+                self.text.set_state(State::Critical);
+                return Err(e);
+            }
         };
 
-        let mut running = 0;
-        match self.qemu_conn.list_all_domains(1 << 4) {
-            Ok(d) => running = d.len() as i64,
-            Err(e) => eprintln!("{}", e),
+        let stopped: i64 = match self
+            .qemu_conn
+            .num_of_defined_domains()
+            .block_error("virt", "unable to get stopped domains")
+        {
+            Ok(d) => d as i64,
+            Err(e) => {
+                self.text.set_state(State::Critical);
+                return Err(e);
+            }
+        };
+
+        let running = match self
+            .qemu_conn
+            .list_all_domains(1 << 4)
+            .block_error("virt", "unable to get running domains")
+        {
+            Ok(d) => d.len() as i64,
+            Err(e) => {
+                self.text.set_state(State::Critical);
+                return Err(e);
+            }
         };
 
         let total = running + stopped + paused;
 
         let mut num_images: i64 = 0;
-        match self.qemu_conn.list_all_storage_pools(1 << 1) {
+        match self
+            .qemu_conn
+            .list_all_storage_pools(1 << 1)
+            .block_error("virt", "could not get list of storage pools from libvirt")
+        {
             Ok(pools) => {
                 for pool in pools {
-                    num_images += pool
+                    num_images += match pool
                         .num_of_volumes()
-                        .expect("could not get number of volumes in pool")
-                        as i64;
+                        .block_error("virt", "could not count volumes in storage pool")
+                    {
+                        Ok(n) => n as i64,
+                        Err(e) => {
+                            self.text.set_state(State::Critical);
+                            return Err(e);
+                        }
+                    }
                 }
             }
-            Err(e) => eprintln!("{}", e),
+            Err(e) => {
+                self.text.set_state(State::Critical);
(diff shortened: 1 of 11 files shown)


src/blocks/libvirt.rs
```diff
@@ -0,0 +1,167 @@
...
+                    .get_uri()
+                    .expect("could not get URI from the currently connected QEMU object"),
+            )
+            .expect("could not re-connect to libvirtd")
+        };
+
+        let mut paused: i64 = 0;
+        match self.qemu_conn.list_all_domains(1 << 5) {
+            Ok(d) => paused = d.len() as i64,
+            Err(e) => eprintln!("{}", e),
```
MaxVerevkin, line 116: Perhaps it would be better to propagate the error instead of logging it?

src/blocks/libvirt.rs
```diff
@@ -0,0 +1,167 @@
...
+        })
+    }
+}
+
+impl Block for Libvirt {
+    fn update(&mut self) -> Result<Option<Update>> {
+        if !self
+            .qemu_conn
+            .is_alive()
+            .expect("unrecoverable error with qemu object")
```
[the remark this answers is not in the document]
dijit: I didn't know that existed, thanks for the advice. :)

src/blocks/libvirt.rs
```diff
@@ -0,0 +1,167 @@
...
+pub struct Libvirt {
+    id: usize,
+    text: TextWidget,
+    format: FormatTemplate,
+    update_interval: Duration,
+    qemu_conn: Connect,
+}
+
+#[derive(Deserialize, Debug, Clone)]
+struct Status {
```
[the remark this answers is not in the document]
dijit: Artefact from trying to do something with JSON, good catch, I will remove it.

src/blocks/libvirt.rs
```diff
@@ -73,51 +73,95 @@ impl ConfigBlock for Libvirt {
...
+                        .unwrap(), /* Only happens if memory corruption; then we should panic anyway */
+                )
+                .block_error("virt", "error in reconnecting to libvirt socket")
+                {
+                    Ok(c) => c,
+                    Err(e) => return Err(e),
+                };
+            }
+            Err(e) => {
+                self.text.set_state(State::Critical);
```
MaxVerevkin, line 107: That's not necessary. When the block returns an error, the bar will switch to "error mode" and just display the error (with critical state's theme).

The difference between `.block_error()` (defined [here](https://github.com/greshake/i3status-rust/blob/b7e4ff2a47031be023c785768041c057a42fac4c/src/errors.rs#L12)) and `.expect()` is that `.expect()` just panics and user can only see something like "status command existed unexpectedly", while `.block_error()` wraps the error with a (hopefully) useful message which you then propagate, which will be eventually printed to the user.

There are plans to improve error handling (see #1371) (actually, it _is_ implemented, but on [a separate branch](https://github.com/greshake/i3status-rust/tree/async)).

src/blocks/libvirt.rs
```diff
@@ -73,51 +73,95 @@ impl ConfigBlock for Libvirt {
...
+                        .qemu_conn
+                        .get_uri()
+                        .block_error("virt", "error in retrieving URI information from libvirt")
+                        .unwrap(), /* Only happens if memory corruption; then we should panic anyway */
+                )
+                .block_error("virt", "error in reconnecting to libvirt socket")
+                {
+                    Ok(c) => c,
+                    Err(e) => return Err(e),
+                };
```
MaxVerevkin, line 104: Rust has a syntactic shortcut for this.

```rust
let a = match b {
    Ok(c) => c,
    Err(e) => return Err(e),
};
```

is the same as

```rust
let a = b?;
```

src/blocks/libvirt.rs
```diff
@@ -73,51 +73,95 @@ impl ConfigBlock for Libvirt {
...
-            )
-            .expect("could not re-connect to libvirtd")
+            Ok(true) => {}
+            Ok(false) => {
+                self.qemu_conn = match Connect::open_read_only(
+                    &self
+                        .qemu_conn
+                        .get_uri()
+                        .block_error("virt", "error in retrieving URI information from libvirt")
+                        .unwrap(), /* Only happens if memory corruption; then we should panic anyway */
```
MaxVerevkin, line 98: Hmm, lets propagate this just in case in the future `.get_uri()` will return error for some other, less critical reasons.

dijit:
Hrm, it seems like it's more likely to crash with the new error handling.

src/blocks/libvirt.rs
dijit: I intentionally avoided the `?` operator as the Error type (`virt::Error::error`) does not go into Error::error properly, but after checking it seems to be working, so I'm not sure how I ended up coming to that conclusion.

Thanks for the suggestion. :) <3

src/blocks/libvirt.rs
```diff
@@ -0,0 +1,157 @@
...
+            .block_error("virt", "error when getting connection status to libvirt")?
+        {
+            self.qemu_conn = Connect::open_read_only(
+                &self
+                    .qemu_conn
+                    .get_uri()
+                    .block_error("virt", "error in retrieving URI information from libvirt")?,
+            )
+            .block_error("virt", "error in reconnecting to libvirt socket")?;
+        };
```
dijit, line 98: I tried moving this to a seperate `Impl` block under libvirt, however I'm not smart enough to make it work with the ownership semantics of `self`.

Basically, this would be a good thing to run before each of our collections, since the collections take a few milliseconds to run and libvirt could become unavailable during that time.

MaxVerevkin: Can you share your attempt here?

src/blocks/libvirt.rs
```diff
@@ -0,0 +1,157 @@
...
+        shared_config: SharedConfig,
+        _: Sender<Task>,
+    ) -> Result<Self> {
+        let text = TextWidget::new(id, 0, shared_config)
+            .with_text("vms")
+            .with_icon("virtual-machine")?;
+
+        let qemu_conn = match Connect::open_read_only(&block_config.qemu_url)
+            .block_error("virt", "error when connecting to libvirt")
+        {
```
MaxVerevkin, line 66: Here you can use `?` too.

src/blocks/libvirt.rs
dijit:
```rust
impl Libvirt {
    fn check_and_reconnect(&self) -> Result<Self> {
        match self
            .qemu_conn
            .is_alive()
            .block_error("virt", "error when getting connection status to libvirt")
        {
            Ok(true) => {}
            Ok(false) => {
                new = self.clone();
                new.qemu_conn = Connect::open_read_only(
                    &self
                        .qemu_conn
                        .get_uri()
                        .block_error("virt", "error in retrieving URI information from libvirt")
                        .unwrap(), /* Only happens if memory corruption; then we should panic anyway */
                )
                .block_error("virt", "error in reconnecting to libvirt socket")?;
                return Ok(new);
            }
            Err(e) => {
                return Err(e);
            }
        };
    }
}
```

MaxVerevkin: You can try
1) Take `&mut self` as an argument instead.
2) Just mutate `self` instead of `clone()`ing.

src/blocks/libvirt.rs
```diff
@@ -0,0 +1,149 @@
...
+        // Check if the connection is still active and re-instatiate it if it's not
+        if !self
+            .qemu_conn
+            .is_alive()
+            .block_error("virt", "error when getting connection status to libvirt")?
+        {
+            self.qemu_conn = Connect::open_read_only(
+                &self
+                    .qemu_conn
+                    .get_uri()
```
MaxVerevkin, line 86: This method returns the same string as `Config::qemu_url`, right? Let's just store it ourselves in `Libvirt`?

src/blocks/libvirt.rs
dijit: Which could then be called by `self = &mut Libvirt::check_and_reconnect(self)?;` or `self.check_and_reconect()` I hope.

MaxVerevkin: > `self.check_and_reconect()`

Yes :)

dijit: oh, I made it work. Nevermind. :)

src/blocks/libvirt.rs
```diff
@@ -24,6 +24,25 @@ pub struct Libvirt {
...
+                    .block_error("virt", "error in retrieving URI information from libvirt")?,
+            )
+            .block_error("virt", "error in reconnecting to libvirt socket")?;
+        };
+        Ok(self)
```
MaxVerevkin, line 42: I'm not sure why you need to return `&mut Self` here.

src/blocks/libvirt.rs
dijit: Sure, why not, I figured I was being clever, but a very tiny bit of memory overhead vs a CPU overhead is probably worthwhile.

src/blocks/libvirt.rs
dijit: I think the preview is not linking to the right line anymore; but assuming you mean the `impl Libvirt {}` block: it's because I modify the struct parameter `qemu_conn` if I reconnect the endpoint.

src/blocks/libvirt.rs
MaxVerevkin: Looking at the implementation of `get_uri()` ([here](https://docs.rs/virt/latest/src/virt/connect.rs.html#641-649) and [here](https://docs.rs/virt/latest/src/virt/lib.rs.html#91-93)), It seems that every call to `get_uri` first allocates a string (on the C side), then allocates a second string (on the Rust side), then deallocates the first string and only then returns an owner Rust string (which shortly after will get deallocated). So, there is a point of time when both strings exist, so we can avoid calling `get_uri` without feeling bad for wasting resources xD

src/blocks/libvirt.rs
MaxVerevkin: I meant that you can return `Result<()>` instead of `Result<&mut Self>` from `check_and_reconnect()`.

dijit: AH! OK. Yes, that's good feedback. Thank you!

ammgws:
> Due to the crate [virt](https://crates.io/crates/virt) using a lot of unsafe, I figure it's best to keep it as an optional include.

It doesn't add any new dependencies and the unsafe code would only be called if you're using the block so I'd say it's OK the other way as well.

MaxVerevkin:
> It doesn't add any new dependencies

Unfortunately it does: https://github.com/greshake/i3status-rust/pull/1471/commits/54223fb9f4cb63a7e943984879cfb537eb992e53

ammgws:
Ah, then I take it back.

MaxVerevkin:
Would be nice if there was a `dlopen` version...

dijit:
Once compiled libvirt-dev is not needed (if that's the question).

MaxVerevkin:
> fix: someone decided it was smart to remove updating of the apt repos 

Can you elaborate? What does this commit fix? Why do you think `apt-get update` is necessary?

MaxVerevkin:
@dijit can you rebase please?

The `Block` trait changed, see the definition here [here](https://github.com/greshake/i3status-rust/blob/dc995ff35ff6364b9ce0f73ce73f815ea96e2ea0/src/blocks.rs#L64-L98). In short, `id()` was removed and `name()` was added.

`block_error` function was removed, see what's available [here](https://github.com/greshake/i3status-rust/blob/master/src/errors.rs). In short, instead of `.block_error(<block>, <msg>)?` use `.error_msg(<msg>)?`.

MaxVerevkin:

PR #1478, opened 2022-05-17
Use full_format if battery is in 'Unknown' status

whenov:
Useful for laptops which have battery conservation mode enabled.

On my laptop (Lenovo Legion 5 2020), the battery reports an 'Unknown' status if the battery is in conservation mode and the capacity is between 50% and 60%. In this state the battery is not charging or discharging, and an extra space is shown in the status bar. This patch makes that space disappear.

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -863,7 +863,7 @@ impl Block for Battery {
                 _ => false,
             };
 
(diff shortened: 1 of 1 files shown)


GladOSkar:
The `Unknown` state can mean many different things, so this is probably not great as a general solution. Maybe an option?

Related: #1254

whenov:
@GladOSkar You are right. I'll send another pull request later.

PR #1486, opened 2022-05-31
battery: add a special case for formatting of zero watts

majewsky:
When a power adapter is connected and the battery is fully charged, there will be no current going in or out of the battery. Therefore the power consumption is reported as 0. Before this commit, `Value::from_float(0).watts()` produces the display string "0.0nW" which looks weird. This change hardcodes "0W" instead.

I considered adding this special case to `struct Value` in general, but that would require some consideration as to what the default prefix should be. (Maybe the default prefix could be something that's configurable via chained builder methods in the same way as the units?)

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -851,8 +851,10 @@ impl Block for Battery {
                     Ok(time) => Value::from_string(format!("{}:{:02}", std::cmp::min(time / 60, 99), time % 60)),
                     _ => Value::from_string("×".into()),
                 },
-                // convert µW to W for display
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
> I considered adding this special case to struct Value in general, but that would require some consideration as to what the default prefix should be. (Maybe the default prefix could be something that's configurable via chained builder methods in the same way as the units?)

I think this is the way to go. As for the default prefix, `Prefix::One` should be OK.

MaxVerevkin:
Closing in favor of #1487

PR #1487, opened 2022-05-31
fix prefix choice for zero-valued floats

majewsky:
A value like `Value::from_float(0).watts()` is currently rendered with the lowest-possible prefix, i.e. as `0.0nW`. This choice of prefix is rather unusual in most practical situations. Choosing no prefix at all, i.e. `0.0W` instead of `0.0nW`, is slightly less arbitrary.

This is a counter-proposal for #1486. It's not exactly the same behavior as over there since we end up with `0.0W` instead of `0W`. As for getting the fractional part dropped, I may make a future PR to add something like "max_precision" added to the formatting syntax if that's something that you folks would find desirable, too.


MaxVerevkin (changes requested):
Thanks for your PR. Please update the code to respect `min_prefix` setting. For example, if the user wants download speed to be displayed in `KB`s, zero should be displayed as `0.0 KB` instead of `0.0 B`.

MaxVerevkin (approved):
LGTM :+1:

PR #1501, opened 2022-06-20
custom: fix cycle

MaxVerevkin:
Do not cycle automatically.

Also I realized that something like
```toml
[[block]]
block = "custom"
shell = "sh"
cycle = ["echo '🐑'", "cat /etc/hostname"]
[[block.click]]
button = "left"
cmd = "<command>"
```
currently will not work because click handler will not pass the event to the block. This can be made configurable, like `pass = true` or something. With `true` as the default.


GladOSkar:
Works! And clicking to cycle also works

PR #1537, opened 2022-06-23
music: add url field and use it as a fallback for $combo placeholder

szclsya:
Some song doesn't have title or artist tag. This PR implements reading `xesam:url` from MPRIS2 and expose it as `$url` placeholder. `$url` is also used by `$combo` now as a fallback when there's no title or artist info, so that there will be at least some info shown about the song.

### DIFF
--- a/src/blocks/music.rs
+++ b/src/blocks/music.rs
@@ -31,7 +31,8 @@
 //! ------------|----------------|------
 //! `artist`    | Current artist | Text
 //! `title`     | Current title  | Text
-//! `combo`     | Resolves to "`$artist[sep]$title"`, `"$artist"`, or `"$title"` depending on what information is available. `[sep]` is set by `separator` option. | Text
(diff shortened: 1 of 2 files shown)


src/blocks/music.rs
```diff
@@ -375,12 +383,14 @@ impl Player {
             player_proxy: proxy,
             title: metadata.title().map(Into::into),
             artist: metadata.artist().map(Into::into),
+            url: metadata.url().map(Into::into),
```
MaxVerevkin, line 386: I think these `.map(Into::into)` are not necessary anymore. Not directly related to your PR, but can you please remove them?

szclsya: Sure, fixed in 5217ccd28e2a32f522531fe531629596ee21571a.

MaxVerevkin:
Thanks!

PR #1538, opened 2022-06-23
Restore support for kbdd keyboard driver

sashomasho:
In the process of migration to async I saw that implementation for the kbdd keyboard driver is missing, so I decided to bring it back using zbus. This implementation is slightly different (and simpler) than the previous one, because before the layouts were parsed from setxkbmap -query command, mainly to save space and avoid the long names which kbdd provides. Now, with the option to map the output of a block to whatever the user wants, I've decided to use the strings from kbdd and to relay on the custom mappings for shorter names as the sway driver.

Also the widget was not added in the bar, so I've fixed that too in this PR.

### DIFF
--- a/src/blocks/keyboard_layout.rs
+++ b/src/blocks/keyboard_layout.rs
@@ -21,7 +21,7 @@
 //!  Key     | Value | Type
 //! ---------|-------|-----
 //! `layout` | Keyboard layout name | String
-//! `variant`| Keyboard variant. Only `localebus` and `sway` are supported so far. | String
+//! `variant`| Keyboard variant. Only `localebus`, `sway` and `kbddbus` are supported so far. | String
 //!
 //! # Examples
 //!
@@ -42,12 +42,21 @@
 //! driver = "localebus"
 //! ```
 //!
-//! Listen to kbdd for changes:
+//! Listen to kbdd for changes, the text is in the following format:
+//! "English (US)" - {$layout ($variant)}
+//! use block.mappings to override with shorter names as shown below.
+//! Also use format = "$layout ($variant)" to see the full text to map,
+//! or you can use:
+//! dbus-monitor interface=ru.gentoo.kbdd
+//! to see the exact variant spelling
 //!
 //! ```toml
 //! [[block]]
 //! block = "keyboard_layout"
 //! driver = "kbddbus"
+//! [block.mappings]
+//! "English (US)" = "us"
+//! "Bulgarian (new phonetic)" = "bg"
 //! ```
 //!
 //! Listen to sway for changes:
@@ -71,6 +80,7 @@
 //! ```
 
 use super::prelude::*;
+use regex::Regex;
 use std::collections::HashMap;
 use swayipc_async::{Connection, Event, EventType};
 use tokio::process::Command;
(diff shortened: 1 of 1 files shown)


src/blocks/keyboard_layout.rs
```diff
@@ -318,3 +330,78 @@ trait LocaleBusInterface {
...
+            ),
+            //fallback, although this should not happen
+            None => (self.current_layout.to_string(), None),
+        };
+        Ok(Info { layout, variant })
+    }
+
+    async fn wait_for_chagne(&mut self) -> Result<()> {
+        select! {
+            Some(event) = self.stream.next() => {
```
MaxVerevkin, line 400: Is there a reason to use `select!` here?

sashomasho: Obviously no - I've updated the PR

src/blocks/keyboard_layout.rs
```diff
@@ -318,3 +330,76 @@ trait LocaleBusInterface {
...
+                Some(caps.get(2).unwrap().as_str().to_string()),
+            ),
+            //fallback, although this should not happen
+            None => (self.current_layout.to_string(), None),
+        };
+        Ok(Info { layout, variant })
+    }
+
+    async fn wait_for_chagne(&mut self) -> Result<()> {
+        if let Some(event) = self.stream.next().await {
```
MaxVerevkin, line 399: Perhaps it's better to throw an error if `.next()` returns `None`?

sashomasho: I agree, probably it's best if this is treated as error. I've updated the PR.

src/blocks/keyboard_layout.rs
```diff
@@ -318,3 +330,93 @@ trait LocaleBusInterface {
...
+    fn current_layout_index(&self) -> zbus::Result<u32>;
+
+    #[dbus_proxy(name = "getLayoutName")]
+    fn current_layout(&self, layout_id: u32) -> zbus::Result<String>;
+}
+
+struct KbddBus {
+    stream: layoutNameChangedStream<'static>,
+    layout: String,
+    variant: Option<String>,
```
MaxVerevkin, line 353: Since you are already storing both layout and variant, you can pack them in one filed of type `Info`

src/blocks/keyboard_layout.rs
```diff
@@ -318,3 +330,93 @@ trait LocaleBusInterface {
...
+        let (layout, variant) = Self::parse_kbdd_layout(&re_layout, &current_layout);
+        Ok(Self {
+            stream,
+            layout,
+            variant,
+            re_layout,
+        })
+    }
+
+    fn parse_kbdd_layout(re_layout: &Regex, layout_name: &str) -> (String, Option<String>) {
```
MaxVerevkin, line 387: Similarly, this function can return `Info`

src/blocks/keyboard_layout.rs
```diff
@@ -71,6 +80,8 @@
 //! ```
 
 use super::prelude::*;
+use regex::Regex;
+use std::collections::HashMap;
```
MaxVerevkin, line 84: `HashMap` is already in `super::prelude`

src/blocks/keyboard_layout.rs
MaxVerevkin: Oh, and by the way, it looks like this function does exactly the same thing as `parse_sway_output()`?

sashomasho: Thanks, good catch, I guess I had to browse the surrounding code a little bit more.

src/blocks/keyboard_layout.rs
sashomasho: I noticed that it appeared after I had to merge, and I saw it's used, so I didn't make the effort to double check why it was shown as a new line.. my bad, sorry. 

I've updated the PR, again..

MaxVerevkin: No problem

MaxVerevkin:
Thanks!

PR #1541, opened 2022-06-24
Set widget in keyboard_layout block

TrueDoctor:
Not sure if this is my bad but I had to add this line to get the keyboard block to show up


MaxVerevkin:
Thanks :smile:

PR #1548, opened 2022-06-27
Add apc_ups battery driver

### LINKED ISSUE #1510: v0.30: Re-implement apcupsd driver for battery block
Added late last year and present in v0.21 onwards, see https://github.com/greshake/i3status-rust/issues/1379 and https://github.com/greshake/i3status-rust/pull/1383

Setting a tentative milestone of v0.30, but I think this driver is likely niche enough that we could probably release v0.30 first and leave re-implementation for a later version. @bim9262 What do you think?

bim9262:
Fixes #1510

### DIFF
--- a/src/blocks/battery/apc_ups.rs
+++ b/src/blocks/battery/apc_ups.rs
@@ -1,0 +1,196 @@
+use super::{BatteryDevice, BatteryInfo, BatteryStatus, DeviceName};
+use crate::blocks::prelude::*;
+use std::io;
+use std::str::FromStr;
+use tokio::net::TcpStream;
+use tokio::time::Interval;
+
+pub(super) struct Device {
+    stream: TcpStream,
+    interval: Interval,
+}
+
+impl Device {
+    pub(super) async fn new(dev_name: DeviceName, interval: Seconds) -> Result<Self> {
+        Ok(Self {
+            stream: TcpStream::connect(dev_name.exact().unwrap_or("localhost:3551"))
+                .await
+                .error("Failed to connect to socket")?,
+            interval: interval.timer(),
+        })
+    }
+
+    async fn get_status(&self) -> Result<HashMap<String, String>> {
+        self.write(b"status").await?;
+
+        let result = self.read_response().await?;
+
+        let mut status_data: HashMap<String, String> = HashMap::new();
+        for line in result.lines() {
+            let (key, value) = line.split_once(':').unwrap();
+            status_data.insert(String::from(key.trim()), String::from(value.trim()));
+        }
+
+        Ok(status_data)
+    }
+
+    async fn write(&self, msg: &[u8]) -> Result<()> {
+        let msg_len = msg.len();
+        if msg_len >= 1 << 16 {
+            return Err(Error::new(
+                "msg is too long, it must be less than 2^16 characters long",
+            ));
+        }
+
+        // Write how many bytes of data are going to be sent (only need the last two octets)
+        // Write the actual message
+        let msgs = [&msg_len.to_be_bytes()[6..], msg];
+
+        for msg in msgs {
+            loop {
+                // Wait for the socket to be writable
+                self.stream
+                    .writable()
+                    .await
+                    .error("Socket closed unexpectedly")?;
+
+                // Try to write data, this may still fail with `WouldBlock`
+                // if the readiness event is a false positive.
+                match self.stream.try_write(msg) {
+                    Ok(_) => {
+                        break;
+                    }
+                    Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => {
+                        continue;
+                    }
+                    Err(e) => {
+                        return Err(Error::new(e.to_string()));
+                    }
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn read_exact(&self, n: usize) -> Result<Vec<u8>> {
+        loop {
+            // Wait for the socket to be readable
+            self.stream
+                .readable()
+                .await
+                .error("Socket closed unexpectedly")?;
+
+            let mut buf = vec![0_u8; n];
+
+            // Try to read data, this may still fail with `WouldBlock`
+            // if the readiness event is a false positive.
+            match self.stream.try_read(&mut buf) {
+                Ok(_) => return Ok(buf),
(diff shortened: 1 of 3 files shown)


src/blocks/battery/apc_ups.rs
```diff
@@ -0,0 +1,196 @@
...
+                        return Err(Error::new(e.to_string()));
+                    }
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn read_exact(&self, n: usize) -> Result<Vec<u8>> {
```
MaxVerevkin, line 76: Did you try [`AsyncReadExt::read_exact()`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read_exact)?

src/blocks/battery/apc_ups.rs
```diff
@@ -0,0 +1,196 @@
...
+                .readable()
+                .await
+                .error("Socket closed unexpectedly")?;
+
+            let mut buf = vec![0_u8; n];
+
+            // Try to read data, this may still fail with `WouldBlock`
+            // if the readiness event is a false positive.
+            match self.stream.try_read(&mut buf) {
+                Ok(_) => return Ok(buf),
```
MaxVerevkin, line 89: I don't think this is correct: `try_read()` is not obligated to fill the buffer completely, i.e. it can return a number less than `n`.

src/blocks/battery/apc_ups.rs
```diff
@@ -0,0 +1,196 @@
...
+                }
+            }
+        }
+    }
+
+    async fn read_response(&self) -> Result<String> {
+        let mut buf = String::new();
+        loop {
+            let read_info = self.read_exact(2).await?;
+            let read_size = ((read_info[0] as usize) << 8) + (read_info[1] as usize);
```
MaxVerevkin, line 104: There is also [`AsyncReadExt::read_u16()`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read_u16)

src/blocks/battery/apc_ups.rs
```diff
@@ -0,0 +1,196 @@
...
+                // Wait for the socket to be writable
+                self.stream
+                    .writable()
+                    .await
+                    .error("Socket closed unexpectedly")?;
+
+                // Try to write data, this may still fail with `WouldBlock`
+                // if the readiness event is a false positive.
+                match self.stream.try_write(msg) {
+                    Ok(_) => {
```
MaxVerevkin, line 60: `try_write()` is not guaranied to write all bytes, use [`write_all()`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#method.write_all) instead.

bim9262:
Thanks for pointing me to `tokio::io::{AsyncReadExt, AsyncWriteExt}`!

src/blocks/battery/apc_ups.rs
```diff
@@ -0,0 +1,178 @@
...
+        Ok(buf)
+    }
+}
+
+#[async_trait]
+impl BatteryDevice for Device {
+    async fn get_info(&mut self) -> Result<Option<BatteryInfo>> {
+        let status_data = self.get_status().await?;
+
+        let capacity = status_data.get_property::<f64>("BCHARGE", "Percent")?;
```
MaxVerevkin, line 127: Capacity should be read after status because if status is "COMMLOST", then this line will throw an error

src/blocks/battery/apc_ups.rs
```diff
@@ -0,0 +1,178 @@
+use std::str::FromStr;
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
```
MaxVerevkin, line 2: These extension traits are already in `blocks::prelude::*`

src/blocks/battery/apc_ups.rs
```diff
@@ -0,0 +1,196 @@
...
+        let status_data = self
+            .get_status()
+            .await
+            .map_err(|e| {
+                debug!("{e}");
+                e
+            })
+            .unwrap_or_default();
+
+        let status_str = status_data.get("STATUS").map_or("COMMLOST", |x| x);
```
bim9262, line 141: @MaxVerevkin , any tips on writing this or the previous let more cleanly?

MaxVerevkin: I think it's good :)

bim9262: Ok the `|x| x` felt kinda hacky to get from the `&String` to a `&str`, but if you think it's kosher, I'm happy!

src/blocks/battery/apc_ups.rs
```diff
@@ -0,0 +1,196 @@
...
+                .into();
+            if read_size == 0 {
+                break;
+            }
+            let mut read_buf = vec![0_u8; read_size];
+            self.0
+                .read_exact(&mut read_buf)
+                .await
+                .error("Could not read from socket")?;
+            buf.extend(String::from_utf8(read_buf));
```
MaxVerevkin, line 91: Probably the last comment. Here `String::from_utf8()` returns `Result`, which happens to implement `IntoIterator`. So this code is equivalent to

```rust
if let Ok(s) = String::from_utf8(read_buf) {
    buf.push_str(&s);
}
```

You probably wanted something like

```rust
buf.push_str(std::str::from_utf8(&read_buf).error(...)?);
```

src/blocks/battery/apc_ups.rs
MaxVerevkin: I'm happy either way :)

src/blocks/battery/apc_ups.rs
MaxVerevkin: fixed in my commit

src/blocks/battery/apc_ups.rs
MaxVerevkin: changed in my commit

MaxVerevkin:
I don't have the necessary hardware to test this, but everything LGTM, thanks!

bim9262:
Thanks @MaxVerevkin, your changes look great. I have confirmed that everything still works :)

PR #1563, opened 2022-07-13
pacman: don't use sh to run fakeroot

MaxVerevkin:
fakeroot can be run directly without sh indirection.

Needs verification that it works not only for me.


ammgws:
Works for me too.

PR #1569, opened 2022-07-15
remove color_space dependency

sonarypt:
Remove lines from 3 files that related to color_space dependency from repository


MaxVerevkin:
This dependency is actually used in `themes.rs`, so you'll have to copy `Rgb` and `Hsv` structs + conversion functions from `color_space` source code. Optionally, it would be preferable to refactor color related code into `src/themes/color.rs`.

PR #1573, opened 2022-07-22
Use API data for AT instead of calculating

jakubkaczor:
At the time of implementing the feature there was no API parameter, but
it was introduced a few months later. According to OpenWeather[^1], they
used Australian Apparent Temperature (AT), so the results should be the
same. Therefore, this commit serves as a cleanup, code simplification.

References:
[^1]: <https://openweather.co.uk/blog/post/new-feels-temperature-openweather-apis>


MaxVerevkin (approved):
This gives me a bit different result, and in my case it seems to be more accurate.

MaxVerevkin:
Thanks!

PR #1577, opened 2022-07-24
weather: Add support for met.no

sigvei:
Refactors code for OpenWeatherMap into its own module, and adds a new
module for met.no.

The api for met.no is documented at [api.met.no](https://api.met.no/weatherapi/locationforecast/2.0/documentation). It is very straightforward to use, but it's data structures are very heavily nested. I think this is mostly right.

There is a separate call to a legends api to fetch the translations of the summary string. This does not really have to be done at every request, but it's not so easy to store data in shared memory inside i3status. I am willing to find a way if this is a deal breaker.


MaxVerevkin:
Haven't tested the changes yet, but can you please split `weather.rs` into `weather.rs`, `weather/open_weather_map.rs` and `weather/met_no.rs`?

sigvei:
> can you please split `weather.rs` into `weather.rs`, `weather/open_weather_map.rs` and `weather/met_no.rs`?

Done.

src/blocks/weather.rs
```diff
@@ -103,7 +112,7 @@ fn default_interval() -> Seconds {
 
 #[derive(Deserialize, Debug)]
 #[serde(tag = "name", rename_all = "lowercase")]
-enum WeatherService {
+pub enum WeatherService {
     OpenWeatherMap {
```
MaxVerevkin, line 116: You can avoid duplicating code by doing this
```rust
#[derive(Deserialize, Debug)]
#[serde(tag = "name", rename_all = "lowercase")]
pub enum WeatherService {
    OpenWeatherMap(open_weather_map::OpenWeatherMapConfig),
    MetNo(met_no::Config),
}
```

src/blocks/weather/open_weather_map.rs
```diff
@@ -0,0 +1,145 @@
+use super::*;
+
+pub const OPEN_WEATHER_MAP_URL: &str = "https://api.openweathermap.org/data/2.5/weather";
+pub const OPEN_WEATHER_MAP_API_KEY_ENV: &str = "OPENWEATHERMAP_API_KEY";
+pub const OPEN_WEATHER_MAP_CITY_ID_ENV: &str = "OPENWEATHERMAP_CITY_ID";
+pub const OPEN_WEATHER_MAP_PLACE_ENV: &str = "OPENWEATHERMAP_PLACE";
```
MaxVerevkin, line 6: `OPEN_WEATHER_MAP_` prefix can now be dropped

src/blocks/weather/open_weather_map.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+#[derive(Deserialize, Debug)]
+struct ApiWeather {
+    main: String,
+    description: String,
+}
+
+pub async fn get(
+    config: OpenWeatherMapConfig,
```
MaxVerevkin, line 70: This should be a reference to avoid unnecessary `clone()`s

src/blocks/weather.rs
```diff
@@ -38,6 +38,17 @@
...
+//! # met.no Options
+//!
+//! Key | Values | Required | Default
+//! ----|--------|----------|--------
+//! `name` | `metno`. | Yes | None
+//! `coordinates` | GPS latitude longitude coordinates as a tuple, example: `["39.236229089090216","9.331730718685696"]`
+//! `lang` | Language code: `en`, `nn` or `nb` | No | `en`
+//! `altitude` | Meters above sea level of the ground | No | Approximated by server
+//!
+//! Met.no does not support location name or apparent temperature.
```
MaxVerevkin, line 50: I suppose we can re-add custom apparent temperature calculation function which was removed in #1573

src/blocks/weather.rs
```diff
@@ -38,6 +38,17 @@
 //! in which case they must be provided in the environment variables
 //! `OPENWEATHERMAP_API_KEY`, `OPENWEATHERMAP_CITY_ID`, `OPENWEATHERMAP_PLACE`.
 //!
+//! # met.no Options
+//!
+//! Key | Values | Required | Default
+//! ----|--------|----------|--------
+//! `name` | `metno`. | Yes | None
+//! `coordinates` | GPS latitude longitude coordinates as a tuple, example: `["39.236229089090216","9.331730718685696"]`
```
MaxVerevkin, line 46:
```suggestion
//! `coordinates` | GPS latitude longitude coordinates as a tuple, example: `["39.236229089090216","9.331730718685696"]` | Required if `autolocate = false` | None
```

ammgws:
@JayGray the refactoring in this PR should make it simpler to add your implementation from #1468

src/blocks/weather.rs
sigvei: Ah, of course! I saw that this was a bad pattern, but couldn't free myself from the struct enums.

src/blocks/weather.rs
```diff
@@ -38,6 +38,17 @@
 //! in which case they must be provided in the environment variables
 //! `OPENWEATHERMAP_API_KEY`, `OPENWEATHERMAP_CITY_ID`, `OPENWEATHERMAP_PLACE`.
 //!
+//! # met.no Options
+//!
+//! Key | Values | Required | Default
+//! ----|--------|----------|--------
+//! `name` | `metno`. | Yes | None
+//! `coordinates` | GPS latitude longitude coordinates as a tuple, example: `["39.2362","9.3317"]` | Required if `autolocate = false` | None
```
MaxVerevkin, line 46: Why did you decide to accept coordinates as strings? Why not just use `f64`?

MaxVerevkin: Same thing with altitude

sigvei: I was worried that it would be confusing to have different configuration standards for openweathermap and metno, and did not want to invalidate old config files by changing the definition for openweathermap. 

For altitude I have no explanation.

MaxVerevkin: @ammgws is this worth a breaking change?

src/blocks/weather/met_no.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+impl Default for ApiLanguage {
+    fn default() -> Self {
+        ApiLanguage::English
+    }
+}
+
+impl From<&Option<&str>> for ApiLanguage {
+    fn from(o: &Option<&str>) -> Self {
```
MaxVerevkin, line 35: This is not used anywhere in the code

src/blocks/weather/met_no.rs
```diff
@@ -0,0 +1,202 @@
+use super::*;
+
+#[derive(Deserialize, Debug)]
+#[serde(tag = "name", rename_all = "lowercase")]
+pub struct Config {
+    coordinates: Option<(String, String)>,
+    altitude: Option<String>,
+    lang: ApiLanguage,
```
MaxVerevkin, line 8: `lang` lacks `#[serde(default)]`

src/blocks/weather/met_no.rs
```diff
@@ -0,0 +1,202 @@
...
+    wind_from_direction: Option<f64>,
+    wind_speed: Option<f64>,
+    relative_humidity: Option<f64>,
+}
+
+const LEGENDS_URL: &str = "https://api.met.no/weatherapi/weathericon/2.0/legends";
+const FORECAST_URL: &str = "https://api.met.no/weatherapi/locationforecast/2.0/compact";
+
+async fn translate(summary: &str, lang: &ApiLanguage) -> Result<String> {
+    let legend: HashMap<String, LegendsResult> = REQWEST_CLIENT
```
MaxVerevkin, line 98: I think this function can be executed just once

sigvei: Yes, indeed. But here I was kind of stumped by the modular architecture. Any global data must be initialized in the loop in weather.rs, and passed by reference to the get function. We could initialize a legends hashmap and then let the get function in met_no check if it is empty, and only then call get_legends to fetch the data and fill in the map. But how can we represent such a shared, mutable object in this async architecture?

 I tried to do this:

```rust
pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
    ...
    let mut legends_store: HashMap<String, met_no::LegendsResult> = HashMap::new();

    loop {
        let data = api
            .recoverable(|| config.service.get(&mut legends_store, &location))
            .await?;
```


But that is disallowed by the compiler: captured variable cannot escape `FnMut` closure body.

Maybe an `Arc<Mutex<HashMap>>` would be allowed? But it seems bad to leek so much of the `met_no` specific logic into the weather.rs file.

sigvei: I do of course realise that I could just initialize it before the loop, if config is the MetNo variant. Maybe that is the best way to do it, even if it feels a bit dirty to initialize outside the met_no module.

MaxVerevkin: I think it's better to store config + state of each service in their own struct. You can look at `battery` or `keyboard_layout` block for examples.

src/blocks/weather.rs
```diff
@@ -140,42 +190,25 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
...
+    // met_no requires a one-time fetch of a set of "legends", which are
+    // translations of symbols into weather strings.
+    let legends_store: met_no::LegendsStore = match config.service {
+        WeatherService::MetNo(_) => met_no::get_legend().await?,
+        _ => HashMap::new(),
+    };
+
     loop {
+        let location = match config.autolocate {
+            true => find_ip_location().await.unwrap_or(None),
```
MaxVerevkin, line 202: This should be executed via `api.recoverable()`

JayGray:
> @JayGray the refactoring in this PR should make it simpler to add your implementation from #1468

Awesome, thx. I will have a look again later.

src/blocks/weather.rs
sigvei: Fixed, see 7aa0069.

src/blocks/weather/met_no.rs
sigvei: Ah, that is better. Fixed now (with the addition of Arc<Mutex>> for the config) in 7aa0069.

src/blocks/weather.rs
```diff
@@ -190,21 +184,21 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
         .new_widget()
         .with_format(config.format.with_default("$weather $temp")?);
 
-    // met_no requires a one-time fetch of a set of "legends", which are
-    // translations of symbols into weather strings.
-    let legends_store: met_no::LegendsStore = match config.service {
-        WeatherService::MetNo(_) => met_no::get_legend().await?,
-        _ => HashMap::new(),
+    let provider: Arc<Mutex<dyn WeatherProvider + Send + Sync>> = match config.service {
```
MaxVerevkin, line 187: why not just `Box`?

MaxVerevkin: Instead of `Arc<Mutex<_>>`

MaxVerevkin: Oh, I see.

src/blocks/weather/met_no.rs
```diff
@@ -109,86 +111,93 @@ fn translate(legend: &LegendsStore, summary: &str, lang: &ApiLanguage) -> String
...
-            | "lightsleet" | "lightsleetshowers" | "lightsnow" | "lightsnowshowers" | "sleet"
-            | "sleetshowers" | "snow" | "snowshowers" => WeatherIcon::Snow,
-            _ => WeatherIcon::Default,
-        },
-    })
+#[async_trait]
+impl WeatherProvider for Config {
+    async fn get_weather(&mut self, location: &Option<LocationResponse>) -> Result<WeatherResult> {
+        if self.legend.is_none() {
+            self.legend = Some(get_legend().await?);
```
MaxVerevkin, line 118: This should be done in `new()` function. This way there will be no need for mutability & mutex.

MaxVerevkin:
```rust
pub(super) struct Service {
    config: Config,
    legend: SomeType,
}

impl Service {
    pub(super) async fn new(config: Config) -> Result<Self> {
        todo!()
    }
}

#[async_trait]
impl WeatherProvider for Service {
    async fn get_weather(&self, location: &Option<LocationResponce>) -> Result<WeatherResult> {
        todo!()
    }
}
```

src/blocks/weather.rs
```diff
@@ -190,21 +184,21 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
...
-        let location = match config.autolocate {
-            true => find_ip_location().await.unwrap_or(None),
-            false => None,
-        };
-
         let data = api
-            .recoverable(|| config.service.get(&legends_store, &location))
+            .recoverable(|| async {
+                let location = match config.autolocate {
+                    true => find_ip_location().await.unwrap_or(None),
```
MaxVerevkin, line 196: Now that it is inside `api.recoverable`, this can/should propagate the error: `true => Some(find_ip_location().await?)`

src/blocks/weather/met_no.rs
sigvei: Even better! Okay, implemented in afe4e23.

src/blocks/weather.rs
sigvei: Okay, done in 631e81a.

src/blocks/weather.rs
```diff
@@ -38,6 +38,17 @@
...
+//! # met.no Options
+//!
+//! Key | Values | Required | Default
+//! ----|--------|----------|--------
+//! `name` | `metno`. | Yes | None
+//! `coordinates` | GPS latitude longitude coordinates as a tuple, example: `["39.2362","9.3317"]` | Required if `autolocate = false` | None
+//! `lang` | Language code: `en`, `nn` or `nb` | No | `en`
+//! `altitude` | Meters above sea level of the ground | No | Approximated by server
+//!
+//! Met.no does not support location name or apparent temperature.
```
MaxVerevkin, line 50: This is outdated

src/blocks/weather/met_no.rs
```diff
@@ -0,0 +1,221 @@
...
+pub(super) struct Service {
+    config: Config,
+    legend: LegendsStore,
+}
+
+impl Service {
+    pub(super) async fn new(config: Config) -> Result<Self> {
+        Ok(Self {
+            config,
+            legend: get_legend().await?,
```
MaxVerevkin, line 23: I think this should be inside of `api.recoverable` so that if the bar is launched before computer gains internet access, the block will eventually get working. So I suppose this `new` function should also accept `api: &mut CommonApi`.

src/blocks/weather.rs
sigvei: Fixed in 123ca9a

src/blocks/weather/met_no.rs
sigvei: Fixed in b216878.

MaxVerevkin (commented):
LGTM :+1: Thanks for quickly responding to feedback!

MaxVerevkin (approved):

ammgws (approved):

PR #1580, opened 2022-07-27
custom_dbus: allow override of interface name with env var

### LINKED ISSUE #1579: Can't use multiple custom_dbus blocks when they're in different bars
Thanks you so much for the work on the async branch. I've been waiting to use multiple dbus blocks for a while :)  I've been testing master to see if I can already do what I'd like.

I have two status bars, top and bottom. When I have multiple dbus blocks in the same bar, everything works fine, but when I move one of them to the other bar, none or only one of the bars will work.

Smallest config to reproduce this with:

`config-top.toml`:
```toml
[[block]]
block = "custom_dbus"
path = "/Foo"
```

`config-bottom.toml`:
```toml
[[block]]
block = "custom_dbus"
path = "/Baz"
```

I'm guessing this has something to do with two different instances of the process fighting over who gets the dbus namespace. An easy fix would be to have a namespace for each bar, although that's not very clean. Another idea could be to make the namespace a setting you can override per block? I'm no expert on dbus itself, so maybe there's another way.

cheers

RobinRamael:
fixes #1579

### DIFF
--- a/src/blocks/custom_dbus.rs
+++ b/src/blocks/custom_dbus.rs
@@ -21,7 +21,7 @@
 //! path = "/my_path"
 //! ```
 //!
-//! Useage:
+//! Usage:
 //! ```sh
 //! # set full text to 'hello' and short text to 'hi'
 //! busctl --user call rs.i3status /my_path rs.i3status.custom SetText ss hello hi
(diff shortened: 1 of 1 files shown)


src/blocks/custom_dbus.rs
```diff
@@ -31,11 +31,18 @@
 //! busctl --user call rs.i3status /my_path rs.i3status.custom SetState s good
 //! ```
 //!
+//! Because it's impossible to publish objects to the same interface from different
+//! processes, having multiple dbus blocks in different bars won't work. As a workaround,
+//! you can set the env var `I3STATUS_DBUS_INTERFACE` to set the interface a bar works on to
```
MaxVerevkin, line 36: Probably better to call it `I3RS_DUBS_INTERFACE` because it is shorter and this is not `i3status`. Also this will be similar to [github's token env variable](https://github.com/greshake/i3status-rust/blob/85cd6cc30ce873b14712dbaa2cd4786d26136add/src/blocks/github.rs#L86).

MaxVerevkin: By the way, not sure but I think technically it is a _name_, not _interface_.

RobinRamael: Ok. I've changed it to I3RS_DBUS_NAME.

MaxVerevkin:
Thanks!

PR #1585, opened 2022-08-03
Add better support for empty batteries

### LINKED ISSUE #1431: battery: "Not charging" state treated as "full"
I have a thikpad with two batteries. The charge level of each battery is always available but in some cases (e.g. battery disabled when charge is less or equal 5%) `/sys/class/power_supply/BAT#/status` will report the state `Not charging`.

I can see that after a recent change of the implementation of battery.rs (bc17fe5112bb7914ae140335569e5c212df83f7b) this case is treated as equivalent to battery full:
https://github.com/greshake/i3status-rust/blob/4c2ddf9a6b108dc51c6546713c67314e9df59242/src/blocks/battery.rs#L940-L943

This causes a minor cosmetic issue, since it will use the wrong icon (and block color) even when the actual charge level can be determined:
bar

svenstaro:
Add `empty_threshold` and `empty_format` to `battery`.
Also change `full_threshold` to `95` as batteries often do not go all the way up to 100%.

Should also fix #1431.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,7 +24,7 @@ block = "net"
 ```
 
 - `kdeconnect` block only supports kdeconnect v20.11.80 and newer (December 2020 and newer)
-
+- `battery` now defaults `full_threshold` to `95` as often batteries never fully charge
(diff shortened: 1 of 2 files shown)


MaxVerevkin (approved):

MaxVerevkin:
Thanks!

PR #1592, opened 2022-08-15
Rework blocks errors

MaxVerevkin:
Implement a way for blocks to report (clickable) errors while still
being functional via `api.set_error()`.

`api.recoverable()` now uses `api.set_error()`.

Errors reported with `api.set_error()` provide the name of the block in
it's message, without the need for an explicit argument, just like
unrecoverable errors.

This makes it trivial to implement [this TODO](https://github.com/greshake/i3status-rust/blob/3eba9745309cd6261d4f7d51c1ee15e300ef674d/src/blocks/custom.rs#L128).

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,16 @@ Key | Description | Default
 `theme` | The [theme](https://github.com/greshake/i3status-rust/blob/master/doc/themes.md#available-themes) that should be used. | `"plain"`
 `[theme.theme_overrides]` | Refer to `Themes and Icons` below. | None
 `scrolling` | The direction of scrolling, either `natural` or `reverse`. | `"reverse"`
+`error_format` | A string to customise how block errors are displayed. See below for available placeholders. | `$short_error_message|X`
+`error_fullscreen_format` | A string to customise how block errors are displayed when clicked. See below for available placeholders. | `$full_error_message`
 `[[block]]` | All blocks that will exist in your bar. | none
+
+Available `error_format` and `error_fullscreen_format` placeholders:
+
+Placeholder         | Value
+--------------------|------
+full_error_message  | The full error message
+short_error_message | The short error message, if available
 
 There are also some optional block-level configuration variables, common to all blocks:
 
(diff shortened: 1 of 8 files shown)


MaxVerevkin:
Here's what I mean by "customizable errors":

```toml
# in config.toml
error_format = "$short_error_message|X"
error_fullscreen_format = "$full_error_message.rot-str(170, 0.2)"
```

https://user-images.githubusercontent.com/34583604/184911730-e6f01d84-7030-493e-815b-7ad941a5efc2.mp4

src/config.rs
```diff
@@ -68,4 +73,12 @@ impl Config {
     fn default_double_click_delay() -> u64 {
         200
     }
+
+    fn default_error_format() -> String {
+        "$short_error_message|X".into()
```
ammgws, line 78: Can you update the table in the README too? For the others as well.

MaxVerevkin: Yeah, I was about to comment that I'm not sure where to document it. Like, there should also be a table of available placeholders and so on. And README is getting a bit cluttered...

ammgws: We could just list it in table for now and figure out how to improve it later.

MaxVerevkin: Okay

README.md
```diff
@@ -31,8 +31,17 @@ Key | Description | Default
...
 `scrolling` | The direction of scrolling, either `natural` or `reverse`. | `"reverse"`
+`error_format` | A string to customise how block errors are displayed. See below for available placeholders. | `$short_error_message|X`
+`error_fullscreen_format` | A string to customise how block errors are displayed when clicked. See below for available placeholders. | `$full_error_message`
 `[[block]]` | All blocks that will exist in your bar. | none
 
+Available `error_format` and `error_fullscreen_format` placeholders:
+
+Placeholder         | Value
+--------------------|------
+full_error_message  | The full error message
```
MaxVerevkin, line 42: An `$icon` would be nice too, I think.

PR #1598, opened 2022-08-23
net: implement ip&ipv6

### LINKED ISSUE #1069: net block: investigate using netlink/nl80211 instead of shelling out

### LINKED ISSUE #1515: v0.30: reimplement ip, ipv6 and bitrate (for wired connections) placeholders for net block
I think ip/ipv6 can be done via netlink, not sure about bitrate though.

### LINKED ISSUE #1595: `net` incorrect default interface
I believe I discovered a bug after installing docker my net block no longer rendered.

```
$ ip route list
default via 192.168.1.1 dev wlp166s0 proto dhcp src 192.168.1.50 metric 600
172.17.0.0/16 dev docker0 proto kernel scope link src 172.17.0.1 linkdown
192.168.1.0/24 dev wlp166s0 proto kernel scope link src 192.168.1.50 metric 600
```

`fn default_interface()` in `netlink.rs` returns `Some("docker0")`. I expected it would return `Some("wlp166s0")` as it is the interface of my _default_ route. 

Reading the source it seems to me `default_interface()` returns whatever netlink returned as the ~~"first"~~ "last"? interface. I can't tell if there is any guarantee this should be the default interface. Not sure if this is intended but it seems like odd behavior. 

Maybe this is related to #1515? #1515 is help wanted, I could look more into this.

MaxVerevkin:
fixes #1515
fixes #1595
fixes #1069
notfixes #587

Needs testing.

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -44,7 +44,7 @@
 //! - `net_down`
 
 use super::prelude::*;
-use crate::netlink::{default_interface, NetDevice};
+use crate::netlink::NetDevice;
 use crate::util;
 use std::time::Instant;
 
@@ -81,17 +81,8 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
     let mut rx_hist = [0f64; 8];
 
     loop {
-        let mut speed_down: f64 = 0.0;
-        let mut speed_up: f64 = 0.0;
-
-        let device = NetDevice::from_interface(match &config.device {
-            Some(i) => i.clone(),
-            None => default_interface().await.unwrap_or_else(|| "lo".into()),
-        })
-        .await;
-
-        match device {
-            Some(device) if !device.is_up().await? => {
+        match NetDevice::new(config.device.as_deref()).await? {
+            Some(device) if !device.iface.is_up => {
                 if config.hide_inactive {
                     api.hide().await?;
                 } else {
@@ -102,8 +93,11 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
             Some(device) => {
                 widget.set_format(format.clone());
 
+                let mut speed_down: f64 = 0.0;
+                let mut speed_up: f64 = 0.0;
+
                 // Calculate speed
-                match (stats, device.read_stats().await) {
+                match (stats, device.iface.stats) {
                     // No previous stats available
                     (None, new_stats) => stats = new_stats,
                     // No new stats available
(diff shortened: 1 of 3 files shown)


MaxVerevkin:
> Needs testing.

VPNs in particular (their presence, their default-ness, their up/down state)

Alextopher:
Attempting to do some more detailed testing. 

Works are expected when there is a single default route:
```
$ ip route
default via 192.168.1.1 dev wlp166s0 proto dhcp src 192.168.1.50 metric 600
172.17.0.0/16 dev docker0 proto kernel scope link src 172.17.0.1 linkdown
192.168.1.0/24 dev wlp166s0 proto kernel scope link src 192.168.1.50 metric 600
```

However, when there are two default routes 
```
$ route
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
default         _gateway        0.0.0.0         UG    100    0        0 enp0s20f0u4c4i2
default         _gateway        0.0.0.0         UG    600    0        0 wlp166s0
172.17.0.0      0.0.0.0         255.255.0.0     U     0      0        0 docker0
172.20.10.0     0.0.0.0         255.255.255.240 U     100    0        0 enp0s20f0u4c4i2
192.168.1.0     0.0.0.0         255.255.255.0   U     600    0        0 wlp166s0
```

the default interfaces function returns the last default route while the kernel seems to prefer using the first.

Alextopher:
Using private internet access VPN (`tun0` interface name) works for me :+1:

MaxVerevkin:
> the default interfaces function returns the last default route while the kernel seems to prefer using the first.

Thanks, fixed.

MaxVerevkin:
> Also was bitrate for wired connections available via netlink?

via netlink but not rtnetlink... see https://github.com/greshake/i3status-rust/issues/1515#issuecomment-1213807759

Alextopher:
I wonder if this creates a minimum supported kernel version

ammgws:
Are you talking about the SSID issue?

ammgws:
I think this should be OK for now

Issue #1594, opened 2022-08-19
weather: add `autolocate_interval` option.


MaxVerevkin:
Calling `find_ip_location()` on each update might not bee needed in many cases.

```toml
autolocate_interval = 3600 # Every hour
autolocate_interval = "once" # Call find_ip_location() only once
```

Alextopher:
@MaxVerevkin I could pick this up

Alextopher:
There is some room for discussion on the behavior. 

Should the block be updated when the auto locate triggers? I personally think the weather block should only change what info is displayed when `interval` triggers.

MaxVerevkin:
> Should the block be updated when the auto locate triggers? I personally think the weather block should only change what info is displayed when interval triggers.

I agree. This wold also prevent the situation when `interval` and `autolocate_interval` align and cause the weather to update twice immediately. (But this could be solved by something like [this](https://github.com/greshake/i3status-rust/blob/master/src/formatting/scheduling.rs)).

Another question: should the auto locate be triggered on `UpdateRequest` (i.e. when a signaled or clicked with `update=true`)? I'm not sure.

Alextopher:
I think the implementation that is the least likely to cause confusion is triggering autolocate on `UpdateRequest`.

PR #1600, opened 2022-08-23
weather: add `autolocate_interval` option

Alextopher:
Issue: #1594

Calling `find_ip_location()` is no longer necessarily executed on every update. 
Behavior is controlled with new `weather.autolocate_interval` option.
if `autolocate_interval` is set while `autolocate` is false auto locate is **not** preformed. Not sure how to make this a 
The default for `autolocate_interval` is to match `interval` (this makes behavior backwards compatible)
Auto locate does not trigger a new render

### DIFF
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -17,7 +17,7 @@
 //! `format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | `"$weather $temp"`
 //! `interval` | Update interval, in seconds. | `600`
 //! `autolocate` | Gets your location using the ipapi.co IP location service (no API key required). If the API call fails then the block will fallback to `city_id` or `place`. | `false`
-//! `autolocate_interval` | Update interval for `autolocate` in seconds or "once" | `600`
+//! `autolocate_interval` | Update interval for `autolocate` in seconds or "once" | `interval`
 //!
 //! # OpenWeatherMap Options
 //!
@@ -105,15 +105,10 @@ struct WeatherConfig {
     service: WeatherService,
     #[serde(default)]
     autolocate: bool,
-    #[serde(default = "default_autolocate_interval")]
-    autolocate_interval: Seconds,
+    autolocate_interval: Option<Seconds>,
 }
 
 fn default_interval() -> Seconds {
-    Seconds::new(600)
-}
-
-fn default_autolocate_interval() -> Seconds {
     Seconds::new(600)
 }
 
@@ -194,11 +189,15 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
 
     if config.autolocate {
         let mut interval = tokio::time::interval(config.interval.0);
-        let mut autolocate_interval = tokio::time::interval(config.autolocate_interval.0);
+        let mut autolocate_interval = tokio::time::interval(match config.autolocate_interval {
+            Some(s) => s.0,
+            None => config.interval.0,
+        });
         let mut location: Option<Coordinates> = api.recoverable(find_ip_location).await?;
 
         loop {
             select! {
+                biased;
                 _ = autolocate_interval.tick() => {
                     location = api.recoverable(find_ip_location).await?;
                 }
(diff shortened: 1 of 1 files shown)


Alextopher:
I think it might be worth creating another case for when `interval == autolocate_interval`. Only using a single timer.

Alextopher:
Now I think this is ready for review

src/blocks/weather.rs
```diff
@@ -185,24 +187,80 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
...
+            let mut autolocate_interval =
+                tokio::time::interval(config.autolocate_interval.unwrap().0);
+            let mut location: Option<Coordinates> = api.recoverable(find_ip_location).await?;
 
-        widget.set_icon(data.icon.to_icon_str())?;
-        widget.set_values(data.into_values());
-        api.set_widget(&widget).await?;
+            loop {
+                select! {
+                    biased; // both timers instantly `tick()` autolocate should always run first
```
MaxVerevkin, line 221: Currently, with `autolocate_interval = "once"` autolocate function is actually called twice. This is because, as you mentioned, both timers instantly tick. This can be solved by delaying the initial tick by one interval.

Also default ["missed tick behavior"](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html) is not the best one.

So I would recommend just pre-computing both the autolocate and weather before the loop, and then create intervals using `config.interval.timer()` (it will delay the first tick and set missed tick behavior).

src/blocks/weather.rs
```diff
@@ -185,24 +187,80 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
...
+                    _ = autolocate_interval.tick() => {
+                        location = api.recoverable(find_ip_location).await?;
+                    }
+                    _ = interval.tick() => {
+                        let data = api
+                            .recoverable(|| async { provider.get_weather(location).await })
```
MaxVerevkin, line 227: `async { x.await }` is equivalent to `x`

src/blocks/weather.rs
```diff
@@ -185,24 +187,86 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
...
+                        let data = api
+                            .recoverable(|| provider.get_weather(location))
+                            .await?;
+                        widget.set_icon(data.icon.to_icon_str())?;
+                        widget.set_values(data.into_values());
+                        api.set_widget(&widget).await?;
+
+                        // both intervals should be reset after a manual sync
+                        autolocate_interval.reset();
+                        interval.reset();
```
MaxVerevkin, line 252: Good catch! I suspect some blocks mistakenly don't do this.

src/blocks/weather.rs
```diff
@@ -185,24 +187,86 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
...
 
-        widget.set_icon(data.icon.to_icon_str())?;
-        widget.set_values(data.into_values());
-        api.set_widget(&widget).await?;
+        if autolocate_interval == config.interval {
+            // In the case where `autolocate_interval` matches `interval` merge both actions.
+            loop {
+                let location: Option<Coordinates> = api.recoverable(find_ip_location).await?;
+                let data = api
+                    .recoverable(|| async { provider.get_weather(location).await })
```
MaxVerevkin, line 202:
```suggestion
                    .recoverable(|| provider.get_weather(location))
```

src/blocks/weather.rs
```diff
@@ -185,24 +187,86 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
...
+                        autolocate_interval.reset();
+                        interval.reset();
+                    },
+                }
+            }
+        }
+    } else {
+        loop {
+            let data = api
+                .recoverable(|| async { provider.get_weather(None).await })
```
MaxVerevkin, line 260: This one too.

Alextopher: These would make good clippy suggestions

MaxVerevkin:
Thanks!

PR #1604, opened 2022-08-28
formatting: add binary SI prefixes

### LINKED ISSUE #1493: Add SI unit prefixes to formatter (base 10; MB, GB, TB)
Hi, I recently added the disk_usage block to my config. My drive has quite exactly 1TB (1000GB, 10**12 Byte) storage. In base-10 this is 1TB, in base-2 this is 931GB.

The formatter seems to use base-2 calculations, while I would like to use base-10 for disk space. It just seems "prettier" for me seeing 1TB instead of 931GB.

Please add new Prefixes for base-10 calculations.

And btw. using "GB", "TB", etc. for base-2 is _legacy_ usage, and should be used for base-10 calculations. The proper names for base-2 nomenclature are "KiB", "MiB", "GiB", "TiB", etc., stemming from e.g. Kibibyte. You expect 1 _Kilo_-meter to be 1000 meters, as an example. The [SI prefixes](https://en.wikipedia.org/wiki/Metric_prefix) require Kilo to be 1000, Mega 1000*1000, etc.. See [Wikipedia](https://en.wikipedia.org/wiki/Byte#Multiple-byte_units) for more info.

Switching GB to base-10 and using GiB for base-2 would be a breaking change, but fixes it for good.

MaxVerevkin:
fixes #1493

(This commit contains some unrelated renames and refactors)

### Examples

`1i` is a special prefix which means "one but binary". `1i` is to `1` as `Ki` is to `K`.  This hack allows users to enable binary prefixes without setting `min_prefix`. If someone has a better idea, please share.

```toml
[[block]]
block = "disk_space"
format = "$available.eng(3,auto,1i)"
```

```toml
[[block]]
block = "net"
format = "$speed_down.eng(3,auto,Ki)$speed_up.eng(3,auto,Ki)"
```

### TODO
- [ ] Document this

### DIFF
--- a/src/blocks/prelude.rs
+++ b/src/blocks/prelude.rs
@@ -2,7 +2,7 @@ pub use super::{BlockEvent::*, CommonApi};
 
 pub use crate::click::MouseButton;
 pub use crate::errors::*;
-pub use crate::formatting::{config::Config as FormatConfig, value::Value};
+pub use crate::formatting::{config::Config as FormatConfig, value::Value, Values};
 pub use crate::util::{default, new_dbus_connection, new_system_dbus_connection};
 pub use crate::widget::{State, Widget};
 pub use crate::wrappers::{Seconds, ShellString};
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -164,7 +164,7 @@ struct WeatherResult {
 }
 
 impl WeatherResult {
-    fn into_values(self) -> HashMap<Cow<'static, str>, Value> {
+    fn into_values(self) -> Values {
         map! {
             "location" => Value::text(self.location),
             "temp" => Value::degrees(self.temp),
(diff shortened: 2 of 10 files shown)


c02y:
Tested it, works expected. 

Question:
What are `auto, 1i`? Using it without `auto` doesn't work, using something other than `1i` like `2i` is not working.

What about using only `3i`?

MaxVerevkin:
> Question:
> What are auto, 1i? Using it without auto doesn't work, using something other than 1i like 2i is not working.

See docs [here](https://greshake.github.io/i3status-rust/i3status_rs/formatting/index.html#eng---format-numbers-using-engineering-notation) and #1605

MaxVerevkin:
In this PR, binary SI prefixes are enabled if `min_prefix` is itself binary.

Issue #1582, opened 2022-07-31
Custom block spacing/padding issues


arathunku:
Hello there!

I've checked several issues regarding spacing but my search-fu is not good enough or there isn't an issue about it yet.

It looks like there's hard-coded empty space push:
https://github.com/greshake/i3status-rust/blob/e203adefbb68e39d81f7ddea135a7fcec58440ef/src/widget.rs

And then when icon is disabled:
https://github.com/greshake/i3status-rust/blob/e203adefbb68e39d81f7ddea135a7fcec58440ef/src/widget.rs#L125

It prevents the user of creating **very** minimal blocks, indistinguishable from tray icons. 

What I'm trying to do is to add few custom blocks (I love the new TOML configuration for click handling!), each outputting 1-3 emojis, displayed next to one another, each with its own cmd 

Currently, it looks like this:

![image](https://user-images.githubusercontent.com/749393/182024770-9a4d56e9-70e2-4ff4-a4aa-15902549bb10.png)

lock icon is custom block that may output 2 emojis, depending on the state

![image](https://user-images.githubusercontent.com/749393/182024937-babcbddc-af4e-453d-98a0-adcbdb1f7c66.png)

Notice the space, sometimes double, between blocks vs icons on the right from tray

What I'd ideally have looked like this:

![image](https://user-images.githubusercontent.com/749393/182024459-0d9e10bb-92b3-4172-9415-370c0374d4e5.png)

Or even less space if possible!!!

There's already formatting but I'm now sure how could it play nicely with custom blocks and those hard-coded(?) paddings within and around the block.

Looks like 0.3 is going to already have few breaking changes, so maybe it would be a good point to change some defaults / add configuration options. I'm happy to work on a PR to make those changes :)

Cheers,
Michal

ammgws:
So I believe the double space comes from the leading space added between `SEPARATOR` and `BLOCK_TEXT` entries as shown in the json below:

```json
...
{"full_text":"SEPARATOR",...},
{"full_text":" ","name":"7",...},
{"full_text":"BLOCK TEXT ","name":"7",...},
{"full_text":"SEPARATOR",...},
```
(removed `color`, `background`, `separator`, `separator_block_width`, `markup` for clarity)

So no matter what you do in your config it's impossible to get rid of that space. I don't remember if it was always like this, it seems like if that space is necessary then we should just include it in the default block format string so it can be overridden. @MaxVerevkin Do you know why we have the space hardcoded atm?

[...]

Issue #1554, opened 2022-07-07
Add option to hide blocks when format string is empty


gregdan3:
i3status is capable of hiding `disk` information when they are not within a threshold, usually by percentage of free space or number of gigabytes of free space. The same goes for `memory` information.

The current behavior of i3status-rust is to fail with ENOENT when a target disk is not mounted, and to always show the format string under all conditions otherwise, even if it is empty.

Example configuration in i3status:
```
disk "/" {
    prefix_type = "decimal"
    format = ""
    format_below_threshold = "/ %avail"
    format_not_mounted = ""
    low_threshold = 10
    threshold_type = "percentage_free"
}

disk "/home/" {
    prefix_type = "decimal"
    format = ""
    format_below_threshold = "⌂ %avail"
    format_not_mounted = ""
    low_threshold = 10
    threshold_type = "percentage_free"
}
memory {
    format = ""
    format_degraded = " %available"
    threshold_degraded = "30%"
}
```

This is one of the few things keeping me from switching to i3status-rust; the ability to hide blocks with empty format strings, and controlling the format string conditionally. This is possible for some blocks, but as far as I can find in the docs, not many of them.

ammgws:
So basically making the `custom` block's `hide_when_empty` a universal option?

gregdan3:
Yup, exactly that @ammgws

MaxVerevkin:
I think #1582 will cover this because if icons and spacing are handled manually by format strings, there will be no other meaning to `format = ""` other than "hide this block".

Issue #1605, opened 2022-08-29
Maybe reconsider formatter arguments?


MaxVerevkin:
Currently positional arguments are used for formatters (e.g. `$num.eng(3,auto,K)`), but I think it's better to discuss this one more time before 0.30 release.

Positional arguments generally work fine, especially for simple cases like `$num.eng(2)` or `$text.rot-str(20)` (but not `$text.str(20)`!). However, it gets a bit weird when all you want is to enable "binary prefixes", such as `"$available.eng(3,auto,1i)"`. Wouldn't `$available.eng(bin:true)` be more clear? But then again, `$num.eng(5,b,Ki)` is way less verbose than `$num.eng(width:5,unit:b,prefix:K,bin:true)`.

One solution can be to mix those two (python style). Most common usecases might only use first positional arguments, while others might require named parameters.

For example, let's take `eng` formatter. It may have the following arguments:
- `width` (number)
- `unit` (Unit)
- `min_prefix` (Prefix)
- `max_prefix` (Prefix)
- `insert_space` (boolean)
- `bin` (boolean)

The positional arguments might be defined in this order: `width`, `unit`, `min_prefix`.

Then this | would mean this
--------------|-------------------------
`.eng(3)` | `width = 3`
`.eng(3,bin:true)` | `width = 3`, `bin = true`
`.eng(min_prefix:K)` | `min_prefix = 3`

This allows for better extensibility in the future. New features can be activated via new named arguments instead of modifiers for existing arguments.

PR #1618, opened 2022-09-09
Drop hard-coded whitespace and render icon as a placeholder

arathunku:
I highly recommend NOT to review this PR with "whitespace" diff rendering disabled :smile: 

This adds $icon placeholder in all blocks with icon. 
It also removes hardcoded whitespace around rendered values and moves that into `format` configuration option.
This allows users to easily skip whitespace or disable $icon by modifying `format` option for a given block.

This should address:
 
- https://github.com/greshake/i3status-rust/issues/1582
- https://github.com/greshake/i3status-rust/issues/1554

The most problematic block for me is "custom", I wasn't sure how to address JSON format so I've added 2 placeholder values - `$text` and  `$short_text`, plus of course `$icon`. For non-json custom block there's just `$text` placeholder.

Personally, I use less than half of the blocks from i3status-rust because I had already written scripts for various things and just hook them in via `custom` block. After changes on this MR I'm very happy how it looks :smile: (obviously) 

Another problem I've encountered while working on this - it's very hard for me to test some blocks because I don't have the hardware/dependency so I'd appreciate tests in the wild with YOUR setup and config - things like kdeconnect, notmuch, battery, etc. Just comment here about any issue you see and I'll try to address the problem.

On the screenshot below marked in red you see 2(!) custom blocks rendering emojis and `format = "$text"` configuration - no blanks, no icons and dropped separators (that will be done on another PR) 

_________________

Future additional work: https://github.com/greshake/i3status-rust/pull/1618#issuecomment-1249491109


src/blocks/temperature.rs
```diff
@@ -18,7 +18,8 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `format` | A string to customise the output of this block. See below for available placeholders | <code>"$average avg, $max max&vert;"</code>
+//! `format` | A string to customise the output of this block when collapsed (default). See below for available placeholders | <code>" $icon "</code>
+//! `format_alt` | A string to customise the output of this block. See below for available placeholders | <code>" $icon {$average avg, $max max&vert;} "</code>
```
MaxVerevkin, line 22: I think this is a good idea, but lets do this in a separate PR

src/blocks/weather.rs
```diff
@@ -166,6 +166,7 @@ struct WeatherResult {
 impl WeatherResult {
     fn into_values(self) -> HashMap<Cow<'static, str>, Value> {
         map! {
+            "icon" => Value::icon(self.icon.to_icon_str().into()),
```
MaxVerevkin, line 169: You forgot the `api.get_icon()` here :)

src/blocks/custom.rs
```diff
@@ -141,7 +148,16 @@ async fn update_bar(
 
 pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
     let config = CustomConfig::deserialize(config).config_error()?;
-    let mut widget = api.new_widget();
+    let mut format_config = FormatConfig { full: config.format.full, short: config.format.short };
+    let format = if config.json {
+        if format_config.short.is_none() {
+            format_config.short = Some(" $icon $short_text ".parse()?);
+        };
+        format_config.with_default(" $icon $text ")?
```
MaxVerevkin, line 156: I think setting the default to `{ $icon|} $text {$short_text |}` would be sufficient?

src/blocks/custom.rs
```diff
@@ -116,20 +123,20 @@ async fn update_bar(
...
-                if let Some(short) = input.short_text {
-                    widget.set_texts(input.text, short);
-                } else {
-                    widget.set_text(input.text);
-                }
+                let short_text = input.short_text.unwrap_or_default();
+                widget.set_values(map!{
+                    "icon" => Value::icon(api.get_icon(&input.icon)?),
+                    "text" => Value::text(input.text),
+                    "short_text" => Value::text(short_text)
```
MaxVerevkin, line 131: Maybe `[if let Some(t) = short_text] "short_text" => Value::text(t),` (just not set `short_text` instead of setting it to an empty string)?

MaxVerevkin:
I think `pomodoro` block will need to have a format config too because currently there is not way to add spaces around it.

MaxVerevkin:
Apart from the comments above, this PR works for me :+1:

MaxVerevkin:
Please update these too: https://github.com/greshake/i3status-rust/blob/c4740bcfdc295df69bf0ced6b3445f6129a8d941/src/config.rs#L77-L84

src/widget.rs
```diff
@@ -136,13 +111,9 @@ impl Widget {
             template.short_text = "<span/>".into();
         }
 
-        let full_cnt = full.len();
-        parts.extend(full.into_iter().enumerate().map(|(i, w)| {
+        parts.extend(full.into_iter().enumerate().map(|(_i, w)| {
```
MaxVerevkin, line 114:
```suggestion
        parts.extend(full.into_iter().map(|w| {
```

src/blocks/custom.rs
arathunku: I don't think I'm able to maintain the same behavior as on master with any configuration but I like your suggestion much more than my conditions...

I'm using the following config for testing:
```toml
[[block]]
block = "custom"
json = true
command = "echo '{\"icon\": \"cpu\", \"text\": \"!\", \"short_text\": \"?\", \"state\": \"Critical\"}'"

[[block]]
block = "custom"
json = true
command = "echo '{\"icon\": \"cpu\", \"text\": \"#\", \"state\": \"Critical\"}'"

[[block]]
block = "custom"
command = "echo '--'"
```

On master this gives me:

```json
[
    {"full_text":"  "},
    {"full_text":"! ","short_text":"<span/>"},
    {"full_text":"<span/>","short_text":"? "},
    {"full_text":"  ","color":"#002B36FF"},
    {"full_text":"# "}
]
```

Testing it with `with_default("{ $icon|} $text {$short_text |}")?` and `[if let Some(t) = input.short_text] "short_text" => Value::text(t)` gives me:

```json
  [
    {"full_text":"  ! "},
    {"full_text":"? "},
    {"full_text":"  # "},
    {"full_text":"  "}
  ]
```

I think the best solution here would be to use `"{ $icon|} $text"` as default but in that case, we need to set default short formatting option when json output `short_text` field is present. I've tried to do that but I'm struggling with mutating `short` FormatConfig field inside `update_bar` without adding a `Copy` trait to FormatConfig :thinking:  :thinking:

arathunku:
I've pushed an update for `hueshift`, missing `get_icon` in `weather` and error formatting. I'll get back later(probably Monday) to `custom` and  'temperature` :smile:

src/blocks/custom.rs
MaxVerevkin: Yeah, you're right. Up until now there weren't a need to set default short text. But I think this should be good:

```rust
let mut widget = api.new_widget().with_format(
    config
        .format
        .with_default("{ $icon|} $text ")?
        .with_short_default("$short_text|")?
);
```

src/config.rs
```diff
@@ -67,18 +67,18 @@ pub struct Config {
...
+        Arc::new("{icon}".into())
     }
 
     fn default_double_click_delay() -> u64 {
         200
     }
 
     fn default_error_format() -> String {
-        "$short_error_message|X".into()
+        " $short_error_message ".into()
```
MaxVerevkin, line 78: `$short_error_message` is not guaranteed to be available

src/blocks/temperature.rs
arathunku: Ok, I'll hardcode collapsed `" $icon "` format as a default and submit this change on another PR to invert the values. I've noticed in other blocks, `format_alt` is usually expanded format.

MaxVerevkin: > I'll hardcode collapsed `" $icon "` format as a default 

Why though? I think this block can function as before (for now): when block is `collapsed`, `$average`, `$max` and `$min` are just unavailable.

src/blocks/custom.rs
arathunku: :thinking: :thinking: :thinking: 

I'm unsure about adding `with_short_default`. `with_default` is FormatConfig method and it returns `Result<Format>`. If I were to add  `with_short_default` and keep it on the same level, we'd have to change return value to `Self` but then it's also not pretty... I've also thought about changing the signature from `with_default(self, default_full: &str) -> Result<Format>` to `with_default(self, default_full: &str, default_short: Option<&str>) -> Result<Format>` and updating ALL but custom block to pass  'None` but decided against because it would make this PR much larger.

I think I've got it:

```toml
[[block]]
block = "custom"
json = true
command = "echo '{\"icon\": \"cpu\", \"text\": \"!\", \"short_text\": \"?\", \"state\": \"Critical\"}'"

[[block]]
block = "custom"
json = true
command = "echo '{\"icon\": \"cpu\", \"text\": \"#\", \"state\": \"Critical\"}'"

[[block]]
block = "custom"
command = "echo 'abc'"
```
Gives me:

```json
    { "full_text": "  ! ", "short_text": "<span/>", "name": "0"},
    { "full_text": "<span/>", "short_text": " ? ", "name": "0"},
    { "full_text": "  # ", "name": "1" },
    { "full_text": " abc ", "name": "2" }
```

It's not pretty but doesn't require that many changes on this PR.
```rust
    let mut format_config = FormatConfig { full: config.format.full, short: config.format.short };
    if format_config.short.is_none() {
        format_config.short = Some(" $short_text |".parse()?);
    };
    let mut widget = api.new_widget().with_format(
        format_config.with_default("{ $icon|} $text ")?
    );
```

MaxVerevkin: Maybe add `with_defaults(full: &str, short: &str)` (similar to `.set_texts()`)?

src/blocks/temperature.rs
arathunku: I'm trying to go from:

image

```
.with_default(" $icon ")?
```

To this:

image

(looks bad with background because I was compiling some other project, usually there's no bg :smile:)

that would be `format = "$icon"` (if format was for collapsed)

We don't have `with_icon` anymore after changes here. I've now realized I still need to revert the change to available values because I had exposed them in collapsed block :see_no_evil:

MaxVerevkin: I think you can do this now with `<normal format>|$icon`

arathunku: Ahhhhhhhh, got it! You're of course right :smile: 

```toml
[[block]]
block = "temperature"

[[block]]
block = "temperature"
collapsed = false
```

```json
  [
    { "full_text": "  "},
    { "full_text": " ", "name": "1" },
    { "full_text": " 34° avg, 49° max ", "name": "1"}
  ]
```

`" $icon{ $average avg, $max max|} "` default format option renders this

src/blocks/custom_dbus.rs
```diff
@@ -63,12 +63,6 @@ struct Block {
 
 #[dbus_interface(name = "rs.i3status.custom")]
 impl Block {
-    async fn set_icon(&mut self, icon: &str) -> fdo::Result<()> {
-        self.widget.set_icon(icon)?;
```
MaxVerevkin, line 67: I guess `custom_dbus` should also have a format option (like `custom`)?

MaxVerevkin:
`net` needs a `format_missing` or `format_unavailable` or something similar because currently it looks like this:

image

src/blocks/battery.rs
```diff
@@ -10,7 +10,7 @@
 //! `device` | The device in `/sys/class/power_supply/` to read from. When using UPower, this can also be `"DisplayDevice"`. Regular expressions can be used. | Any battery device
 //! `driver` | One of `"sysfs"`, `"apc_ups"`, or `"upower"` | `"sysfs"`
 //! `interval` | Update interval, in seconds. Only relevant for `driver = "sysfs"` \|\| "apc_ups"`. | `10`
-//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>"$percentage&vert;"</code>
+//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>" $icon $percentage&vert; "</code>
 //! `full_format` | Same as `format` but for when the battery is full | `""`
 //! `empty_format` | Same as `format` but for when the battery is empty | `""`
 //! `hide_missing` | Completely hide this block if the battery cannot be found. | `false`
```
MaxVerevkin, line 16: Not sure about `hide_missing`, but `hide_full` seems redundant now.

arathunku: `hide_missing` needs to be present because `$icon` value is still provided and the block would be rendered. I'll drop hide_full and noticed another bug, `$icon` needs be set in default values for `full_format` and `empty_format`

arathunku: Regarding `hide_missing`. What do you think of updating both battery and net blocks to similar configuration - `format_missing` instead of `hide_missing`? This would address https://github.com/greshake/i3status-rust/pull/1618#issuecomment-1242766382

MaxVerevkin: I think it's a good idea.

arathunku: Added in https://github.com/greshake/i3status-rust/pull/1618/commits/56828b57776da45a19afcd550f796642c09d7f8b

src/blocks/custom_dbus.rs
arathunku: I've added it named as "format" as in other blocks.

src/blocks/custom.rs
arathunku: Done, added in https://github.com/greshake/i3status-rust/pull/1618/commits/249bc4030017094767b3c39a619d179b78b94ac1

arathunku:
I'll update laters `NEWS` with listed breaking changes :see_no_evil: :see_no_evil: 0.30.0 looks exciting :smile:

src/formatting/config.rs
```diff
@@ -19,21 +19,28 @@ pub struct DummyConfig {
 
 impl Config {
     pub fn with_default(self, default_full: &str) -> Result<Format> {
+        self.with_defaults(default_full, "")
```
MaxVerevkin, line 22: ~~Setting short text to `""` by default is wrong: it will result in blocks being hidden when the bar decides without user's permission.~~

While I was writing that comment I realized two things:
- `short_text` can be set to `""` and it is equvavent to not setting it all all, thus (at the moment) `Format::short` doesn't need to be an `Option`.
- If the user is willing to hide the block to save some space, short text needs to be set to something weird, like `"<span/>"`. That's not ideal.

src/blocks/custom.rs
```diff
@@ -147,12 +147,8 @@ async fn update_bar(
 
 pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
     let config = CustomConfig::deserialize(config).config_error()?;
-    let format_config = FormatConfig {
-        full: config.format.full,
-        short: config.format.short.or(Some(" $short_text |".parse()?))
-    };
     let mut widget = api.new_widget().with_format(
-        format_config.with_default("{ $icon|} $text ")?
+        config.format.with_defaults("{ $icon|} $text ", "{ $icon|}{ $short_text|} ")?
```
MaxVerevkin, line 151: Event if `$short_text` is not available, this string will evaluate to something non-empty, thus the block will be hide-able. So something like this is more correct: `"{ $icon|} $short_text `.

MaxVerevkin:
`bluetooth` doesn't need `hide_disconnected` anymore :tada:

src/blocks/net.rs
```diff
@@ -99,18 +102,14 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
...
-                }
+                widget.set_format(format_missing.clone());
+                api.set_widget(&widget).await?;
             }
             Some(device) if !device.iface.is_up => {
                 if config.hide_inactive {
                     api.hide().await?;
                 } else {
-                    widget.set_text("×".to_string());
+                    widget.set_format(format_missing.clone());
```
MaxVerevkin, line 112: Since both (missing and inactive) cases use `format_missing`, I think it's safe to remove `hide_inactive` too? But I'm not sure if the distinction between these two states has its usecases. If it does, then a new format option is needed :laughing: Otherwise, a better name for `format_missing` is needed.

src/blocks/backlight.rs
```diff
@@ -14,7 +14,7 @@
 //! Key | Values | Default
 //! ----|--------|--------
 //! `device` | The `/sys/class/backlight` device to read brightness information from.  When there is no `device` specified, this block will display information from the first device found in the `/sys/class/backlight` directory. If you only have one display, this approach should find it correctly.| Default device
-//! `format` | A string to customise the output of this block. See below for available placeholders. | `"$brightness"`
+//! `format` | A string to customise the output of this block. See below for available placeholders. | `" $format $brightness "`
```
MaxVerevkin, line 17:
```suggestion
//! `format` | A string to customise the output of this block. See below for available placeholders. | `" $icon $brightness "`
```

src/blocks/battery.rs
```diff
@@ -10,11 +10,10 @@
...
 //! `interval` | Update interval, in seconds. Only relevant for `driver = "sysfs"` \|\| "apc_ups"`. | `10`
-//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>"$percentage&vert;"</code>
-//! `full_format` | Same as `format` but for when the battery is full | `""`
-//! `empty_format` | Same as `format` but for when the battery is empty | `""`
-//! `hide_missing` | Completely hide this block if the battery cannot be found. | `false`
-//! `hide_full` | Hide the block if battery is full | `false`
+//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>" $icon $percentage&vert; "</code>
+//! `full_format` | Same as `format` but for when the battery is full | `" $icon "`
+//! `empty_format` | Same as `format` but for when the battery is empty | `" $icon "`
+//! `format_missing` | Same as `format` if the battery cannot be found. | `" $icon × "`
```
MaxVerevkin, line 16:
```suggestion
//! `format_missing` | Same as `format` if the battery cannot be found. | `" $icon "`
```

MaxVerevkin, line 16: The naming seems inconsistent, we have `{full,empty}_format` so `missing_format` might be better.

MaxVerevkin:
If #1621 gets merged, would it mean that blocks that set `$icon` to a static icon can drop it? I am not sure yet how usable it would be.

src/blocks/battery.rs
arathunku: We also have `format_alt` :smile:  :see_no_evil: I'll change it to `missing_format`

arathunku:
From my understanding of https://github.com/greshake/i3status-rust/pull/1621, that would be possible. There're some blocks(apt, cpu, disk_space, dnf, docker, github, and more) where it would be helpful.

src/blocks/battery.rs
```diff
@@ -10,11 +10,10 @@
 //! `device` | The device in `/sys/class/power_supply/` to read from. When using UPower, this can also be `"DisplayDevice"`. Regular expressions can be used. | Any battery device
 //! `driver` | One of `"sysfs"`, `"apc_ups"`, or `"upower"` | `"sysfs"`
 //! `interval` | Update interval, in seconds. Only relevant for `driver = "sysfs"` \|\| "apc_ups"`. | `10`
-//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>"$percentage&vert;"</code>
-//! `full_format` | Same as `format` but for when the battery is full | `""`
-//! `empty_format` | Same as `format` but for when the battery is empty | `""`
-//! `hide_missing` | Completely hide this block if the battery cannot be found. | `false`
-//! `hide_full` | Hide the block if battery is full | `false`
+//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>" $icon $percentage&vert; "</code>
```
MaxVerevkin, line 13:
```suggestion
//! `format` | A string to customise the output of this block. See below for available placeholders. | `" $icon $percentage "`
```

src/blocks/bluetooth.rs
```diff
@@ -32,7 +32,7 @@
 //! [[block]]
 //! block = "bluetooth"
 //! mac = "00:18:09:92:1B:BA"
-//! hide_disconnected = true
+//! disconnected_format = ""
 //! format = ""
```
MaxVerevkin, line 36:
```suggestion
//! format = " $icon "
```

src/blocks/bluetooth.rs
```diff
@@ -75,44 +75,36 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
...
+                    "name" => Value::text(device.name().await?),
+                    "available" => Value::flag(),
+                };
+                device
+                    .percentage()
+                    .await
+                    .map(|p| values.insert("percentage".into(), Value::percents(p)));
+                if connected {
+                    widget.state = State::Good;
+                    values.insert("connected".into(), Value::flag());
```
MaxVerevkin, line 89: `$connected` is now redundant

src/blocks/bluetooth.rs
```diff
@@ -75,44 +75,36 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
...
-                        widget.state = State::Idle;
-                    }
-                    widget.set_icon(device.icon().await?)?;
-                    widget.set_values(values);
-
-                    api.set_widget(&widget).await?;
+                let mut values = map! {
+                    "icon" => Value::icon(api.get_icon(device.icon().await?)?),
+                    "name" => Value::text(device.name().await?),
+                    "available" => Value::flag(),
```
MaxVerevkin, line 81: Not sure if this flag makes sense anymore

src/blocks/bluetooth.rs
```diff
@@ -14,8 +14,8 @@
 //! ----|--------|--------
 //! `mac` | MAC address of the Bluetooth device | **Required**
 //! `adapter_mac` | MAC Address of the Bluetooth adapter (in case your device was connected to multiple currently available adapters) | `None`
-//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>"$name{ $percentage&vert;}&vert;Unavailable"</code>
-//! `hide_disconnected` | Whether to hide the block when disconnected | `false`
+//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>" $icon $name{ $percentage&vert;}&vert;"</code>
```
MaxVerevkin, line 17:
```suggestion
//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>" $icon $name{ $percentage&vert;} "</code>
```

src/blocks/custom_dbus.rs
```diff
@@ -95,7 +117,11 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
     // will get blocked while trying to send a new message.
     api.event_receiver.close();
 
-    let path = CustomDBusConfig::deserialize(config).config_error()?.path;
+    let config = CustomDBusConfig::deserialize(config).config_error()?;
+    let widget = api.new_widget().with_format(
+        config.format.with_defaults("{ $icon|}{ $text|} ", "{ $icon|}{ $short_text|} ")?
```
MaxVerevkin, line 122:
```suggestion
        config.format.with_defaults("{ $icon|}{ $text|} ", "{ $icon|} $short_text |")?
```

src/blocks/focused_window.rs
```diff
@@ -7,7 +7,7 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>"$title.str(0,21)&vert;"</code>
+//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>" {$title.str(0,21)&vert;} "</code>
 //! `autohide` | Whether to hide the block when no title is available | `true`
```
MaxVerevkin, line 11: `autohide` can be removed :tada:

src/blocks/focused_window.rs
```diff
@@ -7,7 +7,7 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>"$title.str(0,21)&vert;"</code>
+//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>" {$title.str(0,21)&vert;} "</code>
```
MaxVerevkin, line 10:
```suggestion
//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>" $title.str(0,21) &vert;"</code>
```

MaxVerevkin (commented):
I hope this is not too much comments at once :)

src/blocks/bluetooth.rs
arathunku: It's still needed. I wasn't aware how to fully nicely test this but I've found this config: https://sourcegraph.com/github.com/ammgws/dotfiles/-/blob/i3status-rs/.config/i3status-rust/config.toml?L205&subtree=true using it :smile: 

```toml
format = "$connected$available{PS5}|$available{D/C}|N/A"
```

On master I got:

```json
[
    // unknown
    {"full_text":"  ","name":"0"},
    {"full_text":"N/A ","name":"0"},
    // connected
    {"full_text":"  ","name":"1"},
    {"full_text":"PS5 ","name":"1"}
    // known, diconnected
    {"full_text":"  ","name":"1"},
    {"full_text":"D/C","name":"1"}
  ]
```

To fully repro the same behaviour I've used:
```toml
format = " $icon PS5 "
disconnected_format = " $icon$available {D/C} | $icon N/A "
```
This gives me:

```json
  [

    // unknown
    {"full_text":"  N/A ","name":"0"},
    // connected
    {"full_text":"  PS5 ", "name": 1}
    // known, disconnected
    {"full_text":"  ","name":"1"},
    {"full_text":"D/C ","name":"1"}
  ]
```

arathunku:
> I hope this is not too much comments at once :)

@MaxVerevkin, not at all. I'm greatful for your extremly fast feedback on this PR. To be honest, I wasn't expecting this change to have such drastic impact on all the other options related to just... hiding blocks! :smile: I'm still unsure if this the rabbit hole I should have jumped into on this single PR instead of another one, but looks like we're near end here. Once again, thanks a lot for help with pushing this PR.

src/formatting/config.rs
arathunku: I was thinking that user could set `short` to null/nil/undefined value to just "disable" it, but it's not possible with TOML.

> If the user is willing to hide the block to save some space, short text needs to be set to something weird, like "<span/>". That's not ideal.

I guess it would have to be something like `"$hidden"` value that's basically hard-coded to `"<span/>"` or another option for format
```toml
[block.format]
short_hidden = true
```

src/blocks/bluetooth.rs
MaxVerevkin: Yeah, but I though maybe unavailable can be treated just like disconnected. Similar how `net` treats missing and down interfaces the same way.

arathunku: I think we have to agree on the same convention and naming :smile: `disconnected_format` vs `missing_format` naming wise and then if `$available` should be added when the device exists vs adding another format option. From my understanding, that's the only way to make the old behavior possible...

MaxVerevkin: > `disconnected_format` vs `missing_format`

"missing" is shorter but not quite correct :laughing: "disconnected" seems better to me.

> `format = "$connected$available{PS5}|$available{D/C}|N/A"`

@ammgws Do you rely on the `N/A` part of this format? What are you thoughts?

MaxVerevkin:
You can get a list of some of the blocks with outdated docs by running `rg 'format = "[^ "]' src/blocks/ `

src/blocks/bluetooth.rs
ammgws: I think I had that in since at one stage my USB bluetooth adapter was at times not plugged in and I wanted to distinguish between the device being completely unavailable (nothing on dbus) vs the device actually just being in a disconnected state.

MaxVerevkin: Okay, so it has it's usecases. Perhaps it's better to add `unavailable_format`? I don't know how I feel about this whole "flags" idea.

arathunku: Would it be ok for you to tackle this on another PR and leave the flag for now as it is? I think going 100% with "foobar_format" is interesting and would be more explicit.

MaxVerevkin: > Would it be ok for you to tackle this on another PR and leave the flag for now as it is?

Yeah, sure!

arathunku:
Thanks for the regexp! I've also used:

```
rg 'with_defaults?\((.*)\)' src/blocks/`
rg '`format`.*(`".*`)' src/blocks/ -r '$1'
```

arathunku:
Adding top lvl comment about follow-up work based on discussion in this PR, might get easily lost from resolved threads

  - flags vs format options - https://github.com/greshake/i3status-rust/pull/1618#discussion_r969907006
  - hiding short block - https://github.com/greshake/i3status-rust/pull/1618#discussion_r968688067

MaxVerevkin:
> To be honest, I wasn't expecting this change to have such drastic impact on all the other options related to just... hiding blocks!

Yeah... By the way `apt`, `pacman` and `dnf` no longer need `hide_when_uptodate` :laughing:

src/blocks/toggle.rs
```diff
@@ -29,7 +29,6 @@
 //! ```toml
 //! [[block]]
 //! block = "toggle"
-//! text = "4k"
```
MaxVerevkin, line 32: `format = " $icon 4k "`

src/blocks/speedtest.rs
```diff
@@ -44,9 +44,9 @@ struct SpeedtestConfig {
 
 pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
     let config = SpeedtestConfig::deserialize(config).config_error()?;
-    let mut widget = api
-        .new_widget()
-        .with_format(config.format.with_default("$ping$speed_down$speed_up")?);
+    let mut widget = api.
+        new_widget().
+        with_format(config.format.with_default(" $icon $ping$speed_down$speed_up ")?);
```
MaxVerevkin, line 49: You forgot to actually provide `$icon` placeholder.

MaxVerevkin:
Also I think each block that has `$icon` placeholder should explicitly list it in it's docs (like every other placeholder).

MaxVerevkin:
Please update `examples/config.toml`

MaxVerevkin:
I merged master because it had a bug-fix I depend on, I hope it's OK.

arathunku:
Thank you, totally OK! This week is a bit crazy :see_no_evil: I want to address the rest of the comments over the weekend

src/blocks/uptime.rs
```diff
@@ -29,14 +30,16 @@ use tokio::fs::read_to_string;
...
     #[default(60.into())]
     interval: Seconds,
 }
 
 pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
     let config = UptimeConfig::deserialize(config).config_error()?;
-    let mut widget = api.new_widget().with_icon("uptime")?;
+    let mut widget = api.new_widget().with_format(config.format.with_default(" $icon ")?);
 
+    widget.set_values(map!("icon" => Value::icon(api.get_icon("uptime")?)));
```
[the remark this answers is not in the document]
arathunku: I've added `text` value that's equal to current uptime so now format is ` $icon $text `. I was thinking about "$uptime" name but wasn't sure which one will be clearer for this particular block. Anyway, the output now is not missing the icon! :smile:

MaxVerevkin:
`Widget::set_texts()` is not used anymore and can be removed.

MaxVerevkin:
`watson` has broken spacing: 
image

src/blocks/battery.rs
```diff
@@ -24,6 +23,7 @@
 //!
 //! Placeholder  | Value                                                                   | Type              | Unit
 //! -------------|-------------------------------------------------------------------------|-------------------|-----
+//! `icon`       | A static icon                                                           | Icon   | -
```
MaxVerevkin, line 26: This looks confusing and wrong: this icon is not "static" but it represents the state of the battery. Same applies for some other blocks.

arathunku: Uhh, I understood the description differently, more along the lines "a hard reference in code to the icon, see "used icons"" below. I'll improve it

src/blocks/time.rs
```diff
@@ -79,11 +83,14 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
         let full_time = get_time(format, timezone, locale);
         let short_time = format_short.map(|f| get_time(f, timezone, locale));
 
-        if let Some(short) = short_time {
-            widget.set_texts(full_time, short);
+        let format = if let Some(short) = short_time {
+          FormatConfig { full: Some(full_time.parse()?), short: Some(short.parse()?) }.with_default("")?
```
MaxVerevkin, line 87: Since `time` is already special in the way it handles formatting, it might be easier to just substitute `$icon` (or `%icon` to be consistent with other "placeholders"?) in in the beginning?

src/blocks/focused_window.rs
```diff
@@ -23,9 +22,16 @@
 //! [[block]]
 //! block = "focused_window"
 //! [block.format]
-//! full = "$title.rot-str(15)|"
-//! short = "$title.rot-str(10)|"
+//! full = "{ $title.str(0,15) |}"
+//! short = "{ $title.str(0,10) |}"
```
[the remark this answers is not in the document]
arathunku: :thinking: is that correct? I'm getting "Failed to render full text" for windows without tittle otherwise.

src/blocks/notmuch.rs
```diff
@@ -13,6 +13,7 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
+//! `format` | A string to customise the output of this block. See below for available placeholders. | `" $icon{ $name|} $count "`
```
arathunku, line 16: Now name configuration option could be dropped? Format can be used.

src/blocks/time.rs
arathunku: I did it like that and left it after "^icon_" support for 2 reasons:

1. `$icon` for consistency with all other blocks, this block is unique enough. I'd probably even prefer all chrono placeholders as "$" than the reverse
2.  Format supports "^icon_", with "set_text" it will no longer be possible, but this most likely won't be that useful here anyway :smile: 

When you say "substitue", you mean to just "get_time(format, timezone, locale).replace("$icon", ...)" and leave it with `set_text`, or something else?

src/blocks/notmuch.rs
MaxVerevkin: Yes :+1:

src/blocks/time.rs
MaxVerevkin: I meant something like

```rust
let format = config.format.full.replace("%icon", api.get_icon("time")?);
// same for short

loop {
    // main loop
}
```

It just feels weird to parse format string on every iteration. But I'm not completely against it either.

src/blocks/focused_window.rs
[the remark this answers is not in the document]
MaxVerevkin: Hmm, works for me™?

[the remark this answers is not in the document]
MaxVerevkin: There is no reason it should fail w/o them

src/blocks/speedtest.rs
```diff
@@ -23,7 +23,7 @@
 //! [[block]]
 //! block = "speedtest"
 //! interval = 1800
-//! format = "^icon_ping$ping"
+//! format = " ^icon_ping$ping "
```
MaxVerevkin, line 26:
```suggestion
//! format = " ^icon_ping $ping "
```

src/blocks/music.rs
```diff
@@ -131,16 +132,18 @@ struct OwnerChange {
...
     let config = MusicConfig::deserialize(config).config_error()?;
     let dbus_conn = new_dbus_connection().await?;
-    let mut widget = api
-        .new_widget()
-        .with_icon("music")?
-        .with_format(config.format.with_default("$combo.rot-str() $play|")?);
+    let mut widget = api.new_widget().with_format(
+        config
+            .format
+            .with_default(" $icon {$combo.rot-str() $play|} ")?,
```
MaxVerevkin, line 138: This is out of sync with the docs

src/blocks/focused_window.rs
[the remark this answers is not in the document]
arathunku: I must have missed "|" :see_no_evil: you're absolutely right!

src/blocks/time.rs
arathunku: I've simplified the code a little bit but left the parsing in the inner loop because tbh both options sound suboptimal and something like an exception.  

Maybe that's something for https://github.com/greshake/i3status-rust/issues/1605 to consider?

It's a lot of more verbose but aligns with other blocks and placeholders. 

`$timestamp.datetime(" %a %d/%m %R ")` and add Value::datetime + formatter with chrono under the hood.

MaxVerevkin: That might work too :)

So let's leave the code as is and deal with this later (can you update your comment on the follow-up work please?)

arathunku: Done! https://github.com/greshake/i3status-rust/pull/1618#issuecomment-1249491109

src/blocks/music.rs
```diff
@@ -18,7 +18,7 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>"$combo.rot-str() $play&vert;"</code>
+//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>" $icon {$combo.rot-str() $play|} "</code>
```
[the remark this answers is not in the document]
arathunku: I'll fix it here and in other blocks!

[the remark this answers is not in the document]
MaxVerevkin: Note: in order to use `&vert;` the code must be between `<code>` and `</code>` instead  of usual "`"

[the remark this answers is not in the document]
MaxVerevkin: Markdown is weird...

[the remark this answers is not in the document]
arathunku: #TIL should be fixed! `rg "\|\s+`.*&vert;" src/` and `rg "//! `format`.*\|.*\|\}" src/` to the rescue :joy:

src/blocks/music.rs
MaxVerevkin: Now it has double spaces...

MaxVerevkin:
```suggestion
            .with_default(" $icon{ $combo.rot-str() $play|} ")?,
```

or

```suggestion
            .with_default(" $icon {$combo.rot-str() $play |}")?,
```

src/blocks/notify.rs
```diff
@@ -54,21 +55,23 @@ pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
...
+        );
 
-        widget.set_icon(if is_paused { ICON_OFF } else { ICON_ON })?;
-        widget.set_values(if is_paused {
-            map!("paused" => Value::flag())
-        } else {
-            default()
-        });
+        if is_paused {
+            values.insert("paused".into(), Value::flag());
```
MaxVerevkin, line 71: Just a style thing, but this could be written as

```rust
let values = map!(
    "icon" => Value::icon(api.get_icon(if is_paused { ICON_OFF } else { ICON_ON })?),
    [if is_paused] "paused" => Value::flag(),
);
```

src/blocks/custom_dbus.rs
```diff
@@ -12,6 +12,16 @@
...
+//! # Configuration
+//!
+//! Key | Values | Default
+//! ----|--------|--------
+//! `format` | A string to customise the output of this block. | <code>"{ $icon&vert;}{ $text&vert;} "</code>
+//!
+//! Placeholder  | Value                                  | Type   | Unit
+//! -------------|----------------------------------------|--------|---------------
+//! `icon`       | Value of icon set via `SetIcon`        | Icon   | -
+//!
```
MaxVerevkin, line 24: `text` and `short_text` are missing

src/blocks/custom_dbus.rs
```diff
@@ -53,24 +63,40 @@ const DBUS_NAME: &str = "rs.i3status";
...
+        [if let Some(text) = &block.text] "text" => Value::text(text.to_string()),
+        [if let Some(short_text) = &block.short_text] "short_text" => Value::text(short_text.to_string()),
+    })
 }
 
 #[dbus_interface(name = "rs.i3status.custom")]
 impl Block {
     async fn set_icon(&mut self, icon: &str) -> fdo::Result<()> {
-        self.widget.set_icon(icon)?;
+        self.icon = Some(icon.to_string());
```
MaxVerevkin, line 90: The icon should be unset if `icon == ""`

src/blocks/custom.rs
```diff
@@ -116,20 +124,19 @@ async fn update_bar(
...
-                widget.set_icon(&input.icon)?;
-                widget.state = input.state;
                 text_empty = input.text.is_empty();
-                if let Some(short) = input.short_text {
-                    widget.set_texts(input.text, short);
-                } else {
-                    widget.set_text(input.text);
-                }
+                widget.set_values(map!{
+                    "icon" => Value::icon(api.get_icon(&input.icon)?),
```
MaxVerevkin, line 129: Icon should be unset if `icon == ""`

MaxVerevkin:
I think we're close :)

NEWS.md
```diff
@@ -61,12 +61,37 @@ This is a major release which rewrites the core code to be asynchronous.
...
+  format = "cpu:{utilization}"
+  ```
+  needs to be changed to:
+  ```toml
+  [[block]]
+  block = "cpu"
+  format = " $icon $utilization "
+  ```
+
+- **battery**: `hide_missing` option is replaced with `missing_format`. You can set `missing_format = ""` to maintain the behavior
```
arathunku, line 79: Leaving a comment for myself, naive but quick approach of trying to find changes in fields

```
git diff -w --diff-filter=MA -U0 master | rg "^[-+]    \w+:\s"
```

src/blocks/music.rs
arathunku: I've applied the suggestion and fix rest of format docs :|

https://github.com/greshake/i3status-rust/pull/1618/commits/ed585d04772be461639b62769352deb6614b3209

src/blocks/custom.rs
```diff
@@ -1,17 +1,18 @@
...
-//! `state` is optional, it may be Idle, Info, Good, Warning, Critical (default Idle)  
+//! `icon` is optional (default "")
+//! `state` is optional, it may be Idle, Info, Good, Warning, Critical (default Idle)
 //! `short_text` is optional.
 //!
 //! # Configuration
 //!
 //! Key | Values | Default
 //! ----|--------|--------
+//! `format` | A string to customise the output of this block. See below for available placeholders. | `" $icon $text "`
```
MaxVerevkin, line 15:
```suggestion
//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>"{ $icon&vert} $text "</code>
```

MaxVerevkin (approved):

ammgws (approved):

arathunku:
Wooooot! Merged! Thank you for your patience @MaxVerevkin and all the suggestions and fixes. Thank you @ammgws as well, and sorry for breaking your config a little :see_no_evil:

martinpelikan:
I'm a bit late, but wanted to drop by and say thanks for your work on this one. I (accidentally) recompiled with the latest changes, and after fixing some other config breaks, ended up with a bar and config that look much closer to what I wanted. Thanks again!

PR #1619, opened 2022-09-10
docs: Add `ip` and `ipv6` to the `net` block docs

finga:
Add the `ip` and `ipv6` placeholders to the documentation of the `net` block.


MaxVerevkin:
Thanks!

PR #1621, opened 2022-09-13
formatting: embed icons using ^icon syntax

### LINKED ISSUE #1527: Allow theme overrides to link to existing colors
For example in `kdeconnect` block could then do:

```toml
[[block]]
block = "kdeconnect"
format = "$bat_charge $connected{OK}|^phone_disconnected"
```
Where `phone_disconnected` is an icon from the icon file, and `^` is just example syntax for specifying icons

MaxVerevkin:
Resolves #1527

Will work nicely with #1618

Not documented yet. See `net` and `speedtest` for examples.

### DIFF
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -259,7 +259,9 @@ impl CommonApi {
     }
 
     pub fn get_icon(&self, icon: &str) -> Result<String> {
-        self.shared_config.get_icon(icon)
+        self.shared_config
+            .get_icon(icon)
+            .or_error(|| format!("Icon '{}' not found", icon))
     }
 
     /// Repeatedly call provided async function until it succeeds.
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -7,7 +7,7 @@
 //! Key | Values | Default
 //! ----|--------|--------
 //! `device` | Network interface to monitor (as specified in `/sys/class/net/`). Supports regex. | If not set, device will be automatically selected every `interval`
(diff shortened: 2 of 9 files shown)


ammgws:
Do you think `$icon_<iconname>`instead of `^<iconname>` would be more readable?

MaxVerevkin:
> Do you think `$icon_<iconname>`instead of `^<iconname>` would be more readable?

Hmm, maybe. But then it would look like a variable... And as if it could be formatted with some formatter, which it can't. But yeah, `^` seems kinda arbitrary.

ammgws:
> it would look like a variable... And as if it could be formatted with some formatter,

Hmm I guess `^icon_<iconname>` could work then, just seems better to have the prefix for readability.

arathunku:
While working on https://github.com/greshake/i3status-rust/pull/1618 I was actually wondering if it would be possible to have something like `" $icon.music " ` or `" $icon.set("music") "` (this one is annoying because double `"` quotes) but it would change the meaning of `$` and require special casing this in template.rs :sob: 

Going with `^`/`^icon` is really clever. :clap: 

In Elixir there's thing called [sigils](https://elixir-lang.org/getting-started/sigils.html), maybe this could be better for feature extension? `~i()` for icon

MaxVerevkin:
> Hmm I guess `^icon_<iconname>` could work then, just seems better to have the prefix for readability.

Done.

> `~i()` for icon

I think for someone who've never seen it, this is just as cryptic as `^<icon>`...

ammgws (approved):

PR #1626, opened 2022-09-19
fix: make example config work again

just1602:
The example config was giving errors, here's my attempt to fix it with my 15 min experience with i3status-rust.


MaxVerevkin:
This example config targets latest master. Example config for the latest released version (v22.0) is [here](https://github.com/greshake/i3status-rust/blob/v0.22.0/examples/config.toml).

Issue #1635, opened 2022-10-05
Output variable value to notify-send


madhur:
In the net block, I want to popup the current IP address on click, instead of displaying it in block all the time since it consumes real estate.

I tried following but its not working. I tried escaping `$ip` as well. The plain notify send works without any variable interpolation. Is there any solution to this?

Running i3status-rs 0.30.0 (commit 51fd9e5d 2022-10-02)

```
[[block]]
block = "net"
device = "enp5s0"
format = "$icon {$ssid |}{$ip ^icon_net_up $speed_up.eng(3,B,K) ^icon_net_down $speed_down.eng(3,B,K)|Wired connection} "
interval = 5
[[block.click]]
button="left"
cmd = "notify-send -u low -t 5000 $ip"
```

MaxVerevkin:
> I want to popup the current IP address on click

Currently it is not possible to access placeholders such as `$ip` inside of `cmd`.

> instead of displaying it in block all the time since it consumes real estate.

An alternative solution for this usecase might be to set `format_alt = " $icon $ip "`.

madhur:
Got it. That really helps.
I was trying to apply the same `format_alt` to `disk` and `memory` blocks. It seems they are not supported in those blocks.
![image](https://user-images.githubusercontent.com/509431/194713055-7de56663-2a41-4f19-9ccd-03a64a2e591e.png)

Any reason why? Should it not be consistent across the blocks.

MaxVerevkin:
> It seems they are not supported in those blocks. Any reason why?

No one implemented it yet :) If you wish, you can send a PR.

MaxVerevkin:
For the `memory` block: it already has `format_mem` and `format_swap`, so IDK how `format_alt` would fit in.

madhur:
For `disk_space` , raised the PR https://github.com/greshake/i3status-rust/pull/1636

MaxVerevkin:
About the memory block: `format_mem` and `format_swap` already behave as `format` and `format_alt`. So you can have

```toml
[[block]]
block = "memory"
format_mem = " ^icon_memory_mem $mem_used_percents.eng(1) "
format_swap = " ^icon_memory_mem $mem_used / $mem_total "
```

[...]

PR #1636, opened 2022-10-08
Add format_alt option for disk_space block

madhur:
Adds the `format_alt` option for `disk_space` block.

As per discussion in #1635


MaxVerevkin:
Thanks!

PR #1643, opened 2022-10-15
Remove color space

### LINKED ISSUE #1567: remove color_space dependency
>  hsv<->rgb conversion is simple and short enough to just copy it instead of adding yet another dependency

bim9262:
Fixes #1567
Supersedes #1569

### DIFF
--- a/src/themes/color.rs
+++ b/src/themes/color.rs
@@ -6,70 +6,80 @@ use std::fmt;
 use std::ops::Add;
 use std::str::FromStr;
 
-/// An RGB color (red, green, blue).
+/// An RGBA color (red, green, blue, alpha).
 #[derive(Copy, Clone, Debug, Default)]
-pub struct Rgb {
+pub struct Rgba {
     pub r: f64,
     pub g: f64,
     pub b: f64,
+    pub a: u8,
 }
 
-impl Rgb {
-    /// Create a new RGB color.
+impl Rgba {
+    /// Create a new RGBA color.
     ///
     /// `r`: red component (0 to 255).
     ///
     /// `g`: green component (0 to 255).
     ///
     /// `b`: blue component (0 to 255).
+    ///
+    /// `a`: alpha component (0 to 100).
     #[inline]
-    pub fn new(r: f64, g: f64, b: f64) -> Self {
-        Self { r, g, b }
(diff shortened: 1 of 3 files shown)


src/themes/color.rs
```diff
@@ -6,70 +6,80 @@ use std::fmt;
...
+impl Rgba {
+    /// Create a new RGBA color.
     ///
     /// `r`: red component (0 to 255).
     ///
     /// `g`: green component (0 to 255).
     ///
     /// `b`: blue component (0 to 255).
+    ///
+    /// `a`: alpha component (0 to 100).
```
MaxVerevkin, line 27: I think you meant "0 to 255"

src/themes/color.rs
```diff
@@ -6,70 +6,80 @@ use std::fmt;
 use std::ops::Add;
 use std::str::FromStr;
 
-/// An RGB color (red, green, blue).
+/// An RGBA color (red, green, blue, alpha).
 #[derive(Copy, Clone, Debug, Default)]
-pub struct Rgb {
+pub struct Rgba {
     pub r: f64,
```
MaxVerevkin, line 12: Do we need RGB components to be floating point numbers?

bim9262: We don't have to use `f64`s for `Rgba`, but it does require more casting for `Rgba` <-> `Hsva`. I'll put add an additional commit for review.

src/themes/color.rs
```diff
@@ -1,19 +1,193 @@
...
+    /// Create a new RGBA color from the `hex` value.
+    ///
+    /// ```let cyan = Rgba::from_hex(0xffffff);```
+    pub fn from_hex(hex: u32) -> Self {
+        Self {
+            r: (hex >> 24) as u8,
+            g: (hex >> 16) as u8,
+            b: (hex >> 8) as u8,
+            a: hex as u8,
+        }
```
MaxVerevkin, line 42:
```suggestion
        let [r, g, b, a] = hex.to_be_bytes();
        Self { r, g, b, a }
```

bim9262: @MaxVerevkin, that looks good. I'll amend my commit.

MaxVerevkin:
Thanks!

Issue #1328, opened 2021-11-04
Add option to configure in rust by turning blocks & bar into a library crate


LordMZTE:
I'm currently looking to switch from a homemade bar to i3status-rust, but I think that I'd prefer configuring i3status-rs in rust, as opposed to toml.

This would (assuming we add a good API) also make it easy to implement custom blocks. The way I'm seeing this work, is to create a `lib.rs` file in i3status-rs, and have all the modules that wouldn't be used only in the current, toml-based version be declared in `lib.rs`, in order to have them be available in the library crate. This crate could then also be pushed to crates.io.

The user would then create a binary crate, and add the i3status-rust library to `Cargo.toml`. The main function could look something like this:
```rust
use i3status_rust::{blocks::{ClockBlock, BatteryBlock}, Bar};
fn main() {
    let mut bar = Bar::new();

    bar.add_block(ClockBlock::new());
    bar.add_block(BatteryBlock::new());

    bar.run();
}
```

Of course this is just a rough idea of how the API could look.

There seems to be a project that already works like this called [i3monkit](https://github.com/38/i3monkit), but it seems to be unmaintained, and doesn't have nearly as many blocks as i3status-rust, and also doesn't seem to support styling as fancy as i3status-rust does or callbacks.

I would love to contribute to i3status-rust, but I've currently not used it or looked at the code, so it might take me some time to get used to it. 😅

GladOSkar:
Related: #756 (But very different)

ammgws:
I don't really understand what you are gaining by doing configuration in source or for creating a custom binary as in your example. If you're going that far you might as well just keep a personal fork and do whatever you want there?

LordMZTE:
> I don't really understand what you are gaining by doing configuration in source or for creating a custom binary as in your example. If you're going that far you might as well just keep a personal fork and do whatever you want there?

I think this approach would make it more convenient. In a personal fork, you'd have to rip out the whole config part of i3status-rust first. With a library crate, this would be made much easier and we could easily integrate it, as the binary crate could just call the same API as a cutom binary would, but based on the config file.

Here's some reasons I think someone might prefer configuring in rust:
- different syntax (some might prefer it over toml)
- less runtime overhead (no config loading)
- greater customizability (ability to add custom blocks by simply implementing a trait)
- custom callbacks in custom blocks

MaxVerevkin:
[...]

PR #1652, opened 2022-10-30
refactor: avoid toml::Value manipulation

MaxVerevkin:
Now each block's `run` function directly accepts a config struct instead of `toml::Value`. Might help with #1328

### DIFF
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -2,20 +2,14 @@
 
 pub mod prelude;
 
-use crate::formatting::config::Config as FormatConfig;
-use crate::themes::ThemeOverrides;
 use crate::BoxedFuture;
 use futures::future::FutureExt;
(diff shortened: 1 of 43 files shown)


MaxVerevkin:
Advantages:
- cleaner code
- if/when #1328 gets implemented, there will be no need to construct an intermediate `toml::Value`

Disadvantages:
- a broken block configuration will break the entire bar and not just one block.

MaxVerevkin:
@ammgws any thoughts on this?

ammgws (approved):

MaxVerevkin:
> a broken block configuration will break the entire bar and not just one block.

I think this can be fixed. We can have

```rs
struct BlocksList(Vec<Result<BlockConfig, Error>>);
```

with manual `Deserialize` impl which doesn't propagate errors and always succeeds.

PR #1662, opened 2022-11-15
Add support for SwayNotificationCenter driver

bim9262:
See https://github.com/ErikReider/SwayNotificationCenter

### DIFF
--- a/src/blocks/notify.rs
+++ b/src/blocks/notify.rs
@@ -8,7 +8,7 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `driver` | Which notifications daemon is running. Available drivers are: `"dunst"` | `"dunst"`
+//! `driver` | Which notifications daemon is running. Available drivers are: `"dunst"` and `"swaync"` | `"dunst"`
 //! `format` | A string to customise the output of this block. See below for available placeholders. | `" $icon "`
 //!
 //! Placeholder | Value                                      | Type   | Unit
@@ -49,6 +49,7 @@ struct NotifyConfig {
 enum DriverType {
     #[default]
     Dunst,
+    SwayNC,
 }
 
 pub async fn run(config: toml::Value, mut api: CommonApi) -> Result<()> {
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
This causes extremely high cpu usage for me.

For some reason, `self.changes.next().await;` never stops. But without `.cache_properties(zbus::CacheProperties::No)` the opposite happens - `self.changes.next().await;` never returns.

image

bim9262:
Ouch 🙈 Good catch. From your experience with dbus any idea what's causing the problem when using the cache? (I believe lazy caching is the default)

MaxVerevkin:
> From your experience with dbus any idea what's causing the problem when using the cache?

The only time I had issues with caching was kdeconnect, so I had to [disable](https://github.com/greshake/i3status-rust/blob/89c4ec7ca77169c0a571bf4e68f5302db6c952aa/src/blocks/kdeconnect.rs#L191-L211) it. If I remember correctly, that was because kdeconnect doesn't emit standard signals when properties change and uses custom signals instead.

But I've never had a `PropertyStream` that spams events as fast as it can. Might be `zbus` bug?

MaxVerevkin:
@bim9262 I pushed a commit that should fix it, can you test it? The issue was that `swaync` doesn't emit `PropertiesChanged` and instead provides it's own `OnDndToggle`. IDK why.

bim9262:
I ended up using the SetDnd and GetDnd properties as well as the Subscribe signal. The cpu usage is back down to normal.

MaxVerevkin:
> as well as the Subscribe signal

How is it different from OnDndToggle?

bim9262:
Subscribe is triggered when the count, dnd, or when the visibility of the notification center changes. I was thinking that in a future PR we could set the widget state based on if there are any new notifications or not.

Much like @ammgws noted in https://github.com/dunst-project/dunst/issues/765 where paused was not being emitted displayedLength doesn't either.

MaxVerevkin (approved):

MaxVerevkin:
Thanks!

bim9262:
You're welcome!

PR #1664, opened 2022-11-17
backlight: Add option to hide if no backlight supported

BourgondAries:
I'm using this feature on multiple machines, both laptops and stationary. I do not want different configuration files for these systems, however, backlight will fail on reading the backlight device directory on systems without backlight, yielding an ugly red error message.

This commit allows one to hide the backlight bar if that happens.

### DIFF
--- a/src/blocks/backlight.rs
+++ b/src/blocks/backlight.rs
@@ -21,6 +21,7 @@
 //! `cycle` | The brightnesses to cycle through on each click | `[minimum, maximum]`
 //! `root_scaling` | Scaling exponent reciprocal (ie. root) | `1.0`
 //! `invert_icons` | Invert icons' ordering, useful if you have colorful emoji | `false`
+//! `hide_on_unavailable` | Hide this block when the backlight device does not exist | `false`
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
You can get the same result with

```toml
error_format = ""
```

ammgws:
Or use "if_command"

BourgondAries:
@MaxVerevkin @ammgws Oh didn't know these existed! Thanks.

ammgws:
Let us know your final block config since could be useful for others who come across the same problem

BourgondAries:
> Let us know your final block config since could be useful for others who come across the same problem

```
[[block]]
block = "backlight"
format = " $icon {$brightness.eng(4) |}"
error_format = ""
```

I think `error_format` just wasn't easily discoverable for me, to me it was in a sea of global configs, while I was looking inside the block's specific config. I think I did this because I saw `[icons] table:` in the readme and skipped over the visually grouped-looking "block" of content (including the error_format section) straight to `Further documentation:`.

Issue #1519, opened 2022-06-22
Overriding default click actions


ammgws:
With the config below, right clicking on the sound block will trigger the default mute action as well as executing random_command2.

```toml
[[block]]
block = "sound"
[[block.click]]
button = "left"
cmd = "random_command1"
[[block.click]]
button = "right"
cmd = "random_command2"
```

Should the default click action just be overridden?

Also, should there be a way to remap or use the default action with another button, without reimplementing it via a command yourself?
e.g.
```toml
[[block]]
block = "sound"
[[block.click]]
button = "left"
cmd = "{block_default_left}"
```
( I just made up the syntax to illustrate the point)

instead of having to do
```toml
[[block]]
block = "sound"
[[block.click]]
button = "left"
cmd = "pamixer --mute"
```
(That is just a simple example, the bluetooth block disconnect is probably less trivial)

MaxVerevkin:
> Also, should there be a way to remap or use the default action with another button, without reimplementing it via a command yourself?

I imagine something like this could be implemented:

```toml
[[block]]
block = "sound"
# Left-click will toggle mute and _will not_ run custom command defined for right-click
[[block.click]]
button = "left"
redirect = "right"
# Right-click will run a command and not toggle mute
[[block.click]]
button = "right"
cmd = "my_cmd"
pass = false
```

ammgws:
Hmm, should `pass` be `false` by default?

MaxVerevkin:
> Hmm, should we make pass be false by default?

That sounds like a good idea

ammgws:
I just noticed that when overriding the left mouse button click for `music` that it also overrides the click actions for the play/pause/prev "widgets", which IIRC didn't happen with v0.22 and earlier. I don't think there is a way around this atm, so I guess just have to document it?

MaxVerevkin:
Some more thoughts.

Suppose each block has an `enum` of "actions". For example, a `sound` block has

```rust
pub enum Action {
    Mute,
    IncVolBy(u8),
    DecVolBy(u8),
}
```

Instead of receiving a stream of "click events" each block receives a stream of it's "actions":

```rust
event = api.event() => match event {
    UpdateRequest => { ... }
    ActionRequest(action) => match action {
        Action::Mute => { ... }
        Action::IncVolBy(step) => { ... }
        Action::DecVolBy(step) => { ... }
    }
}
```

Then the user can assign any action to any button:

```toml
[[block]]
block = "sound"
[...]

PR #1665, opened 2022-11-17
Notification counts and show messages

bim9262:
Left-clicking on this block will show notifications.
Right-clicking on this block will enable/disable notifications.

Add flag so that when there are notifications you could show the number of events.
Set widget state if there are notifications.

Depends on dunst-project/dunst#1127 for dunst notification count.

Please let me know if you have any thoughts on the design, format variable names,  if there's any other clever way to not need the `notifications` flag to achieve the same goal, or widget states.

### DIFF
--- a/src/blocks/notify.rs
+++ b/src/blocks/notify.rs
@@ -18,6 +18,7 @@
 //! Action          | Default button
 //! ----------------|---------------
 //! `toggle_paused` | Left
+//! `show`          | -
 //!
 //! # Examples
 //!
@@ -34,6 +35,19 @@
 //! [[block]]
 //! block = "notify"
 //! format = " $icon {($notification_count.eng(1)) |}"
+//! ```
+//! How to remap actions
+//!
+//! ```toml
+//! [[block]]
+//! block = "notify"
+//! driver = "swaync"
+//! [[block.click]]
+//! button = "left"
+//! action = "show"
+//! [[block.click]]
+//! button = "right"
+//! action = "toggle_paused"
 //! ```
 //!
 //! # Icons Used
@@ -98,6 +112,9 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
                 Action(a) if a == "toggle_paused" => {
                     driver.set_paused(!is_paused).await?;
                 }
+                Action(a) if a == "show" => {
+                    driver.notification_show().await?;
+                }
                 _ => (),
             }
         }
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
I like the idea, but I'm a bit worried about changing click behavior. Hopefully #1519 will improve the situation.

bim9262:
> I like the idea, but I'm a bit worried about changing click behavior. 

I understand the worry about changing the click behavior. It seems like it's more natural to left click to see existing/old notifications and right click to mute notifications.

> Alternatively, $notification_count can be present only when it's nonzero.

Thanks, that's definitely a lot less messy looking.

bim9262:
I added in the waitingCount to the total `notification_count` too. (FWIW this matches what swaync provides in its `NotificationCount`)

I was thinking about the states... If you're paused/dnd does that mean that the state should be more severe (warning or critical) than if you weren't paused/dnd?

bim9262:
@MaxVerevkin, I tested this with #1686 and that seems like a good solution for allowing actions to be remapped.

bim9262:
dunst-project/dunst#1127 has been merged!
I left the `show` action unmapped and left the left mouse mapped to `toggle_paused`. Let me know if there's anything you think that should be changed/reworked.

src/blocks/notify.rs
```diff
@@ -29,12 +29,20 @@
 //! block = "notify"
 //! format = " $icon {$paused{Off}|On} "
 //! ```
+//! How to use `notification_count`
+//!
+//! ```toml
+//! [[block]]
+//! block = "notify"
+//! format = " $icon {($notification_count.eng(1))|} "
```
MaxVerevkin, line 37: To avoid double spaces when notification_count is zero:
```suggestion
//! format = " $icon {($notification_count.eng(1)) |}"
```

src/blocks/notify.rs
```diff
@@ -11,14 +9,16 @@
...
+//! Placeholder          | Value                                      | Type   | Unit
+//! ---------------------|--------------------------------------------|--------|-----
+//! `icon`               | Icon based on notification's state         | Icon   | -
+//! `notification_count` | The number of notification (omitted if 0)  | Number | -
+//! `paused`             | Present only if notifications are disabled | Flag   | -
 //!
 //! Action          | Default button
 //! ----------------|---------------
 //! `toggle_paused` | Left
+//! `show`          | -
```
MaxVerevkin, line 21: An example of how to map this action might be useful, since it's a new feature.

MaxVerevkin:
Also, should we add a note in the docs that notification_count does not work properly for dunst <=1.9.0?

ammgws:
Yes

src/blocks/notify.rs
bim9262: Added an example :)

MaxVerevkin (approved):

MaxVerevkin:
Thanks for working on this and fixing `dunst` along the way!

bim9262:
You're welcome!

PR #1679, opened 2022-12-16
corrected 'unit' to 'alert_unit' in block disk_space in examples/config.toml

pyrrhull:
It's a minor thing but it prevents i3status-rs from displaying anything but an error.


MaxVerevkin:
Thanks

PR #1685, opened 2022-12-21
Add `not_charging_format` config option

### LINKED ISSUE #1416: Battery: add `not_charging_format` config option
Battery config
```toml
[[block]]
block = "battery"
format = "{percentage:2}"
device = "BAT0"
allow_missing = true
```

i3status-rust version: `0.21.2`
Linux kernel version: `5.15.17`

image

This is all I see
image

Using `upower` also has the same behaviour.

Let me know if there anything I can do to debug this issue.

benjamb:
It is possible for a battery to be in the "Not charging" state, in multi-battery setups especially, where both batteries may be partially drained, but only one will be being charged at once. In such cases the battery block incorrectly reports the non-charging battery as full.

In order to not break existing behaviour, and also given as mentioned in continue to report "Not charging" batteries to be full by default.

Fixes #1416.


MaxVerevkin (approved):

MaxVerevkin:
Thanks!

PR #1698, opened 2023-01-10
formatting: implement named arguments instead of positional

### LINKED ISSUE #1605: Maybe reconsider formatter arguments?

### LINKED ISSUE #1624: Formatting: `rot-str`: add an option to disable padding

### LINKED ISSUE #1655: pango escaping on by default?
At some point recently it seems like escaping pango characters has been added to format strings by default, but as far as I can tell it breaks being able to use pango formatting in the first place.

e.g. a custom block that outputs `"<span color='red'>my custom text here</span>"` is pango escaped so displays literally in the bar without being rendered by pango
```json
{"full_text":"","color":"#424242FF","separator":false,"separator_block_width":0,"markup":"pango"},{"full_text":" &lt;span color=&#39;red&#39;&gt;my custom text here&lt;/span&gt; ","color":"#FFFFFFFF","background":"#424242FF","name":"0","separator":false,"separator_block_width":0,"markup":"pango"}],
```

MaxVerevkin:
Closes #1605
Closes #1624
Closes #1655

```toml
format = " $speed_down.eng(3,_B,K) $frequency.eng(4) $ssid.rot-str(10,0.2) "
```

Becomes

```toml
format = " $speed_down.eng(w:3,p:K,hide_unit:true) $frequency.eng(w:4) $ssid.str(max_w:10, rot_interval:0.2) "
```

Also use `nom` for parsing. The downside of this is that error messages are even worse now. But this should be fixable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -881,6 +881,7 @@ dependencies = [
  "neli",
  "neli-wifi",
  "nix 0.26.2",
+ "nom",
  "notmuch",
  "once_cell",
  "regex",
@@ -1157,6 +1158,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2a60c7ce501c71e03a9c9c0d35b861413ae925bd979cc7a4e30d060069aaac8d"
 
 [[package]]
+name = "minimal-lexical"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "68354c5c6bd36d73ff3feceb05efa59b6acb7626617f4962be322a825e61f79a"
+
+[[package]]
 name = "mio"
 version = "0.8.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1248,6 +1255,16 @@ dependencies = [
  "memoffset 0.7.1",
  "pin-utils",
  "static_assertions",
+]
+
+[[package]]
+name = "nom"
+version = "7.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d273983c5a657a70a3e8f2a01329822f3b8c8172b73826411a55751e404a0a4a"
+dependencies = [
+ "memchr",
+ "minimal-lexical",
 ]
 
 [[package]]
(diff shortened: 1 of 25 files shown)


GladOSkar:
I think i liked the old, concise version better. But this seems fine as well, not gonna die on that hill

MaxVerevkin:
> I think i liked the old, concise version better.

The old version is nice, simple and compact, but it is also not extensible. For example I don't know how to implement `$x.str(pango:true)` with it, apart from introducing a dedicated formatter which behaves like `str` but does not escape pango. Also hacks like ` _B` for "hidden bytes with a space" are questionable (but compact, which is nice).

Speaking about dedicated formatters, with this new syntax `rot-str` can be merged into `str` by adding "rot_interval" argument to `str`. This would close #1624.

MaxVerevkin:
>  I don't know how to implement $x.str(pango:true) with it, apart from introducing a dedicated formatter which behaves like str but does not escape pango

On a second thought, making it a separate formatter seems like a good idea because min/max width and rotating doesn't work for unescaped pango anyway.

ammgws:
Seems good so far

MaxVerevkin:
I'll mark this PR as ready once I fix error reporting.

ammgws (approved):

Issue #1641, opened 2022-10-13
Announcements: Breaking changes


MaxVerevkin:

MaxVerevkin:
#1640

**temperature**: `collapsed` option is removed and now you can use `format_alt = " $icon "` to maintain the behavior

MaxVerevkin:
#1644

**memory**: `clickable`, `display_type`, `format_mem` and `format_swap` are removed and now you can use `format` and `format_alt` to maintain the behavior

MaxVerevkin:
#1686

- `pass` has been removed form click handler config options.
- `apt` `dnf` `pacman`: Clicking on the block does not cause an update by default anymore. To replicate that behavior use
  ```toml
  [[block.click]]
  # Updates the block on left click
  button = "left"
  update = true
  ```
- `rofication`: Clicking on the block does not open `rofication-gui` by default anymore. To replicate that behavior use
  ```toml
  [[block.click]]
  button = "left"
  cmd = "rofication-gui"
  ```

MaxVerevkin:
#1698

- Instead of positional arguments, format templates now use named arguments, i.e. `$speed_down.eng(w:3,p:K,hide_unit:true)` instead of `$speed_down.eng(3,_B,K)`. See documentation [here](https://greshake.github.io/i3status-rust/i3status_rs/formatting/index.html).
- `rot-str` formatter has been removed. To replicate its behavior, use `rot_interval` argument of `str`. For example `$ssid.str(max_w:10, rot_interval:0.2)`.

MaxVerevkin:
#1713 replaces `cpu` icon with`cpu_low`, `cpu_med` and `cpu_high`. Don't forget to run `./install.sh`. If you were overriding `cpu` icon, you now have to override these three icons instead.

MaxVerevkin:
#1718 

- **music**: `seek_step` was replaced with `seek_step_secs`.

MaxVerevkin:
#1723 changes **time** block formatting and removes `locale` option.

### Before

```toml
[[block]]
block = "time"
locale = "fr_BE"
format = " $icon %d/%m %R "
```

### After

```toml
[[block]]
block = "time"
format = " $icon $timestamp.datetime(f:'%a %Y-%m-%d %R %Z', l:fr_BE) "
```

MaxVerevkin:
[...]

PR #1713, opened 2023-02-03
replace cpu icon with cpu_{low,med,high}

MaxVerevkin:
Only for `material-nf`, free variants of `awesome` don't have these icons.


c02y:
I'm using `icons = "awesome6"`, now `format = "$icon $utilization "` for `block = "cpu"` doesn't work:

image

MaxVerevkin:
> I'm using `icons = "awesome6"`, now `format = "$icon $utilization "` for `block = "cpu"` doesn't work:
> 
> image

See https://github.com/greshake/i3status-rust/issues/1641#issuecomment-1417304664

c02y:
Thanks, it works.

PR #1716, opened 2023-02-06
Add service_status block

### LINKED ISSUE #1712: feature: service status via dbus integration
On devices using `systemd`: To continuously check if a service is running (i.e. `tor.service`) one could use `dbus` to query for the status of a certain service.
There already exists a project interfacing the [dbus in Rust](https://github.com/diwic/dbus-rs).

Currently a working solution without dbus could look like this, which should be replaced with propper dbus querying.

```toml
[[block]]
block = "custom"
command = "systemctl status tor 1&>/dev/null && echo TOR|| echo -"
```

bim9262:
Resolves #1712

### DIFF
--- a/src/blocks/service_status.rs
+++ b/src/blocks/service_status.rs
@@ -1,0 +1,173 @@
+//! Display the status of a service
+//!
+//! Right now only `systemd` is supported.
+//!
+//! # Configuration
+//!
+//! Key | Values | Default
+//! ----|--------|--------
+//! `driver` | Which init system is running the service. Available drivers are: `"systemd"` | `"systemd"`
+//! `service` | The name of the service | **Required**
+//! `active_format` | A string to customise the output of this block. See below for available placeholders. | `" $service active "`
+//! `inactive_format` | A string to customise the output of this block. See below for available placeholders. | `" $service inactive "`
+//! `active_state` | A valid [`State`] | [`State::Idle`]
+//! `inactive_state` | A valid [`State`]  | [`State::Critical`]
+//!
+//! Placeholder    | Value                     | Type   | Unit
+//! ---------------|---------------------------|--------|-----
+//! `service`      | The name of the service   | Text   | -
+//!
+//! # Example
+//!
+//! Example using an icon:
+//!
+//! ```toml
+//! [[block]]
+//! block = "service_status"
+//! service = "cups"
+//! active_format = " ^icon_tea "
+//! inactive_format = " no ^icon_tea "
+//! ```
+//!
+//! Example overriding the default `inactive_state`:
+//!
+//! ```toml
+//! [[block]]
+//! block = "service_status"
+//! service = "shadow"
+//! active_format = ""
+//! inactive_format = " Integrity of password and group files failed "
+//! inactive_state = "Warning"
+//! ```
+//!
+
+use super::prelude::*;
+use zbus::dbus_proxy;
+use zbus::PropertyStream;
+
+#[derive(Deserialize, Debug, Default)]
+#[serde(default)]
+pub struct Config {
+    driver: DriverType,
+    service: String,
+    active_format: FormatConfig,
+    inactive_format: FormatConfig,
+    active_state: Option<State>,
+    inactive_state: Option<State>,
+}
+
+#[derive(Deserialize, Debug, SmartDefault)]
+#[serde(rename_all = "snake_case")]
+enum DriverType {
+    #[default]
+    Systemd,
+}
+
+pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
+    api.event_receiver.close();
+
+    let mut widget = Widget::new();
+    let active_format = config.active_format.with_default(" $service active ")?;
+    let inactive_format = config.inactive_format.with_default(" $service inactive ")?;
+
+    let active_state = config.active_state.unwrap_or(State::Idle);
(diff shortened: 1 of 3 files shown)


src/blocks/service_status.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        widget.set_values(map!(
+            "service" =>Value::text(config.service.clone()),
+            "active_state" =>Value::text(active_state_str.to_string()),
+        ));
+        api.set_widget(&widget).await?;
+
+        select! {
+            x = driver.wait_for_change() => x?,
+        }
```
MaxVerevkin, line 73: You might want to either call `api.event()` and ignore it, or close the `api.event_receiver` channel, otherwise the bar will deadlock eventually.

src/blocks/service_status.rs
```diff
@@ -0,0 +1,143 @@
...
+//! Key | Values | Default
+//! ----|--------|--------
+//! `driver` | Which init system is running the service. Available drivers are: `"systemd"` | `"systemd"`
+//! `service` | The name of the service | **Required**
+//! `format` | A string to customise the output of this block. See below for available placeholders. | `" $service $active_state "`
+//!
+//! Placeholder    | Value                     | Type   | Unit
+//! ---------------|---------------------------|--------|-----
+//! `service`      | The name of the service   | Text   | -
+//! `active_state` | The status of the service | Text   | -
```
MaxVerevkin, line 16: 1. I think it's better to just have two formats: `active_format` and `inactive_format`.
2. Since `service` is just a string from the config, is it really necessary?

src/blocks/service_status.rs
```diff
@@ -0,0 +1,143 @@
...
+        let dbus_conn = new_system_dbus_connection().await?;
+
+        let encoded_service = format!("{service}.service")
+            .chars()
+            .map(|c| {
+                if c.is_ascii_alphanumeric() {
+                    c.to_string()
+                } else {
+                    format!("_{:02x}", c as u8)
+                }
```
MaxVerevkin, line 99: Can you please add a comment explaining what is going on here? Also, are you sure about this `char as u8` cast? By the way, instead of `.collect::<Vec<String>>().join("")` you can do `.collect::<String>()`.

src/blocks/service_status.rs
bim9262: I think having `active_format` and `inactive_format` is fine (or perhaps `format_active` and `format_inactive`). What would you make them default to?
I like having `" $service $active_state "` as a default for both. That way if you don't want to mess with the config to define the format if you don't want to (or  just want to set it to `""` to hide it).

In order to allow that default we'd have to use either:
* `format!(" {} $active_state ", config.service.clone()).as_str()`
* `" $service $active_state "`

I prefer just keeping `$service` as a placeholder.


Here are a few example configs that I though someone might use:

```toml
[[block]]
block = "service_status"
service = "cups"

[[block]]
block = "service_status"
service = "cups"
format_active = ""
format_inactive = " cups stopped running "
```

MaxVerevkin: I don't really see the point of `$active_state`. The defaults could be
```toml
format_active = " $service active "
format_inactive = " $service inactive  "
```
or even
```toml
format_active = " $service "
format_inactive = " $service "
```
because the state is duplicated by `widget.state`.

MaxVerevkin:
What do you think about adding `inactive_state` and `active_state` options? Inactive service doesn't necessarily indicate a failure. I know that this cat be achieved with theme overrides, but this would probably be more convenient.

bim9262:
> What do you think about adding `inactive_state` and `active_state` options? Inactive service doesn't necessarily indicate a failure. I know that this cat be achieved with theme overrides, but this would probably be more convenient.

That sounds like a good idea. I'll give it a try.
Any thoughts on a user set icon option? For example you might use a printer for cups (or maybe a teacup if you're funny).

MaxVerevkin:
> Any thoughts on a user set icon option? For example you might use a printer for cups (or maybe a teacup if you're funny).

Already possible:

```toml
active_format = " ^icon_tea "
inactive_format = " no ^icon_tea "
```

bim9262:
Whoops... Looks like I closed this accidently.

MaxVerevkin:
@leongross did you try this PR?

ammgws:
So this only works with system services and not user ones?

src/blocks/service_status.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+struct SystemdDriver {
+    proxy: UnitProxy<'static>,
+    active_state_changed: PropertyStream<'static, String>,
+}
+
+impl SystemdDriver {
+    async fn new(service: String) -> Result<Self> {
+        let dbus_conn = new_system_dbus_connection().await?;
```
ammgws, line 113: can we support user services too?

bim9262: I did try that, but user services are removed from dbus (the path is gone) when the service is stopped. I couldn't find a nice way to discover that it had come back. Perhaps if we did a filter like we do in the music block... but it feels like overkill if you are only interested in 1 service.

ammgws: Interesting. Guess if it is going to require a flag in the config then can be revisited in the future without having to figure it out now.

leongross:
@MaxVerevkin Yes, I tested it and it works on my machine!

PR #1717, opened 2023-02-06
apt: Add option to ignore waiting phased updates

sndels:
Apt seems to almost always hold a few phased updates back, for Ubuntu 22.04 at least. This makes the update indicator less useful. The logic appears to be that [anything under 100% might be held back based on the machine ID](https://salsa.debian.org/apt-team/apt/-/commit/c5bc86d45e003905ef411146e66b414d26fb1ff8). I added the option to filter out potentially held back packages from the update count.

The implementation just parses the package name from the apt list lines that were previously tallied raw, e.g.
```
gnome-initial-setup/jammy-updates 42.0.1-1ubuntu2.1 amd64 [upgradable from: 42.0.1-1ubuntu2]
```
finds the phase % from apt-cache if it's there, e.g.
```
$ apt-cache policy grub-efi-amd64-bin
grub-efi-amd64-bin:
  Installed: 2.06-2ubuntu10
  Candidate: 2.06-2ubuntu14
  Version table:
     2.06-2ubuntu14 500 (phased 24%)
        500 http://fi.archive.ubuntu.com/ubuntu jammy-updates/main amd64 Packages
 *** 2.06-2ubuntu10 500
        500 http://security.ubuntu.com/ubuntu jammy-security/main amd64 Packages
        100 /var/lib/dpkg/status
     2.06-2ubuntu7 500
        500 http://fi.archive.ubuntu.com/ubuntu jammy/main amd64 Packages
```
and then filters out any packages that don't have a phase % or have a % other than 100.

As a disclaimer, I don't have a lot of exprience with tokio so there might be a cleaner way to express the `filter_map`/`Result`-mess.

### DIFF
--- a/src/blocks/apt.rs
+++ b/src/blocks/apt.rs
@@ -14,6 +14,7 @@
 //! `format_up_to_date` | Same as `format`, but for when no updates are available. | `" $icon $count.eng(w:1) "`
 //! `warning_updates_regex` | Display block as warning if updates matching regex are available. | `None`
 //! `critical_updates_regex` | Display block as critical if updates matching regex are available. | `None`
+//! `ignore_phased_updates` | Doesn't include potentially held back phased updates in the count. | `false`
 //!
 //! Placeholder | Value                       | Type   | Unit
 //! ------------|-----------------------------|--------|------
@@ -66,6 +67,7 @@ pub struct Config {
     format_up_to_date: FormatConfig,
     warning_updates_regex: Option<String>,
     critical_updates_regex: Option<String>,
+    ignore_phased_updates: bool,
 }
 
 pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
@@ -121,7 +123,12 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
 
     loop {
         let updates = get_updates_list(config_file.to_str().unwrap()).await?;
-        let count = get_update_count(&updates);
+        let count = get_update_count(
+            config_file.to_str().unwrap(),
+            config.ignore_phased_updates,
+            &updates,
+        )
+        .await?;
 
         widget.set_format(match count {
             0 => format_up_to_date.clone(),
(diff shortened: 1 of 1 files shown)


src/blocks/apt.rs
```diff
@@ -192,3 +198,54 @@ fn get_update_count(updates: &str) -> usize {
...
+                    Ok(true) => None,
+                    Ok(false) => Some(ResultBool(Ok(true))),
+                    Err(e) => Some(ResultBool(Err(e))),
+                }
+            })
+            .collect::<Vec<ResultBool>>()
+            .await
+            .into_iter()
+            .map(|ResultBool(v)| v)
+            .collect::<Result<Vec<bool>>>()?;
```
MaxVerevkin, line 225: I think this is a bit over-engineered. Writing this using a simple `for` loop is simpler, more straight-forward and avoids two unnecessary `.collect`s:

```rs
let mut cnt = 0;

for update_line in updates.lines().filter(|line| line.contains("[upgradable")) {
    if !is_phased_update(config_path, update_line).await? {
        cnt += 1;
    }
}

Ok(cnt)
```

sndels: Ah yes, derp. Good point.

src/blocks/apt.rs
```diff
@@ -192,3 +198,38 @@ fn get_update_count(updates: &str) -> usize {
 fn has_matching_update(updates: &str, regex: &Regex) -> bool {
     updates.lines().any(|line| regex.is_match(line))
 }
+
+async fn get_update_count_ignore_phased(config_path: &str, updates: &str) -> Result<usize> {
+    let mut cnt = 0;
+
+    for update_line in updates.lines().filter(|line| line.contains("[upgradable")) {
+        if !is_phased_update(config_path, update_line).await? {
```
MaxVerevkin, line 206: You can also merge `get_update_count` and `get_update_count_ignore_phased` by doing something like

```rust
if !ignore_phased_updates || !is_phased_update(config_path, update_line).await? {
    cnt += 1;
}
```

sndels: Makes sense. I don't know why I expected weirdness with an async term in an or-chain, but looks like it does get skipped you'd expect :+1:

MaxVerevkin (approved):

MaxVerevkin:
Thanks!

sndels:
Thank you! Nice to have this integrated :+1:

PR #1720, opened 2023-02-09
allow icon configs to contain lists instead of strings, remove icons such as bat_N

MaxVerevkin:
This is a breaking change but it will make adding "animated" icons a non-breaking change in the future.

### DIFF
--- a/files/icons/awesome5.toml
+++ b/files/icons/awesome5.toml
@@ -1,40 +1,26 @@
 # FontAwesome 5: https://fontawesome.com/icons?d=gallery&p=2&m=free
-backlight_empty = "\U0001f315"
-backlight_full = "\U0001f311"
-backlight_1 = "\U0001f314"
-backlight_2 = "\U0001f314"
-backlight_3 = "\U0001f314"
-backlight_4 = "\U0001f314"
-backlight_5 = "\U0001f313"
-backlight_6 = "\U0001f313"
-backlight_7 = "\U0001f313"
-backlight_8 = "\U0001f313"
-backlight_9 = "\U0001f313"
-backlight_10 = "\U0001f312"
-backlight_11 = "\U0001f312"
-backlight_12 = "\U0001f312"
-backlight_13 = "\U0001f312"
+backlight = [
+    "\U0001f315",
+    "\U0001f314",
+    "\U0001f313",
+    "\U0001f312",
+    "\U0001f311",
+]
 bat_charging = "\uf1e6"
-bat_discharging = "\uf242"
 bat_not_available = "\uf244"
-bat_10 = "\uf244"
-bat_20 = "\uf243"
-bat_30 = "\uf243"
-bat_40 = "\uf243"
-bat_50 = "\uf242"
-bat_60 = "\uf242"
-bat_70 = "\uf241"
-bat_80 = "\uf241"
-bat_90 = "\uf241"
-bat_full = "\uf240"
+bat = [
+	"\uf244",
+	"\uf243",
+	"\uf242",
+	"\uf241",
+	"\uf240",
+]
 bell = "\uf0f3"
 bell-slash = "\uf1f6"
 bluetooth = "\uf294"
 calendar = "\uf073"
 cogs = "\uf085"
-cpu_low = "\uf3fd" # fa-tachometer-alt (other variations of this icon are not free)
-cpu_med = "\uf3fd" # fa-tachometer-alt
-cpu_high = "\uf3fd" # fa-tachometer-alt
+cpu = "\uf3fd" # fa-tachometer-alt (other variations of this icon are not free)
 cpu_boost_on = "\uf205"
 cpu_boost_off = "\uf204"
 disk_drive = "\uf0a0"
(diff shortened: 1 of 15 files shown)


files/icons/material.toml
```diff
@@ -81,13 +76,17 @@ toggle_on = "\ue837" # radio_button_on
 unknown = "\ueb8b" # question_mark | TODO: broken?
 update = "\ue8d7" # system_update_alt
 uptime = "\ue425" # timer
-volume_empty = "\ue04e" # volume_mute
-volume_full = "\ue050" # volume_up
-volume_half = "\ue04d" # volume_down
+volume_empty = [
```
bim9262, line 79: This should just be `volume`.

files/icons/material.toml
```diff
@@ -81,13 +76,17 @@ toggle_on = "\ue837" # radio_button_on
...
+    "\ue04e", # volume_mute
+    "\ue04d", # volume_down
+    "\ue050", # volume_up
+]
 volume_muted = "\ue04f" # volume_off
-microphone_full = "\ue029" # mic
-microphone_half = "\ue029" # mic
-microphone_empty = "\ue02a" # mic_none
+microphone_empty = [
```
bim9262, line 85: This should just be `microphone`.

files/icons/awesome5.toml
```diff
@@ -80,13 +66,13 @@ toggle_on = "\uf205"
 unknown = "\uf128"
 update = "\uf062"
 uptime = "\uf2f2"
-volume_empty = "\uf026"
-volume_full = "\uf028"
-volume_half = "\uf027"
+volume_empty = [
```
bim9262, line 69: This should just be `volume`.

files/icons/material.toml
```diff
@@ -1,41 +1,36 @@
...
-backlight_5 = "\ue3a8" # brightness_3
-backlight_6 = "\ue3a8" # brightness_3
-backlight_7 = "\ue3a9" # brightness_4
-backlight_8 = "\ue3a9" # brightness_4
-backlight_9 = "\ue3aa" # brightness_5
-backlight_10 = "\ue3aa" # brightness_5
-backlight_11 = "\ue3ab" # brightness_6
-backlight_12 = "\ue3ab" # brightness_6
-backlight_13 = "\ue3c8" # brightness_7
+backlight_empty = [
```
bim9262, line 3: This should just be `backlight`.

bim9262 (commented):
It seems like there's a few mismatches on the names in the icon files.

src/blocks/sound/pulseaudio.rs
```diff
@@ -350,9 +350,10 @@ impl Client {
     }
 
     fn send_update_event() {
-        for tx in &*EVENT_LISTENER.lock().unwrap() {
-            tx.blocking_send(()).unwrap();
-        }
```
ammgws, line 355: Was this a clippy lint?

ammgws (commented):
On mobile

src/blocks/sound/pulseaudio.rs
MaxVerevkin: No. I don't remember what I did, but this unwrap caused a panic. So instead of panicking this code will remove the listener from the list if channel was closed.

ammgws:
So for some blocks like sound,  the cut off points for each icon were nonlinear, do we want to try keep that behavior? From what I saw now what was previously 0..20 will be 0..33 if there are 3 icons in the list

ammgws:
On mob on

MaxVerevkin:
> So for some blocks like sound,  the cut off points for each icon were nonlinear, do we want to try keep that behavior? From what I saw now what was previously 0..20 will be 0..33 if there are 3 icons in the list

I'm okay either way. To restore previous behavior we can change `[a,b,c]` to `[a,b,b,b,c]`.

files/icons/awesome6.toml
```diff
@@ -1,40 +1,26 @@
...
+]
 bell = "\uf0f3"
 bell-slash = "\uf1f6"
 bluetooth = "\uf294"
 calendar = "\uf073"
 cogs = "\uf085"
-cpu_low = "\uf624" # fa-gauge (fa-gauge-{min,max} are not free)
-cpu_med = "\uf624" # fa-gauge
-cpu_high = "\uf624" # fa-gauge
+cpu = "\uf624" # fa-gauge (fa-gauge-{min,max} are not free)
```
bim9262, line 23: You might consider making this:

```toml
cpu = [ # fa-gauge-{min,max} are not free
    "\uf624", # fa-gauge
    "\uf624", # fa-gauge
    "\uf625", # fa-gauge-high
]
```

That way the gauge will go high at 66% which looks appropriate for the icon `fa-gauge-high`

src/blocks/sound/pulseaudio.rs
ammgws: Can you make it a separate commit so that we don't accidentally regress in the future?

ammgws:
> I'm okay either way. To restore previous behavior we can change `[a,b,c]` to `[a,b,b,b,c]`.

This way allows the user more customisation but could be a bit confusing - guess it could be OK as long as we document it or add comment in the icon config file

src/blocks/sound/pulseaudio.rs
MaxVerevkin: Pushed to master.

MaxVerevkin:
It took me seven months to realize that the current `^icon_` syntax does not support specifying an icon from that progression...

PR #1710, opened 2023-02-02
Add playerctld support to music block

bim9262:
playerctld registers itself like any other mpris media player so I wanted to make sure it wasn't showing up on the list.

I ended up making it a default `interface_name_exclude`, but perhaps it would be better to hard code in as interface that should be ignored so that if a user sets `interface_name_exclude` playerctld is still ignored.  Let me know if you have any thoughts on that.

PR #1722, opened 2023-02-11
MPRIS music volume

bim9262:
This is just a draft for now.  It add's individual volume controls per player. I keep my master volume at 100% and set each player volume independently.

It depends on changes in ~~#1710~~, ~~#1718~~ and ~~#1720~~.

As those are merged I will rebase this to de-duplicate the duplicated commits.


bim9262:
@ammgws, each MPRIS player has its own volume, it's not just playerctld related: https://specifications.freedesktop.org/mpris-spec/2.2/Player_Interface.html

bim9262:
Rebased to resolve the conflicts with #1718 too.

PR #1723, opened 2023-02-11
Add datetime value parser

### LINKED ISSUE #1708: time block - drop parsing format in the inner loop, maybe add datetime Value with formatter
https://github.com/greshake/i3status-rust/pull/1618#discussion_r980158956

bim9262:
Resolves #1708 ~~(well will)~~

I added a new parsers for the current time.

~~Issues to resolve~~
~~* icon doesn't work~~
~~* Instead of parsing the format each iteration it seems that a `StrftimeItems` could be used, but the lifetimes eluded me so far.~~

### DIFF
--- a/src/blocks/time.rs
+++ b/src/blocks/time.rs
@@ -31,29 +31,22 @@ use chrono::Utc;
 use chrono_tz::Tz;
 
 use super::prelude::*;
-use crate::formatting::config::DummyConfig;
 
 #[derive(Deserialize, Debug, SmartDefault)]
 #[serde(default)]
 pub struct Config {
-    format: DummyConfig,
+    format: FormatConfig,
     #[default(1.into())]
     interval: Seconds,
     timezone: Option<Tz>,
 }
 
 pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
-    let mut widget = Widget::new();
-
-    let format = config
-        .format
-        .full
-        .as_deref()
-        .unwrap_or(" $icon $timestamp.datetime() ");
-
-    let format_short = config.format.short.as_deref().unwrap_or_default();
-
-    widget.set_format(FormatConfig::default().with_defaults(format, format_short)?);
+    let mut widget = Widget::new().with_format(
+        config
+            .format
+            .with_default(" $icon $timestamp.datetime() ")?,
+    );
 
     let timezone = config.timezone;
 
--- a/src/blocks/time.rs
+++ b/src/blocks/time.rs
@@ -4,14 +4,14 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `format` | Format string. See [chrono docs](https://docs.rs/chrono/0.3.0/chrono/format/strftime/index.html#specifiers) for all options. | `" $icon %a %d/%m %R "`
+//! `format` | Format string. See [chrono docs](https://docs.rs/chrono/0.3.0/chrono/format/strftime/index.html#specifiers) for all options. | `" $icon $timestamp.datetime() "`
 //! `interval` | Update interval in seconds | `10`
 //! `timezone` | A timezone specifier (e.g. "Europe/Lisbon") | Local timezone
-//! `locale` | Locale to apply when formatting the time | System locale
 //!
-//! Placeholder   | Value                                       | Type   | Unit
-//! --------------|---------------------------------------------|--------|-----
-//! `icon`        | A static icon                               | Icon   | -
+//! Placeholder   | Value                                       | Type     | Unit
+//! --------------|---------------------------------------------|----------|-----
+//! `icon`        | A static icon                               | Icon     | -
+//! `timestamp`   | The current time                            | Datetime | -
 //!
 //! # Example
 //!
@@ -19,17 +19,15 @@
 //! [[block]]
 //! block = "time"
 //! interval = 60
-//! locale = "fr_BE"
 //! [block.format]
-//! full = " $icon %d/%m %R "
-//! short = " $icon %R "
+//! full = " $icon $timestamp.datetime(f:'%a %Y-%m-%d %R %Z', l:fr_BE) "
+//! short = " $icon $timestamp.datetime(f:%R) "
 //! ```
 //!
 //! # Icons Used
 //! - `time`
 
-use chrono::offset::{Local, Utc};
-use chrono::Locale;
+use chrono::Utc;
 use chrono_tz::Tz;
 
 use super::prelude::*;
@@ -42,7 +40,6 @@ pub struct Config {
     #[default(1.into())]
     interval: Seconds,
     timezone: Option<Tz>,
-    locale: Option<String>,
 }
 
 pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
@@ -52,14 +49,13 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
         .format
         .full
         .as_deref()
-        .unwrap_or(" $icon %a %d/%m %R ");
-    let format_short = config.format.short.as_deref();
+        .unwrap_or(" $icon $timestamp.datetime() ");
+
+    let format_short = config.format.short.as_deref().unwrap_or_default();
+
+    widget.set_format(FormatConfig::default().with_defaults(format, format_short)?);
 
     let timezone = config.timezone;
-    let locale = match config.locale.as_deref() {
-        Some(locale) => Some(locale.try_into().ok().error("invalid locale")?),
-        None => None,
-    };
 
     let mut timer = config.interval.timer();
 
@@ -70,13 +66,10 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
             unsafe { tzset() };
         }
 
-        let full_time = get_time(format, timezone, locale);
-        let short_time = format_short
-            .map(|f| get_time(f, timezone, locale))
-            .unwrap_or_else(|| "".into());
-
-        widget.set_format(FormatConfig::default().with_defaults(&full_time, &short_time)?);
-        widget.set_values(map!("icon" => Value::icon(api.get_icon("time")?)));
+        widget.set_values(map!(
+            "icon" => Value::icon(api.get_icon("time")?),
+            "timestamp" => Value::datetime(Utc::now(), timezone)
+        ));
 
         api.set_widget(&widget).await?;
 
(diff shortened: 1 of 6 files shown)


MaxVerevkin:
> Instead of parsing the format each iteration it seems that a StrftimeItems could be used, but the lifetimes eluded me so far.

Ah, self referential structs... AFAIK there are a few approaches: use some library that will do some magic, leak the string to get static lifetime, write a wrapper type (like `StrftimeItemsOwned`) with a bit of `unsafe` which will "leak" (`Box::into_raw`) string on creation and will free to on drop (`Box::from_raw`). I'm okay with either the first or third option.

MaxVerevkin:
I _think_ this is correct, as long as it is in its own module: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ffad275ea8783f05068a850ac351195c

bim9262:
@MaxVerevkin, why does it matter if it's in its own module?

MaxVerevkin:
> why does it matter if it's in its own module?

This way it can be proven that no code accidentally leaks static reference, because public API would only provide access to borrowed items.

Actually, never mind. Any `chrono::format::Item` can be made `'static`:

```rust
fn make_static_item(item: Item<'_>) -> Item<'static> {
    mach item {
        Item::Literal(str) => Item::OwnedLiteral(str.into()),
        // And so on
    }
}

fn parse(str: &str) -> Vec<Item<'static>> {
    StrftimeItems::new(str).map(make_static_item).collect()
}
```

src/blocks/time.rs
```diff
@@ -70,13 +80,8 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
...
         }
 
-        let full_time = get_time(format, timezone, locale);
-        let short_time = format_short
-            .map(|f| get_time(f, timezone, locale))
-            .unwrap_or_else(|| "".into());
-
-        widget.set_format(FormatConfig::default().with_defaults(&full_time, &short_time)?);
         widget.set_values(map!("icon" => Value::icon(api.get_icon("time")?)));
+        widget.set_values(map!("timestamp" => Value::timestamp(timezone)));
```
MaxVerevkin, line 84:
```suggestion
        widget.set_values(map! {
            "icon" => Value::icon(api.get_icon("time")?),
            "timestamp" => Value::timestamp(timezone),
        });
```

This resolves

> icon doesn't work

MaxVerevkin:
> Actually, never mind. Any chrono::format::Item can be made 'static:

I even think this could be upstreamed to `chrono`

src/blocks/time.rs
bim9262: Thanks, I noticed that this morning and groaned it my idiocy.

src/formatting/parse.rs
```diff
@@ -82,14 +82,19 @@ fn alphanum1(i: &str) -> IResult<&str, &str, PError> {
...
 }
 
 // `key:val`
 fn parse_arg(i: &str) -> IResult<&str, Arg, PError> {
     map(
         separated_pair(alphanum1, cut(char(':')), cut(arg1)),
-        |(key, val)| Arg { key, val },
+        |(key, val)| Arg {
+            key,
+            val: val.trim(),
```
bim9262, line 96: Arg vals couldn't have spaces before, but they can now, so I added this trim, but if there's a nicer way to do that let me know.

bim9262:
Thanks @MaxVerevkin for your help getting the static item stuff worked out!! I added docs the the formatting page as well.

src/formatting/parse.rs
MaxVerevkin: I think this is okay. Alternatively this can be done inside `arg1`.

Also, should `.fmt(key: )` be accepted? Maybe? Should this represent an empty string or a whitespace?

What do you think about requiring escaped spaces in arguments? Like `.datetime(f:%a\ %b)`.

bim9262:
This should be tagged as breaking (removal of the locale option and addition of the new datetime formatter)

src/blocks/time.rs
```diff
@@ -42,7 +39,6 @@ pub struct Config {
     #[default(1.into())]
```
MaxVerevkin, line 39: We can now use normal `FormatConfig` instead of `DummyConfig`. I think we can even remove `DummyConfig`.

MaxVerevkin: github doesn't allow comments on unedited lines for some reason :angry:

src/formatting/formatter.rs
```diff
@@ -392,20 +432,76 @@ impl Formatter for FixFormatter {
...
+
+        Ok(Self { items: Some(items) })
+    }
+}
+
+impl Formatter for TimestampFormatter {
+    fn format(&self, val: &Value) -> Result<String> {
+        let items = self.items.clone().unwrap();
+        match val {
+            Value::Number { .. } | Value::Text(_) | Value::Icon(_) | Value::Flag => unreachable!(),
```
MaxVerevkin, line 484: This code is not actually `unreachable`. The reason this is true for `FlagFormatter` is because there is no way for user to explicitly specify flag formatter.

MaxVerevkin: The good news is that this code can be simplified by implementing `Value::type_name(&self) -> &'static str`:

```rust
match val {
    Value::Timestamp(timezone) => todo!(),
    other => Err(Error::new_format(format!(
        "{} cannot be formatted with 'datetime' formatter",
        other.type_name(),
    ))),
}
```

bim9262:
```rs
Err(Error::new_format(format!(
    "{:?} cannot be formatted with 'datetime' formatter",
    other,
)))
```

Also seems to work

src/formatting/parse.rs
bim9262: Maybe using single quotes would be best as it would allow spaces to be preserved:

```
.fmt(key:' ') 
.datetime(f:'%a %b')
```

src/formatting/formatter.rs
MaxVerevkin: The problem with this is that it prints the internals of the `Value`, which is an implementation detail and kind off irrelevant, making the error message longer that it needs to be.

src/formatting/parse.rs
MaxVerevkin: That might be the best way :+1:

src/formatting/formatter.rs
```diff
@@ -112,14 +119,35 @@ pub fn new_formatter(name: &str, args: &[Arg]) -> Result<Box<dyn Formatter>> {
...
+            let mut format = None;
+            let mut locale = None;
+            for arg in args {
+                match arg.key {
+                    "format" | "f" => {
+                        format = Some(arg.val.parse().error("format value must be a string")?);
+                    }
+                    "locale" | "l" => {
+                        locale = arg.val.parse().ok();
+                    }
```
MaxVerevkin, line 140:
```suggestion
                    "format" | "f" => {
                        format = Some(arg.val.to_string());
                    }
                    "locale" | "l" => {
                        locale = Some(arg.val.to_string());
                    }
```

bim9262:
Updates to the parser still to come...

src/formatting.rs
```diff
@@ -59,6 +59,13 @@
 //!
 //! No arguments.
 //!
+//! ## `datetime` - Display datetime
+//!
+//! Argument               | Description                                                                                               |Default value
+//! -----------------------|-----------------------------------------------------------------------------------------------------------|-------------
+//! `format` or `f`        | [chrono docs](https://docs.rs/chrono/0.3.0/chrono/format/strftime/index.html#specifiers) for all options. | `"%a %d/%m %R"`                                      | `5`
```
MaxVerevkin, line 66:
```suggestion
//! `format` or `f`        | [chrono docs](https://docs.rs/chrono/0.3.0/chrono/format/strftime/index.html#specifiers) for all options. | `"%a %d/%m %R"`
```

src/formatting/formatter.rs
```diff
@@ -386,15 +400,64 @@ impl Formatter for FixFormatter {
...
+        .map(make_static_item)
+        .collect();
+
+        Ok(Self { items: Some(items) })
+    }
+}
+
+impl Formatter for DatetimeFormatter {
+    fn format(&self, val: &Value) -> Result<String> {
+        let items = self.items.clone().unwrap();
```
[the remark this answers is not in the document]
MaxVerevkin: Also, why `clone()`?

src/formatting/formatter.rs
```diff
@@ -386,15 +400,64 @@ impl Formatter for FixFormatter {
...
+        match val {
+            Value::Datetime(timezone) => Ok(match timezone {
+                Some(tz) => Utc::now()
+                    .with_timezone(tz)
+                    .format_with_items(items.iter())
+                    .to_string(),
+                None => Local::now().format_with_items(items.iter()).to_string(),
```
MaxVerevkin, line 455: I think we should store `DateTime<Utc>` in `Value::Datetime`. It's unlikely that `Datetime` will be ever used to display something other than the current time, but still.

src/formatting/formatter.rs
```diff
@@ -386,27 +405,88 @@ impl Formatter for FixFormatter {
...
     }
 }
 
+impl Formatter for DatetimeFormatter {
+    fn format(&self, val: &Value) -> Result<String> {
+        self._format(val)
+    }
+}
+
+impl Formatter for once_cell::sync::Lazy<DatetimeFormatter> {
```
bim9262, line 474: I couldn't find a way around this :(

MaxVerevkin: You can convert `&Lazy<T>` to `&T` when choosing the default formatter:

```diff
diff --git a/src/formatting/value.rs b/src/formatting/value.rs
index 11b89fcd..e86e1049 100644
--- a/src/formatting/value.rs
+++ b/src/formatting/value.rs
@@ -127,7 +127,7 @@ impl Value {
         match &self.inner {
             ValueInner::Text(_) | ValueInner::Icon(_) => &formatter::DEFAULT_STRING_FORMATTER,
             ValueInner::Number { .. } => &formatter::DEFAULT_NUMBER_FORMATTER,
-            ValueInner::Datetime { .. } => &formatter::DEFAULT_DATETIME_FORMATTER,
+            ValueInner::Datetime { .. } => &*formatter::DEFAULT_DATETIME_FORMATTER,
             ValueInner::Flag => &formatter::DEFAULT_FLAG_FORMATTER,
         }
     }
```

src/formatting/formatter.rs
```diff
@@ -386,27 +405,88 @@ impl Formatter for FixFormatter {
...
+        Item::Space(str) => Item::OwnedSpace(str.into()),
+        Item::OwnedSpace(boxed) => Item::OwnedSpace(boxed),
+        Item::Numeric(numeric, pad) => Item::Numeric(numeric, pad),
+        Item::Fixed(fixed) => Item::Fixed(fixed),
+        Item::Error => Item::Error,
+    }
+}
+
+impl DatetimeFormatter {
+    fn new(format: String, locale: Option<String>) -> Result<Self> {
```
MaxVerevkin, line 434: `format` and `locale` can be `&str`s instead of `String`s.

bim9262:
Hopefully that does it 🤞!

MaxVerevkin (approved):

PR #1725, opened 2023-02-16
Feature: VPN block

chisade:
Hello,
I created a small block to support [NordVPN](https://nordvpn.com/) network connections.

Main features are:
- show state of current network connection (connected / disconnected)
- toggle connection on left click
- different formats for connected and disconnected state
- optionally include the name of the country you are connected to
- customize widget state (highlighting color) for connected and disconnected state

Best regards,

Chisade


cfsmp3:
@chisade You may want to show a country flag (or support it at least), take a look at `country_flag_from_iso_code`
Example usage in block external_ip

src/blocks/nordvpn.rs
```diff
@@ -0,0 +1,219 @@
...
+        return Ok(Status::Disconnected);
+    } else if line_status.ends_with("Connected") {
+        match line_country {
+            Some(country_line) => {
+                let country = country_line
+                    .split(": ")
+                    .collect::<Vec<&str>>()
+                    .last()
+                    .unwrap()
+                    .to_string();
```
MaxVerevkin, line 179: There's [`rsplit`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.rsplit) method on `str` that does this in fewer steps.

src/blocks/nordvpn.rs
```diff
@@ -0,0 +1,219 @@
...
+        match *self {
+            Status::Connected(_) => "net_vpn",
+            Status::Disconnected => "net_wired",
+            Status::Error => "net_down",
+        }
+    }
+}
+
+pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
+    api.set_default_actions(&[(MouseButton::Left, None, "toggle")])
```
MaxVerevkin, line 96: Please document "toggle" action.

MaxVerevkin:
I think it might be a good idea to make this block more generic by renaming it to just "vpn" and adding a "driver" option (for now with "nordvpn" being the only option). See the recent `service_status` (#1716) which implements a similar approach.

src/blocks/nordvpn.rs
chisade: Would you consider [this](https://github.com/greshake/i3status-rust/pull/1725/commits/2e502dcd794ea28a7c16ebce0badc26e3ad5d0ee#diff-3f790987dd77e1f06baa6dcaf001d026de91ede112d783a6eb2e37b73fbc1885R2) a documentation?
If not, please suggest how you would document this.

src/blocks/nordvpn.rs
chisade: See [here](https://github.com/greshake/i3status-rust/pull/1725/commits/2e502dcd794ea28a7c16ebce0badc26e3ad5d0ee#diff-3f790987dd77e1f06baa6dcaf001d026de91ede112d783a6eb2e37b73fbc1885R251)

chisade:
@MaxVerevkin I implemented the block in a more generic way as suggested using "drivers".

src/blocks/nordvpn.rs
chisade: Otherwise consider [this](https://github.com/greshake/i3status-rust/pull/1725/commits/4670d15801dc6caba8162d06e1b8c1446f62d2d0#diff-3f790987dd77e1f06baa6dcaf001d026de91ede112d783a6eb2e37b73fbc1885R21-R24)

MaxVerevkin: > Otherwise consider [this](https://github.com/greshake/i3status-rust/pull/1725/commits/4670d15801dc6caba8162d06e1b8c1446f62d2d0#diff-3f790987dd77e1f06baa6dcaf001d026de91ede112d783a6eb2e37b73fbc1885R21-R24)

Yes, exposing the name of the action lets users remap it to different buttons.

src/blocks/vpn.rs
```diff
@@ -0,0 +1,300 @@
...
+}
+
+#[derive(Deserialize, Debug)]
+#[serde(default)]
+pub struct Config {
+    driver: DriverType,
+    interval: Seconds,
+    format_connected: FormatConfig,
+    format_disconnected: FormatConfig,
+    state_connected: Option<String>,
```
MaxVerevkin, line 90: I think it's better to have `State` instead of `Option<String>` and let `serde` deserialize handle it.

src/blocks/vpn.rs
```diff
@@ -0,0 +1,300 @@
...
+    state_disconnected: Option<String>,
+}
+
+impl Default for Config {
+    fn default() -> Config {
+        Config {
+            driver: DriverType::Nordvpn,
+            interval: 10.into(),
+            format_connected: FormatConfig::from_str(" VPN: $icon ").unwrap(),
+            format_disconnected: FormatConfig::from_str(" VPN: $icon ").unwrap(),
```
MaxVerevkin, line 100: This is unnecessary, the defaults are already set using `.with_default("...")?`

src/blocks/vpn.rs
```diff
@@ -0,0 +1,300 @@
...
+    driver: DriverType,
+    interval: Seconds,
+    format_connected: FormatConfig,
+    format_disconnected: FormatConfig,
+    state_connected: Option<String>,
+    state_disconnected: Option<String>,
+}
+
+impl Default for Config {
+    fn default() -> Config {
```
MaxVerevkin, line 95: Instead of manually implementing `Default` you can use `SmartDefault` macro. Example: https://github.com/greshake/i3status-rust/blob/c6a68b5d959019aebbda837db195bd5531b18b5a/src/blocks/cpu.rs#L52-L59

src/blocks/vpn.rs
```diff
@@ -0,0 +1,300 @@
...
+            Status::Error => (),
+        }
+        Ok(())
+    }
+}
+
+impl FromStr for State {
+    type Err = ();
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
```
MaxVerevkin, line 291: `State` already implements `Deserialize`, so this is unnecessary. With b3e52f6 it should support lowercase names too.

src/blocks/vpn.rs
chisade: I applied changes according to your suggestion.

src/blocks/vpn.rs
chisade: I applied changes according to your suggestion.

src/blocks/vpn.rs
chisade: I applied changes according to your suggestion.

src/blocks/vpn.rs
chisade: I applied changes according to your suggestion.

src/blocks/vpn.rs
```diff
@@ -0,0 +1,270 @@
...
+//! Possible values for `state_connected` and `state_disconnected`:
+//!
+//! ```
+//! Warning
+//! Critical
+//! Good
+//! Info
+//! Idle
+//! ```
+//! `Option::None` defaults to `idle`.
```
MaxVerevkin, line 55: This line doesn't provide much information and feels out of place.

src/blocks/vpn.rs
```diff
@@ -0,0 +1,270 @@
+//! Current connection status for vpn networks
+//! This widget toggles the connection on left click.
+//!
```
MaxVerevkin, line 3: Inserting an empty line after the first line will help `rustdoc` generate a better page at https://greshake.github.io/i3status-rust/i3status_rs/blocks/index.html

src/blocks/vpn.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    loop {
+        let status = driver.get_status().await?;
+
+        widget.set_values(map!(
+            "icon" => Value::icon(api.get_icon(status.icon())?),
+            "country" => Value::text(status.country().to_string()),
+            "flag" => Value::text(status.flag().to_string()),
+
+        ));
```
MaxVerevkin, line 148: I think it's better to only add `country` and `flag` if they exist, i.e. if `status` is `Connected`.

src/blocks/vpn.rs
```diff
@@ -0,0 +1,275 @@
...
+            _ => "",
+        }
+    }
+
+    fn country(&self) -> &str {
+        match self {
+            Status::Connected { country, .. } => country.as_str(),
+            _ => "",
+        }
+    }
```
MaxVerevkin, line 124: These functions are not necessary anymore because you can just do 
```rust
Status::Connected { country, country_flag } => {
                widget.set_values(map!(
                        "icon" => Value::icon(api.get_icon(status.icon())?),
                        "country" => Value::text(country),
                        "flag" => Value::text(country_flag),
                ));
                widget.set_format(format_connected.clone());
                config.state_connected
            }
```

chisade: Right!
I removed them.

MaxVerevkin (approved):
Looks good to me :+1:

Haven't tested it though because I don't use vpn.

chisade:
I'm using this block daily and for me it's working as expected.
It would definitely be nice if someone else could test this though.

totikom:
As mentioned in #1867, I've started working on adding Mullvad support. I've already implemented it, but on my machine it fails with `Failed to render full test (Placeholder 'icon' not found)`. I thought, that it is a bug in my code, but turns out NordVPN driver exhibits the same behavior on my machine.

PR #1728, opened 2023-02-18
Theme material-nf3

bim9262:
[**Update material-nf**](https://github.com/greshake/i3status-rust/commit/2225b77edcf240551105cf8a91c2fa5ea0534465) 

The `nf-mdi-*` icons have been deprecated. This commit exchanges them
for their `nf-md` counterparts.

[**Add thermometer progression for material-nf**](https://github.com/greshake/i3status-rust/commit/11778c6638c26f94113fd3b626098aff7f7fcab2)

### DIFF
--- a/files/icons/material-nf.toml
+++ b/files/icons/material-nf.toml
@@ -1,9 +1,5 @@
 # Material from NerdFont
 # https://www.nerdfonts.com/cheat-sheet
-#
-# Note: avoid using icons in the range 0xf500-0xfd46,
-# as they will cause the block to render backwards
-# until https://github.com/ryanoasis/nerd-fonts/issues/365 is fixed
 backlight = [
     "\ue38d", # nf-weather-moon_new
     "\ue3d4", # nf-weather-moon_alt_waxing_gibbous_6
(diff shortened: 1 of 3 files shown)


ammgws:
In reference to this: https://github.com/ryanoasis/nerd-fonts/issues/1096 ?

What was the difference between nf-mdi and nf-md? I can't remember...

bim9262:
Yes, I think that they moved the old mdi icons to a new region so they renamed them all to just md so the names wouldn't overlap as well. I will note that the new icons do render a bit smaller than the old ones.

ammgws:
> I will note that the new icons do render a bit smaller than the old ones.

Apparently...
https://github.com/ryanoasis/nerd-fonts/issues/1061

I imagine getting a few bug reports from that alone..

files/icons/material-nf.toml
```diff
@@ -21,91 +21,95 @@ backlight = [
...
+    "\U000f0082", # nf-md-battery_90
+    "\U000f0079", # nf-md-battery
 ]
-bell = "\uf599" # nf-mdi-bell
-bell-slash = "\uf59a" # nf-mdi-bell_off
-bluetooth = "\uf5ae" # nf-mdi-bluetooth
-calendar = "\uf5ec" # nf-mdi-calendar
-cogs = "\uf992" # nf-mdi-settings
+bell = "\U000f009c" # nf-md-bell_outline
+bell-slash = "\U000f009b" # nf-md-bell_off
```
MaxVerevkin, line 39: Why outline? And why only for `bell` but not `bell-slash`?

bim9262: It matched what had been done in fa. I think that it's easier to tell the difference when one is the outline and one is filled. (I did the same with the toggles)

MaxVerevkin: Now these two icons are not perfectly aligned: 

https://user-images.githubusercontent.com/34583604/219849036-6fb130df-5d96-4dcf-a68d-7e9c63696bba.mp4

MaxVerevkin: (does this mp4 work for anyone?)

ammgws: > (does this mp4 work for anyone?)

image

MaxVerevkin: It works in `mpv` :(

bim9262: That's interesting, it doesn't move for me


https://user-images.githubusercontent.com/1033745/219849545-422d04b5-7354-4f3a-aa19-27dfc8fda973.mp4

MaxVerevkin: image

image

MaxVerevkin: It shifts, but less... Anyway, this is not a big deal, just a bit weird.

Finii:
I will fix 1061 before v3.0.0 goes out.

bim9262:
Perhaps this should become a new theme, `material-nf3`?

ammgws:
So if nerd fonts v3.0.0 is going to not just deprecate but also remove the codepoints that we are currently using then we should probably create `material-nf3`.

bim9262:
v2 does write glyphs outside of the PUA. As noted on https://github.com/ryanoasis/nerd-fonts/issues/365:

> It is the problem that font_patcher writes glyphs from Material into the range: U+F500 - U+FD46. This range overlaps areas that should not use for such purpose.

One of the last comments from @Finii:

> Yes, according to plan this will come (glyphs removed) with v3.0.0, this was just one necessary intermediate step.

So, if I understand correctly, some of the codepoints we use will in today's material-nf will be removed in v3.

Finii:
@bim9262 

Citation from https://github.com/ryanoasis/nerd-fonts/releases

Screenshot 2023-02-23 at 08 41 01

bim9262:
Thanks @Finii!!

@ammgws, I'll rename this `material-nf3`

MaxVerevkin:
> we should probably create material-nf3

Is this a temporal solution until the size of new icons is fixed?

bim9262:
Looks great @Finii, thanks!!

If anyone else wants to test and hasn't built a font before, this is what I did on arch:

```
sudo pacman -S fontforge
cd /tmp
git clone --depth 1 https://github.com/ryanoasis/nerd-fonts.git
cd nerd-fonts
./font-patcher -s -ext ttf -c -out ~/.local/share/fonts/ src/unpatched-fonts/NerdFontsSymbolsOnly/NerdFontsSymbols\ 2048\ EM\ Nerd\ Font\ Complete\ Blank.sfd
```
Restart i3status-rust

bim9262:
@MaxVerevkin, @ammgws, 

Anything that should be done before this is merged?

Finii:
> If anyone else wants to test and hasn't built a font before, this is what I did on arch:
> 
> ```
> sudo pacman -S fontforge
> cd /tmp
> git clone --depth 1 https://github.com/ryanoasis/nerd-fonts.git
> cd nerd-fonts
> ./font-patcher -s -ext ttf -c -out ~/.local/share/fonts/ src/unpatched-fonts/NerdFontsSymbolsOnly/NerdFontsSymbols\ 2048\ EM\ Nerd\ Font\ Complete\ Blank.sfd
> ```

Cloning the repo even with `depth 1` is rather time consuming. I would recomment using the [docker image](https://hub.docker.com/r/nerdfonts/patcher) (well, that needs some docker setup), or

```diff
 sudo pacman -S fontforge
 cd /tmp
-git clone --depth 1 https://github.com/ryanoasis/nerd-fonts.git
+mkdir nerd-fonts
 cd nerd-fonts
+curl -LO https://github.com/ryanoasis/nerd-fonts/raw/master/FontPatcher.zip 
+unzip FontPatcher.zip
 ./font-patcher -s -ext ttf -c -out ~/.local/share/fonts/ src/unpatched-fonts/NerdFontsSymbolsOnly/NerdFontsSymbols\ 2048\ EM\ Nerd\ Font\ Complete\ Blank.sfd
```

ammgws:
> @MaxVerevkin, @ammgws,
> 
> Anything that should be done before this is merged?

I was thinking perhaps we should instead keep `material-nf` pointed towards the latest version, and instead introduce `material-nf2.3` as a legacy option instead.

bim9262:
We may want to wait until the next nerd-fonts release so that we don't have complaints about the small icons. (I imagine most people haven't installed them from src)

bim9262:
Just saw https://github.com/greshake/i3status-rust/blob/4e45727dbb078f080d636207089ba42ca3928955/files/icons/material-nf.toml#L73-L79. We might just end up with this commit done piecemeal. Do we want to keep `nf-md-*` out of `material-nf.toml` still? It feels like we'll muddle the line between `material-nf2.3.toml` and `material-nf.toml`.

MaxVerevkin:
>  Do we want to keep nf-md-* out of material-nf.toml still?

If I understand it correctly, the reason we don't just switch to nf-md-* is icons' size. I didn't notice any scaling issues with the wifi icons, so I added them. If the scaling issues will be resolved, I see no point in splitting `material-nf` in two themes.

ammgws:
The split was aimed at people using specific distros where they might have nerdfonts in the repo but not i3status-rust, and end up with the former lagging behind the latter which would be updated by the user. But I just checked the nerd-fonts repo and it seems like it's not even in Debian or Ubuntu repos (the usual suspects) so maybe we don't need to keep the legacy icon set?

bim9262:
All of the codepoints we're using (both nf-mdi and nf-md) are available in 2.3, so as long as the major repos that have nerdfonts are at least that up to date it shouldn't be required to keep two icon sets around.

ammgws:
Ok that makes things simpler then

bim9262:
v3.0.0 has been released, congratulations @Finii!

MaxVerevkin:
Yay, `Nerd Font Propo` makes spacing even without `icons_format` hacks :tada:

PR #1733, opened 2023-02-19
Fix UPower when device is missing

### LINKED ISSUE #1495: v0.30: battery block regression when device is missing (UPower)
When device is not available the block errors, but it should just be hidden since `hide_missing = true`:

image

```toml
[[block]]
block = "battery"
driver = "upower"
device = "battery_ps_controller_battery_d0obcoc1oe6o57o5c"
format = "PS5: $percentage|PS5: N/A"
full_format = "PS5: $percentage"
hide_missing = true
[block.theme_overrides]
alternating_tint_bg = "#FFFFFFFF"
alternating_tint_fg = "#FFFFFFFF"
# TODO: change to confirm menu
[[block.click]]
button = "left"
cmd = "allow_only_one_instance sway_setup_tv_gaming"
```

bim9262:
Fixes #1495

### DIFF
--- a/src/blocks/battery/upower.rs
+++ b/src/blocks/battery/upower.rs
@@ -1,5 +1,6 @@
+use tokio::try_join;
 use zbus::fdo::{PropertiesChangedStream, PropertiesProxy};
-use zbus::{zvariant, Connection, MatchRule, MessageStream};
+use zbus::{zvariant, Connection};
 use zvariant::{ObjectPath, Structure, Value};
 
 use super::{BatteryDevice, BatteryInfo, BatteryStatus, DeviceName};
@@ -14,7 +15,7 @@ struct DeviceConnection {
 
 impl DeviceConnection {
     async fn new(dbus_conn: &Connection, device: &DeviceName) -> Result<Option<Self>> {
-        let device_connection_info = if device.exact() == Some("DisplayDevice") {
+        let device_conn_info = if device.exact() == Some("DisplayDevice") {
             let path: ObjectPath = "/org/freedesktop/UPower/devices/DisplayDevice"
                 .try_into()
                 .unwrap();
@@ -59,7 +60,7 @@ impl DeviceConnection {
             res
         };
 
-        Ok(match device_connection_info {
+        Ok(match device_conn_info {
             Some((device_path, device_proxy)) => {
                 let changes = PropertiesProxy::builder(dbus_conn)
                     .destination("org.freedesktop.UPower")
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
It looks like the block still errors when device is not found during startup?

bim9262:
Yes, it will still error if the device is not present at startup. Any ideas on how to avoid that while still keeping the regex matching on the device name?

MaxVerevkin:
> Any ideas on how to avoid that while still keeping the regex matching on the device name?

I think `bluetooth` does something similar.

MaxVerevkin:
There is also `battery/sysfs.rs` which implements battery re-scanning.

src/blocks/battery/upower.rs
```diff
@@ -57,86 +56,176 @@ impl Device {
...
+            }
+            None => {
+                if let Some(msg) = self.device_added_stream.next().await {
+                    let msg = msg.unwrap();
+                    let body: Structure = msg.body().unwrap();
+                    let fields = body.fields();
+                    if let Some(device_connection) =
+                        DeviceConnection::new(&self.dbus_conn, &self.device).await?
+                    {
+                        if fields[0] == Value::ObjectPath(device_connection.device_path.clone()) {
```
MaxVerevkin, line 221: Is this check actually necessary? The fact that `DeviceConnection::new` returned `Some` is enough I think.

src/blocks/battery/upower.rs
```diff
@@ -57,86 +56,176 @@ impl Device {
...
+        }
     }
 
     async fn wait_for_change(&mut self) -> Result<()> {
-        self.changes.next().await;
+        match &mut self.device_conn {
+            Some(device_connection) => {
+                select! {
+                    _ = device_connection.changes.next() => {},
+                    Some(msg)= self.device_removed_stream.next() => {
```
MaxVerevkin, line 202: I think we should also call `self.device_added_stream.next()` and discard its result, although I am not sure what happens if `MessageStream` runs out of its buffer space. Similarly `self.device_removed_stream.next()` should be called even if `self.device_conn` is `None`.

src/blocks/battery/upower.rs
```diff
@@ -57,86 +56,176 @@ impl Device {
...
-            _ => BatteryStatus::Unknown,
-        };
+        match &self.device_conn {
+            None => Ok(None),
+            Some(device_connection) => {
+                let capacity = device_connection
+                    .device_proxy
+                    .percentage()
+                    .await
+                    .error("Failed to get capacity")?;
```
MaxVerevkin, line 147: If this error happens it probably means that device suddenly became unavailable. This can be either just assumed (bluetooth block [does this](https://github.com/greshake/i3status-rust/blob/c3fc0f0d50871089c6718d335c1234533cd4ecbd/src/blocks/bluetooth.rs#L259-L267)) or the error variant could be checked (I haven't checked which error variant is returned in this case).

src/blocks/battery/upower.rs
```diff
@@ -57,86 +56,176 @@ impl Device {
...
+                .interface("org.freedesktop.UPower")
+                .and_then(|x| x.path("/org/freedesktop/UPower"))
+                .and_then(|x| x.member("DeviceAdded"))
+                .unwrap()
+                .build(),
+            &dbus_conn,
+            None,
+        )
+        .await
+        .error("Failed to add match rule")?;
```
MaxVerevkin, line 111: Can we use `zbus` generated proxy instead?

MaxVerevkin: Like this:

```rust
        let upower_proxy = UPowerProxy::new(&dbus_conn)
            .await
            .error("Failed to create UPowerProxy")?;

        let device_added_stream = upower_proxy
            .receive_device_added()
            .await
            .error("Failed to create stream")?;

        let device_removed_stream = upower_proxy
            .receive_device_removed()
            .await
            .error("Failed to create stream")?;
```

src/blocks/battery/upower.rs
```diff
@@ -57,86 +57,138 @@ impl Device {
...
+                select! {
+                    _ = device_conn.changes.next() => {},
+                    _ = self.device_added_stream.next() => {},
+                    Some(msg) = self.device_removed_stream.next() => {
+                        let body:  Structure = msg.body().unwrap();
+                        let fields = body.fields();
+                        if fields[0]==Value::ObjectPath(device_conn.device_path.clone()){
+                            self.device_conn = None;
+                        }
+
```
MaxVerevkin, line 174: `zbus` also generates some code for this

```suggestion
                        let args = msg.args().unwrap();
                        if args.device().as_ref() == device_connection.device_path {
                            self.device_conn = None;
                        }
```

src/blocks/battery/upower.rs
```diff
@@ -57,86 +57,138 @@ impl Device {
...
+        }
     }
 
     async fn wait_for_change(&mut self) -> Result<()> {
-        self.changes.next().await;
+        match &mut self.device_conn {
+            Some(device_conn) => {
+                select! {
+                    _ = device_conn.changes.next() => {},
+                    _ = self.device_added_stream.next() => {},
```
MaxVerevkin, line 167: As is, this will trigger block updates on "device_added" event even though it is irrelevant. So the entire `select!` must be wrapped in a `loop` which will break only when a relevant event is received.

PR #1738, opened 2023-02-20
Fix typos

bim9262:
Might want to consider vending a pre commit hook to help prevent so many typos in the future or a github check.

### DIFF
--- a/src/blocks/cpu.rs
+++ b/src/blocks/cpu.rs
@@ -201,11 +201,11 @@ impl CpuTime {
         let idle = u64::from_str(s.next()?).ok()?;
         let iowait = u64::from_str(s.next()?).ok()?;
         let irq = u64::from_str(s.next()?).ok()?;
-        let softirq = u64::from_str(s.next()?).ok()?;
+        let soft_irq = u64::from_str(s.next()?).ok()?;
 
         Some(Self {
             idle: idle + iowait,
-            non_idle: user + nice + system + irq + softirq,
+            non_idle: user + nice + system + irq + soft_irq,
         })
     }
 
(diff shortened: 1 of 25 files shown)


ammgws:
Not sure the .vscode specific stuff should be in the repo.

Can you add some info to CONTRIBUTING.md as well?

bim9262:
I'd be interested in hearing if there's another spell checker you'd prefer, before I add documentation on whatever tool seems like the best fit. I did take a look at https://crates.io/crates/cargo-spellcheck, but found the setup (hunspell) to be more difficult. cspell can be used in vscode (as an extension) or as a standalone utility from npm.

cfsmp3:
Personally I think a PR that fixes existing typos would be good to merge, and table the rest for now.

I suspect many contributors don't use vscode at least for this specific project.

bim9262:
Removed cspell config.

src/blocks/cpu.rs
```diff
@@ -201,11 +201,11 @@ impl CpuTime {
         let idle = u64::from_str(s.next()?).ok()?;
         let iowait = u64::from_str(s.next()?).ok()?;
         let irq = u64::from_str(s.next()?).ok()?;
-        let softirq = u64::from_str(s.next()?).ok()?;
+        let soft_irq = u64::from_str(s.next()?).ok()?;
 
         Some(Self {
             idle: idle + iowait,
-            non_idle: user + nice + system + irq + softirq,
+            non_idle: user + nice + system + irq + soft_irq,
```
cfsmp3, line 208: I'd leave softirq as written. See for example:

https://archive.kernel.org/oldlinux/htmldocs/kernel-hacking/basics-softirqs.html

bim9262: Thanks the extra context!

PR #1747, opened 2023-02-22
Look for assets  (themes / icons) in a directory relative to the running binary

cfsmp3:
I found my freshly built from source i3status-rs looking for .toml files in places that are correct for proper installs but incorrect for people (i.e developers) testing in-place from a cloned repo .

This diffs just checks for files/ relative to the running binary so devs can trust that they're not using stale resource from previous installs.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,3 +1,4 @@
+use std::env::current_exe;
 use std::fs::File;
 use std::io::{self, BufReader, Read};
 use std::path::{Path, PathBuf};
@@ -26,6 +27,20 @@ pub fn find_file(file: &str, subdir: Option<&str>, extension: Option<&str>) -> O
     // Try full path
     if file.exists() {
         return Some(file);
+    }
+
+    // Developer testing live? Check project path
+    if let Ok(exe_path) = current_exe() {
+        if let Some(parent) = exe_path.parent() {
+            let mut path = parent.join("..").join("..").join("files");
+            if let Some(subdir) = subdir {
+                path.push(subdir);
+            }
+            path.push(&file);
+            if path.exists() {
+                return Some(path);
+            }
+        }
     }
 
     // Try XDG_CONFIG_HOME (e.g. `~/.config`)


MaxVerevkin:
Should we check that `parent` is either "release" or "debug" and `parent.parent` is "target"?

Also an alternative could be a `I3RS_RUNTIME` env var or similar which would be checked before `XDG_CONFIG_HOME`.

MaxVerevkin:
Or we can rename `files` to `files/i3status-rust` so that `$XDG_DATA_HOME` could be used.

ammgws:
> Or we can rename `files` to `files/i3status-rust` so that `$XDG_DATA_HOME` could be used.

I like this idea

cfsmp3:
> > Or we can rename `files` to `files/i3status-rust` so that `$XDG_DATA_HOME` could be used.
> 
> I like this idea

How does this help when i3status-rust is cloned in a random location, built, and just run from there? For example in my case it's `~/codebase/i3status-rs` . I was getting errors about the "bar" icon not being defined - reason being that it was using a .toml location from some old install that was using the official install places.

I understand that will be the case for most people, but not devs (well, I guess, I don't know :-) )

MaxVerevkin:
> How does this help when i3status-rust is cloned in a random location, built, and just run from there?

You could run `XDG_DATA_HOME=files cargo r`

cfsmp3:
> You could run `XDG_DATA_HOME=files cargo r`

I don't love this - I mean, that stuff is not something a casual dev cloning and building would know to do :-)

tpoliaw:
Themes/icons are already looked for relative to the current working directory before trying in the `XDG_*_DIR` locations. For testing, you can set (eg) `theme = "files/icons/awesome6.toml"` in the config file you're testing with and it'll pick them up without any changes when you run `cargo run`.

cfsmp3:
Yes but you don't have that in the config, do you? You just have
"awesome5".

tpoliaw:
Fair enough but (for an outsider's two cents) if you're mixing a development build with a live config file, I'd second the `XDG_DATA_HOME` approach.

cfsmp3:
> Fair enough but (for an outsider's two cents) if you're mixing a development build with a live config file, I'd second the `XDG_DATA_HOME` approach.

It's not really about dev build but about someone testing a new version from clone when an old one installed normally (meaning in the standard places) exists in the system. 

Either we check that the version of things is what we expect or we look for the files in the right locations.

MaxVerevkin:
@cfsmp3 Do you know any other project that does what you propose? AFAIK everyone just expects you to either install the runtime files or point to them using an env var.

cfsmp3:
> @cfsmp3 Do you know any other project that does what you propose? AFAIK everyone just expects you to either install the runtime files or point to them using an env var.

I don't. This is the only I care about these days though :-)

I guess it's fair. I'm closing the PR but when someone says "I'm getting icon not found errors" remember this :-)

PR #1754, opened 2023-02-22
fix typo for 'play' and 'preferred'

leongross:
small typo fixes in `music.rs` block

### DIFF
--- a/src/blocks/music.rs
+++ b/src/blocks/music.rs
@@ -84,7 +84,7 @@
 //! interface_name_exclude = [".*kdeconnect.*", "mpd"]
 //! ```
 //!
-//! Click anywhere to paly/pause:
+//! Click anywhere to play/pause:
 //!
 //! ```toml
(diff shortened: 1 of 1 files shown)


src/blocks/music.rs
```diff
@@ -94,7 +94,7 @@
 //! action = "play_pause"
 //! ```
 //!
-//! Scroll to change the player volume, use the forward and back buttons to seek:
+//! Scroll to change the player volume, use the forward and backward buttons to seek:
```
MaxVerevkin, line 97: For some reason, `back` is the actual name of that button, so I think it's better to leave this one.

leongross: Okay, then maybe this is a separate naming issue we should discuss another time. For now, I will revert his change.

MaxVerevkin (approved):

MaxVerevkin:
Thanks for contribution!

PR #1759, opened 2023-02-23
Implement cycling through timezones in "time" block on click

### LINKED ISSUE #1658: Cycle through timezones in "time" block on click
Unless I didn't read through the docs close enough, this currently isn't possible

omkar-mohanty:
Fixes #1658

### DIFF
--- a/src/blocks/time.rs
+++ b/src/blocks/time.rs
@@ -13,6 +13,10 @@
 //! `icon`        | A static icon                               | Icon     | -
 //! `timestamp`   | The current time                            | Datetime | -
 //!
+//! Action   | Default button
+//! ---------|---------------
+//! `next_timezone` | Left
+//!
 //! # Example
 //!
 //! ```toml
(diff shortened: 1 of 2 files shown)


src/blocks/time.rs
```diff
@@ -38,17 +38,29 @@ pub struct Config {
     format: FormatConfig,
     #[default(1.into())]
     interval: Seconds,
-    timezone: Option<Tz>,
+    timezones: Option<Vec<Tz>>,
```
MaxVerevkin, line 41: This is a breaking change but it doesn't need to be. See #1452 for very similar example. TL;DR: You can create an enum `Timezone` which is either `Tz` or `Vec<Tz>`. `#[serde(untagged)]` will handle the deserialization.

src/blocks/time.rs
```diff
@@ -38,17 +38,29 @@ pub struct Config {
...
 }
 
 pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
+    // "toggle" changes the current displayed timezone to the timezone next in the list.
+    // for example in the list ["Asia/Kolkata", "Europe/London"] if the current displayed timezone is
+    // "Europe/London" on "toggle" the timezone changes to "Asia/Kolkata"
+    api.set_default_actions(&[(MouseButton::Left, None, "toggle")]).await?;
```
MaxVerevkin, line 48: Perhaps `next_timezone` would be a better name? New actions can be added later and "toggle" is a bit too generic.

Also please document this action. For example, see [net docs](https://greshake.github.io/i3status-rust/i3status_rs/blocks/net/index.html).

omkar-mohanty: Thanks for the review! I added the documentation for `next_timezone` action.

omkar-mohanty:
@MaxVerevkin I think i have implemented all the suggestions can I get a review? Thanks!

MaxVerevkin (approved):

PR #1764, opened 2023-02-23
sound: Use Regex as key in mappings

### LINKED ISSUE #1525: sound: handle both pipewire and pulseaudio device names when mapping
Pulseaudio sep char is a colon `_`, where as in Pipewire it is an underscore `:`.
Thus when using `[block.mappings]` on a system where you might switch between the two backends from time to time, you need to specify two different mappings for the same device:

```toml
[[block]]
block = "sound"
driver = "pulseaudio"
[block.mappings]
# pulseaudio
"alsa_output.pci-0000_00_1b.0.analog-stereo" = "headphones"
# pipewire-pulse
"alsa_output.pci-0000:00:1b.0.analog-stereo" = "headphones"
```

Instead we could just replace all colons with underscores in output_name here so either way works:

https://github.com/greshake/i3status-rust/blob/2aca056dafe58693009d353ecbdd010291803dbf/src/blocks/sound.rs#L158-L163

elbertronnie:
Fixes #1525 

## Changes
- Added a new dependency serde_with. It is used to serialize and deserialize Vec of tuples as Map.
- `Config` struct stores a Vec of tuples instead of HashMap since ordering of regex will be important.
- A new configuration property `mappings_use_regex` is added. It will let `mappings` match using regex
- The replacement is regex aware and can contain capture groups


ammgws:
My initial idea was to just convert the device name and mapping name to the same format (by replacing either `:` or `_`) so using regex for that seems a bit overkill.

But if we instead accept regex for the mapping itself, configs could become a lot more verstatile.

For example my config is currently quite long, but with regex it would probably only be a couple lines:
```toml
[block.mappings]
# pulseaudio
"alsa_output.pci-0000_00_1b.0.analog-stereo" = ""
"alsa_output.pci-0000_00_1f.3.analog-stereo" = ""
"alsa_output.pci-0000_00_1f.3.analog-stereo.3" = ""
"alsa_output.usb-Harman_Multimedia_JBL_Pebbles_1.0.0-00.analog-stereo" = "U"
"alsa_output.usb-Creative_Technology_Ltd_Sound_Blaster_Play__3_00145592-00.analog-stereo" = "D"
"bluez_sink.20_74_CF_26_25_D0.a2dp_sink" = "BT"
"alsa_output.pci-0000_01_00.1.hdmi-stereo-extra2" = "TV"
"alsa_output.pci-0000_01_00.1.hdmi-stereo-extra3" = "TV"
"alsa_output.pci-0000_01_00.1.hdmi-stereo-extra4" = "TV"
"alsa_output.pci-0000_01_00.1.hdmi-stereo-extra5" = "TV"
# pipewire-pulse
"alsa_output.pci-0000:00:1b.0.analog-stereo" = ""
"alsa_output.pci-0000:00:1f.0.analog-stereo.3" = ""
```

ammgws (approved):
Need to update docs too

jak0b:
Perhaps it would be a better idea to simply leave `mappings` as a collection of keys and values (HashMap), which wouldn't introduce any breaking changes? `mappigs_regex` would then be added for regular expression mappings, so a sample config file with mappings and regular expression mappings would look like that:
```toml
[block.mappings]
"bluez_output.08_EB_ED_E0_77_37.1" = "BLT"
"alsa_output.usb-bestechnic_HUAWEI_USB-C_HEADSET_02963100000000000000000000000-00.analog-stereo" = "USB-C"
[block.mappings_regex]
"alsa_output\\.pci-0000_00_1f\\.3-platform-skl_hda_dsp_generic\\.?\\d*\\.HiFi__hw_sofhdadsp_3__sink" = "DP-3"
"alsa_output\\.pci-0000_00_1f\\.3-platform-skl_hda_dsp_generic\\.?\\d*\\.HiFi__hw_sofhdadsp_4__sink" = "DP-4"
"alsa_output\\.pci-0000_00_1f\\.3-platform-skl_hda_dsp_generic\\.?\\d*\\.HiFi__hw_sofhdadsp_5__sink" = "DP-5"
"alsa_output\\.pci-0000_00_1f\\.3-platform-skl_hda_dsp_generic\\.?\\d*\\.HiFi__hw_sofhdadsp__sink" = "In"
"alsa_output\\.usb-MOTU_M2_M2\\d{10}-00\\.?\\d*\\.HiFi__hw_M2__sink" = "M2 (HiFi)"
"alsa_output\\.usb-MOTU_M2_M2\\d{10}-00\\.?\\d*\\.Direct__hw_M2__sink" = "M2 (Direct)"
```
So the config struct would look like that:
```rust
#[serde_as]
#[serde(default)]
pub struct Config {
    format: FormatConfig,
    driver: KeyboardLayoutDriver,
    #[default(60.into())]
    interval: Seconds,
    sway_kb_identifier: Option<String>,
    mappings: Option<HashMap<String, String>>,
    #[serde_as(as="Option<Map<RegexAsString, _>>")]
    mappings_regex: Option<Vec<(Regex, String)>>,
}
```
and the `loop` would first check for `HashMap` mappings and after that for possible regular expression matches:
```rust
        if let Some(mappings) = &config.mappings {
            if let Some(mapped) = mappings.get(&output_name) {
                output_name = mapped.clone();
            } 
        }

        if let Some(mappings_regex) = &config.mappings_regex {
            if let Some((regex, mapped_regex)) = mappings_regex.iter().find(|&(regex, _)| regex.is_match(&output_name)) {
                output_name = regex.replace(&output_name, mapped_regex).into_owned();
            }
        }
```
, which would effectively make regular expressions override `HashMap` mappings in case of a match.

ammgws:
That's a good point, bit wary of breaking changes now that v0.30 is out..

@MaxVerevkin @bim9262 What do you think about splitting it up?

bim9262:
I'd think you could leave everything in `mappings` and bite the cost of convering them to regex once. The only special symbol in the names should be a `.`, but even if it's not escaped it will match the literal `.`.

(I'm traveling so don't have access to a computer to confirm anything).

MaxVerevkin:
> The only special symbol in the names should be a `.`, but even if it's not escaped it will match the literal `.`

Yeah, is this a breaking change in practice?

bim9262:
> > The only special symbol in the names should be a `.`, but even if it's not escaped it will match the literal `.`
> 
> Yeah, is this a breaking change in practice?

I'd say that's not a breaking change, although it's a new behavior.

jak0b:
You're absolutely right @bim9262 using regular expressions in `mappings` in the sound module is very unlikely to break something as sound card names never contain any regular expression reserved characters with the notable exception of `.`, but as @elbertronnie correctly mentioned `mappings` are also used in the `keyboard_layout` module and here the introduction of of regex would constitute a breaking change in my use case.
On sway the keyboard layout I use is called `English (intl., with AltGr dead keys)` and it being a regular expression wouldn't match itself because of the capturing group characters `(` and `)`.

So IMHO we should either allow a breaking change to occur in the both modules for the sake of consistency or add just add `mappigs_regex` fields in both modules' config structs.

I also think that having regex in the `keyboard_layout` modules is quite useful, as it allows for mapping multiple keyboard layouts with a single emoji flag as show here:
```toml
[[block]]
block = "keyboard_layout"
driver = "sway"
format = "$layout"
[block.mappings_regex]
"Polish\\.*" = "🇵🇱"
"English\\.*" = "🇺🇸"
```

ammgws:
Another option to avoid breaking/suprising config behaviour could be to add an option like "mappings_use_regex" to thte blocks in question and default it to `false`.

jak0b:
> Another option to avoid breaking/surprising config behaviour could be to add an option like "mappings_use_regex" to thte blocks in question and default it to `false`.

You mean something like that:
```toml
[[block]]
block = "sound"
step_width = 3
device_kind = "sink"
format = "$icon $output_name $volume"
mappings_use_regex = true
[block.mappings]
"bluez_output\\.08_EB_ED_E0_77_37\\.1" = "BLT"
"alsa_output\\.usb-bestechnic_HUAWEI_USB-C_HEADSET_02963100000000000000000000000-00\\.analog-stereo" = "USB-C"
"alsa_output\\.pci-0000_00_1f\\.3-platform-skl_hda_dsp_generic\\.?\\d*\\.HiFi__hw_sofhdadsp_3__sink" = "DP-3"
"alsa_output\\.pci-0000_00_1f\\.3-platform-skl_hda_dsp_generic\\.?\\d*\\.HiFi__hw_sofhdadsp_4__sink" = "DP-4"
"alsa_output\\.pci-0000_00_1f\\.3-platform-skl_hda_dsp_generic\\.?\\d*\\.HiFi__hw_sofhdadsp_5__sink" = "DP-5"
"alsa_output\\.pci-0000_00_1f\\.3-platform-skl_hda_dsp_generic\\.?\\d*\\.HiFi__hw_sofhdadsp__sink" = "In"
"alsa_output\\.usb-MOTU_M2_M2\\d{10}-00\\.?\\d*\\.HiFi__hw_M2__sink" = "M2 (HiFi)"
"alsa_output\\.usb-MOTU_M2_M2\\d{10}-00\\.?\\d*\\.Direct__hw_M2__sink" = "M2 (Direct)"
```
But wouldn't this be less flexible, since it would loose the possibility of having non-regexp entries?

MaxVerevkin:
> But wouldn't this be less flexible, since it would loose the possibility of having non-regexp entries?

I thought the idea was to eventually have only regexes. Like battery and net's "device" option is only regex.

bim9262:
> > But wouldn't this be less flexible, since it would loose the possibility of having non-regexp entries?
> 
> I thought the idea was to eventually have only regexes. Like battery and net's "device" option is only regex.

And backlight's `device` soon too 🤞

elbertronnie:
I have added `mappings_use_regex`.
For now I am assuming that not having non-regex entries is won't be a problem since they could be converted to regex somewhat easily.
Also during the release of 0.31, we should set the default to `true` if the aim is to have regexes only.

MaxVerevkin:
> Also during the release of 0.31, we should set the default to true if the aim is to have regexes only.

And then remove it in 0.32.

src/blocks/sound.rs
```diff
@@ -52,13 +52,14 @@
...
 //! `mappings` | Map `output_name` to custom name. | `None`
+//! `mappings_use_regex` | Let `mappings` match using regex instead of string equality. | `false`
 //!
 //! Placeholder          | Value                             | Type   | Unit
 //! ---------------------|-----------------------------------|--------|---------------
 //! `icon`               | Icon based on volume              | Icon   | -
 //! `volume`             | Current volume. Missing if muted. | Number | %
 //! `output_name`        | PulseAudio or ALSA device name    | Text   | -
-//! `output_description` | PulseAudio device description, will fallback to `output_name` if no description is available and will be overwritten by mappings (mappings will still use `output_name`) | Text | -
+//! `output_description` | PulseAudio device description, will fallback to `output_name` if no description is available and will be overwritten by mappings or mappings_regex (mappings and mappings_regex will still use `output_name`) | Text | -
```
ammgws, line 62: `mappings_regex` still remaining here

ammgws (approved):

src/blocks/sound.rs
```diff
@@ -174,16 +185,39 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
         )?),
     };
 
+    let mappings = config.mappings.map(|m| {
+        if config.mappings_use_regex {
+            Mappings::Regex(
+                m.into_iter()
+                    .filter_map(|(key, val)| Some((Regex::new(&key).ok()?, val)))
+                    .collect(),
```
MaxVerevkin, line 193: This silently skips invalid regexes. Is it intentional? (you can `.collect()` an iterator of results into `Result`)

elbertronnie: Actually it was intentional. My reasoning was that should we crash the whole thing just becuase one of the regex was incorrect?
And even if we get a result back, are we allowed to just `unwrap()` it? Otherwise the whole mapping will have to be skipped which is worse in my opinion.

MaxVerevkin: > should we crash the whole thing just becuase one of the regex was incorrect?

Not the whole thing, only this block.

> are we allowed to just unwrap() it?

No. Return from the block with this error.

I think it's better to let users know they have an error in their config rather than silently skip them. I imagine it being annoying to debug...

MaxVerevkin:
> The replacement is regex aware and can contain capture groups

Please mention this in the docs too. And by the way, just curious, do you have a usecase for this (replacement with capture groups) which you can share?

elbertronnie:
> Please mention this in the docs too. And by the way, just curious, do you have a usecase for this (replacement with capture groups) which you can share?

None in my mind so far. I initially thought if people wanted to reformat the name or cut some parts of the name, this might be useful.

MaxVerevkin (approved):

PR #1767, opened 2023-02-23
cpu: make frequency optional

kujeger:
This avoids a panic on systems where cpu frequency is unavailable.

(Such as on this M1 mac running Asahi Linux)

### DIFF
--- a/src/blocks/cpu.rs
+++ b/src/blocks/cpu.rs
@@ -117,7 +117,7 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
             "barchart" => Value::text(barchart),
             "frequency" => Value::hertz(freq_avg),
             "utilization" => Value::percents(utilization_avg * 100.),
-            "max_frequency" => Value::hertz(freqs.iter().copied().max_by(f64::total_cmp).unwrap()),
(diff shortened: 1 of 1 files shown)


src/blocks/cpu.rs
```diff
@@ -117,7 +117,7 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
             "barchart" => Value::text(barchart),
             "frequency" => Value::hertz(freq_avg),
             "utilization" => Value::percents(utilization_avg * 100.),
-            "max_frequency" => Value::hertz(freqs.iter().copied().max_by(f64::total_cmp).unwrap()),
+            "max_frequency" => Value::hertz(freqs.iter().copied().max_by(f64::total_cmp).unwrap_or_default()),
```
MaxVerevkin, line 120: I would prefer to be explicit and make `frequency` and `max_frequency` optional (similar to `boost`, with a note in the docs). I.e.

```suggestion
            [if !freqs.is_empty()] "max_frequency" => Value::hertz(freqs.iter().copied().max_by(f64::total_cmp).unwrap()),
```

kujeger: Thanks for the suggestion! I've implemented it and also updated the relevant doc. :tada:

MaxVerevkin (approved):

MaxVerevkin:
Thanks!

PR #1772, opened 2023-02-24
Fix more typos

bim9262:
I was looking into another spell check tool (https://crates.io/crates/typos) and found some more errors.

https://pre-commit.com seem like it will be a useful tool for making sure collaborators have code.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,7 +70,7 @@ to do this is by running `cargo fmt`.
 
 ## Maintainership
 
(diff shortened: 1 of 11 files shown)


MaxVerevkin:
I never used pre-commit... Will it require all contributors to install something additional?

bim9262:
> I never used pre-commit... Will it require all contributors to install something additional?

Only if they wanted to run the checks locally, it can also be run as a GitHub action https://github.com/pre-commit/action. I've seen it used by projects like clap.

PR #1776, opened 2023-02-24
blocks/watson: Avoid error on race condition while reading state

### LINKED ISSUE #1775: Watson stuck in "Unable to deserialize state"
While changing watson's state it happens sometimes, that the block ends up with `"Unable to deserialize state"` and no action in watson resolves that. The block is stuck then and the only way out is restarting `i3status-rust`.

freswa:
While reading state, neither we nor watson seem to perfom locking of the state file. Thus, parsing the state file can fail unexpectedly. This would lead to an error in the block, making it unusable until i3status-rust is restartet. This change introduces the empty state as a default to avoid this behavior.

fixes #1775

### DIFF
--- a/src/blocks/watson.rs
+++ b/src/blocks/watson.rs
@@ -90,7 +90,7 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
         let state = read_to_string(&state_path)
             .await
             .error("Failed to read state file")?;
-        let state = serde_json::from_str(&state).error("Unable to deserialize state")?;
+        let state = serde_json::from_str(&state).unwrap_or(WatsonState::Idle {});
         match state {
             state @ WatsonState::Active { .. } => {
                 widget.state = State::Good;


MaxVerevkin:
This wasn't happening in v0.22?

freswa:
It happened with 0.22 as well but not as frequent.

MaxVerevkin:
Maybe we should add a delay between inotify update and the read? And/or read again on failure?

freswa:
The file is read once in the timer interval anyways, that's why I decided to use this minimal approach.
Retrying reads is another rabbithole imo. How many retries? What if the interval is lower than $interval?
But if that's a requirement, I can go that route.

Also from a technical perspecive the empty state is correct here, since that's the default behavior of watson as well.

The sane approach would be either a watson api or a file locking approach. But both would require changes in watson itself.

src/blocks/watson.rs
```diff
@@ -90,7 +90,7 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
         let state = read_to_string(&state_path)
             .await
             .error("Failed to read state file")?;
-        let state = serde_json::from_str(&state).error("Unable to deserialize state")?;
+        let state = serde_json::from_str(&state).or(serde_json::from_str("{}")).error("Unable to deserialize state")?;
```
MaxVerevkin, line 93: Instead of parsing a static string, it's better to `.unwrap_or` with a default value

MaxVerevkin:
> But if that's a requirement, I can go that route.

No, just a thought.

By the way, I haven't looked into watson's code, but does this mean it's writing the file by parts? I mean, if it was just a single write, then this wouldn't happen?

freswa:
They use this[ temporary file mechanism](https://github.com/TailorDev/Watson/blob/master/watson/utils.py#L217). Which is called [here](https://github.com/TailorDev/Watson/blob/master/watson/watson.py#L131). Looking at the code I think we can't read half way written files here.
But since I'm missing a good way to reproduce and the handling on our side is subpar since it requires restarting, I'd still like to set the default here until we  found a good way to handle watson state.

MaxVerevkin:
How often does this happen?

I run 

```zsh
for i in {1..1000}
do
watson start y && sleep 0.1 && watson stop && sleep 0.1
done
```

and still couldn't reproduce this. I was just interested in what exactly is read by the block that can't be parsed. Could you capture the value of `state` when it is unparseable?

freswa:
I replaced the line with `let state = serde_json::from_str(&state).error(format!("Failed to parse state: {}", &state))?;` and played a bit with watson.
image
The mail icon is the next block already. So I guess the file is empty at the time of reading it.

freswa:
> How often does this happen?

I'd say every other day at the moment, while having 3-10 tasks a day.

MaxVerevkin:
Also, does it happen when you start task or stop?

> Try using watson stop and watson restart instead of start and stop. That seems to trigger the issue way better

Still couldn't reproduce

MaxVerevkin:
By the way, what version of watson do you use?

freswa:
> Also, does it happen when you start task or stop?

I only remember this issue ocurring on `stop`.

> By the way, what version of watson do you use?

2.1.0

MaxVerevkin (approved):
If this workaround works good enough for you, I'm okay with it.

PR #1779, opened 2023-02-24
Stop find_files requiring a specific extension

tpoliaw:
I have my config in `~/.config/i3/status` next to i3 config and the latest upgrade fails with `Configuration file '~/.config/i3/status' not found`, and it took a while to track down the cause. It looks like a recent change forces all files to have a `'.toml` extension even if they exist without it or have a different extension.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -17,13 +17,7 @@ use crate::errors::*;
 ///
 /// Automatically append an extension if not presented.
 pub fn find_file(file: &str, subdir: Option<&str>, extension: Option<&str>) -> Option<PathBuf> {
-    // Set (or update) the extension
-    let mut file = PathBuf::from(file);
-    if let Some(extension) = extension {
-        file.set_extension(extension);
-    }
-
-    // Try full path
+    let file = PathBuf::from(file);
     if file.exists() {
         return Some(file);
     }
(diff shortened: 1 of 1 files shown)


src/util.rs
```diff
@@ -58,13 +52,28 @@ pub fn find_file(file: &str, subdir: Option<&str>, extension: Option<&str>) -> O
...
-    if usr_share_path.exists() {
-        return Some(usr_share_path);
+    if let Some(file) = exists_with_extension(&usr_share_path, extension) {
+        return Some(file);
     }
 
     None
 }
 
+fn exists_with_extension<'a>(file: &'a Path, extension: Option<&str>) -> Option<PathBuf> {
```
tpoliaw, line 62: Not sure about the name, almost wants to be something like `exists_with_or_without_extension` but less of a mouthful

src/util.rs
```diff
@@ -17,14 +17,8 @@ use crate::errors::*;
...
-    // Set (or update) the extension
-    let mut file = PathBuf::from(file);
-    if let Some(extension) = extension {
-        file.set_extension(extension);
-    }
-
-    // Try full path
-    if file.exists() {
+    let file = PathBuf::from(file);
+    if let Some(file) = exists_with_extension(&file, extension) {
```
MaxVerevkin, line 21: I think it's okay to require full paths to include the extension, so this can be just `file.exists()`.

src/util.rs
MaxVerevkin: Perhaps `exists_with_opt_extension`?

src/util.rs
tpoliaw: That makes sense.

src/util.rs
tpoliaw: Thanks. Feels a bit like it's duplicating the signature but it's better than anything I can come up with.

PR #1780, opened 2023-02-24
Add pre-commit

bim9262:


bim9262:
The pre-commit action until #1772 is not resolved

bim9262:
There are now a few duplicate checks in pre-commit vs rust/rustfmt and rust/clippy.
Also rust/check runs cargo check which covers a subset of what cargo clippy covers. Is there any reason to keep that?

MaxVerevkin (approved):

MaxVerevkin:
> Is there any reason to keep that?

I think no.

_typos.toml
```diff
@@ -0,0 +1,16 @@
+files.extend-exclude = [
+    "target/",
+    "man/*",
+]
+
+[default.extend-identifiers]
+hellow554 = "hellow554"
```
MaxVerevkin, line 7: What is this?

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,13 @@
+name: pre-commit
+
+on: [push, pull_request]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
```
MaxVerevkin, line 8: I think you forgot to install build dependencies.

_typos.toml
bim9262: So I just went back and added a bit of documentation on the difference between identifiers and words. Basically typos thinks hellow is a misspelling of hello, but instead of saying hellow is fine, we say that hellow554 is fine, but still complain about hellow in general. (words can't include numbers, but identifiers can)

.github/workflows/pre-commit.yml
bim9262: pre-commit pulls in all of the dependencies you need to run its checks (not only in github actions but also when you use it locally), and then caches them using `actions/cache@v3`.

bim9262: Oh do you mean?:

```
- name: Get required packages
  run: sudo apt-get install libsensors-dev libssl-dev libpulse-dev libnotmuch-dev
```

MaxVerevkin: I mean dependencies of `i3status-rust`, `sudo apt-get install libsensors-dev libssl-dev libpulse-dev libnotmuch-dev`.

MaxVerevkin:
I think you forgot to remove `rust/clippy`

bim9262:
I added a check for cargo test too.

bim9262:
Sorry for so much churn on this... it seems to be working well now though, including making sure that the man pages are up to date.

MaxVerevkin:
How does manpage generation work? If run locally, it modifies the commit, I suppose. But what if it's run in CI?

MaxVerevkin:
> Sorry for so much churn on this.

Messing with CI is always like this :laughing:

bim9262:
If you run it locally and it differs it aborts the commit so you can review the changes made, before adding the changes and making the commit.
If you fail during CI, then it will show up as failed. For example here is a run I did (with an old version of pandoc that caused it to fail): https://github.com/bim9262/i3status-rust/actions/runs/4271529102/jobs/7436057027

bim9262:
Perhaps a check should be added to man/generate.sh to make sure that version pandoc v3 is being used...

bim9262:
Instead of having just one big run of pre-commit we can actually run each test individually so that everything is run in parallel and feel more like the CI does today.

MaxVerevkin:
So it will force everyone who edits the docs to run manpage generator? Not sure about this...

bim9262:
It would, maybe we can set that test to only run on tagged commits (ie a version release)

bim9262:
Here's an example for release `vFAKE`: https://github.com/bim9262/i3status-rust/actions/runs/4271907305/jobs/7436702832 and https://github.com/bim9262/i3status-rust/actions/runs/4271907304. You can see that I made a commit that changed the docs, but  had a stale manpage. The second one has all of the other checks that would be run on all push/pull_requests.

MaxVerevkin:
I'm not very familiar with git tags. You created tag `vFAKE` and it failed CI. How would you fix it? Do you delete that tag and create it again?

bim9262:
That's a good point. You can push delete a tag, `git push -u origin :tag_name`,  and then create a new tag.
Maybe it would be better to understand the process that you follow when making a release.

bim9262:
Can you explain why you are cautious about having everyone update the manpage when they update the documentation? I think it makes it easier to review changes to the man page if they are made alongside to the changes to the docs.

MaxVerevkin:
> Maybe it would be better to understand the process that you follow when making a release.

I just use GitHub's "create tag on publish"

image

> Can you explain why you are cautious about having everyone update the manpage when they update the documentation?

I'm not against this, just wanted to start a discussion, maybe someone has opinions on this.

bim9262:
@ammgws any though regarding having developers having to update the man page when they update the docs in the source code? It could be easily be tested in a gh action

ammgws:
> Maybe it would be better to understand the process that you follow when making a release.

I think we should focus on this and formalise the release process which includes generating the manpages.

They are autogenerated from the docs and IMO not that easy to read anyway, so wouldn't want to bother contributors with it:
image

bim9262:
Sounds good, for now I'll remove the man page related sections from this PR and we can always bring it back in later. (I'll put those changes in it's own draft PR in case we want to use any of what I wrote so far)

ammgws (approved):

MaxVerevkin:
Yay, fmt check works again :)

PR #1785, opened 2023-02-26
Add back missing default icons

### LINKED ISSUE #1782: Add back volume_muted in src/icons.rs
The volume_muted icon has been removed from src/icons.rs in https://github.com/greshake/i3status-rust/commit/98af6386cfb2021677892404365a0b6c3602709b, and now muting volume with icons set to "none" leads to error "Icon 'volume_muted' not found".

This is a breaking change with icons set to none. Please add volume_muted back to src/icons.rs. (Besides the block crashing, it is also very useful to know if the volume is muted, which worked before when "VOL MUTED" was used as text.

To reproduce:

[icons]
icons = "none"
[[block]]
block = "sound"
\# Breaks with volume muted.

bim9262:
Fixes #1782

`volume_muted` and `microphone_muted` were accidentally removed in 98af638.
xrandr was still using `backlight_full` instead of just `backlight`.


ammgws:
Good catch!

Issue #1786, opened 2023-02-26
Segmentation fault in `sound` and `temperature`


GladOSkar:
After updating to current master (`i3status-rs 0.30.2 (commit 76e19f29 2023-02-22)`) and fixing up my config, i got:

```sh
$ i3status-rs
{"version": 1, "click_events": true}
[
fish: Job 1, 'i3status-rs' terminated by signal SIGSEGV (Address boundary error)
```
sometimes also:
```sh
$ i3status-rs
{"version": 1, "click_events": true}
[
[],
fish: Job 1, 'i3status-rs' terminated by signal SIGSEGV (Address boundary error)
```

Minimal configs to reproduce:
```toml
[[block]]
block = "sound"
```
and
```toml
[[block]]
block = "temperature"
```
both create the segfault, most other blocks don't.

This happens both when compiling in `debug` and in `release` mode.

I attached the LLVM debugger and got this stack trace (`sound` block config):

https://gist.github.com/GladOSkar/3352afe39d26f052d4ba6610a369a5a9#file-i3status-rs_segfault-L139
(Segfault in Thread 3, Line 139 ff.)

Looks like a read from `0x0`. Going further is a bit beyond me, especially with all that threading and stuff, help would be appreciated.

My system:
- Distribution: Void
- Kernel: `Linux [...] 6.1.12_1 #1 SMP PREEMPT_DYNAMIC [...] x86_64 GNU/Linux`
- libc: musl
- DE: sway
- rustc: `1.67.1 (d5a82bbd2 2023-02-07)`
- cargo: `1.67.1 (8ecd4f20a 2023-01-10)`

MaxVerevkin:
Can you try https://github.com/greshake/i3status-rust/issues/401#issuecomment-812456518?

ammgws:
> libc: musl

worked before with this?>

GladOSkar:
@MaxVerevkin that works!

@ammgws yes never was a problem

GladOSkar:
Ohh i just remembered another thing i changed. I switched from void's `rust` package to `rustup` (because void's package is [stuck on 1.64.0](https://github.com/void-linux/void-packages/issues/40805)).

[...]

PR #1787, opened 2023-02-26
Document install settings for non-glibc systems

GladOSkar:
See #1786

### DIFF
--- a/doc/manual_install.md
+++ b/doc/manual_install.md
@@ -7,6 +7,13 @@
 
 Compilation is only tested with very recent stable versions of `rustc`. If you use a distro with older Rust packages, consider using [rustup](https://rustup.rs/) to install a newer toolchain.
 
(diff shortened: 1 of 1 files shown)


doc/manual_install.md
```diff
@@ -7,6 +7,13 @@
 
 Compilation is only tested with very recent stable versions of `rustc`. If you use a distro with older Rust packages, consider using [rustup](https://rustup.rs/) to install a newer toolchain.
 
+On systems using alternative (non-glibc) C standard libraries like `musl`, `cargo` must be configured to be able to link it statically. Otherwise, blocks needing to link to system libraries like `temperature`, `sound` (for pulseaudio) and maybe others will not be usable due to segmentaion faults. To configure `cargo`, just add this to your `~/.cargo/config.toml`:
```
MaxVerevkin, line 10: I believe this config _disables_ static linking.

GladOSkar: Yes of course, that makes more sense. I was confused by the flag, but apparently the target flags start with `+` or `-` depending on whether they enable or disable, it's not a gnu style cli option

GladOSkar:
@MaxVerevkin no, but i only switched to void last summer and have barely compiled any rust since then

PR #1791, opened 2023-02-27
backlight: dbus fallback to sysfs

### LINKED ISSUE #1760: brightness: fix support for systems without `org.freedesktop.login1`
currently on systems that don't implement `org.freedesktop.login1` brightness is completely broken - any action results in `Failed to send D-Bus message (Cause: org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.login1 was not provided by any .service files)`

however, the following config doesn't seem to do anything:
```
[[block]]
block = "backlight"
format = " $brightness "
[[block.click]]
button = "left"
cmd = "brightnessctl set 100%"
[[block.click]]
button = "right"
cmd = "brightnessctl set 10%"
[[block.click]]
button = "wheel_up"
cmd = "brightnessctl set 10+%"
[[block.click]]
button = "wheel_down"
cmd = "brightnessctl set 10-%"
```

bim9262:
Fixes #1760

If you are on a system without `org.freedesktop.login1`, setting the brightness was failing. This change enables fallback to write to the brightness file.

### DIFF
--- a/man/i3status-rs.1
+++ b/man/i3status-rs.1
@@ -210,7 +210,8 @@ it works under both X11 and Wayland.
 The block uses \f[V]inotify\f[R] to listen for changes in the
 device\[cq]s brightness directly, so there is no need to set an update
 interval.
-This block uses DBus to set brightness level using the mouse wheel.
+This block uses DBus to set brightness level using the mouse wheel, but
+will fallback to sysfs if \f[V]systemd-logind\f[R] is not used.
 .SS Root scaling
 .PP
 Some devices expose raw values that are best handled with nonlinear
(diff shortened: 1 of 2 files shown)


bim9262:
@ptrcnull, can you try this out to confirm if it resolves the issue for you?

MaxVerevkin:
Can you please also add a note about write permissions and udev rules ([v0.22](https://github.com/greshake/i3status-rust/blob/a4b46d7a113cacfeba64898ebc6398e31e87c914/doc/blocks.md?plain=1#L173-L183))?

ptrcnull:
it works, thanks!

MaxVerevkin (approved):

PR #1792, opened 2023-02-27
kdeconnect: choose the first avilable device

### LINKED ISSUE #1612: kdeconnect: choose first available device
When I don't specify a `device_id` in the kdeconnect block it doesn't choose the first device paired and reachable. 

`kdeconnect-cli --list-available` shows one device. When I use the id from there and set it as `device_id` it works.

`kdeconnect-cli --list-devices` shows 7 paired devices but only the 5. device is paired and reachable

omkar-mohanty:
Fixes #1612

### DIFF
--- a/src/blocks/kdeconnect.rs
+++ b/src/blocks/kdeconnect.rs
@@ -45,8 +45,7 @@
 //! - `phone`
 //! - `phone_disconnected`
 
-use tokio::sync::mpsc;
-use zbus::dbus_proxy;
+use zbus::{dbus_proxy, SignalStream};
 
 use super::prelude::*;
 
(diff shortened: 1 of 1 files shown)


omkar-mohanty:
@MaxVerevkin is this the correct direction to go for? I am a bit unsure if all the cases are handles.

src/blocks/kdeconnect.rs
```diff
@@ -84,14 +84,32 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
...
+                    api.hide().await?;
+                    continue;
+                }
+            };
+
+            let (tx_new, rx_new) = mpsc::channel(8);
+
+            rx = rx_new;
+            
+            device = Device::new(&dbus_conn, tx_new, &id).await?;
```
MaxVerevkin, line 109: I haven't tested this, but it looks like a task with is spawned in `Device::new()` will panic after `rx = rx_new`, which drops `rx`.

MaxVerevkin:
> @MaxVerevkin is this the correct direction to go for? I am a bit unsure if all the cases are handles.

I think a better way to organize things is to do something similar to `bluetooth.rs` and `battery/upower.rs`.

omkar-mohanty:
@MaxVerevkin I organized `kdeconnect.rs` like `bluetooth.rs`, I think it should be ready now.

MaxVerevkin:
When device is not found, the block should wait until a new device is added, then it can try searching again. Right now there is a hot loop...

---

And to be honest, I don't quite understand why I decided to use a tokio task + channels to watch for updates when I was porting this block. It doesn't make sense. Instead a `Device` can contain three streams which are `.next().await`ed in `wait_for_change`.

omkar-mohanty:
@MaxVerevkin for some reason `cargo fmt` does not seem to work in my local machine

MaxVerevkin:
> @MaxVerevkin for some reason `cargo fmt` does not seem to work in my local machine

`cargo fmt` unfortunately does not format files which are included via macros (in our case the `define_blocks!` macro in `blocks.rs`), so if you don't use `rust_analyzer`, you have to run `rustfmt --edition 2021 <file>`.

omkar-mohanty:
@MaxVerevkin is the PR close to what you have in mind?

MaxVerevkin:
@omkar-mohanty CPU usage is still 100%

MaxVerevkin:
Now CPU usage is 100% even if the device is available

PR #1796, opened 2023-02-27
net: try to fetch ssid from NL80211_BSS_INFORMATION_ELEMENTS

### LINKED ISSUE #1607: net block: No SSID and signal strength for wifi device
* Version: `0.22.0`
* Kernel: `5.19.4-arch1-1`
* Block configuration:
```toml
[[block]]
block = "net"
device = "wlp2s0"
format = "{ssid} {signal_strength} {speed_down;K}"
interval = 5
```
* `nmcli dev wifi list` shows SSID and strength
* `iw` is installed and
  * `iw dev wlp2s0 info` shows info, but no SSID
  * `iw dev wlp2s0 link` shows the SSID

The net block with the above configuration doesn't show the SSID or signal strength. The down speed is shown, however. I use a very similar setup and config on other laptops and it works there. Just installed it on a new laptop, where it doesn't work. Any ideas?

MaxVerevkin:
I don't know what `NL80211_BSS_INFORMATION_ELEMENTS` is, but on my machine the first two bytes don't make sense, then goes SSID up to byte 1.

Fixes #1607

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1200,9 +1200,9 @@ dependencies = [
 
 [[package]]
 name = "neli-wifi"
-version = "0.4.0"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4714dd86601f4da56faffa07bc24c684048d01575f064e2256cef45c5722a10b"
+checksum = "bba6803a3cd5cd89aff307e237c0d3f45ee5e3cd7e2980eae0b7f61d22217476"
 dependencies = [
  "neli",
  "neli-proc-macros",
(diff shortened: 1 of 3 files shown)


ammgws:
Used here too 

https://github.com/i3/i3status/blob/main/src/print_wireless_info.c

Stebalien:
This is cycling through random visible SSIDs, not returning the SSID of the current network.

MaxVerevkin:
> This is cycling through random visible SSIDs, not returning the SSID of the current network.

Should be fixed in da349b1

cfsmp3:
It's fixed for me at least.

Stebalien:
Fixed for me too. Thanks!

Issue #1761, opened 2023-02-23
external_ip: add option to choose between IPv4 and IPv6 (if available)


hgigas:
```
[[block]]
block = "external_ip"
format = "$ip"
```

shows: `2a02:9130:fd....`

kagaroth:
This is not a bug. In dual stack you use IPv6 first then you fallback to IPv4 and because the site / server used to test and show your public IP uses both IPv4 and IPv6 and because you also have an IPv6 set on your active connection the you use that v6 IP for communicating with it and that is what is shown in the output of that block.

hgigas:
I'm sorry, it isn't a bug.

What's my ip web page shows me both IPv6 and IPv4. How to display only IPv4 in `external_ip` block?

kagaroth:
Put this is `/etc/hosts`:
```
104.26.8.44 ipapi.co
104.26.9.44 ipapi.co
172.67.69.226 ipapi.co
```
This should force the use of IPv4 when talking to ipapi.co as this disable recursive DNS resolution for ipapi .co IP. That way your machine will know that ipapi.co is the 3 IPv4 listed above in `/etc/hosts`.

Before:
```
{
    "ip": "2a02:2f0f:b2ff:ffff::5679:5aad",
    "network": "2a02:2f0f:b000::/38",
    "version": "IPv6",
...
}
```
After:
```
{
    "ip": "86.121.90.173",
    "network": "86.121.88.0/22",
    "version": "IPv4",
...
}
```

GladOSkar:
@kagaroth's solution works but is a bit messy (means no other application can obtain ipv6 from ipapi.com anymore). We should probably provide an option for this.

hgigas:
That would be wonderful, thank you very much.

Thank you a lot @kagaroth for the fast reply

cfsmp3:
> That would be wonderful, thank you very much.
> 
Could you try this, see if it does what you want and then we'll see how to make it properly?

Edit `main.rs` and add on line 56 (after the user_agent one), this:

```
 .local_addr("0.0.0.0:0".parse().unwrap())
```
so it looks like this:

```
   reqwest::Client::builder()
        .user_agent(APP_USER_AGENT)
         .local_addr("0.0.0.0:0".parse().unwrap())
        .timeout(REQWEST_TIMEOUT)
        .build()
        .unwrap()
``` 
 
This forces ipv4 in all calls, but for testing at least it should be good enough...

hgigas:
[...]

PR #1801, opened 2023-03-01
Allow showing or forcing IPv4 in `external_ip`

GladOSkar:
This is a first idea for how we could do IPv4 in `external_ip` to solve #1761.

In the first commit, i just added a `use_ipv4` option that makes the block use IPv4.

But here's where i need some advice: Thinking about it, it might be more useful to have an `$ipv4` placeholder, which would allow showing both at the same time (without requiring 2 blocks).

I implemented this in the second commit. It works, but has the disadvantage of all other info (country, city, ASN, etc.) still belonging to the original (potentially-non-v4) IP. The only alternative i see here would be doubling all format placeholders (`v4_country`, `v4_city`, `v4_asn`), which isn't very nice either. There's also the disadvantage of both requests being made even if only `$ipv4` is used.

Or is it just easier to go with commit 1 and say that if users want both they should use 2 blocks (optionally with `merge_with_next`)?

Feedback appreciated! :)

### DIFF
--- a/src/blocks/external_ip.rs
+++ b/src/blocks/external_ip.rs
@@ -7,6 +7,7 @@
 //! `format` | A string to customise the output of this block. See below for available placeholders. | `" $ip $country_flag "`
 //! `interval` | Interval in seconds for automatic updates | `300`
 //! `with_network_manager` | If 'true', listen for NetworkManager events and update the IP immediately if there was a change | `true`
+//! `use_ipv4` | If 'true', use IPv4 for obtaining all info | `false`
 //!
 //!  Key | Value | Type | Unit
 //! -----|-------|------|------
@@ -74,6 +75,8 @@ pub struct Config {
     interval: Seconds,
     #[default(true)]
     with_network_manager: bool,
+    #[default(false)]
+    use_ipv4: bool,
 }
 
 pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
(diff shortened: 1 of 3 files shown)


src/blocks/external_ip.rs
```diff
@@ -211,8 +216,8 @@ struct IPAddressInfo {
 }
 
 impl IPAddressInfo {
-    async fn new() -> Result<Self> {
-        let info: Self = REQWEST_CLIENT
+    async fn new(client: &Lazy<reqwest::Client>) -> Result<Self> {
```
MaxVerevkin, line 219: Just `&reqwest::Client` will work fine, since `Lazy<T>` dereferences to `T`.

GladOSkar:
@MaxVerevkin any opinion on which implementation you'd prefer?

MaxVerevkin:
> @MaxVerevkin any opinion on which implementation you'd prefer?

No, not really. I'm not sure when or why the other information (country, city, etc.) differs between ipv4 and ipv6 requests. In case this information does change, a `force_ipv4 = true` would probably be a better solution. If it doesn't change, then I'm OK with both `$ipv4` placeholder and `force_ipv4` (or `use_ipv4`).

GladOSkar:
Depending on VPN/CG-NAT configurations the information can absolutely vary between v4 and v6. But thinking about it now, there's nothing that prevents me from allowing both forcing v4 and adding the `$ipv4` placeholder.

MaxVerevkin:
> there's nothing that prevents me from allowing both forcing v4 and adding the $ipv4 placeholder.

I think it makes it very easy to accidentally make two API calls when it is not needed. For example 

```toml
[[block]]
block = "external_ip"
format = " $ipv4 "
```

GladOSkar:
Yes, hence the note in the docs. I think it's worth it considering it enables both usecases as well as possiblr IMO

MaxVerevkin (approved):
I just don't know how useful is it to see ipv4 but info from ipv6. But I don't know much about networking anyway.

GladOSkar:
That's probably not very useful. `$ipv4` is only for showing both v4 and v6 on the same block. But i think you've convinced me, it's weird. And 2 blocks for 2 IP versions is perfectly valid. Also, it's nice to have only one way to do a thing and not multiple

MaxVerevkin (approved):

PR #1802, opened 2023-03-01
backlight: Add regex device name matching

### LINKED ISSUE #1771: backlight: Control multiple/all devices
It would be nice if there was a way to specify that you want control the backlight of more than one monitor at a time. 

Any thoughts on how to extend the config in a way which doesn't break the default behavior, if no device is specified, it to use the first one that can be found? Perhaps `device="ALL"` or `device="*"`

bim9262:
Fixes #1771

There's still no device rescanning to detect if a device has been removed/added.
I left the default behavior when no device is specified to match only the first device, but would like others thoughts if we should make it match all devices by default (that's what I'd suggest). Most people who didn't specify a device already probably only have a single monitor so matching everything won't make an impact to them.

### DIFF
--- a/src/blocks/backlight.rs
+++ b/src/blocks/backlight.rs
@@ -13,7 +13,7 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `device` | The `/sys/class/backlight` device to read brightness information from.  When there is no `device` specified, this block will display information from the first device found in the `/sys/class/backlight` directory. If you only have one display, this approach should find it correctly.| Default device
+//! `device` | A regex to match against `/sys/class/backlight` devices to read brightness information from (can match 1 or more devices). When there is no `device` specified, this block will display information for all devices found in the `/sys/class/backlight` directory. | Default device
 //! `format` | A string to customise the output of this block. See below for available placeholders. | `" $icon $brightness "`
 //! `step_width` | The brightness increment to use when scrolling, in percent | `5`
 //! `minimum` | The minimum brightness that can be scrolled down to | `5`
@@ -43,6 +43,12 @@
 //! device = "intel_backlight"
 //! ```
 //!
+//! # calibright
+//!
+//! Additional display brightness calibration can be set in `$XDG_CONFIG_HOME/calibright/config.toml`
+//! See [https://github.com/bim9262/calibright] for more details.
+//! This block will override any global config set in `$XDG_CONFIG_HOME/calibright/config.toml`
+//!
 //! # D-Bus Fallback
 //!
 //! If you don't use `systemd-logind` i3status-rust will attempt to set the brightness
@@ -63,67 +69,27 @@
 //! # Icons Used
 //! - `backlight` (as a progression)
 
-use std::cmp::max;
-use std::ops::Range;
-use std::path::{Path, PathBuf};
-
-use inotify::{Inotify, WatchMask};
-use tokio::fs::{read_dir, OpenOptions};
-use tokio::io::AsyncWriteExt;
+use calibright::{CalibrightBuilder, CalibrightConfig, DeviceConfig};
 
 use super::prelude::*;
-use crate::util::read_file;
-
-make_log_macro!(debug, "backlight");
-
-#[zbus::dbus_proxy(
-    interface = "org.freedesktop.login1.Session",
-    default_service = "org.freedesktop.login1",
-    default_path = "/org/freedesktop/login1/session/auto"
-)]
-trait Session {
-    fn set_brightness(&self, subsystem: &str, name: &str, brightness: u32) -> zbus::Result<()>;
-}
-
-/// Location of backlight devices
-const DEVICES_PATH: &str = "/sys/class/backlight";
-
-/// Filename for device's max brightness
-const FILE_MAX_BRIGHTNESS: &str = "max_brightness";
-
-/// Filename for current brightness.
-const FILE_BRIGHTNESS: &str = "actual_brightness";
-
-/// amdgpu drivers set the actual_brightness in a different scale than
-/// [0, max_brightness], so we have to use the 'brightness' file instead.
-/// This may be fixed in the new 5.7 kernel?
-const FILE_BRIGHTNESS_AMD: &str = "brightness";
-
-/// set the requested brightness level
-const FILE_BRIGHTNESS_WRITE: &str = "brightness";
-
-/// Range of valid values for `root_scaling`
-const ROOT_SCALDING_RANGE: Range<f64> = 0.1..10.;
 
 #[derive(Deserialize, Debug, SmartDefault)]
 #[serde(default)]
 pub struct Config {
     device: Option<String>,
     format: FormatConfig,
-    #[default(5)]
-    step_width: u8,
-    #[default(5)]
-    minimum: u8,
-    #[default(100)]
-    maximum: u8,
-    cycle: Option<Vec<u8>>,
-    #[default(1.0)]
-    root_scaling: f64,
+    #[default(5.0)]
+    step_width: f64,
+    #[default(5.0)]
+    minimum: f64,
+    #[default(100.0)]
+    maximum: f64,
+    cycle: Option<Vec<f64>>,
     invert_icons: bool,
-    #[default(1.0)]
-    ddcci_sleep_multiplier: f64,
-    #[default(10)]
-    ddcci_max_tries_write_read: u8,
+    //Calibright config settings
+    root_scaling: Option<f64>,
+    ddcci_sleep_multiplier: Option<f64>,
+    ddcci_max_tries_write_read: Option<u8>,
 }
 
 pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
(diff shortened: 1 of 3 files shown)


bim9262:
Here's a video showing the behavior legend: (L)eft, (R)ight, A(ll). I started with both monitors at 5 percent then cycled the left monitor to 100 (making the average brightness go become 52%), then I decremented the brightness using the All block which caused all of the screens to be set 47%, 42%, etc.


https://user-images.githubusercontent.com/1033745/222021900-531b3dee-f940-4a51-8cb8-ffa9fa2f7ace.mp4

src/blocks/backlight.rs
```diff
@@ -126,6 +129,27 @@ pub struct Config {
     ddcci_max_tries_write_read: u8,
 }
 
+pub async fn join_all_accept_single_ok<I, T>(iter: I) -> Result<Vec<T>>
```
bim9262, line 132: Probably should come up with a better name. It's a join all, but only results in an error if there is not at least 1 ok value.

GladOSkar:
Would it make sense to allow calibrating 2 monitors against each other? Like specifying that 80% for one is equal to 100% for the other, so that they're the same brightness when changing all? Or would that be unnecessarily complicated?

bim9262:
It would make sense, perhaps they could be set the same way that the mapping are done for the pulseaudio block. I'll look into it

bim9262:
Legend: (L)eft, (R)ight, A(ll), sc(ale) n(o)s(cale)
The blocks labeled raw are the actual brightness, without any scaling
The R ns block is just there so that it's easier to read (it is exactly the same as R raw).

You can see that the sc and ns block are within 1 of each other because of rounding.

https://user-images.githubusercontent.com/1033745/222347614-e9cb3134-550b-48d5-94c0-8ce74ba3f236.mp4

MaxVerevkin:
Now I wonder if there is a CLI app that can do the same (to control brightness of multiple devices via keybindings) :smile:

GladOSkar:
Yes this does not seem like something a status bar should have to implement ^^

bim9262:
I'm sure this isn't a complete list of [backlight utilities](https://wiki.archlinux.org/title/backlight#Backlight_utilities), but I didn't see any mention that as a capability in my cursory look.

bim9262:
I'm thinking about calling it `calibright`

bim9262:
I'd lean towards making a `calibright` lib and bin. It seems odd to have a brightness daemon. Almost no other brightness utility runs as a daemon either. Do you have any strong reasons against calling a library (provided the dependencies don't blow up the dependencies of `i3status-rust`, they should be a subset of `i3status-rust`'s dependencies)?

MaxVerevkin:
Hm, yes, that's probably better.

MaxVerevkin:
But this way the configuration will need to be repeated in multiple places.

bim9262:
I think only three config parameters need to be duplicated,  `root_scaling`, `ddcci_sleep_multiplier`, `ddcci_max_tries_write_read` (I don't plan on duplicating the calibration settings because that really should be set for `calibright`.

I was thinking that `backlight` block's config would look like:

```rust
#[serde(default)]
pub struct Config {
    device: Option<String>,
    format: FormatConfig,
    #[default(5.0)]
    step_width: f64,
    #[default(5.0)]
    minimum: f64,
    #[default(100.0)]
    maximum: f64,
    cycle: Option<Vec<f64>>,
    invert_icons: bool,
    //Calibright config settings
    root_scaling: Option<f64>,
    ddcci_sleep_multiplier: Option<f64>,
    ddcci_max_tries_write_read: Option<u8>,
}
```

Note how those three values I mentioned have become `Option`s so that only if you specify them in your `i3status-rust` config would they do anything (they would be treated as the default values unless they are overridden in the `calibright` config).

src/blocks/backlight.rs
```diff
@@ -63,67 +63,38 @@
...
-    ddcci_sleep_multiplier: f64,
-    #[default(10)]
-    ddcci_max_tries_write_read: u8,
+    //Calibright config settings
+    root_scaling: Option<f64>,
+    ddcci_sleep_multiplier: Option<f64>,
+    ddcci_max_tries_write_read: Option<u8>,
+}
+
+impl Error {
```
bim9262, line 89: @MaxVerevkin , any suggestions on mapping errors from one crate to another?
I am basically using the same error structure as i3status-rs, except without the block.

src/blocks/backlight.rs
```diff
@@ -140,239 +111,107 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
...
-
         loop {
             select! {
-                _ = file_changes.next() => break,
+                // Calibright can recover from errors, just keep reading the next event.
+                _ = calibright.next() => {
+                    block_error = calibright
+                        .get_brightness()
+                        .await
+                        .map_or_else(
```
bim9262, line 169: It would be nice to not repeat this map_or_else pattern

bim9262 (commented):
I still haven't committed/released anything for calibrigt... Just need to add the bin and config parser.

ETA: I do have dynamic device discovery/removal though so that's really nice :)

src/blocks/backlight.rs
bim9262: or now:

```rust
.map_err(Error::from_calibright)
.err();
```

src/blocks/backlight.rs
MaxVerevkin: Perhaps using a structured error enum (with `thiserror`) would be better for a library? This would allow crate users to distinguish between error variants.

MaxVerevkin:
> I still haven't committed/released anything for calibrigt... Just need to add the bin and config parser.

I can contribute that if you wish.

src/blocks/backlight.rs
MaxVerevkin: But in general, to map errors just implement `From<calibright::Error> for Error`.

MaxVerevkin:
> I do have dynamic device discovery/removal though so that's really nice :)

If only ddcci-driver-linux also supported that :smiling_face_with_tear:

src/blocks/backlight.rs
bim9262: I'd tried that but got ```multiple `impl`s satisfying `errors::Error: From<_>` found```

> > I still haven't committed/released anything for calibrigt... Just need to add the bin and config parser.
> 
> I can contribute that if you wish.

I made a simple cli and got config parsing sorted out!

bim9262:
> > I do have dynamic device discovery/removal though so that's really nice :)
> 
> If only ddcci-driver-linux also supported that smiling_face_with_tear

I have unloaded and reloaded the ddcci module so many times...

src/blocks/backlight.rs
```diff
@@ -63,67 +63,39 @@
...
+    ddcci_sleep_multiplier: Option<f64>,
+    ddcci_max_tries_write_read: Option<u8>,
+}
+
+impl Error {
+    fn from_calibright(err: CalibrightError) -> Self {
+        Self {
+            kind: ErrorKind::Other,
+            message: Some(err.to_string().into()),
+            // TODO: get the cause?
```
bim9262, line 94: Any tips on converting the error to the `cause`?

MaxVerevkin: `message` is supposed to be a short description of what when wrong. Given that `calibright`'s error `Display` implementation just forwards to the underlying error, you can just use `.error()` as you would with any other library.

bim9262:
Thank you for your help and patience @MaxVerevkin  and @GladOSkar!
I'm proud to introduce my first crate: https://crates.io/crates/calibright
Let me know if there's anything that should be fixed/changed either here or in calibright itself (https://github.com/bim9262/calibright)

src/blocks/backlight.rs
```diff
@@ -13,7 +13,7 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `device` | The `/sys/class/backlight` device to read brightness information from.  When there is no `device` specified, this block will display information from the first device found in the `/sys/class/backlight` directory. If you only have one display, this approach should find it correctly.| Default device
+//! `device` | A regex to match against `/sys/class/backlight` devices to read brightness information from (can match 1 or more devices). When there is no `device` specified, this block will display information for all devices found in the `/sys/class/backlight` directory. | Default device
```
bim9262, line 16: Maybe the default should be listed as `"."` or All devices.

bim9262:
I wish I'd named the crates calibright(bin) and calibright-core(or something for the lib...)

bim9262:
@MaxVerevkin, do you have thoughts on the rename? I feel like it's ugly to rename the crates, but it's also odd having the binary name be `calibright_cli`. Although you probably will just bind it to the keys once and then not think about it any more.

MaxVerevkin:
> but it's also odd having the binary name be `calibright_cli`

IMO it's okay

bim9262:
Alright, I'll just leave it as-is.

MaxVerevkin:
> I do have dynamic device discovery/removal though so that's really nice :)

Just tested it and it works for me too ;) Do you think it makes sense to add `missing_format`? Right now I work around it using `error_format = ""`.

bim9262:
I think sounds good, I'll give that a go soon Although it could be done disjointly from this PR.

bim9262:
I couldn't find a way to have a missing_format without completely losing the error message. If that's really the case (this was the first time I've really messed with the set_error stuff, so maybe I missed something), I think setting `error_format = ""` might be the best approach if you don't want to see the error.

MaxVerevkin:
> Although it could be done disjointly from this PR.

Sure.

> I couldn't find a way to have a missing_format without completely losing the error message.

Having `calibright.get_brightness()` return `Result<Option<f64>>` would make it easier...

> I think setting error_format = "" might be the best approach if you don't want to see the error.

Maybe. But it feels like a workaround. And I don't want to silent _all_ errors.

bim9262:
> Maybe. But it feels like a workaround. And I don't want to silent *all* errors.

Fair enough, I'll take another stab at it

MaxVerevkin (approved):

ammgws:
@bim9262 Anything else left to do?

bim9262:
Other than the merge conflicts it's ready to go 🙂

MaxVerevkin:
Interestingly this PR increases binary size by 1M :eyes:

src/blocks/backlight.rs
MaxVerevkin: Yes

PR #1810, opened 2023-03-05
Battery: Use driver recommended icon when capacity is not reliable (WIP)

cfsmp3:
Upower can explicitly say "don't use this charge percentage [we call it capacity], it's unreliable" in some cases. But at least it provides a recommended icon.

This diff [WIP] passes both "capacity is reliable" and "icon to use" from upower upstream so we can display the correct icon.

TO-DO:
- Icon mapping. Don't know if we should just import all the possible icons, or map with what we have (I believe it's not enough though).
- Maybe introduce a new format if the capacity is not reliable? Probably on top of displaying the correct info we don't want to display the capacity value at all.

@MaxVerevkin can you take a look and see if it's reasonably rustic?

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -110,8 +110,8 @@ enum BatteryDriver {
 }
 
 #[derive(Clone, Debug)]
-enum DriverIcon {
-    Upower(String),
+enum BatteryLevel {
+    Upower(u32),
 }
 
 pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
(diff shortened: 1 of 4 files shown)


MaxVerevkin:
Why would we parse/map the provided icon when we know that it is based on "battery level", which is available?

cfsmp3:
> Why would we parse/map the provided icon when we know that it is based on "battery level", which is available?

Good point, dumb idea. Changed.

Question on how do those progressive icons work: I see that for the battery we have this:

 bat = [                                                                         
     "\uf244",                                                                   
     "\uf243",                                                                   
     "\uf242",                                                                   
     "\uf241",                                                                   
     "\uf240",                                                                   
 ]   

How can I see what each of them represent? (i.e. how do I print the icon for uf244 somewhere, sorry for the newbie question). I assume it's battery low to full but let me check :-)

The icon_value passed to get_icon_in_progression is a float from 0 to 1, right? So for a collection of 5 icons, the first one would go 0 to 0.2, the next 0.2+ to 0.4 and so on.

cfsmp3:
Some of these pre-commits checks are wrong:

MaxVerevkin:
> Some of these pre-commits checks are wrong:
> 
> 18 | LOW = 3, | ^^^ help: consider making the acronym lowercase, except the initial letter (notice the capitalization): `Low`
> 
> That is not an acronym.

It isn't, but it still should be `Low` because it is an enum variant.

MaxVerevkin:
> Question on how do those progressive icons work: I see that for the battery we have this:
> 
> bat = [ "\uf244", "\uf243", "\uf242", "\uf241", "\uf240", ]
> 
> How can I see what each of them represent? (i.e. how do I print the icon for uf244 somewhere, sorry for the newbie question). I assume it's battery low to full but let me check :-)
> 
> The icon_value passed to get_icon_in_progression is a float from 0 to 1, right? So for a collection of 5 icons, the first one would go 0 to 0.2, the next 0.2+ to 0.4 and so on.

Correct. The idea was that different icon sets have different number of battery icons (battery is just an example here). Also switching from a static icon (e.g. `bat_charging`) to a dynamic icon is not a breaking change.

> i.e. how do I print the icon for uf244 somewhere, sorry for the newbie question

That's not a newbie question, that's unwritten docs for `get_icon_in_progression` :)

cfsmp3:
Probably overkill. 
Closing.

Issue #1826, opened 2023-03-12
hueshift step not working with `wl-gammarelay-rs`


joeftiger:
I am using [`wl-gammarelay-rs`](https://github.com/MaxVerevkin/wl-gammarelay-rs) in conjunction with the `hueshift` block.
However, I noticed that the `step = 100` (default) property is not being applied correctly, as the temperature is only inc-/decreased by `500`. Changing the `step` size doesn't do anything for it.

bim9262:
@joeftiger, would you be interested in creating a PR with the fix? I think that this would be a good first issue!

marcybelardo:
Hi @bim9262 , I was looking around issues and came across this. I think the issue was calling `max` on config.step instead of `min`, so instead of returning values 500K or lower, it would always return values of 500 or higher. I can make a PR for this now if that's ok~

PR #1827, opened 2023-03-13
fix: hueshift block step now maxes at 500

marcybelardo:
`max` was returning the larger value between `config.step` and 500, basically defaulting step to 500 or higher. `min` will return the smaller of `config.step` and 500, meaning the value will never go above 500, but may be lower.

Fix for #1826


MaxVerevkin (approved):
Thanks!

PR #1843, opened 2023-03-22
Force apt output to be in default language

gpayer:
If your system language is set to something else than English, then the upgradable package count failes.
E.g. in German the key word ist "aktualisierbar".
Fortunately this is easily fixable by setting the env variable LANG to C.


MaxVerevkin:
Thanks!

PR #1852, opened 2023-04-04
docs: fix Switch GTK Theme example

gwena:
Fix the Switch GTK Theme example in the examples/README.md to conform to the new syntax of i3status-rust 0.30.0


MaxVerevkin:
Thanks!

PR #1853, opened 2023-04-04
Don't break the entire bar when block's config has errors

MaxVerevkin:
Also improve the error message when unknown/disabled block is given.

But, it currently requires `block = "..."` to be the first field of block config, because some necessary structs are in `serde::_private`, so only derived imps can implement proper tagged enum deserialization.

### DIFF
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -31,11 +31,12 @@ pub mod prelude;
 
 use crate::BoxedFuture;
 use futures::future::FutureExt;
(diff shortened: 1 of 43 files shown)


ammgws:
> But, it currently requires `block = "..."` to be the first field of block config

I think that might work out OK for most users, and if not then at least they get an error message in the bar

MaxVerevkin:
So okay for v0.31?

PR #1856, opened 2023-04-08
sound: add active_port specifier and let it be mapped

ammgws:


src/blocks/sound.rs
```diff
@@ -60,6 +73,7 @@
 //! `volume`             | Current volume. Missing if muted. | Number | %
 //! `output_name`        | PulseAudio or ALSA device name    | Text   | -
 //! `output_description` | PulseAudio device description, will fallback to `output_name` if no description is available and will be overwritten by mappings (mappings will still use `output_name`) | Text | -
+//! `active_port`        | Active port, will be `""` if no information available (Same as information in Ports section of `pactl list cards`) | Text | -
```
MaxVerevkin, line 76: Why `""` instead of just making it absent?

src/blocks/sound.rs
```diff
@@ -37,6 +37,19 @@
...
+//!
+//! ```toml
+//! [[block]]
+//! block = "sound"
+//! driver = "pulseaudio"
+//! device_kind = "source"
+//! format = " $icon { $volume|}$active_port "
+//! [block.mappings]
+//! "alsa_input.pci-0000_00_1f.3.analog-stereo analog-input-rear-mic" = ""
+//! "alsa_input.pci-0000_00_1f.3.analog-stereo analog-input-front-mic" = "ERR!"
```
MaxVerevkin, line 50: I'm a bit confused. Is your `active_port` "analog-input-rear-mic" or the entire string?

MaxVerevkin:
IMO a separate option for `active_port_mappings` would be better. For example, this would allow to set default values for active port, like "headset-output" -> "H", ".*" -> "". Also it would be less confusing.

src/blocks/sound.rs
ammgws: The former

src/blocks/sound.rs
ammgws: You're right, should just make it absent

ammgws:
That would be more useful, yes.

src/blocks/sound.rs
MaxVerevkin: Then how does it work?

MaxVerevkin: Oh, it doesn't.

MaxVerevkin:
@ammgws should we revert it for now to release v0.30.7?

PR #1861, opened 2023-04-12
formatting.rs: fix `rot_interval` default in docs

OddBloke:
0.5 _was_ the default, before rotation was merged into `.str()`.


MaxVerevkin (approved):
Thanks!

Issue #1872, opened 2023-04-21
Conflict in the dbus signal reading process in kdeconnect block for battery and network refresh


Atreyagaurav:
There is a "refreshed" signal emitted from kdeconnect plugin for cell network, as well as for battery strength that kdeconnect block is using. 

This signal is the one currently being used by the kdeconnect block to refresh the value for battery status:

    dbus-monitor "type='signal',sender='org.kde.kdeconnect',interface='org.kde.kdeconnect.device.battery'"

The signal is emitted here:

https://invent.kde.org/network/kdeconnect-kde/-/blob/master/plugins/battery/batteryplugin.cpp#L124

There is a signal with the same name with a different interface, if you turn on the connectivity plugin you can get the signal:

    dbus-monitor "type='signal',sender='org.kde.kdeconnect',interface='org.kde.kdeconnect.device.connectivity_report'"

This is emitted here:

https://invent.kde.org/network/kdeconnect-kde/-/blob/master/plugins/connectivity-report/connectivity_reportplugin.cpp#L54

So I wanted to add the connectivity status on the kdeconnect block of the i3status-rs as well. Then I ran into the problem due to the macro used to create the functions as it makes the function names and other things based on the signal name which makes these two have the same name.

```diff
+#[dbus_proxy(
+    interface = "org.kde.kdeconnect.device.connectivity_report",
+    default_service = "org.kde.kdeconnect"
+)]
+trait ConnectivityDbus {
+    // this here doesn't work as the macro uses the "refreshed" name for objects and ends up with duplicate names
+    #[dbus_proxy(signal, name = "refreshed")]
+    fn refreshed(&self, network_type: String, network_strength: i32) -> zbus::Result<()>;
+
+    #[dbus_proxy(property, name = "cellularNetworkStrength")]
+    fn cellular_network_strength(&self) -> zbus::Result<i32>;
+
+    #[dbus_proxy(property, name = "cellularNetworkType")]
+    fn cellular_network_type(&self) -> zbus::Result<String>;
+}
```
Full commit: https://github.com/greshake/i3status-rust/commit/d3b25de840dbd735a47ae14cf5dfd97cc36f5524 

![image](https://user-images.githubusercontent.com/46195892/233510174-5df6c07a-2d01-4541-8529-39b7be446ca1.png)

The image shows one error, there are a ton of error, all due to the conflicting name.

[...]

PR #1873, opened 2023-04-21
Add connectivity report (cell network) on kdeconnect block

Atreyagaurav:
Closes https://github.com/greshake/i3status-rust/issues/1872


bim9262:
@Atreyagaurav, I'm going to mark this as draft while you work out how to handle the disabled sim. Also check out the fmt and clippy checks as they aren't passing.

Atreyagaurav:
I wasn't able to check the fmt as `cargo fmt` was ok, and rust analyzer was giving me errors about `async fn`. But I checked the contributing.md, and ran `rustfmt 2021` should be ok now.

For the SIM disable stuff looks like https://invent.kde.org/network/kdeconnect-kde/-/tree/master/plugins/connectivity-report
says the signal strength is `signalStrength (int) [0..=4]: The signal
strength` but it seems to be -1 when the SIM is disabled. It is probably emitted from something else (strength value changed signal) as I didn't get any signal when I disabled the sim when I did: 

```
dbus-monitor "type='signal',sender='org.kde.kdeconnect',interface='org.kde.kdeconnect.device.connectivity_report'"
```
Now it works when I disable the sim. And should work also when the strength is zero.

src/blocks/kdeconnect.rs
```diff
@@ -34,13 +36,14 @@
 //! ```toml
 //! [[block]]
 //! block = "kdeconnect"
-//! format = " $icon {$bat_icon $bat_charge|}{ $notif_icon|} "
+//! format = " $icon {$bat_icon $bat_charge|}{ $notif_icon|} {$network_icon $network|}"
```
bim9262, line 39: I think we would want this to be:

```
//! format = " $icon {$bat_icon $bat_charge |}{$notif_icon |}{$network_icon $network |}"
```

src/blocks/kdeconnect.rs
```diff
@@ -129,6 +137,22 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
                     }
                 }
 
+                let network = device.network().await;
+                if let Some(net) = network.0 {
+                    let network_strength = network.1.unwrap_or(0);
+                    values.insert(
+                        "network_icon".into(),
+                        Value::icon(api.get_icon("net_wireless")?),
```
bim9262, line 145: It would be nice (but not a must) if this were an icon progression.

There's some code in this PR: https://github.com/greshake/i3status-rust/pull/1728/files#diff-f32290632a1d310d9d47a251e3ce1ed66f1a6943421ee2e44da0c8bfb5e118caR277-R286

that can be used for an icon progression if the scale isn't 0%-100% (0f64..=1f64)

src/blocks/kdeconnect.rs
```diff
@@ -256,6 +295,16 @@ impl Device {
             .map(|n| n.len())
             .ok()
     }
+
+    async fn network(&self) -> (Option<String>, Option<i32>) {
+        (
```
bim9262, line 300: I'd use `tokio::join!` here and have the function signature:

```rust
async fn network(&self) -> (Result<String>, Result<i32>)
```

src/blocks/kdeconnect.rs
```diff
@@ -129,6 +137,22 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
                     }
                 }
 
+                let network = device.network().await;
```
bim9262, line 140: You can do something like this so you don't have to recall what .0 and .1 are in the tuple:

```rust
let (cellular_network_type, cellular_network_strength) = device.network().await;
```

src/blocks/kdeconnect.rs
```diff
@@ -129,6 +137,22 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
...
+                if let Some(net) = network.0 {
+                    let network_strength = network.1.unwrap_or(0);
+                    values.insert(
+                        "network_icon".into(),
+                        Value::icon(api.get_icon("net_wireless")?),
+                    );
+
+                    if network_strength <= 0 {
```
bim9262, line 148: A comment here what the range of networks strengths are would be helpful

bim9262 (commented):
I tested your changes out (took me a bit since I hadn't used kdeconnect before), and results look good! ☺️

Just a few things I'd change!

src/blocks/kdeconnect.rs
Atreyagaurav: Oh Yeah, the Icon Progression is something I tried first and got really frustrated because it didn't work. Turns out the Font Awesome 6's network related progress icons were for paid version only, and my arch install didn't have that. I haven't looked at other icon themes yet. If we can make it work for some, then I'll modify it to be that case. I looked at the network block and in fontawesome6 it was not a progression.

I'll try to use nordic font(?) and see if I can make it work.

bim9262: material-nf is the only font with the net_wireless progression.

src/blocks/kdeconnect.rs
Atreyagaurav: I don't get this. I used the same format as that in the battery one few lines above this. Can I leave it for someone to change it when they change both. Or do you mean something that's specifically different in the cellular network that would require me to do that? I've never used tokio, so I don't know what I'll be doing.

src/blocks/kdeconnect.rs
Atreyagaurav: I shifted the space of network part before the `|` but I didn't touch the other parts that were there already. Should I modify those?

src/blocks/kdeconnect.rs
Atreyagaurav: Should be fixed now. I didn't test using the nordic-nf myself though.

src/blocks/kdeconnect.rs
bim9262: I would actually change the battery block as well... We can pull of the async code up higher in the code and perform it all at the same time. This is what I'd change (in regard to improving async performance):


```diff
diff --git a/src/blocks/kdeconnect.rs b/src/blocks/kdeconnect.rs                                                                                                                                                                                                
index ffaffec7..a289279f 100644                                                                                                                                                                                                                                 
--- a/src/blocks/kdeconnect.rs                                                                                                                                                                                                                                  
+++ b/src/blocks/kdeconnect.rs                                                                                                                                                                                                                                  
@@ -48,6 +48,7 @@                                                                                                                                                                                                                                               
 //! - `phone`                                                                                                                                                                                                                                                  
 //! - `phone_disconnected`                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                
+use futures::TryFutureExt;                                                                                                                                                                                                                                     
 use tokio::sync::mpsc;                                                                                                                                                                                                                                         
 use zbus::dbus_proxy;                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                
@@ -112,8 +113,13 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {                                                                                                                                                                       
                 values.insert("icon".into(), Value::icon(api.get_icon("phone")?));                                                                                                                                                                             
                 values.insert("connected".into(), Value::flag());                                                                                                                                                                                              
                                                                                                                                                                                                                                                                
-                let (level, charging) = device.battery().await;                                                                                                                                                                                                
-                if let Some(level) = level {                                                                                                                                                                                                                   
+                let (                                                                                                                                                                                                                                          
+                    (level, charging),                                                                                                                                                                                                                         
+                    (cellular_network_type, cellular_network_strength),                      
+                    notif_count,                                                                                              
+                ) = tokio::join!(device.battery(), device.network(), device.notifications());                                  
+                                                                                                                               
+                if let Ok(level) = level {      
                     values.insert("bat_charge".into(), Value::percents(level));                                               
                     values.insert(                                                                                             
                         "bat_icon".into(), 
@@ -137,23 +143,21 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {                                      
                     }         
                 }                                    
                                               
-                let network = device.network().await;                                                                          
-                if let Some(net) = network.0 {                                                                                
-                    let network_strength = network.1.unwrap_or(0);                                                            
+                if let Ok(cellular_network_type) = cellular_network_type {                                                    
                     values.insert(                                                                                             
                         "network_icon".into(),                                                                                
                         Value::icon(api.get_icon("net_wireless")?),                                                           
                     );                        
                                                                                                                               
-                    if network_strength <= 0 {                                                                                
+                    if cellular_network_strength <= 0 {                                                                        
                         widget.state = State::Critical;                                                                       
                         values.insert("network".into(), Value::text("×".into()));                                              
                     } else {                                                                                                  
-                        values.insert("network".into(), Value::text(net));                                                    
+                        values.insert("network".into(), Value::text(cellular_network_type));                                   
                     }                     
                 }                                                                                                             
                                                            
-                let notif_count = device.notifications().await.unwrap_or(0);                                                   
+                let notif_count = notif_count.unwrap_or(0);                                                                    
                 if notif_count > 0 {                 
                     values.insert("notif_count".into(), Value::number(notif_count));                                          
                     values.insert(                                                                                            
@@ -277,14 +281,15 @@ impl Device {                                                                                            
         self.device_proxy.name().await.ok()
     }                                              
                                                                                                                               
-    async fn battery(&self) -> (Option<u8>, bool) {
-        (            
+    async fn battery(&self) -> (Result<u8>, bool) {
+        tokio::join!(                                  
             self.battery_proxy                         
                 .charge()                                                                                                     
-                .await                          
-                .ok()                                                                                                         
-                .map(|p| p.clamp(0, 100) as u8),                                                                              
-            self.battery_proxy.is_charging().await.unwrap_or(false),                                                           
+                .map_ok(|p| p.clamp(0, 100) as u8)
+                .map_err(|_| Error::new("Could not get charge")),                                                              
+            self.battery_proxy                                                                                                
+                .is_charging()                             
+                .map_ok_or_else(|_| false, |x| x),
         )                                                                                                                     
     }                             
                                   
@@ -296,13 +301,14 @@ impl Device {         
             .ok()
     }                                                                                                                          
                                                    
-    async fn network(&self) -> (Option<String>, Option<i32>) {                                                                 
-        (                                             
-            self.connectivity_proxy.cellular_network_type().await.ok(),                                                        
+    async fn network(&self) -> (Result<String>, i32) {
+        tokio::join!(                                          
+            self.connectivity_proxy                                                                                            
+                .cellular_network_type()                       
+                .map_err(|_| Error::new("Could not get cellular_network_type")),                                               
             self.connectivity_proxy                            
                 .cellular_network_strength()                   
-                .await                                         
-                .ok(),                                         
+                .map_ok_or_else(|_| 0, |x| x),                 
         )                                                      
     }                                                                                                                          
 }             
```

src/blocks/kdeconnect.rs
```diff
@@ -34,13 +36,14 @@
 //! ```toml
 //! [[block]]
 //! block = "kdeconnect"
-//! format = " $icon {$bat_icon $bat_charge|}{ $notif_icon|} "
+//! format = " $icon {$bat_icon $bat_charge|}{ $notif_icon|}{$network_icon $network |}"
```
bim9262, line 39: Format still doesn't look quite right.

You want a ` ` before each `|` and no leading ` ` inside of the `{}`

Like:

`format = " $icon {$bat_icon $bat_charge |}{$notif_icon |}{$network_icon $network |}"`

src/blocks/kdeconnect.rs
bim9262: I show how you can do this in the diff above.

src/blocks/kdeconnect.rs
bim9262: Ah, yes. I think that was there was a mistake before you started your change.

Atreyagaurav:
It said it can't apply the patch, I did copy and paste the most things. And it refused to compile as the `notif_count` was now `Option<usize>`, so I changed the function for it. Should be ok.

src/blocks/kdeconnect.rs
Atreyagaurav: Ok, I also changed the default on the code, not only in the doc. I just didn't want to modify what was already there, but if you wanna change it then it should be correct now.

src/blocks/kdeconnect.rs
bim9262: They aren't nordic-nf, but material-nf :) This change looks good though!

bim9262 (commented):
Otherwise I don't see anything I'd change.
Thanks for your hard work on this @Atreyagaurav  and your patience as we review it :)

src/blocks/kdeconnect.rs
```diff
@@ -129,7 +142,26 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
...
+                            cellular_network_strength.clamp(0, 4) as f64 / 4.0,
+                        )?),
+                    );
+
+                    // network strength is 0..=4 from docs of
+                    // kdeconnect/plugins/connectivity-report, and I
+                    // got -1 for disabled SIM (undocumented)
+                    if cellular_network_strength <= 0 {
+                        widget.state = State::Critical;
+                        values.insert("network".into(), Value::text("X".into()));
```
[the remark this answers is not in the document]
MaxVerevkin: Ideally this should be an `icon` which would be override-able. But that's a theme for future improvement.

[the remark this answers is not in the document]
bim9262: Do you think it would be better to omit network if it's missing and then let the user specify the fallback in the format?

MaxVerevkin:
From https://invent.kde.org/network/kdeconnect-kde/-/tree/master/plugins/connectivity-report

> signalStrengths (object<string, object>): Maps each SIM (subscription ID) to the following object:
    networkType (string):
        One of "5G", "LTE", "CDMA", "EDGE", "GPRS", "GSM", "HSPA", "UMTS", "CDMA2000", "iDEN", "Unknown"
    signalStrength (int) [0..=4]: The signal strength

Does the dbus interface allow us to get the information about all sims?

>  ```rust
> fn cellular_network_strength(&self) -> zbus::Result<i32>;
> ```

Does this return the strength of the first sim? Or the average?

bim9262:
> Does the dbus interface allow us to get the information about all sims?

Looks like it's just the first sim: https://invent.kde.org/network/kdeconnect-kde/-/blob/master/plugins/connectivity-report/connectivity_reportplugin.cpp#L45

src/blocks/kdeconnect.rs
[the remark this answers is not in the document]
Atreyagaurav: I had the unicode character "×" before, but I then changed it to X lol. I didn't even know it was being used in the net block.

Yeah, I'll change it back. And I think we might as well add a percentage from network strength like in netblock as the network icon isn't progression in most icon theme.

Atreyagaurav:
Maybe we can submit a pull request to the plugin. Doesn't make sense to just discard the information if kdeconnect app sends the signal for all. But it might complicate things now with backward compatibility. IDK.

Atreyagaurav:
I made the "$network" give network percentage and "$network_type" give the type of network. Having the percentage might be desirable when there's no progression. I think that's most of what I'd want in this block. We can improve it for a few things but we might have to change upstream, either the plugin in kdeconnect app or desktop for those.

bim9262:
I think that `$network` might be confusing for network strength (both in name and value) perhaps `$network_strength` would be better. Also I don't think that representing number of bars as a percent is really ideal. Maybe doing bars (up to 4) would be a good option.

MaxVerevkin:
> Also I don't think that representing number of bars as a percent is really ideal

Why though?

src/blocks/kdeconnect.rs
```diff
@@ -34,22 +37,29 @@
...
 //! block = "kdeconnect"
-//! format = " $icon {$bat_icon $bat_charge|}{ $notif_icon|} "
+//! format = " $icon {$bat_icon $bat_charge |}{$notif_icon |}{$network_icon$network $network_type |}"
 //! bat_good = 101
 //! ```
 //!
 //! # Icons Used
 //! - `bat` (as a progression)
 //! - `bat_charging` (as a progression)
+//! - `net_wireless` (as a progression)
```
MaxVerevkin, line 47: It might be a good idea to use specialized icons, such as `nf-md-network_strength*`. Font awesome also has some signal icons, but I'm not sure if they're free...

Atreyagaurav: I used,
- https://fontawesome.com/icons/signal?f=classic&s=solid on fontawecome6
- https://www.nerdfonts.com/cheat-sheet and `nf-md-network_strength*` from it for material-nf
- https://fonts.google.com/icons?selected=Material+Symbols+Outlined:network_cell:FILL@0;wght@400;GRAD@0;opsz@48&icon.query=network+cell for material theme
- https://fontawesome.com/v4/icon/wifi for awesome4
- https://fontawesome.com/v5/icons/wifi?f=classic&s=solid for awesome5

Turns out I don't have those fonts installed, and when I tried to install some and test I only got unicode boxes, so I can't verify other ones work. The awesome6 one work as I've verified it.
Please check my latest commit and verify those if you can.

Atreyagaurav:
> Also I don't think that representing number of bars as a percent is really ideal. Maybe doing bars (up to 4) would be a good option.

If the icon theme supports it, then we'll have the progression on the symbol itself. But without that I think the percent is easiest to understand compared to other numbers (as people might not know 4 is maximum here, even that seems debatable as I've been using it for a while, and I've never seen it give 100% even though the document says it's 0..=4).

I did think about using something that's free and available here: https://github.com/greshake/i3status-rust/issues/1886
But I think that might be a good idea to implement globally in a different pull request for all other percentage related values than for just this one. And it's also easier if we use bar, as we'd know the maximum value is 100.

bim9262:
> > Also I don't think that representing number of bars as a percent is really ideal
> 
> Why though?

"There’s no industry standard that ties cell phone signal strength to the number of bars on your phone. It's up to each carrier or phone manufacturer to decide what's 1, 2, 3, 4, or full bars on their service. So, 2 bars on AT&T could be 3 bars on T-Mobile, or 4 bars on Verizon." https://www.wilsonamplifiers.com/blog/how-to-read-cell-phone-signal-strength-the-right-way/

For example on a phone with 5 bars you might see:

Signal Strength | General Results
-- | --
-50 to -79 dBm | Considered great signal (4 to 5 bars)
-80 to -89 dBm | Considered good signal (3 to 4 bars)
-90 to -99 dBm | Considered average signal (2 to 3 bars)
-100 to -109 dBm | Considered poor signal (1 to 2 bars)
-110 to -120 dBm | Considered very poor signal (0 to 1 bar)


In order to calculate something as a percentage you need to be able to say what the $max$ is in a linear scale:

$$\frac{max-current}{max}$$

but for a logarithmic scale like dBm, we need something like:

$$ \frac{\log(current) - log(min)}{log(max)-log(min)} $$

On my phone I observed that -117 [dBm](https://en.wikipedia.org/wiki/Mobile_phone_signal) is 1 bar while -118 dBm is 0 bars, but the change of 1 dBm doesn't represent a 25% drop in network strength. If we say the $max=-50, min=-120$ then -118 dBm is 1.92% strength while -117 dBm is 2.89% strength, however if we say that $max=-44, min=-140$ (this seem to be the full spectrum of signal strength) then -118 dBm is 14.77% strength while -117 dBm is 15.51% strength.

To me it seems misleading to say the strength is 0%, 25%, 50%, 75%, or 100%

src/blocks/kdeconnect.rs
bim9262: `net_strength` can be `\uf012` for all of the font-awesome versions. It was introduced in version 1.0.0. https://fontawesome.com/v4/icon/signal.
material appears to have a progression available: https://fonts.google.com/icons?selected=Material+Symbols+Outlined:signal_cellular_0_bar:FILL@0;wght@400;GRAD@0;opsz@48&icon.query=signal+cellular
For material-nf you put 6 elements in the progression, but there are only 5 states (0..=4). I'd drop `nf-md-network_strength_off_outline`

bim9262:
> So for our fix, do you want to just remove the percentage thing for now and be satisfied as the icon is progression? Or do you want me to add the vertical bar thing (▁ ▂ ▃ ▄ ▅ ▆ ▇: as a progression) instead of percentage as format variable?

I'll defer to @MaxVerevkin since I was the original descent, if the reason I gave was reason enough to not include the network strength.

src/blocks/kdeconnect.rs
Atreyagaurav: Ohh, I somehow didn't find it by searching "network" "cell" like in v6, so I thought it wasn't there. I'll change that.

I'll add the material one too.

Oh! In my mind I was thinking of 5 states `0..=4` and then -1 as unavailable, but I guess we make that 0 to make it into progression. Might be better to add 1 to the value obtained and then turn it into `0..=5`. SIM disabled (-1) and no/very weak signal (0), might warrant a difference.

MaxVerevkin:
> > So for our fix, do you want to just remove the percentage thing for now and be satisfied as the icon is progression? Or do you want me to add the vertical bar thing (▁ ▂ ▃ ▄ ▅ ▆ ▇: as a progression) instead of percentage as format variable?
> 
> I'll defer to @MaxVerevkin since I was the original descent, if the reason I gave was reason enough to not include the network strength.

Personally I don't mind if we expose network strength as percentage (we already do so for wifi strength, with somewhat arbitrary conversion from dBm to percents). As for the vertical bar thing, I think it is best to implement it as a formatter option (#1886) (this could be useful for wifi and cpu too). Note that it doesn't really matter if we have strength as a percentage or a number of bars, both could be formatted as "vertical bars" (but the latter would require an additional argument to set the "max value" to 4).

> To me it seems misleading to say the strength is 0%, 25%, 50%, 75%, or 100%

Having more precision would be better, sure.

> Maybe doing bars (up to 4) would be a good option.

Do you mean a number from 0 to 4? Or graphical representation?

Atreyagaurav:
Yeah, people could choose to not use the percentage. And we can possibly add in the documentation that it's not the correct representation, but it can later be used for other representation. And we're working on what information we have, the bars not being accurate representation is mostly the upstream problem.

It's trivial to remove that, so if you guys think we should remove it, just let me know. At least for my usecase I only want to know whether the network is very bad or not. So for me it works if it shows me 0% or 25%.

This seems to be where the signal strength is converted to number of bars for KDE connect Android app:

https://invent.kde.org/network/kdeconnect-android/-/blob/master/src/org/kde/kdeconnect/Plugins/ConnectivityReportPlugin/ConnectivityReportPlugin.java#L145

So I think even if the networks use different ways, as long as the same library is used and the strength given by the device is absolute value, it's consistent? That could also be the reason even when my phone shows 3 or 4 bars, as the actual value isn't changed much the kdeconnect block is stuck at 75% for me. I'm not knowledgeable in this, just my 2 cents.

files/icons/awesome6.toml
```diff
@@ -47,6 +47,7 @@ net_bridge = "\uf0e8"
 net_down = "\uf019"
 net_loopback = "LO"
 net_modem = "\uf095"
+net_strength = "\uf012"
```
MaxVerevkin, line 50: It's unfortunate that font awesome free has no "no signal" icon, but we can at least use the cross symbol here. Something like

```toml
net_strength = [
        "<x symbol>",
        "\uf012",
        "\uf012",
        "\uf012",
        "\uf012",
        "\uf012",
]
```

src/blocks/kdeconnect.rs
```diff
@@ -238,23 +291,35 @@ impl Device {
...
+    }
+
+    async fn network(&self) -> (Result<String>, i32) {
+        tokio::join!(
+            self.connectivity_proxy
+                .cellular_network_type()
+                .map_err(|_| Error::new("Could not get cellular_network_type")),
+            self.connectivity_proxy
+                .cellular_network_strength()
+                .map_ok_or_else(|_| 0, |x| x),
```
MaxVerevkin, line 321: Is that equivalent to `.unwrap_or(0)`? Also, wouldn't `-1` be a better default?

files/icons/awesome6.toml
Atreyagaurav: The `network_type` has "×" (and critical state) if the signal is zero or less, so if you have icon + network type then it'll be conveyed. I think that way we see × after signal icon instead of just × which won't have context. I think that's better, what do you think?

src/blocks/kdeconnect.rs
Atreyagaurav: I think it's the equivalent for that, I copied it from the @bim9262 's suggestion. 

I think we actually will never need to worry about this as the dbus signal will always send those two values in a pair. If the `network_type` is not `Ok(t)` we'll just skip it, and if it's present then `network_strength` will have value. 0 is the minimum hence I used it, The -1 part is an undocumented signal that the kdeconnect plugin seems to send if it can't read network status. 

I guess now that we're adding `1` to the value to get the progression icon it makes sense to make it `-1`.

bim9262:
> Do you mean a number from 0 to 4? Or graphical representation?

Sorry, I meant a graphical representation.

files/icons/awesome6.toml
Atreyagaurav: Preview of current state when the network is zero (tested by disabling the SIM)
image

bim9262: I think that looks good. @MaxVerevkin, I'll be traveling for the next 2 weeks if you'd like to take over the reviews/merge.

files/icons/awesome4.toml
```diff
@@ -43,6 +43,7 @@ net_bridge = "\uf0e8" # fa-sitemap
 net_down = "\u2b07"
 net_loopback = "LO"
 net_modem = "\uf095" # fa-phone
+net_strength = "\uf012"
```
MaxVerevkin, line 46: Just a though: maybe a better name would be `net_cellular`?

MaxVerevkin (approved):

files/icons/awesome4.toml
Atreyagaurav: Done

MaxVerevkin:
Thanks @Atreyagaurav for addressing all the feedback and thanks @bim9262 for reviewing!

src/blocks/kdeconnect.rs
```diff
@@ -129,7 +143,31 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
...
+                            (cellular_network_strength + 1).clamp(0, 5) as f64 / 5.0,
+                        )?),
+                    );
+                    values.insert(
+                        "network_strength".into(),
+                        Value::percents(cell_network_percent),
+                    );
+
+                    if cellular_network_strength <= 0 {
+                        widget.state = State::Critical;
```
sersorrel, line 164: it would be really nice if things like this could be included in the changelog, or at least in the documentation :) I just spent half an hour trying to figure out why my kdeconnect block was yelling at me for my phone only having 75% battery remaining, because I didn't realise it would now yell at me if my phone is not connected to the cell network. (incidentally, the phone _is_ connected to the cell network, kdeconnect just doesn't know that.)

I also don't see a way of disabling this other than patching i3status-rust – that's fine, I do that anyway, but again it would be nice if that could be mentioned somewhere.

Atreyagaurav: So it does that even when you don't use the cellular network plugin from kdeconnect? if that's the case then it'd be a bug. 

IF you don't use kdeconnect network plugin then that code line should be ignored.

Atreyagaurav: I'd tried it before, but I tried it again, and if you uncheck the connectivity monitor the block simply disappears.

image
image

image
image

MaxVerevkin: Also an option can be added to select the source of block's state: battery, network or both.

sersorrel: ahh, cool, I didn't realise there was a kdeconnect plugin I could disable. thanks :)

Issue #1867, opened 2023-04-15
Support Mullvad in vpn widget


totikom:
Right now vpn widget supports only Nordvpn, but it should be trivial to adapt it's code to Mullvad.
I've started working on it and hopefully create a PR today or tomorrow.

totikom:
I've found that  cb994547 seems to break vpn widget entirely (at least for disconnected state). I'll try to fix it for both Mullvad (WIP) and NordVPN drivers.

MaxVerevkin:
@totikom thanks for reporting! If you find a way to fix it, please submit the fix in a separate PR from the feature one.

PR #1876, opened 2023-04-24
Fixed vpn widget rendering

totikom:
this fix the issue, mentioned in #1867.
The bug was caused by the order, in which `Widget::set_values` and `Widget::set_format` methods were called.

By the way, semantics of `Widget::set_values` is very unintuitive: it sets new values for `State::Format` case and silently does nothing on other cases.

Should I create a PR, which will change this behavior to something more intuitive?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,18 @@
+# i3status-rust 0.31.0 [unreleased]
+
+### New Blocks and Features
+
+* Battery: added `charging_format` config option.
+
+### Bug Fixes and Improvements
+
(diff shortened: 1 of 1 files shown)


totikom:
> I think I fixed this in [55e9334](https://github.com/greshake/i3status-rust/commit/55e93340f24c5fe18fec8f4be9572158d9a571ff) by allowing `.set_values()` before `.set_format()`, can you verify please?

Yep, I'll check it.
If it fixes the issue, I'll change #1877 so it will not depend on my fix.

PR #1889, opened 2023-05-01
sound: reconnect to pa

### LINKED ISSUE #334: Reconnection to pulseaudio
How to reproduce: run i3status-rust with the following config and then run `pulseaudio -k`.

```toml
[[block]]
block = "sound"
```

Pulseaudio is restarted (since it a normal behavior for a daemon), but no reconnection is made. I don't really know, is it a significant issue, since it can be fixed with an i3 restarting. But still no notes that something went wrong.

MaxVerevkin:
Fixes #334

Ideally this should be reflected in the UI, but for now I suppose it is better than just stalling the block.

Also, apparently there is a bug in `libpulse_binding` as it is possible to segfault w/o `unsafe`.

### DIFF
--- a/src/blocks/sound/pulseaudio.rs
+++ b/src/blocks/sound/pulseaudio.rs
@@ -13,6 +13,7 @@ use crossbeam_channel::{unbounded, Sender};
 
 use std::cmp::{max, min};
 use std::convert::{TryFrom, TryInto};
+use std::io;
 use std::os::fd::RawFd;
 use std::sync::Mutex;
 use std::thread;
@@ -49,7 +50,7 @@ struct Connection {
 
 struct Client {
     send_req: Sender<ClientRequest>,
-    wake_tx: RawFd,
+    ml_waker: MainloopWaker,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


MaxVerevkin:
Is lowlevel PA protocol documented? Is there something like https://wayland-book.com/protocol-design/wire-protocol.html but for pulseaudio?

src/blocks/sound/pulseaudio.rs
```diff
@@ -182,78 +198,106 @@ impl Connection {
 impl Client {
     fn new() -> Result<Client> {
         let (send_req, recv_req) = unbounded();
+        let (wake_rx, wake_tx) = nix::unistd::pipe().unwrap();
+
+        extern "C" fn wake_cb(
```
bim9262, line 203: Could you add some comments from here through line 210? I found it pretty difficult to appreciate what's happening.

bim9262:
While I wasn't able to grok lines 201-210 the change looked good overall and my preliminary testing looks good too. Now if only players I use would also reconnect...

src/blocks/sound/pulseaudio.rs
MaxVerevkin: Yeah, sure! I will probably refactor this code into something like `struct MainloopWaker`.

PR #1893, opened 2023-05-08
Added pacman_command config option

I60R:
Allows to avoid using `fakeroot` if AUR helper is available


MaxVerevkin (approved):

I60R:
Can you reproduce this error with `paru` (a newer AUR helper a replacement for `yay`)?

src/blocks/pacman.rs
```diff
@@ -150,12 +151,13 @@ static PACMAN_DB: Lazy<PathBuf> = Lazy::new(|| {
...
-    pub format_up_to_date: FormatConfig,
-    pub warning_updates_regex: Option<String>,
-    pub critical_updates_regex: Option<String>,
+    interval: Seconds,
+    format: FormatConfig,
+    format_singular: FormatConfig,
+    format_up_to_date: FormatConfig,
+    warning_updates_regex: Option<String>,
+    critical_updates_regex: Option<String>,
+    pacman_command: Option<String>,
```
MaxVerevkin, line 160: Probably a merge/rebase error, please make them `pub`.

I60R: Okay, wait a few minuts

I60R: But it seems that they don't need to be `pub`

MaxVerevkin: You can technically use `i3status-rust` as a library and configure your bar from rust, which requires all config options to be public.

bim9262:
> Can you reproduce this error with `paru` (a newer AUR helper a replacement for `yay`)?

I had the same error with `paru`, but I think that was actually because of the svn to git transition altering the dbs.
Now with both `paru`/`yay` I see the following:

```bash
$ paru -Qu --dbpath=/var/lib/pacman/sync; echo $?
1
$  paru -Qu
jupyter-nbclient 0.7.4-2 -> 0.8.0-1
jupyter-server-fileid 0.8.0-2 -> 0.9.0-1
libheif 1.16.1-2 -> 1.16.2-1
typos 1.14.10-1 -> 1.14.11-1
wlroots-hidpi-xprop-git 0.17.0.20230522.165315-1 -> latest-commit
yay 12.0.4-1 -> 12.0.5-1
```

I think we might also want to suggest `-n` be used so that packages from AUR are excluded.

```bash
$ paru -Qun
jupyter-nbclient 0.7.4-2 -> 0.8.0-1
jupyter-server-fileid 0.8.0-2 -> 0.9.0-1
libheif 1.16.1-2 -> 1.16.2-1
typos 1.14.10-1 -> 1.14.11-1
```

I60R:
@bim9262 It seems that on my machine packages from AUR already were excluded without `-n` option. Must downgrade some packages from either AUR and pacman and check again

I60R:
I figured it out: an appropriate `aur_command` would be `paru -Qum` and appropriate `pacman_command` would  be `paru -Qun`

bim9262:
Another option for an appropriate `aur_command` would be `paru -Qua` (I don't think it matters which is used)

bim9262:
I'm not so sure that running this without the fakeroot is actually really all that desirable, unless there's something else updating `/var/lib/pacman/sync` on ones system.

src/blocks/pacman.rs
```diff
@@ -304,20 +312,26 @@ pub async fn run(config: Config, mut api: CommonApi) -> Result<()> {
...
-    if !has_command("fakeroot").await? {
-        Err(Error::new("fakeroot not found"))
-    } else {
-        Ok(())
-    }
+async fn get_pacman_available_updates(pacman_command: &Option<String>) -> Result<String> {
+    let Some(pacman_command) = pacman_command else {
+        return get_pacman_available_updates_fakeroot().await
+    };
+    let stdout = Command::new("sh")
```
bim9262, line 324: If a pacman_command is supplied then there's no updating the `sync` dir (neither `/var/lib/pacman/sync` nor `/tmp/checkup-db-i3statusrs/sync`).

bim9262:
@I60R should we close out this PR or are you still working on an alternate approach?

PR #1908, opened 2023-05-26
Add shell.nix for dev environment

cortices:
Take 3. Should be just the commit for this feature. sorry about all the extra emails you must be getting!

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,16 @@ In addition, some blocks have debug logs that can be enabled like so: `RUST_LOG=
 
 We welcome new contributors! Take a gander at [CONTRIBUTING.md](CONTRIBUTING.md).
 
+### Nix environment
+
+`shell.nix` is provided to configure the development environment for `i3status-rust`.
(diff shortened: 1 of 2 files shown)


ammgws:
I'm on Arch, but this feels out of place... Why does it need to be upstream?

bim9262:
I'm also on arch, but you're right it does feel out of place... but I don't know much about how nix works

bim9262:
Perhaps https://github.com/NixOS/nixpkgs would be more appropriate?

cortices:
nixpkgs isn't really the right place, that's for packages (of which there already is one in home-manager instead).
shell.nix file is to set up the requisite development dependencies. It shouldn't require changes -- it just tells nixOS what is needed to build a rust project, without necessarily having a rust env already on the machine. if you don't want it in the project, you don't have to keep it.

MaxVerevkin:
Any thoughts on this being in the wiki instead?

PR #1909, opened 2023-05-30
use is_some_and(..) instead of map_or(false, ..)

MaxVerevkin:
- Easier to read and understand
- A bit shorter
- Requires Rust 1.70 (1 June, 2023)


cfsmp3:
Is this work it requiring a rust version that is not packaged in the major distros?
I know anyone can just fetch 1.70 and use rustup instead of packaged rust, but some distro (debian, ubuntu) aren't there yet...

MaxVerevkin:
> Is this work it requiring a rust version that is not packaged in the major distros?

It's not released yet :) I'll wait a few weeks before merging this.

PR #1910, opened 2023-05-31
Allow custom separator for rotating text

### LINKED ISSUE #1818: [music] Make the rotating text sepratator customizable
Actually, the character inserted between the end and the start of the rotating text is `|` (cf. https://github.com/greshake/i3status-rust/blob/master/src/formatting/formatter.rs#L185).

Could we consider make it customizable?

bim9262:
Fixes #1818

### DIFF
--- a/src/formatting/formatter.rs
+++ b/src/formatting/formatter.rs
@@ -16,6 +16,7 @@ use crate::escape::CollectEscaped;
 const DEFAULT_STR_MIN_WIDTH: usize = 0;
 const DEFAULT_STR_MAX_WIDTH: usize = usize::MAX;
 const DEFAULT_STR_ROT_INTERVAL: Option<f64> = None;
+const DEFAULT_STR_ROT_SEP: Option<String> = None;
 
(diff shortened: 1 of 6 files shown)


src/formatting/formatter.rs
```diff
@@ -191,13 +199,14 @@ impl Formatter for StrFormatter {
                 let width = text.chars().count();
                 Ok(match (self.rot_interval_ms, self.init_time) {
                     (Some(rot_interval_ms), Some(init_time)) if width > self.max_width => {
-                        let width = width + 1; // Now we include '|' at the end
+                        let rot_separator = self.rot_separator.clone().unwrap_or("|".to_string());
+                        let width = width + rot_separator.len(); // Now we include `rot_separator` at the end
```
MaxVerevkin, line 203: This should be `.chars().count()` to work better with unicode (although ideally we should probably count [graphemes](https://docs.rs/unicode-segmentation/latest/unicode_segmentation/trait.UnicodeSegmentation.html#tymethod.graphemes)).

MaxVerevkin (approved):

PR #1912, opened 2023-06-04
retry 3 times before displaying an error

### LINKED ISSUE #1838: Recover from "Failed to render full text"
From time to time, the `net` block of my status line gets stuck in

```
Error in net: Failed to render full text. (Cause: Placeholder 'ip' not found)
```

This persists until I restart i3status with a sway config reload.

### LINKED ISSUE #1891: feat: retry policy on `external_ip` block
https://github.com/greshake/i3status-rust/blob/49499c1ca624228c4b9e88b62b36cd0fa5e2dce5/src/blocks/external_ip.rs#L230

In my case the message "Failed to request current location" appears longer than the ip. Maybe basic 3 retries...

MaxVerevkin:
Closes #1891
Closes #1838

### DIFF
--- a/cspell.yaml
+++ b/cspell.yaml
@@ -159,7 +159,6 @@ words:
   - wlroots
   - wlsunset
   - wofi
-  - Wttr
   - wttr
   - xclip
   - xesam
--- a/cspell.yaml
+++ b/cspell.yaml
@@ -23,6 +23,7 @@ words:
   - aplay
   - armv
   - arphrd
+  - backon
   - BCHARGE
   - bluez
   - bokmaal
@@ -158,8 +159,8 @@ words:
   - wlroots
   - wlsunset
   - wofi
-  - wttr
   - Wttr
+  - wttr
   - xclip
   - xesam
   - xkbswitch
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -165,6 +165,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
+name = "backon"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0c1a6197b2120bb2185a267f6515038558b019e92b832bb0320e96d66268dcf9"
+dependencies = [
+ "fastrand",
+ "futures-core",
+ "pin-project",
+ "tokio",
+]
+
+[[package]]
 name = "base64"
 version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -941,6 +953,7 @@ version = "0.31.6"
 dependencies = [
  "async-once-cell",
  "async-trait",
+ "backon",
  "calibright",
  "chrono",
  "chrono-tz",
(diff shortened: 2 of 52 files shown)


bim9262:
So this will change the behaviour from trying again every 5 seconds (and showing an error) to trying three times with 1 and 2 second delays between attempts before displaying an error and then sleeping for 30 seconds? That 30 second window seems like it's potentially a really long timeout depending on what failed.

cspell.yaml
```diff
@@ -158,8 +159,8 @@ words:
   - wlroots
   - wlsunset
   - wofi
-  - wttr
   - Wttr
```
bim9262, line 162: looks like the tools we used to sort the list disagree about the order of capitalized words. I just checked though and we can remove `Wttr` since `wttr` is there too.

MaxVerevkin: I used nvim's `:sort i`.

> I just checked though and we can remove Wttr since wttr is there too.

Awesome.

MaxVerevkin:
> So this will change the behaviour from trying again every 5 seconds (and showing an error) to trying three times with 1 and 2 second delays between attempts before displaying an error and then sleeping for 30 seconds?

Yes.

> That 30 second window seems like it's potentially a really long timeout depending on what failed.

True. It can be reverted back to five.

MaxVerevkin:
By the way, I noticed README states that 

> When the state is "Error", a short error will be displayed in the block. The full message can be toggled by clicking on the block (overrides any click actions defined in the config). **The block will be restarted after error_interval has elapsed.**

Which is not what is happening right now. But maybe we can literally restart the failed block and get rid of all the `.recoverable()` calls? Then only the blocks that need retry policy (basically those that make web requests) will have it. It wasn't possible before async rewrite because some blocks used to spawn threads, but it is not as issue any more. Any thoughts?

bim9262:
I like the sound of that 👍

MaxVerevkin:
One annoying thing is that `blocks::CommonApi` is not `Clone` because of `event_receiver`, which is needed for the `custom_dbus` block, which doesn't even use `event_receiver`.

ammgws:
We could perhaps store state for those on disk if it must absolutely respect `"once"`

MaxVerevkin:
Never mind, the last autolocate result can be stored in a static mutex.

MaxVerevkin:
Implementing #1838 turned out to be easier than I thought. However I'm not convinced that https://github.com/greshake/i3status-rust/pull/1912/commits/453bc9de3f7db2c4181d0b4a561ca1922f3a7f73 doesn't break anything, testing is needed. But I haven't noticed any issues.

bim9262:
So far my testing hasn't shown any issues caused by https://github.com/greshake/i3status-rust/commit/453bc9de3f7db2c4181d0b4a561ca1922f3a7f73.

Do you think that using `action.as_ref()` instead of `&*action` is easier to read? (or is there some other downside to that's not not obvious...)

```diff
-                Some(action) = actions.recv() => match &*action {
+                Some(action) = actions.recv() => match action.as_ref() {
```

MaxVerevkin:
> Do you think that using action.as_ref() instead of &*action is easier to read?

I personally don't see much difference, but if you do, we can change it.

> or is there some other downside to that's not not obvious...

I guess it is just a matter of preference.

bim9262:
I don't mind either way, but I think that maybe `as_ref` will be a bit more intuitive to newcomers who wish to contribute.

bim9262:
lgtm @MaxVerevkin

PR #1919, opened 2023-06-30
Added premade configurations for Catppuccin theme set

sthurston99:
Created new theme toml files under files/themes for Catppuccin Mocha, Macchiato, Frappe, and Latte. Styled them as if using the base tone of each theme as the background for the i3bar. Compiled and tested each theme and readability is good, all the colors render properly, and should be ready to add to the rest of the existing themes.


files/themes/ctp-latte.toml
```diff
@@ -0,0 +1,13 @@
+idle_bg = "#ccd0da"
+idle_fg = "#4c4f69"
+info_bg = "#1e66f5"
+info_fg = "#eff1f5"
+good_bg = "#40a02b"
+good_fg = "#eff1f5"
+warning_bg = "#df8eld"
```
MaxVerevkin, line 7: This is not valid base-16. A typo?

sthurston99: yep looks like, just pushed a fix

MaxVerevkin:
Could you also update `doc/themes.md`?

gen-screenshots/swayconfig_i3rs
```diff
@@ -4,6 +4,7 @@ bar {
     status_padding 0
     tray_output none
     font pango:DejaVu Sans Mono, Font Awesome 6 Free 10
+    #font pango:DejaVu Sans Mono for Powerline, Font Awesome 6 Free 10
```
bim9262, line 7: I'm not sure why, but I had an extra space between the separator and the text, unless I used the Powerline version of DejaVu.

MaxVerevkin (approved):

PR #1924, opened 2023-07-23
Update missing default icons/remove old overrides

bim9262:
Add script to verify that the default icons in src/icons.rs match files/icons/*.toml

Remove unused default icons
    
Removed:
* bat_discharging
* bat_empty
* bat_full
    
Updated `doc/themes.md` to remove the unused icons that were being "overridden"

Add default net_cellular icons


doc/themes.md
```diff
@@ -89,10 +89,14 @@ warning_bg = { link = "good_bg" }
 [icons]
 icons = "awesome6"
 [icons.overrides]
-bat = " | | "
-bat_full = " |X| "
-bat_charging = " |^| "
-bat_discharging = " |v| "
+bat = [
```
bim9262, line 92: We could also just make this `bat = ["|E|", "|_|", "|=|", "|F|"]` like in the `examples/config.toml`. I didn't see this at first because I forgot to rebase for I started.

bim9262:
Just went back and added a pre-commit hook so that we can make sure we keep the defaults and all of the icon files in sync. Sorry for the spammy edits...

src/icons.rs
```diff
@@ -20,16 +20,19 @@ impl From<&'static str> for Icon {
     }
 }
 
+impl From<Vec<&'static str>> for Icon {
```
MaxVerevkin, line 23: Since the memory from `Vec` is not reused, this can be `&[&str]` instead of `Vec<&'static str>`.

verify_icon_files.sh
```diff
@@ -0,0 +1,21 @@
...
+
+TMPDIR="$(mktemp -d)"
+trap 'rm -rf -- "$TMPDIR"' EXIT
+
+EXITCODE=0
+
+for f in files/icons/*.toml; do 
+    filename=$(basename $f)
+    echo == Verifying $f ==
+    comm -3 <(sed -n '/impl Default for Icons {/, /}/p' src/icons.rs | awk -F '"' '/=>/ {print $2}' | sort) <(awk '!/^#/ && /=/ {print $1}' $f | sort) > $TMPDIR/$filename
```
MaxVerevkin, line 11: Can't we just reuse the same temp file?

src/icons.rs
bim9262: When I tried that I got:

```
error[E0277]: the trait bound `Icon: From<[&str; 6]>` is not satisfied
   --> src/util.rs:159:41
    |
159 |           $map.insert($key.into(), $value.into());
    |                                           ^^^^ the trait `From<[&str; 6]>` is not implemented for `Icon`
    |
   ::: src/icons.rs:32:14
    |
32  |           Self(map! {
    |  ______________-
33  | |             "backlight" => "BRIGHT",
34  | |             "bat" => "BAT",
35  | |             "bat_charging" => "CHG",
...   |
106 | |             "xrandr" => "SCREEN"
107 | |         })
    | |_________- in this macro invocation
    |
    = help: the following other types implement trait `From<T>`:
              <Icon as From<&'static str>>
              <Icon as From<&[&str]>>
    = note: required for `[&str; 6]` to implement `std::convert::Into<Icon>`
    = note: this error originates in the macro `map` (in Nightly builds, run with -Z macro-backtrace for more info)
    ```

Comment (bim9262) [reply src/icons.rs:23]: I tried `["..", ".."]` and `&["..", ".."]`, and neither worked, but `["..", ".."].as_slice()` and `["..", ".."].as_ref()` work.

I thought `&["..", ".."]` and  `["..", ".."].as_ref()` were equivalent.

Comment (bim9262) [reply src/icons.rs:23]: Oh, we could do this:

```rust
impl<const N: usize> From<[&str; N]> for Icon {
    fn from(value: [&str; N]) -> Self {
        Self::Progression(value.iter().map(|s| s.to_string()).collect())
    }
}
```
or
```rust
impl<const N: usize> From<&[&str; N]> for Icon {
    fn from(value: &[&str; N]) -> Self {
        Self::Progression(value.iter().map(|s| s.to_string()).collect())
    }
}
```

Using `["..", ".."]` and `&["..", ".."]` respectively. The former looks cleaner, but if it's better (performance?) to pass the ref we can use the latter.

MaxVerevkin: > but if it's better (performance?) to pass the ref

I think in our case of 6 elements it doesn't matter :)

MaxVerevkin (approved):

Issue #1774, opened 2023-02-24
sound: don't look up icons when not needed


ptrcnull:
with the following config:
```
[[block]]
block = "sound"
format = " $output_name($volume) "
step_width = 5
```

sometimes i get `Icon 'volume_muted' not found`, despite the icon not being used in the format at all

MaxVerevkin:
This particular error was fixed in #1785. As for not looking up unused icons, this might be implemented as an optimization in the future.

PR #1926, opened 2023-07-28
Defer icon lookup until formatting

bim9262:
See #1774

Let me know if you think this is a good approach to only set the required values needed based on the format that's being used.

### DIFF
--- a/src/formatting/value.rs
+++ b/src/formatting/value.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use super::formatter;
 use super::unit::Unit;
 use super::Metadata;
@@ -13,7 +15,7 @@ pub struct Value {
 #[derive(Debug, Clone)]
 pub enum ValueInner {
     Text(String),
-    Icon(String),
+    Icon(Cow<'static, str>, Option<f64>),
     Number { val: f64, unit: Unit },
     Datetime(DateTime<Utc>, Option<Tz>),
     Flag,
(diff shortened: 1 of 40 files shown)


MaxVerevkin:
I think this kind of optimization has potential. In particular if we could actually skip expensive computation of values that are not used. For example, in `net` block, we may entirely skip IP address fetch if it is not present in the format.

However I am not sure if doing so for `backlight` or `apt` makes much sense. `brightness` and `count` are most certainly present - why would anyone use this block but not use it's _only_ feature?

As for icons, instead of resolving them immediately (as in `Value::icon(api.get_icon("gpu")?)`), we can do so inside a formatter (so this becomes `Value::icon("gpu")`). This way unused icons are never actually fetched, which takes care of #1774.

bim9262:
> As for icons, instead of resolving them immediately (as in `Value::icon(api.get_icon("gpu")?)`), we can do so inside a formatter (so this becomes `Value::icon("gpu")`). This way unused icons are never actually fetched, which takes care of https://github.com/greshake/i3status-rust/issues/1774.

I like that! I'll give that a go. We'll need to handle the icon progressions in the formatter too...

src/formatting/value.rs
```diff
@@ -13,7 +13,7 @@ pub struct Value {
 #[derive(Debug, Clone)]
 pub enum ValueInner {
     Text(String),
-    Icon(String),
+    Icon(String, Option<f64>),
```
MaxVerevkin, line 16: We can use `Cow<'static, str>` here since most of the time icon names are static

MaxVerevkin:
Looks great!

src/formatting/value.rs
bim9262: Just added that in :cow:!

MaxVerevkin (approved):

Issue #1930, opened 2023-08-05
error with reading memory usage


ahumblescientist:
here is what i3status-rust outputs:
![2023-08-05_23-32](https://github.com/greshake/i3status-rust/assets/138120547/b09ae97e-c507-484a-aa61-e0163666182e)

sudo ps_mem ouput:
![2023-08-05_23-33ps](https://github.com/greshake/i3status-rust/assets/138120547/98f5e270-7979-4803-9158-0dc72596849f)
(sorry for bad english)

ammgws:
Hmm, that percentage seems pretty sus. Thought we had fixed memory calcs a while back.

Do you use ZFS?

ahumblescientist:
> Hmm, that percentage seems pretty sus. Thought we had fixed memory calcs a while back.
> 
> Do you use ZFS?

nope

bim9262:
@ahumblescientist does the value in your status bar match the output of `free -g`?
On my computer `free -g` and `ps_mem` report different amounts of used memory. I'm pretty sure `ps_mem` just doesn't count everything.

ahumblescientist:
> @ahumblescientist does the value in your status bar match the output of `free -g`? On my computer `free -g` and `ps_mem` report different amounts of used memory. I'm pretty sure `ps_mem` just doesn't count everything.

no, it says used 2GB, unlike i3status-rs which says that the used is 12GB

ammgws:
What does htop show?

bim9262:
Can you share the format you are using for the memory block?

bim9262:
So you are showing how much is available, not how much is being used

ahumblescientist:
@bim9262 
this is the dumbest i have ever felt to be honest, i rlly didnt notice that

bim9262:
Haha no worries, glad we got it sorted out

ammgws:
Possible cause of confusion could be that the percentage shown in the default format string is the total used %.

Perhaps should change

`$mem_avail.eng(prefix:M)/$mem_total.eng(prefix:M)($mem_total_used_percents.eng(w:2))`

to

`$mem_avail.eng(prefix:M)/$mem_total.eng(prefix:M)($mem_avail_percents.eng(w:2))`

?

bim9262:
I think that `used/total(used%)` would be a better default. :man_shrugging:

ahumblescientist:
@ammgws 
[...]

PR #1932, opened 2023-08-12
Update default memory format

bim9262:
Based on conversation in #1930

### DIFF
--- a/src/blocks/memory.rs
+++ b/src/blocks/memory.rs
@@ -4,7 +4,7 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `format` | A string to customise the output of this block when in "Memory" view. See below for available placeholders. | `" $icon $mem_avail.eng(prefix:M)/$mem_total.eng(prefix:M)($mem_total_used_percents.eng(w:2)) "`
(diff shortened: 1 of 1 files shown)


MaxVerevkin (approved):

src/blocks/memory.rs
```diff
@@ -86,7 +86,7 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
         .await?;
 
     let mut format = config.format.with_default(
-        " $icon $mem_avail.eng(prefix:M)/$mem_total.eng(prefix:M)($mem_total_used_percents.eng(w:2)) ",
+        " $icon $mem_used.eng(prefix:Mi)/$mem_total.eng(prefix:Mi)($mem_used_percents.eng(w:2)) ",
```
bim9262, line 89: Just wanted to double check, do you think using the mem_ued and base 2 units is the way to go?

I tried to do some research about what is used (base 2 or 10) and I believe RAM and disk size are usually reported with base 2 units while network speed is measured in base 10 units.

MaxVerevkin: Probably. At least, `htop` does so.

PR #1935, opened 2023-08-14
amd_gpu: select device automatically

### LINKED ISSUE #1928: amd_gpu: select device automatically
Default card was apparently renamed from `card0` to `card1` (kernel 6.4.8-arch1-1).

Instead of defaulting to `card0` we can `read_dir` and select the first directory starting with "card".

MaxVerevkin:
Resolves #1928

### DIFF
--- a/src/blocks/amd_gpu.rs
+++ b/src/blocks/amd_gpu.rs
@@ -4,7 +4,7 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `device` | The device in `/sys/class/drm/` to read from. | `"card0"`
+//! `device` | The device in `/sys/class/drm/` to read from. | Any AMD card
 //! `format` | A string to customise the output of this block. See below for available placeholders. | `" $icon $utilization "`
 //! `format_alt` | If set, block will switch between `format` and `format_alt` on every click | `None`
 //! `interval` | Update interval in seconds | `5`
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
> I guess you only have one GPU and it is card1?

Yes.

> Wonder if we can check any of the other properties in /sys/class/drm/card0x to prefer ones that say 'AMD' or something.

That would be nice. `rg -i amd /sys/class/drm/card1/device/` showed a few results, `/sys/class/drm/card1/device/uevent` is probably the best. On my system it has a line `DRIVER=amdgpu`.

I'm not aware of a way to check if a given GPU is integrated or not.

src/blocks/amd_gpu.rs
```diff
@@ -101,29 +111,69 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
...
+            }
+
+            let mut path = entry.path();
+            path.push("device");
+
+            let Ok(uevent) = read_file(path.join("uevent")).await else {
+                continue;
+            };
+
+            if uevent.contains("DRIVER=amdgpu") {
```
ammgws, line 148: Maybe niche but if using proprietary drivers for whatever reason then this might be different?

Another option could be to use the PCI_ID, 1002 should always be AMD: https://admin.pci-ids.ucw.cz/read/PC/1002

MaxVerevkin: Do other drivers provide the same sysfs interface?

> Another option could be to use the PCI_ID, 1002 should always be AMD

Sounds good.

ammgws (approved):

PR #1936, opened 2023-08-24
switch to std::sync::mpsc from crossbeam-channel

MaxVerevkin:
Since Rust 1.72, Sender<T> implements Sync if T: Send, which lets us use it instead of crossbeam-channel - one less dependency.

Bumps MSRV to 1.72.

### DIFF
--- a/src/blocks/sound/pulseaudio.rs
+++ b/src/blocks/sound/pulseaudio.rs
@@ -9,8 +9,6 @@ use libpulse_binding::mainloop::standard::{IterateResult, Mainloop};
 use libpulse_binding::proplist::{properties, Proplist};
 use libpulse_binding::volume::{ChannelVolumes, Volume};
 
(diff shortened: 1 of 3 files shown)


src/blocks/sound/pulseaudio.rs
```diff
@@ -238,11 +236,11 @@ impl Client {
                 }
 
                 loop {
+                    use std::sync::mpsc::TryRecvError;
```
bim9262, line 239: Should this be at the top of the file? Is there any impact of having it in the body (of a loop)?

MaxVerevkin: `tokio::sync::mpsc::error::TryRecvError` has the same name and we use both std and tokio channels in this file, so I thought minimizing the scope of this `use` might be a good idea.

Issue #1561, opened 2022-07-10
How does pacman block work?


c02y:
I thought it would check update in the background and notify me in the status bar when updates are available. But it seems it is not working like this.

My config:
```
[[block]]
block = "pacman"
format = "{pacman:1}/{aur:1}"
aur_command = "paru -Qua"
interval = 60
hide_when_uptodate = true
on_click = 'fakeroot pacman -Qu --dbpath /tmp/checkup-db-chz/ | rg -v "[ignored]" | rofi -dmenu'
```

It never notifies me that I got updates, then I manually check it using command `checkupdates`, I got two newer packages available. `on_click` also displays no package.

So basically I have to manually update my pacman database, then pacman block for i3status-rust will display those updates in the bar?

ammgws:
> I thought it would check update in the background and notify me in the status bar when updates are available.

That is what it does for me. Try removing the aur stuff and see if it makes a difference? Otherwise we might have to add some debug statements to see what is going on.

c02y:
> Try removing the aur stuff

I'm sorry, I don't understand this part, what exactly should I remove about aur stuff?

ammgws:
Like this

```toml
[[block]]
block = "pacman"
format = "{pacman:1}"
interval = 60
```

c02y:
> Like this
> 
> ```toml
> [[block]]
> block = "pacman"
> format = "{pacman:1}"
> interval = 60
> ```

Oh, you mean remove aur part in the config. Just tried it, nothing, it still shows 0 updates, and I haven't synced my db or updated my Arch Linux for several days since I started to use pacman block, now `checkupdates` command in terminal lists 43 new packages available.

Out of i3status-rust topic, do you happen to know how to regularly sync the pacman db in the background using some kind of service or timer. There is a file `/usr/lib/systemd/system/pacman-filesdb-refresh.service` from `pacman-contrib` package, but I failed to activate it:
```
>> sudo systemctl start pacman-filesdb-refresh.service
[sudo] password for user:
>> sudo systemctl status pacman-filesdb-refresh.service
○ pacman-filesdb-refresh.service - Refresh pacman files databases
     Loaded: loaded (/usr/lib/systemd/system/pacman-filesdb-refresh.service; static)
     Active: inactive (dead)

Jul 12 13:10:03 archlinux systemd[1]: pacman-filesdb-refresh.service: Deactivated successfully.
[...]

PR #1938, opened 2023-08-29
pacman: create a checkup-db dir per user

MaxVerevkin:
See https://github.com/greshake/i3status-rust/issues/1561#issuecomment-1697865332

This is a breaking change.


bim9262 (approved):
lgtm

MaxVerevkin:
> What part of this is breaking though?

This broke my script that shows outdated packages when clicking on the block (there is an example in the docs).

PR #1941, opened 2023-09-03
Add weather icons for night

bim9262:
Also split off Fog/Mist from Cloudy

One thing I did consider, but didn't do was to make the weather a progression:

So this:

```toml
weather_fog = "\ue313" # nf-weather-fog
weather_fog_night = "\ue346" # nf-weather-night_fog
```

becomes:

```toml
weather_fog = [
    "\ue313", # nf-weather-fog
     "\ue346", # nf-weather-night_fog
]
```

This would also mean that there's less clutter for fonts that don't differentiate between day and night:

```toml
weather_fog = "\uf0c2" # fa-cloud
weather_fog_night = "\uf0c2" # fa-cloud
```

becomes:

```toml
weather_fog = "\uf0c2" # fa-cloud
```

Let me know what you think about that :)


MaxVerevkin:
> This would also mean that there's less clutter for fonts that don't differentiate between day and night:

> Let me know what you think about that :)

Sounds nice! The only problem I can imagine is that this would make the night variants unnameable using the current `^icon_` syntax, but this applies to all progressions and after seven months no one has complained (https://github.com/greshake/i3status-rust/pull/1720#issuecomment-1704334119), so...

bim9262:
I'm not sure why someone would want to... But if you only wanted to show the night version you could override the icon in the config

MaxVerevkin:
> I'm not sure why someone would want to... But if you only wanted to show the night version you could override the icon in the config

I meant if someone wanted to reference the night version from their custom block or in some other block, for some reason. But since this applies to all progressions I is not much of a concern.

Also one weird thing about using a progression to represent day/night variant is that the order is not that intuitive.

Anyway, feel free to choose the option you think is better.

bim9262:
> Also one weird thing about using a progression to represent day/night variant is that the order is not that intuitive.

I think that's reason enough to leave it as is.

src/blocks/weather/met_no.rs
```diff
@@ -194,9 +201,11 @@ fn weather_to_icon(weather: &str) -> WeatherIcon {
         | "lightsleetandthunder"
         | "lightrainandthunder"
         | "lightsnowandthunder"
-        | "lightssleetshowersandthunder"
-        | "lightssnowshowersandthunder"
-        | "lightrainshowersandthunder" => WeatherIcon::Thunder,
+        | "lightssleetshowersandthunder" // There's a typo in the api it will be fixed in the next version to the following entry
+        | "lightsleetshowersandthunder"
```
MaxVerevkin, line 205: Can you please point me to where you found this info?

bim9262: https://api.met.no/weatherapi/locationforecast/2.0/documentation

> Note: There is a typing error in lightssleetshowersandthunder and lightssnowshowersandthunder (extra "s" after "light"). Unfortunately, correcting this would mean breaking existing applications, so it has been postponed to the next version of weathericon/locationforecast.

MaxVerevkin (approved):
lgtm

PR #1942, opened 2023-09-03
Make the backlight icons consistent

bim9262:
Make it so that the moon is filling in from left to right as the brightness increases. This matches what emoji and material-nf does too.


MaxVerevkin:
It's... complicated. See #573.

bim9262:
I see the argument made about the issue being with the single color of the font...

with :

```toml
[[block]]
block = "backlight"
[block.theme_overrides]
idle_bg = { link = "good_bg" }
idle_fg = { link = "good_fg" }
```

I'll close this out and leave well enough alone. I'll revert the change to emoji so at least it's the same as font awesome.

Issue #1485, opened 2022-05-29
Always use "lat, lon" for OpenWeatherMap requests


Donatzsky:
Using anything other than `lat, lon` for the current weather API is deprecated. Also, city name and zip-code can both be rather inaccurate (which affects `autolocate`).

From https://openweathermap.org/current :
> We highly recommend you using our [Geocoding API](https://openweathermap.org/api/geocoding-api). This API uses a much more accurate geographical database and supports both direct and reverse geocoding.
> 
> Please note that built-in API requests by city name, zip-codes and city id will be deprecated soon.

Deprecating `city id` does seem a little strange, though.

I suggest making the following changes:

- [ ] Keep using ipapi.co for `autolocate`, but use the coordinates instead of city name (#1593).
- [ ] Use the OMW geocoding API to convert `place` into `lat, lon`. Add a note in docs that this will consume API calls. Also, make it clear the format is `city, country code`.
- [ ] Add a `zip` key. Again, this consumes an API call.
- [ ] Keep `city_id` for now, but add a note that it may stop working in the future.

PR #1944, opened 2023-09-05
Add weather forecast

### LINKED ISSUE #1267: Future weather
The weather widget only supports fetching the current weather. It could be useful to add support for forecasts, something the openweathermap API seems to support as well.

Configuration wise this could simply be done as a single option, something like `forecast`, defaulting to `0` for the current weather, or taking a number in some time scale. The API has the [/onecall endpoint](https://openweathermap.org/api/one-call-api) returns forecasts for up to seven days, but I would guess most users are only interested in either the next few hours, or something more specific, such as "tomorrow afternoon".

The current implementation seems to be using the `/weather` endpoint, but that one does not support hourly forecasts in the free tier, unlike the `/onecall` endpoint. Switching to *only* use the `/onecall` endpoint would probably make the implementation easier, but it does reduce the number of daily calls. However, even for the free tier, it still allowed for 30,000 per month / 1,000 per day, which would probably be enough for most users even if they reuse the same API key on multiple computers.

The big difference between the two is that the `/onecall` endpoint does not support getting location by city id. It *only* accepts coordinates. As far as I could tell, they don't have an API for converting a city ID into coordinates. They do have a geocoding API, but that takes `{zip code},{country code}`. It would, regardless, be a breaking change, but that still seems preferable than supporting different config options for different endpoints, depending on whether the user wants forecasts or not.

If this sounds at all useful, I'd be happy to give implementing this a shot myself. If so I would also like to add support for showing UV Index, but since that's included in the data that's already being fetched I didn't feel like it warranted it's own issue.

bim9262:
Fixes #1267

### DIFF
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -15,6 +15,7 @@
 //! ----|--------|--------
 //! `service` | The configuration of a weather service (see below). | **Required**
 //! `format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | `" $icon $weather $temp "`
+//! `format_alt` | If set, block will switch between `format` and `format_alt` on every click | `None`
 //! `interval` | Update interval, in seconds. | `600`
 //! `autolocate` | Gets your location using the ipapi.co IP location service (no API key required). If the API call fails then the block will fallback to service specific location config. | `false`
 //! `autolocate_interval` | Update interval for `autolocate` in seconds or "once" | `interval`
@@ -33,12 +34,15 @@
 //! `zip` | OpenWeatherMap 'By {zip code},{country code}' search query. See [here](https://openweathermap.org/api/geocoding-api#direct_zip). Consumes an additional API call | Yes* | None
 //! `units` | Either `"metric"` or `"imperial"`. | No | `"metric"`
 //! `lang` | Language code. See [here](https://openweathermap.org/current#multi). Currently only affects `weather_verbose` key. | No | `"en"`
+//! `forecast_hours` | How many hours should be forecast (must be increments of 3 hours, max 120 hours) | No | 12
 //!
 //! One of `coordinates`, `city_id`, `place`, or `zip` is required. If more than one are supplied, `coordinates` takes precedence over `city_id` which takes precedence over `place` which takes precedence over `zip`.
 //!
 //! The options `api_key`, `city_id`, `place`, `zip`, can be omitted from configuration,
 //! in which case they must be provided in the environment variables
 //! `OPENWEATHERMAP_API_KEY`, `OPENWEATHERMAP_CITY_ID`, `OPENWEATHERMAP_PLACE`, `OPENWEATHERMAP_ZIP`.
+//!
+//! Forecasts are only fetched if forecast_hours > 0 and the format has keys related to forecast.
 //!
 //! # met.no Options
 //!
@@ -48,23 +52,38 @@
 //! `coordinates` | GPS latitude longitude coordinates as a tuple, example: `["39.2362","9.3317"]` | Required if `autolocate = false` | None
 //! `lang` | Language code: `en`, `nn` or `nb` | No | `en`
 //! `altitude` | Meters above sea level of the ground | No | Approximated by server
+//! `forecast_hours` | How many hours should be forecast | No | 12
 //!
 //! Met.no does not support location name.
 //!
 //! # Available Format Keys
 //!
-//!  Key              | Value                                                              | Type   | Unit
-//! ------------------|--------------------------------------------------------------------|--------|-----
-//! `icon`            | Icon representing the weather                                      | Icon   | -
-//! `location`        | Location name (exact format depends on the service)                | Text   | -
-//! `temp`            | Temperature                                                        | Number | degrees
-//! `apparent`        | Australian Apparent Temperature                                    | Number | degrees
-//! `humidity`        | Humidity                                                           | Number | %
-//! `weather`         | Textual brief description of the weather, e.g. "Raining"           | Text   | -
-//! `weather_verbose` | Textual verbose description of the weather, e.g. "overcast clouds" | Text   | -
-//! `wind`            | Wind speed                                                         | Number | -
-//! `wind_kmh`        | Wind speed. The wind speed in km/h                                 | Number | -
-//! `direction`       | Wind direction, e.g. "NE"                                          | Text   | -
+//!  Key                                         | Value                                                                         | Type   | Unit
+//! ---------------------------------------------|-------------------------------------------------------------------------------|--------|-----
+//! `location`                                   | Location name (exact format depends on the service)                           | Text   | -
+//! `icon{,_ffin}`                               | Icon representing the weather                                                 | Icon   | -
+//! `weather{,_ffin}`                            | Textual brief description of the weather, e.g. "Raining"                      | Text   | -
+//! `weather_verbose{,_ffin}`                    | Textual verbose description of the weather, e.g. "overcast clouds"            | Text   | -
+//! `temp{,_{favg,fmin,fmax,ffin}}`              | Temperature                                                                   | Number | degrees
+//! `apparent{,_{favg,fmin,fmax,ffin}}`          | Australian Apparent Temperature                                               | Number | degrees
+//! `humidity{,_{favg,fmin,fmax,ffin}}`          | Humidity                                                                      | Number | %
+//! `wind{,_{favg,fmin,fmax,ffin}}`              | Wind speed                                                                    | Number | -
+//! `wind_kmh{,_{favg,fmin,fmax,ffin}}`          | Wind speed. The wind speed in km/h                                            | Number | -
+//! `direction{,_{favg,fmin,fmax,ffin}}`         | Wind direction, e.g. "NE"                                                     | Text   | -
+//!
+//! You can use the suffixes noted above to get the following:
+//!
+//! Suffix    | Description
+//! ----------|------------
+//! None      | Current weather
+//! `_favg`   | Average forecast value
+//! `_fmin`   | Minimum forecast value
+//! `_fmax`   | Maximum forecast value
+//! `_ffin`   | Final forecast value
+//!
+//! Action          | Description                               | Default button
+//! ----------------|-------------------------------------------|---------------
+//! `toggle_format` | Toggles between `format` and `format_alt` | Left
 //!
 //! # Example
 //!
@@ -74,11 +93,13 @@
 //! [[block]]
 //! block = "weather"
 //! format = " $icon $weather ($location) $temp, $wind m/s $direction "
+//! format_alt = " $icon_ffin Forecast (9 hour avg) {$temp_favg ({$temp_fmin}-{$temp_fmax})|Unavailable} "
 //! [block.service]
 //! name = "openweathermap"
 //! api_key = "XXX"
 //! city_id = "5398563"
 //! units = "metric"
+//! forecast_hours = 9
 //! ```
 //!
 //! # Used Icons
@@ -99,6 +120,8 @@ use std::fmt;
 use std::sync::{Arc, Mutex};
 use std::time::Instant;
 
+use crate::formatting::Format;
+
 use super::prelude::*;
 
 pub mod met_no;
@@ -115,6 +138,7 @@ pub struct Config {
     pub interval: Seconds,
     #[serde(default)]
     pub format: FormatConfig,
+    pub format_alt: Option<FormatConfig>,
     pub service: WeatherService,
     #[serde(default)]
     pub autolocate: bool,
@@ -127,8 +151,11 @@ fn default_interval() -> Seconds {
 
 #[async_trait]
 trait WeatherProvider {
-    async fn get_weather(&self, autolocated_location: Option<Coordinates>)
-        -> Result<WeatherResult>;
+    async fn get_weather(
+        &self,
+        autolocated_location: Option<Coordinates>,
+        need_forecast: bool,
+    ) -> Result<WeatherResult>;
 }
 
 #[derive(Deserialize, Debug)]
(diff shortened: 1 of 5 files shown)


src/blocks/weather.rs
```diff
@@ -47,23 +51,34 @@
...
+//! `weather_verbose`   | Textual verbose description of the weather, e.g. "overcast clouds" | Text   | -
+//! `wind`              | Wind speed                                                         | Number | -
+//! `wind_kmh`          | Wind speed. The wind speed in km/h                                 | Number | -
+//! `direction`         | Wind direction, e.g. "NE"                                          | Text   | -
+//! `temp_forcast`      | Temperature forcast                                                | Number | degrees
+//! `apparent_forcast`  | Australian Apparent Temperature forcast                            | Number | degrees
+//! `humidity_forcast`  | Humidity forcast                                                   | Number | %
+//! `wind_forcast`      | Wind speed forcast                                                 | Number | -
+//! `wind_kmh_forcast`  | Wind speed forcast. The wind speed in km/h                         | Number | -
+//! `direction_forcast` | Wind direction forcast, e.g. "NE"                                  | Text   | -
```
bim9262, line 77: Notable omission is weather_forcast / weather_verbose_forcast. I suppose the most common in the sampled forecasts could be used...

src/blocks/weather.rs
```diff
@@ -32,13 +33,16 @@
...
 //! `lang` | Language code. See [here](https://openweathermap.org/current#multi). Currently only affects `weather_verbose` key. | No | `"en"`
+//! `forcast_hours` | How many hours should be forcast (must be increments of 3 hours, max 120 hours) | No | 0
 //!
 //! One of `city_id`, `place` or `coordinates` is required. If more than one are supplied, `city_id` takes precedence over `place` which takes place over `coordinates`.
 //!
 //! The options `api_key`, `city_id`, `place` can be omitted from configuration,
 //! in which case they must be provided in the environment variables
 //! `OPENWEATHERMAP_API_KEY`, `OPENWEATHERMAP_CITY_ID`, `OPENWEATHERMAP_PLACE`.
 //!
+//! Forcasts are disabled by default as they consume another api call
```
bim9262, line 44: I'm not sure that it's really a big deal to make this extra call... What do you think?

src/blocks/weather/open_weather_map.rs
```diff
@@ -1,13 +1,15 @@
 use super::*;
 use chrono::Utc;
+use serde::{de, Deserializer};
 
-pub(super) const URL: &str = "https://api.openweathermap.org/data/2.5/weather";
+pub(super) const CURRENT_URL: &str = "https://api.openweathermap.org/data/2.5/weather";
```
bim9262, line 5: #1267 mentions using /onecall, but this requires a different api key.

bim9262: Both of these call still allow city_id and place still, but #1485 should be tackled next.

bim9262:
Oh, also the forecast values are 0 if there's no forecast data. Perhaps we can display the value as missing some other way, or omit the keys from the map entirely.

src/blocks/weather.rs
```diff
@@ -331,6 +382,34 @@ fn convert_wind_direction(direction_opt: Option<f64>) -> &'static str {
     }
 }
 
+// Compute the average wind speed and direction
+fn average_wind(winds: &Vec<Wind>) -> Wind {
+    let mut opposite = 0.0;
+    let mut adjacent = 0.0;
```
MaxVerevkin, line 388: I think "north" and "east" are more descriptive

MaxVerevkin:
> omit the keys from the map entirely.

This is probably the best option, as users will be able to customize the output to their liking. It just needs to be documented very explicitly.

src/blocks/weather.rs
```diff
@@ -331,6 +382,34 @@ fn convert_wind_direction(direction_opt: Option<f64>) -> &'static str {
     }
 }
 
+// Compute the average wind speed and direction
+fn average_wind(winds: &Vec<Wind>) -> Wind {
```
MaxVerevkin, line 386:
```suggestion
fn average_wind(winds: &[Wind]) -> Wind {
```

src/blocks/weather.rs
```diff
@@ -331,6 +382,34 @@ fn convert_wind_direction(direction_opt: Option<f64>) -> &'static str {
...
+        Wind {
+            speed: 0.0,
+            degrees: None,
+        }
+    } else {
+        opposite /= count;
+        adjacent /= count;
+        Wind {
+            speed: opposite.hypot(adjacent),
+            degrees: Some(opposite.atan2(adjacent).to_degrees()),
```
MaxVerevkin, line 408: The range of atan2 is (-pi, pi], but convert_wind_direction expects the range [0, tau].

bim9262: [atan](https://doc.rust-lang.org/std/primitive.f64.html#method.atan): `Computes the arctangent of a number. Return value is in radians in the range [-pi/2, pi/2]`
[atan2](https://doc.rust-lang.org/std/primitive.f64.html#method.atan2):
```
Computes the four quadrant arctangent of self (y) and other (x) in radians.

x = 0, y = 0: 0
x >= 0: arctan(y/x) -> [-pi/2, pi/2]
y >= 0: arctan(y/x) + pi -> (pi/2, pi]
y < 0: arctan(y/x) - pi -> (-pi, -pi/2)
```

bim9262: So we can fix that by adding a `.rem_euclid(360.0)` on there?

src/blocks/weather.rs
```diff
@@ -331,6 +392,34 @@ fn convert_wind_direction(direction_opt: Option<f64>) -> &'static str {
...
+    if count == 0.0 {
+        Wind {
+            speed: 0.0,
+            degrees: None,
+        }
+    } else {
+        north /= count;
+        east /= count;
+        Wind {
+            speed: north.hypot(east),
```
bim9262, line 417: Oh we can remove a division, by just dividing by the hypotenuse by the count insead of the north and south components.

```math
\displaylines{
\sqrt{(xa)^2+(xb)^2)} = xc^2\\
\sqrt{x^2(a^2+b^2)} = xc^2\\
x\sqrt{a^2+b^2} = xc^2\\
\sqrt{a^2+b^2} = c^2
}
```

src/blocks/weather.rs
MaxVerevkin: The function seems correct, but you swapped "north" and "east". When angle is 0, the direction is north, but in this function it is east.

src/blocks/weather.rs
MaxVerevkin: I think it is a good idea to avoid unnecessary API calls. However we can automatically enable the forecast if the format template contains any templates that rely on it.

src/blocks/weather/met_no.rs
```diff
@@ -2,13 +2,15 @@ use super::*;
...
-#[serde(tag = "name", rename_all = "lowercase")]
+#[derive(Deserialize, Debug, SmartDefault)]
+#[serde(tag = "name", rename_all = "lowercase", deny_unknown_fields, default)]
 pub struct Config {
     coordinates: Option<(String, String)>,
     altitude: Option<String>,
     #[serde(default)]
     lang: ApiLanguage,
+    #[default(12)]
+    forcast_hours: usize,
```
MaxVerevkin, line 13: Docs state that the default is `0`.

bim9262:
I'll be AFK for the next couple days, but I'll look again on Monday

src/blocks/weather/met_no.rs
bim9262: The docs say 12 hours default for met.no and 0 hours default for owm (because of the extra api call). I can set the default to 12 though (and update the doc). I'll add in a check to see if the forecast needs to be fetched based on the format.

src/blocks/weather.rs
bim9262: I think the function looks right... I'll add some tests, but it seems like if you average the wind at 0 degrees the result is also at 0 degrees, not 90 degrees (East)

src/blocks/weather/met_no.rs
MaxVerevkin: > The docs say 12 hours default for met.no and 0 hours default for owm 

Sorry, didn't notice the difference.

src/blocks/weather.rs
MaxVerevkin: The function result _is_ correct, I was referring to local variables "north" and "east". They are supposed to represent the components of the wind, so when the direction is 0 degrees and speed is 1, `north` should be `1.0` and `east` should be `0.0`. Right now it's the opposite.

src/blocks/weather.rs
```diff
@@ -224,11 +293,28 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
...
-            _ = api.wait_for_update_request() => ()
+            _ = api.wait_for_update_request() => (),
+            Some(action) = actions.recv() => match action.as_ref() {
+                    "toggle_format" => {
+                        if let Some(ref mut format_alt) = format_alt {
+                            std::mem::swap(format_alt, &mut format);
+                        }
+                    }
+                    _ => (),
+                }
```
MaxVerevkin, line 304: I'm not sure if fetching the weather on each `toggle_foramt` is a good idea. It consumes API calls and most importantly makes the block less responsive (there is a noticeable delay before the block changes its format).

src/blocks/weather.rs
bim9262: 🤦‍♂️

src/blocks/weather.rs
```diff
@@ -215,20 +284,43 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
...
+fn need_forecast(format: &Format, format_alt: &Option<Format>) -> bool {
+    fn has_forecast_key(format: &Format) -> bool {
+        format.contains_key("temp_forecast")
+            || format.contains_key("apparent_forecast")
+            || format.contains_key("humidity_forecast")
+            || format.contains_key("wind_forecast")
+            || format.contains_key("wind_kmh_forecast")
+            || format.contains_key("direction_forecast")
+    }
+    has_forecast_key(format) || format_alt.as_ref().map_or(false, has_forecast_key)
```
MaxVerevkin, line 321: By the way, we now have `Option::is_some_and(f)` which is equivalent to `Option::map_or(false, f)` which may be more readable.

src/blocks/weather.rs
```diff
@@ -215,20 +284,43 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
             None
         };
 
-        let fetch = || provider.get_weather(location);
+        let fetch = || provider.get_weather(location, need_forecast(&format, &format_alt));
```
MaxVerevkin, line 287: Not that it is particularly expensive, but `need_forecast(&format, &format_alt)` can be computed once outside the loop.

src/blocks/weather.rs
MaxVerevkin: Also `icon_forecast`. I'm not familiar with openweathermap's API (and I couldn't even found the docs for `https://api.openweathermap.org/data/2.5/forecast` on their site), but at least, from my understanding, it is easy to implement it in metno? I'm not against having some placeholders available for some providers but not other.

bim9262: The api doc is here: https://openweathermap.org/forecast5

both owm and metno have descriptions of the weather for each interval, but not a summary. That's why I had omitted the `weather_forecast`, `weather_verbose_forecast`, and `icon_forecast`.

Do you think taking the most common descriptions over the sampled intervals is good enough or should it be left out as it is perhaps misleading? For example another strategy would be to show the worst/most severe weather but that is a bit nebulous :drum:

MaxVerevkin: Ha! I just realized that my understanding of this feature is completely wrong. I thought that `forecast_hours = 3` will show me the weather _after_ three hours. What it actually does is it shows the "average" weather throughout the next three hours :facepalm:

bim9262: Do you have a better key name or description we can use to clarify?

MaxVerevkin: Maybe we can have something like `x_forecast_{avg,min,max,final}`? This allows for `{icon,weather,weather_verbose}_forecast_final`. That's a lot of placeholders, but we don't have to list all of them in the docs, just explain how they are formed with a few examples.

bim9262: What's the average of tornado and clear skies? Strong winds?

MaxVerevkin: IDK. I don't think providing min/max/avg for weather/icon is a good idea.

bim9262: I agree

MaxVerevkin: Any thoughts no this? I can imagine

```toml
format_alt = " $icon_ff in 12 hours: $temp_fa($temp_fmin-$temp_fmax) "
```

being useful. Here `_ff` stands for "forecast final", `_fa` for "forecast average" and `_f{min,max}` for "forecast {min,max}". That's just one way to name them.

bim9262: I think this is a good idea. 
`icon`, `weather`, and `weather_verbose` can only be suffixed with `_ffin`.
`temp`, `apparent`, `humidity`, `wind`, `wind_kmh`, `direction` can be suffixed with `_f{fin,avg,min,max}`

src/blocks/weather.rs
```diff
@@ -48,23 +52,46 @@
...
+//! `wind_kmh{,_{favg,fmin,fmax,ffin}}`          | Wind speed. The wind speed in km/h                                            | Number | -
+//! `direction{,_{favg,fmin,fmax,ffin}}`         | Wind direction, e.g. "NE"                                                     | Text   | -
+//! `temp_forecast{,_{favg,fmin,fmax,ffin}}`      | Temperature forecast (omitted if no forecast available)                        | Number | degrees
+//! `apparent_forecast{,_{favg,fmin,fmax,ffin}}`  | Australian Apparent Temperature forecast (omitted if no forecast available)    | Number | degrees
+//! `humidity_forecast{,_{favg,fmin,fmax,ffin}}`  | Humidity forecast (omitted if no forecast available)                           | Number | %
+//! `wind_forecast{,_{favg,fmin,fmax,ffin}}`      | Wind speed forecast (omitted if no forecast available)                         | Number | -
+//! `wind_kmh_forecast{,_{favg,fmin,fmax,ffin}}`  | Wind speed forecast. The wind speed in km/h (omitted if no forecast available) | Number | -
+//! `direction_forecast{,_{favg,fmin,fmax,ffin}}` | Wind direction forecast, e.g. "NE" (omitted if no forecast available)          | Text   | -
```
MaxVerevkin, line 78:
```suggestion
```

MaxVerevkin (approved):
Looks good.

Thanks for working on this!

bim9262:
Thanks @MaxVerevkin!

PR #1946, opened 2023-09-20
Simplify some error handling and make it more structured

MaxVerevkin:
- Remove `kind` and `block` fields from `Error`.
- Removed `block` field is replaced by `BlockError` struct.
- Represent formatting errors (such as "placeholder not found") as an enum with thiserror macro.
- Merge `ResultExt` and `OptionExt` into `ErrorContext`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1089,6 +1089,7 @@ dependencies = [
  "signal-hook-tokio",
  "smart-default",
  "swayipc-async",
+ "thiserror",
  "tokio",
  "toml 0.8.0",
  "unicode-segmentation",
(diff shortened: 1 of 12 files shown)


MaxVerevkin:
Notable things that where broken and still are:
- Invalid block-specific configuration does not bring down the entire bar. But errors in the "common" config options (such as `merge_with_next = 1` do). This is inconsistent. It might be okay to fail completely in both cases.
- Invalid click-handler button (such as `button = "not a valid button"`) does not result in any sort of error.

bim9262:
I think these changes look good. Let's create new issues for the things that you mentioned that are still broken (assuming there aren't already issues open)

bim9262 (approved):

PR #1948, opened 2023-10-01
Use lat/lon for Open Weather Map

### LINKED ISSUE #1485: Always use "lat, lon" for OpenWeatherMap requests

bim9262:
Fixes #1485

### DIFF
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -16,7 +16,7 @@
 //! `service` | The configuration of a weather service (see below). | **Required**
 //! `format` | A string to customise the output of this block. See below for available placeholders. Text may need to be escaped, refer to [Escaping Text](#escaping-text). | `" $icon $weather $temp "`
 //! `interval` | Update interval, in seconds. | `600`
-//! `autolocate` | Gets your location using the ipapi.co IP location service (no API key required). If the API call fails then the block will fallback to `city_id` or `place`. | `false`
+//! `autolocate` | Gets your location using the ipapi.co IP location service (no API key required). If the API call fails then the block will fallback to service specific location config. | `false`
 //! `autolocate_interval` | Update interval for `autolocate` in seconds or "once" | `interval`
 //!
 //! # OpenWeatherMap Options
@@ -27,17 +27,18 @@
 //! ----|--------|----------|--------
 //! `name` | `openweathermap`. | Yes | None
 //! `api_key` | Your OpenWeatherMap API key. | Yes | None
-//! `city_id` | OpenWeatherMap's ID for the city. | Yes* | None
-//! `place` | OpenWeatherMap 'By city name' search query. See [here](https://openweathermap.org/current) | Yes* | None
 //! `coordinates` | GPS latitude longitude coordinates as a tuple, example: `["39.2362","9.3317"]` | Yes* | None
+//! `city_id` | OpenWeatherMap's ID for the city. (Deprecated) | Yes* | None
+//! `place` | OpenWeatherMap 'By {city name},{state code},{country code}' search query. See [here](https://openweathermap.org/api/geocoding-api#direct_name). Consumes an additional API call | Yes* | None
+//! `zip` | OpenWeatherMap 'By {zip code},{country code}' search query. See [here](https://openweathermap.org/api/geocoding-api#direct_zip). Consumes an additional API call | Yes* | None
 //! `units` | Either `"metric"` or `"imperial"`. | No | `"metric"`
 //! `lang` | Language code. See [here](https://openweathermap.org/current#multi). Currently only affects `weather_verbose` key. | No | `"en"`
 //!
-//! One of `city_id`, `place` or `coordinates` is required. If more than one are supplied, `city_id` takes precedence over `place` which takes place over `coordinates`.
+//! One of `coordinates`, `city_id`, `place`, or `zip` is required. If more than one are supplied, `coordinates` takes precedence over `city_id` which takes precedence over `place` which takes precedence over `zip`.
 //!
-//! The options `api_key`, `city_id`, `place` can be omitted from configuration,
+//! The options `api_key`, `city_id`, `place`, `zip`, can be omitted from configuration,
 //! in which case they must be provided in the environment variables
-//! `OPENWEATHERMAP_API_KEY`, `OPENWEATHERMAP_CITY_ID`, `OPENWEATHERMAP_PLACE`.
+//! `OPENWEATHERMAP_API_KEY`, `OPENWEATHERMAP_CITY_ID`, `OPENWEATHERMAP_PLACE`, `OPENWEATHERMAP_ZIP`.
 //!
 //! # met.no Options
 //!
@@ -201,8 +202,10 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
     let format = config.format.with_default(" $icon $weather $temp ")?;
 
     let provider: Box<dyn WeatherProvider + Send + Sync> = match &config.service {
-        WeatherService::MetNo(config) => Box::new(met_no::Service::new(config)?),
-        WeatherService::OpenWeatherMap(config) => Box::new(open_weather_map::Service::new(config)),
+        WeatherService::MetNo(service_config) => Box::new(met_no::Service::new(service_config)?),
+        WeatherService::OpenWeatherMap(service_config) => {
+            Box::new(open_weather_map::Service::new(config.autolocate, service_config).await?)
+        }
     };
 
     let autolocate_interval = config.autolocate_interval.unwrap_or(config.interval);
--- a/src/blocks/weather/open_weather_map.rs
+++ b/src/blocks/weather/open_weather_map.rs
@@ -2,9 +2,11 @@ use super::*;
 use chrono::Utc;
 
 pub(super) const URL: &str = "https://api.openweathermap.org/data/2.5/weather";
+pub(super) const GEO_URL: &str = "https://api.openweathermap.org/geo/1.0";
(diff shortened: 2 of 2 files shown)


MaxVerevkin:
Do you think it makes sense to remove useless spaces/newlines in `open_weather_map_cities.json` like `met_no_legends.json`?

bim9262:
If there's ever an update to the files it would be very hard to read the diff (as a sanity check) if it's all on a single line. Since the line breaks are not a factor binary size when using bincode I'd leave the line breaks there. Also where did you find the met_no_legends? I can add it to the update script.

bim9262:
I'm considering restructuring the json file to keep the keep the id as a number so that we can pack it smaller by not having strings as keys (need to do some testing though)

MaxVerevkin:
> If there's ever an update to the files it would be very hard to read the diff (as a sanity check) if it's all on a single line

Good point.

> Also where did you find the met_no_legends?

Here: https://web.archive.org/web/20230328155110/https://api.met.no/weatherapi/weathericon/2.0/legends.

MaxVerevkin:
> to keep the keep the id as a number

Yeah, I just wanted to ask why you converted them to strings, as it doesn't work is some cases. For example the last entry of the file is

```json
  "7.12969E+7": {
    "lon": 33.92,
    "lat": 44.59
  }
```

MaxVerevkin:
Is it necessary to prefetch this info? In case with met.no we had to because they disabled the API. The 1M binary increase is unfortunate.

bim9262:
There's two errors like that, but they come from the source file. I opened a support case with OWM...

Based on the errors in the other city list they publish they accidentally made them floats (those same IDs ended with .0 in that format)

bim9262:
> Is it necessary to prefetch this info? In case with met.no we had to because they disabled the API. The 1M binary increase is unfortunate.

We could fetch it on demand. We'd probably want to give the lowest precedence to city_id since it would become the most expensive to load.

src/blocks/weather/open_weather_map.rs
```diff
@@ -79,24 +95,184 @@ struct ApiWeather {
...
+                        .map(|sys_time| {
+                            DateTime::<Utc>::from(sys_time)
+                                .format("%a, %d %b %Y %H:%M:%S GMT")
+                                .to_string()
+                        });
+
+                    let cities_request = if let Ok(timestamp) = cities_temp_timestamp.as_ref() {
+                        debug!("{}", timestamp);
+                        REQWEST_CLIENT.get(CITIES_URL).header(
+                            reqwest::header::IF_MODIFIED_SINCE,
```
bim9262, line 200: This isn't being respected by owm's server. Curl is smart enough to handle it, but reqwest/hyper doesn't yet.

bim9262:
This is now with live fetching. It caches the result of the file to disk and compares the last modified date on the server with the local modification date. (owm does not respect the `If-Modified-Since` header).

Switched over to the larger city list (as described at https://openweathermap.org/bulk#list):
List of major cities across the globe - 22,635 cities. You can download the full list of cities [here](https://bulk.openweathermap.org/sample/current.city.list.json.gz)
Extended cities list across the globe - 209,579 cities. You can download the full list of cities [here](https://bulk.openweathermap.org/sample/city.list.json.gz)

src/blocks/weather/open_weather_map.rs
```diff
@@ -79,24 +94,198 @@ struct ApiWeather {
...
+    id: f32,
+    coord: CityCoord,
+}
+#[derive(Deserialize, Debug)]
+struct CityCoord {
+    lat: f64,
+    lon: f64,
+}
+
+static LOCATION_CACHE: Mutex<Option<String>> = Mutex::new(None);
```
MaxVerevkin, line 109: Does this prevent from having two blocks which show weather in different places?

src/blocks/weather/open_weather_map.rs
```diff
@@ -79,24 +94,198 @@ struct ApiWeather {
...
+            None => match self.config.city_id.as_ref() {
+                Some(city_id) => {
+                    let city_id = &city_id.parse().error("city_id must be numeric")?;
+
+                    let mut cities_temp_path = env::temp_dir();
+                    let user = env::var("USER");
+                    cities_temp_path.push(format!(
+                        "city.list.min.json.gz-{}",
+                        user.as_deref().unwrap_or("no-user")
+                    ));
```
MaxVerevkin, line 208: Perhaps it's better to use `dirs::cache_dir()` instead of temp?

MaxVerevkin: By the way, maybe extract this cache and fetching logic into a different function?

src/blocks/weather/open_weather_map.rs
bim9262: Yes... Good catch.

src/blocks/weather.rs
```diff
@@ -105,7 +107,7 @@ pub mod open_weather_map;
 
 const IP_API_URL: &str = "https://ipapi.co/json";
 
-static LAST_AUTOLOCATE: Mutex<Option<AutolocateResult>> = Mutex::new(None);
+static LAST_AUTOLOCATE: Lazy<Mutex<Option<AutolocateResult>>> = Lazy::new(|| Mutex::new(None));
```
MaxVerevkin, line 110: Why this change? This mutex is not locked across `.await`s. In cases like this [tokio advises to use ordinary mutexes](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use).

src/blocks/weather/open_weather_map.rs
```diff
@@ -79,24 +94,233 @@ struct ApiWeather {
...
+#[derive(Eq, Hash, PartialEq, Debug)]
+enum LocationKeys {
+    CityId(String),
+    Place(String),
+    Zip(String),
+}
 
-        let location_query = autolocated
+static LOCATION_CACHE: Lazy<Mutex<HashMap<LocationKeys, String>>> =
+    Lazy::new(|| Mutex::new(HashMap::new()));
```
MaxVerevkin, line 117: Why can't we just store the cached location string in the `Service` struct?

MaxVerevkin:
> we don't update the list or correct issues due to it being mainly open source

I can't make sense of this sentence...

> So perhaps we should continue to send the city_id, but query the geo api for place and zip.

If they won't actually remove support for this, and it makes our code simpler, then why not.

bim9262:
> > we don't update the list or correct issues due to it being mainly open source
> 
> I can't make sense of this sentence...
> 
> > So perhaps we should continue to send the city_id, but query the geo api for place and zip.

Keep the code I wrote for the place name and zip codes in this pr, but use the city_id like we were doing originally

src/blocks/weather.rs
bim9262: I did that because of the use super::* in own, but can use the two mutex types with an alias or qualify wich mutex to use

src/blocks/weather/open_weather_map.rs
bim9262: I tried that initially, but could not figure out how to fix the error caused when you make the service mutable. The line above the retry becomes a FnMut, and complains: `Captured variable cannot escape FnMut closure body`

MaxVerevkin:
> Keep the code I wrote for the place name and zip codes in this pr, but use the city_id like we were doing originally

I was referring to their rationale for not fixing city IDs. "open source" == "unmaintained"? But yes, let's use city_id as we were doing it.

src/blocks/weather/open_weather_map.rs
MaxVerevkin: Oh, yeah, that's an annoying error. Can we construct the location string in `Service::new()` then? It can take `autolocate: bool` and do the necessary stuff if it is `false`.

src/blocks/weather.rs
MaxVerevkin: If it is possible to get rid of `LOCATION_CACHE` then this will resolve itself, I guess.

MaxVerevkin (approved):
Looks good

bim9262:
> > Keep the code I wrote for the place name and zip codes in this pr, but use the city_id like we were doing originally
> 
> I was referring to their rationale for not fixing city IDs. "open source" == "unmaintained"? But yes, let's use city_id as we were doing it.

I was half asleep when I read your message. Yeah I don't understand  "open source" == "unmaintained" either.

This was the response about where the data comes form:

>We get our city database from https://www.geonames.org/. However, you can't find the weather IDs there, IDs were generated by us. Since the list has been deprecated and was not updated for a while some city IDs could be missed or point to other coordinates.
>
>We strongly recommend using geographical coordinates instead of city IDs and using Geocoder API: https://openweathermap.org/api/geocoding-api

PR #1949, opened 2023-10-05
xrandr: support multiple outputs

### LINKED ISSUE #1947: xrandr $resolution returns "primary" for multiple monitors
I think these lines should handle multiple monitors

https://github.com/greshake/i3status-rust/blob/40bc766708a32cc3c2072fdb6637ddf7e900c49e/src/blocks/xrandr.rs#L182-L189

since the output could be something like `eDP1 connected 1920x1080+0+0` and `HDMI2 connected primary 2560x1080+1920+0`

MaxVerevkin:
Fixes #1947


bim9262 (approved):
lgtm

PR #1950, opened 2023-10-06
Use autolocate city for met_no location

bim9262:


src/blocks/weather.rs
```diff
@@ -153,7 +153,7 @@ fn default_interval() -> Seconds {
 trait WeatherProvider {
     async fn get_weather(
         &self,
-        autolocated_location: Option<Coordinates>,
+        autolocated_location: &Option<Coordinates>,
```
MaxVerevkin, line 156: Usually it is better to use `Option<&T>` instead of `&Option<T>`. Doesn't matter as much in this case though.

MaxVerevkin (approved):

src/blocks/weather/met_no.rs
```diff
@@ -284,7 +284,9 @@ impl WeatherProvider for Service<'_> {
         };
 
         Ok(WeatherResult {
-            location: "Unknown".to_string(),
+            location: location
+                .as_ref()
```
MaxVerevkin, line 288: I'm surprised clippy doesn't catch this, but this `.as_ref()` is not needed.

MaxVerevkin: I suppose most `Option<&T>::as_ref()` calls are useless, except for rare cases when `Option<&&T>` is actually needed. I may look into adding this to clippy in the future.

PR #1951, opened 2023-10-06
Fail if click handler refers to unknown button

### LINKED ISSUE #1918: Invalid `button` in a click handler does not result in an error
```toml
[[block.click]]
button = "nonexisting"
```

does not error, it it probably should.

MaxVerevkin:
Fixes #1918

### DIFF
--- a/src/protocol/i3bar_event.rs
+++ b/src/protocol/i3bar_event.rs
@@ -40,7 +40,14 @@ fn unprocessed_events_stream(invert_scrolling: bool) -> BoxedStream<I3BarEvent> 
                 button: MouseButton,
             }
 
-            let event: I3BarEventRaw = serde_json::from_str(line).unwrap();
+            let event: I3BarEventRaw = match serde_json::from_str(line) {
+                Ok(event) => event,
+                Err(err) => {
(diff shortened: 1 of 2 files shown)


MaxVerevkin:
Technically a breaking change but realistically it should be okay.

src/protocol/i3bar_event.rs
```diff
@@ -37,7 +37,7 @@ fn unprocessed_events_stream(invert_scrolling: bool) -> BoxedStream<I3BarEvent>
             #[derive(Deserialize)]
             struct I3BarEventRaw {
                 instance: Option<String>,
-                button: MouseButton,
+                button: i64,
```
bim9262, line 40: What's the benefit of using an i64 instead of keeping it as a `MouseButton`?

MaxVerevkin: I wanted to handle the case in which bar sends a click event with button we don't support / know about. Now I see that is it better to leave this as a `MouseButton` and gracefully handle deserialization errors instead of unwrapping.

bim9262 (approved):
Looks good :)

PR #1968, opened 2023-11-16
Add basic support for non Gregorian calendars

### LINKED ISSUE #1966: Alternative calendar support in time block
There are many calendars other than the Gregorian calendar that are used in the world. For example, the ecmascript and web platform support [19 of them](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getCalendars#supported_calendar_types). Would you accept a PR that add support for these calendars in the time block (or a new block) using the [icu4x](https://github.com/unicode-org/icu4x) crate?

HKalbasi:
fix #1966

### DIFF
--- a/src/blocks/time.rs
+++ b/src/blocks/time.rs
@@ -29,6 +29,21 @@
 //! short = " $icon $timestamp.datetime(f:%R) "
 //! ```
 //!
+//! # Non Gregorian calendars
+//!
+//! You can use calendars other than the Gregorian calendar by adding the calendar specifier in the locale string. When using
+//! this feature you can't use chrono style format string, and you should use one of the options provided by
+//! the `icu4x` crate: `short`, `medium`, `long`, `full`.
+//!
+//! ## Example
+//!
+//! ```toml
+//! [[block]]
+//! block = "time"
+//! interval = 60
+//! format = "$timestamp.datetime(locale:'fa_IR-u-ca-persian', f:'full')"
+//! ```
+//!
 //! # Icons Used
 //! - `time`
 
--- a/src/blocks/time.rs
+++ b/src/blocks/time.rs
@@ -35,6 +35,8 @@
 //! this feature you can't use chrono style format string, and you should use one of the options provided by
 //! the `icu4x` crate: `short`, `medium`, `long`, `full`.
 //!
+//! ** Only available using feature `icu_calendar`. **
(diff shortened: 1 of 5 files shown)


HKalbasi:
CI failure is due `locid` which is a module in the `icu` crate. What should I do about it?

MaxVerevkin:
> CI failure is due `locid` which is a module in the `icu` crate. What should I do about it?

You can edit `cspell.yaml` to add this new word.

HKalbasi:
I replaced the `icu` crate with `icu_calendar` and `icu_datetime`. Now 18 dependencies are added, 14 of them are icu internal crates.

src/blocks/time.rs
```diff
@@ -31,6 +31,22 @@
 //!
 //! # Icons Used
```
bim9262, line 32: I'd keep this after both examples

src/formatting/formatter.rs
```diff
@@ -535,25 +561,40 @@ impl DatetimeFormatter {
...
+                    },
+                }
+                .to_string(),
+                DatetimeFormatter::Icu { locale, length } => {
+                    let date = icu_calendar::Date::try_new_iso_date(
+                        datetime.year_ce().1 as i32,
+                        datetime.month() as u8,
+                        datetime.day() as u8,
+                    )
+                    .expect("Current date should be a valid date");
```
bim9262, line 588: I think we should use `.error` instead of `.expect`.

HKalbasi:
I found out that I didn't handle the timezone (by using the patch on my machine in a full day), and fixed that in the latest commit.

MaxVerevkin:
Binary size change: stripped 17M -> 31M, unstripped 24M -> 39M. That's too much... Would you be okay to put this feature behind a feature flag?

HKalbasi:
I made it optional with a feature flag and I'm happy with that. Instead of / in addition to that, we can split the [locale data from the binary](https://github.com/unicode-org/icu4x/blob/main/docs/tutorials/data_management.md), though it will complicate things.

MaxVerevkin (approved):

bim9262 (approved):
@HKalbasi I have used the approach you mentioned to split the locale data, and while it's doable it's a lot of work to slim down the binary size.

PR #1974, opened 2023-12-16
blocks/toggle: Allow customizing State when on/off

### LINKED ISSUE #1923: Toggle critical_bg not working
From my understanding based on here: https://github.com/greshake/i3status-rust/blob/v0.31.8/src/blocks/toggle.rs#L129

If the status is "on" then `idle_bg` is used, and if the status is "off" then `critical_bg` is used.

The determintor from `command_state` is whether the resulting command has text or if it's an empty string (https://github.com/greshake/i3status-rust/blob/v0.31.8/src/blocks/toggle.rs#L20).

However, the tokio docs, if I'm reading it correctly, says that it's based on the exit code.

I'm not familiar w/ tokio and only a beginner at Rust, so maybe I'm missing something here, but the `critical_bg` override is never applied when `command_state` returns an empty string or has exit code of 1.

Here's my current block:

```toml
[[block]]
block = "toggle"
format = " $icon "
command_state = "nmcli connection show | grep wg0 | awk '{print $4}' | grep -v 'wg0'"
command_on = "nmcli connection down wg0"
command_off = "nmcli connection up wg0"
[block.theme_overrides]
idle_bg="#005500"
critical_bg="#000000"
```

What I've also done is move the state command to a shell script and either do `exit 0` if the awk value == wg0 or `exit 1` if it's not.

image

So what I would expect to see is when the toggle is in this state, the bg is a red color.  But instead it stays in idle bg regardless.

The toggle works, so the commands are fine and I can see it reflect in the NetworkManager applet and in the output of nmcli.  So it's more so something is not working when trying to set the color.

softmoth:
Fixes #1923.

### DIFF
--- a/src/blocks/toggle.rs
+++ b/src/blocks/toggle.rs
@@ -2,7 +2,7 @@
 //!
 //! You can add commands to be executed to disable the toggle (`command_off`), and to enable it
 //! (`command_on`). If these command exit with a non-zero status, the block will not be toggled and
-//! the block state will be changed to give a visual warning of the failure. You also need to
+//! the block state will be changed to `critical` to give a visual warning of the failure. You also need to
 //! specify a command to determine the state of the toggle (`command_state`). When the command outputs
 //! nothing, the toggle is disabled, otherwise enabled. By specifying the interval property you can
 //! let the command_state be executed continuously.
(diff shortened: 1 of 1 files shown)


softmoth:
Is it preferable to use `State::Info` instead of `::Good`? That allows for end-user customization but makes less of a statement by default.

MaxVerevkin:
Maybe thin can be a config option? Like `state_on = "good"` `state_off = "idle"` (`State` implements deserialize) (naming can be different). Also, should we try to preserve current behavior by default?

softmoth:
Thanks for the hint on deserialize, I'll do that, and try keeping state_on = "idle" by default.

MaxVerevkin (approved):

PR #1976, opened 2023-12-17
blocks/focused_window: Close code block in markdown doc

softmoth:
This problem manifests in the man page when looking at the *next* block's documentation. Currently that is the `github` block, which is almost entirely subsumed into this `focus_window` Example toml.


MaxVerevkin:
Thanks! I wish `rustdoc` would catch this...

PR #1981, opened 2024-01-07
Add Wireguard support to vpn block

ppx17:
Adds a new driver for the VPN block to support Wireguard connections.

I couldn't find any way to control Wireguard without using root permissions, so I hope this route with sudo is OK. The docs include a sudoers configuration that should be safe if your Wireguard configuration files are locked down properly, but please correct me if I'm mistaken here.

The driver supports the same toggle action on click as the other drivers, but I've reduced the country code and flag to empty strings, as they don't make a lot of sense in this context.

We can update the Connected enum to support a state without country information included, but I tried to change as few things as possible on the existing drivers.

### DIFF
--- a/src/blocks/vpn/wireguard.rs
+++ b/src/blocks/vpn/wireguard.rs
@@ -1,0 +1,95 @@
+use std::process::Stdio;
+
+use async_trait::async_trait;
+use nix::unistd::getuid;
+use tokio::process::Command;
+
+use crate::blocks::prelude::*;
+
+use super::{Driver, Status};
+
+pub struct WireguardDriver {
+    interface: String,
+}
+
+impl WireguardDriver {
+    pub async fn new(interface: String) -> WireguardDriver {
+        WireguardDriver { interface }
+    }
+}
+
+const SUDO_CMD: &'static str = "/usr/bin/sudo";
+const WG_QUICK_CMD: &'static str = "/usr/bin/wg-quick";
+const WG_CMD: &'static str = "/usr/bin/wg";
+
+#[async_trait]
+impl Driver for WireguardDriver {
+    async fn get_status(&self) -> Result<Status> {
+        let status = run_wg(vec!["show", self.interface.as_str()]).await;
+
+        match status {
+            Ok(status) => {
+                if status.contains(format!("interface: {}", self.interface).as_str()) {
+                    Ok(Status::Connected {
+                        country: "".to_owned(),
+                        country_flag: "".to_owned(),
+                    })
+                } else {
+                    Ok(Status::Disconnected)
+                }
+            }
+            Err(_) => Ok(Status::Error),
+        }
+    }
+
+    async fn toggle_connection(&self, status: &Status) -> Result<()> {
+        match status {
+            Status::Connected { .. } => {
+                run_wg_quick(vec!["down", self.interface.as_str()]).await?;
+            }
+            Status::Disconnected => {
+                run_wg_quick(vec!["up", self.interface.as_str()]).await?;
+            }
+            Status::Error => (),
+        }
+        Ok(())
+    }
+}
+
+async fn run_wg(args: Vec<&str>) -> Result<String> {
+    let stdout = make_command(should_use_sudo(), WG_CMD)
+        .args(&args)
+        .output()
+        .await
+        .error(format!("Problem running wg command: {args:?}"))?
+        .stdout;
+    let stdout =
+        String::from_utf8(stdout).error(format!("wg produced non-UTF8 output: {args:?}"))?;
+    Ok(stdout)
+}
+
+async fn run_wg_quick(args: Vec<&str>) -> Result<()> {
+    make_command(should_use_sudo(), WG_QUICK_CMD)
+        .args(&args)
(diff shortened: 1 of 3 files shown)


MaxVerevkin:
> The driver supports the same toggle action on click as the other drivers, but I've reduced the country code and flag to empty strings, as they don't make a lot of sense in this context.
> 
> We can update the Connected enum to support a state without country information included, but I tried to change as few things as possible on the existing drivers.

I would prefer to make them optional, and on include them at all when using this driver.

src/blocks/vpn/wireguard.rs
```diff
@@ -0,0 +1,95 @@
...
+        .output()
+        .await
+        .error(format!("Problem running wg command: {args:?}"))?
+        .stdout;
+    let stdout =
+        String::from_utf8(stdout).error(format!("wg produced non-UTF8 output: {args:?}"))?;
+    Ok(stdout)
+}
+
+async fn run_wg_quick(args: Vec<&str>) -> Result<()> {
```
MaxVerevkin, line 71:
```suggestion
async fn run_wg_quick(args: &[&str]) -> Result<()> {
```

src/blocks/vpn/wireguard.rs
```diff
@@ -0,0 +1,95 @@
...
+            Status::Disconnected => {
+                run_wg_quick(vec!["up", self.interface.as_str()]).await?;
+            }
+            Status::Error => (),
+        }
+        Ok(())
+    }
+}
+
+async fn run_wg(args: Vec<&str>) -> Result<String> {
```
MaxVerevkin, line 59:
```suggestion
async fn run_wg(args: &[&str]) -> Result<String> {
```

ppx17:
> > The driver supports the same toggle action on click as the other drivers, but I've reduced the country code and flag to empty strings, as they don't make a lot of sense in this context.
> > We can update the Connected enum to support a state without country information included, but I tried to change as few things as possible on the existing drivers.
> 
> I would prefer to make them optional, and on include them at all when using this driver.

Alright, I've implemented a separate 'ConnectedToCountry' state that includes the country and flag. It felt a little cleaner than still having them as Optional in drivers that don't care. Then again, there are now two separate states that both represent connected, not sure what the best approach would be? Happy to switch to a single connected state with optional attributes as well.

MaxVerevkin:
I think having optional fields is slightly better. First of all, maybe in the future we will add more fields and having one "connected" state will make things easier. Next, it would eliminate duplicated `match` arms when we don't care about the country info. Finally, optionals work nicely with `map!` macro:

```rust
map! {
	"icon" => Value::icon(status.icon()),
	[if let Some(c) = country] "country" => Value::text(c.into()),
	[if let Some(f) = country_flag] "flag" => Value::text(f.into()),
}
```

ppx17:
Fair points, and neat trick with map! The PR is updated. :+1:

c3Ls1US:
To clarify, you're invoking the `sudo` command internally to bring up a WireGuard interface? That signals a red flag to me and this would be the only block that would do it. What other options have you explored? 

If I remember correctly, `NetworkManager` allows bringing up and tearing down network interfaces without sudo permissions so it's possible you can accomplish this with WireGuard interfaces too. I think that would be preferred.

ppx17:
Hmm, I don't really see a proper way to do this, even if there is a route through NetworkManager, then that is probably still not how most people have configured their current Wireguard connections, as it doesn't seem to support the default /etc/wireguard/wg0.conf settings.

Looking at the source of `wg-quick`, it [seems](https://git.zx2c4.com/wireguard-tools/tree/src/wg-quick/linux.bash#n85) to be a pretty hard requirement. Open to any other pointers to look at though!

c3Ls1US:
`wg-quick` creates a WireGuard tunnel and sets up a firewall. Those two actions at least are inherently privileged operations that requires root. As far as I can tell, there's no way around that for non-root users and I believe `NetworkManager` is able to do this because its daemon is running as root. Same with other WireGuard clients such as Cloudflare's `warp-cli`.

In my opinion, accomplishing this feature set would be relatively straightforward through `NetworkManager`, safer and more ideal as it would integrated at the system level via `libnm`. Then for every other network manager, just don't allow managing the WG interface.

Besides `sudo`, there's also the `polkit` framework. I don't have an opinion on it, but perhaps that's a possibility?

Kabouik:
Just chiming in to voice  the interest from a simple user's perspective for a Wireguard driver for the VPN block, although I have no solution to the issues raised above.

Kabouik:
In the mean time, for those who would be interested in at least showing the VPN status in `i3status-rust` without necessarily operating on `wg-quick` directly, the following custom block works:

```toml
[[block]]
block = "custom"
interval = 5  # Check every 5 seconds
cycle = ["curl -s https://ipinfo.io | jq -r '.country'", "curl -s https://ipinfo.io | jq -r '.city + \" (\" + .country +\")\"'"]
[[block.click]]
button = "right"
cmd = "footclient -H -o 'font=Iosevka Medium Extended:size=8' -a floating -W 65x30 curl -s https://ipinfo.io | jq '.'"
```

Issue #1973, opened 2023-12-13
Add a new privacy indicator


emersion:
Use PipeWire to figure out when the microphone, camera, or screen capture is in use.

Ref https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/2840

bim9262:
I've started work a privacy indicator. I checked my setup though and my camera doesn't appear to be used by pipewire (although I see it listed). Applications are using it directly (the webcam shows up in `v4l`).

Other drivers:

- [ ] `pipewire` (audio capture)
- [ ] `pipewire` (screen capture)
- [ ] `pipewire` (webcam)
- [ ] `pulseaudio` (audio capture)?

emersion:
Note, more applications should move to PipeWire for camera as the ecosystem moves to libcamera. But yeah, it's still the early days.

MaxVerevkin:
Is there something left not implemented in #1973?

bim9262:
This issue was about pipewire monitoring. If someone wants to add another driver another issue can be opened.

emersion:
Thank you for working on this!

bim9262:
You're welcome, it was a fun block to work on

kevenwyld:
I'm trying to use this (Because it's really cool!) however I noticed the docs say `pipewire Options (requires the pipewire feature to be enabled)`.  Unfortunately it doesn't mention which feature as far as I can tell and audio/screencapture don't seem to be working.  I read through the PR but I'm still not quite sure what I should be enabling or if these features should work like I'm expecting.  Is there additional documentation I should be looking at or does anyone here have any suggestions?

Thanks!

MaxVerevkin:
@kevenwyld you have to compile with `-F pipewire` flag.

kevenwyld:
> @kevenwyld you have to compile with `-F pipewire` flag.

That worked perfectly, thanks!

PR #1982, opened 2024-01-10
Add privacy monitor block

bim9262:
See #1973


bim9262:
@MaxVerevkin do you have any tips/recommendations for working with pipewire like setting up a global pipewire client that can be accessed by multiple threads?

I did attempt to make a global client like there is for the pipewire sound module, but ended up getting stuck.

MaxVerevkin:
> I did attempt to make a global client like there is for the pipewire sound module, but ended up getting stuck.

What problems did you encounter? I would imagine if we want to share pipewire connection between blocks, we would need a global Mutex (or RwLock) with the state (nodes, links),  global `Mutex<Vec<Sender<_>>` for event listeners and a global Lazy to spawn the thread only once. This seems similar to pulseoudio.

bim9262:
I was having a deadlock issue, but I got it sorted out now :)

Looks like libpipewire 1.0 is required which won't be released on ubuntu until April.

c3Ls1US:
While not directly related to this functionality, since `pipewire` will be added as a dependency, I think this would also be a good opportunity to support it as a sound driver as well

MaxVerevkin:
> Looks like a newer version of libpipewire is required than is available in the runner image. Might need to add a step to build it from source.

Would depending on an older version of `pipewire` crate fix this?

MaxVerevkin:
> While not directly related to this functionality, since `pipewire` will be added as a dependency, I think this would also be a good opportunity to support it as a sound driver as well

Does it have any advantages over pulseaudio driver, like any additional useful features not available today?

bim9262:
Perhaps you wouldn't need to run pipewire-pulse (I haven't tried this though)

src/blocks/privacy.rs
```diff
@@ -0,0 +1,201 @@
...
+//! Action          | Description                               | Default button
+//! ----------------|-------------------------------------------|---------------
+//! `toggle_format` | Toggles between `format` and `format_alt` | Left
+//!
+//! # Example
+//!
+//! ```toml
+//! [[block]]
+//! block = "privacy"
+//! merge_with_next = true
```
bim9262, line 55: Do you think that we should allow one or more driver to be used by the same block instead of just using two and merging them?

src/blocks/privacy.rs
```diff
@@ -0,0 +1,201 @@
...
+//! block = "privacy"
+//! [block.driver]
+//! name = "pipewire"
+//! exclude_input = ["openrgb"]
+//! ```
+//!
+//! # Icons Used
+//! - `microphone`
+//! - `volume`
+//! - `xrandr`
```
bim9262, line 69: Reusing `xrandr` icon is convenient for screen share, but perhaps we should make it be it's own dedicated icon?

Cargo.toml
```diff
@@ -17,6 +17,7 @@ members = [".", "xtask"]
 [features]
 default = ["pulseaudio"]
 pulseaudio = ["libpulse-binding"]
+pipewire = ["dep:pipewire"]
```
bim9262, line 20: I assume we don't want to make pipewire default since it's not as widespread as pulse,right?

ammgws:
> Perhaps you wouldn't need to run pipewire-pulse (I haven't tried this though)

Not sure if it's worth the maintenance burden since Pipewire implements the Pulseaudio api, so as you said all those users need to do is install pipewire-pulse if they want to use the sound block

src/blocks/privacy.rs
MaxVerevkin: Maybe we don't need a `driver` option and can guess the driver based on format placeholders used? For example `format = " $icon_video |"` would use `v4l` and `pipewire` (if enabled and available). Driver-specific options can be added in the main config if needed.

MaxVerevkin: Also, wouldn't merging two privacy blocks result in potentially duplicated icons?

src/blocks/privacy/v4l.rs
```diff
@@ -0,0 +1,155 @@
...
+            .await
+            .error("Unable to get next device in /proc")?
+        {
+            let proc_path = proc_path.path();
+            let fd_path = proc_path.join("fd");
+            if let Ok(mut fd_paths) = read_dir(fd_path).await {
+                while let Ok(Some(fd_path)) = fd_paths.next_entry().await {
+                    if let Ok(link_path) = fd_path.path().read_link() {
+                        if self.devices.contains_key(&link_path) {
+                            if let Ok(mut file) = File::open(proc_path.join("comm")).await {
```
MaxVerevkin, line 117: Let's use `let Ok(...) = ... else { continue };` to avoid this level of nesting :)

src/blocks/privacy.rs
bim9262: `merge_with_next = true` could result in multiple camera icons if you have two different applications that access a camera through a different mechanism (v4l + pipewire).
If we allow multiple drivers to be used in the same block we can merge the results from all of the drivers.

bim9262:
> > Looks like a newer version of libpipewire is required than is available in the runner image. Might need to add a step to build it from source.
> 
> Would depending on an older version of `pipewire` crate fix this?

I'm not sure, but I can run a test in another branch.

src/blocks/privacy.rs
```diff
@@ -0,0 +1,212 @@
...
+            });
+        }
+
+        widget.set_values(values);
+
+        api.set_widget(widget)?;
+
+        select! {
+            _ = api.wait_for_update_request() => (),
+            _ = select_all(drivers.iter_mut().map(|driver| driver.wait_for_change())) =>(),
```
MaxVerevkin, line 203: Do you think it is worth it to keep track which driver has an update and only poll this specific driver?

bim9262: I don't think that it really matters for these two drivers. `v4l`'s `get_info` is just reading from `/proc` and `pipewire`'s `get_info` is just querying the list of nodes/links that are already discovered. Since we don't have to do any expensive I/O, I think it's ok as-is.

MaxVerevkin: Alright, lets implement this when/if we have more drivers.

c3Ls1US:
> > Perhaps you wouldn't need to run pipewire-pulse (I haven't tried this though)
> 
> Not sure if it's worth the maintenance burden since Pipewire implements the Pulseaudio api, so as you said all those users need to do is install pipewire-pulse if they want to use the sound block

I'm not a PipeWire expert so I cannot provide an informed opinion, but it appears that there are downsides by going through `pipewire-pulse`. 

A few years ago, `mpv` provided direct support for the PipeWire audio output driver, stating [reduced complexity and richer integration as the primary reasons](https://github.com/mpv-player/mpv/pull/9587/commits/ebb0739d329dc65be13364db5aa82efae3981b37). I can't comment if it'll provide "richer integration" for this project, but certainly targeting PipeWire directly should result in less latency for users because the compatibility layer(s)/intermediate steps would be bypassed.

Aside from that, would it _really_ be a maintenance burden? Along with Wayland, PipeWire in Linux looks to become a standard component to not be a future burden so I believe supporting PipeWire natively will be useful.

bim9262:
There may be an advantage for mpv as a source of audio to use pipewire over pulse, but any latency issues between pipewire and pipewire-pulse should be moot for our use case. They only real advantage is that you wouldn't need one thread for pipewire and another for pulse within the status bar. 

I'm not sure that it's really much of an extra burden to support pipewire for audio info since it's 90% of the way there already.

bim9262:
After much investigation... I have determined that the pipewire sound block isn't quite as nice as using the pulseaudio version. The main reason is that there's no default sink/source (as far as I can tell). Also if you use a mix of tools and set the pipewire volume it doesn't update the pulseaudio volume, but updating the pulseaudio volume updates both. This leads to odd situations where pipewire thinks the volume is one thing and pulseaudio thinks it's another.

I suppose if you were completely pulseaudio free (which I kinda doubt at this point being all that feasible) then it wouldn't be an issue.

src/blocks/sound/pipewire.rs
```diff
@@ -0,0 +1,643 @@
...
+                        node.channels,
+                        proxies.borrow_mut().proxies_t.get(&node.proxy_id),
+                    ) {
+                        let volume = volume as f32 / NORMAL;
+                        let volume = volume * volume * volume;
+
+                        let mut pod_data = Vec::new();
+                        let mut pod_builder = PodBuilder::new(&mut pod_data);
+
+                        // let mut frame: ::std::mem::MaybeUninit<libspa::sys::spa_pod_frame> = ::std::mem::MaybeUninit::uninit();
```
bim9262, line 198: This comment is the equivalent of the `builder_add!` invocation below, but that requires a new feature (that's enabled by the patch in `Cargo.toml`)

src/blocks/sound/pipewire.rs
```diff
@@ -0,0 +1,643 @@
...
+    fn new() -> Result<Client> {
+        thread::Builder::new()
+            .name("privacy_pipewire".to_string())
+            .spawn(Client::main_loop_thread)
+            .error("failed to spawn a thread")?;
+
+        Ok(Client::default())
+    }
+
+    fn spawn_idle_listener(
```
bim9262, line 273: This loop runs when the main loop isn't busy. Not sure if we can spin up an async context instead of looping over the `command_listeners` over and over

Cargo.toml
```diff
@@ -93,3 +97,9 @@ lto = "thin"
 [profile.release-debug-info]
 inherits = "release"
 debug = true
+
+[patch.crates-io]
+pipewire = { git = "https://github.com/bim9262/pipewire-rs.git", branch = "builder_array" }
+pipewire-sys = { git = "https://github.com/bim9262/pipewire-rs.git", branch = "builder_array"}
+libspa = { git = "https://github.com/bim9262/pipewire-rs.git", branch = "builder_array" }
+libspa-sys = { git = "https://github.com/bim9262/pipewire-rs.git", branch = "builder_array" }
```
MaxVerevkin, line 105: Do you plan to upstream these changes?

src/blocks/sound/pipewire.rs
```diff
@@ -0,0 +1,645 @@
...
+                        continue;
+                    }
+                    i += 1;
+                    let Ok(command) = recv_result else {
+                        continue;
+                    };
+                    return command.execute(client, proxies);
+                }
+            }
+            std::thread::sleep(Duration::from_millis(1));
```
MaxVerevkin, line 295: Why is this sleep() needed?

src/blocks/sound/pipewire.rs
MaxVerevkin: Does it run right before the main loop goes idle (e.g. before calling `poll()` or equivalent), or does it run repeatedly until main loop has real work to do? If the former, then what wakes the loop when a command is send? If the latter, well, this is bad.

MaxVerevkin:
> After much investigation... I have determined that the pipewire sound block isn't quite as nice as using the pulseaudio version. The main reason is that there's no default sink/source (as far as I can tell). Also if you use a mix of tools and set the pipewire volume it doesn't update the pulseaudio volume, but updating the pulseaudio volume updates both. This leads to odd situations where pipewire thinks the volume is one thing and pulseaudio thinks it's another.
> 
> I suppose if you were completely pulseaudio free (which I kinda doubt at this point being all that feasible) then it wouldn't be an issue.

So.. is it worth adding 600 loc? Pulseaudio-free setups don't seem too realistic to me, plus people who use pipewire with pulseaudio compat may set the backend to "pipewire" and, as you described, get mixed results.

Cargo.toml
bim9262: I do. I have opened an issue as I can't create a fork (permission issue) on freedesktop's gitlab instance

src/blocks/sound/pipewire.rs
bim9262: It runs repeatedly until the main loop has real work to do, which I agree is bad.

src/blocks/sound/pipewire.rs
bim9262: Since the loop runs repeatedly, the sleep prevents pegging the cpu

bim9262:
> So.. is it worth adding 600 loc? Pulseaudio-free setups don't seem too realistic to me, plus people who use pipewire with pulseaudio compat may set the backend to "pipewire" and, as you described, get mixed results.

I think not. I can drop it especially the way the idle loop works for processing commands is less than ideal.

src/blocks/privacy.rs
MaxVerevkin: I don't have a specific opinion in this

Cargo.toml
MaxVerevkin: Yeah, I think so.

MaxVerevkin:
> Looks like a newer version of libpipewire is required than is available in the runner image. Might need to add a step to build it from source.

- This should be mentioned in `doc/manual_install.md`. 
- A bit weird idea: maybe we should just switch to Arch on the CI? I did so in https://github.com/MaxVerevkin/ewc because I couldn't install a developer version of some package for some reason.

bim9262:
> * This should be mentioned in `doc/manual_install.md`.

Updated the doc

> * A bit weird idea: maybe we should just switch to Arch on the CI? I did so in https://github.com/MaxVerevkin/ewc because I couldn't install a developer version of some package for some reason.

I update the workflows to use the archlinux image :)

bim9262:
Interesting that clippy and cargo-test fail since they worked when I pushed them to master on my fork.
It says:

```
All changes made by hooks:
diff --git a/Cargo.lock b/Cargo.lock
index 515d7ca..8a321be 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2875,7 +2875,7 @@ dependencies = [
  "cfg-expr",
  "heck",
  "pkg-config",
- "toml 0.8.8",
+ "toml",
  "version-compare",
 ]
 ```
 
 It doesn't make any sense why that would happen though since there are two different versions of toml in our package.

**Comment by bim9262** [comment]
> It doesn't make any sense why that would happen though since there are two different versions of toml in our package.

There *where* two versions of toml, but there aren't now. Rebasing and updating the Cargo.lock file fixed the issue (looks like the wrong line was git merged in).

File: src/blocks/privacy.rs [hunk 10]
```diff
@@ -0,0 +1,246 @@
...
+    AudioSink,
+    Video,
+    Webcam,
+    Unknown,
+}
+
+// {type: {name: {reader}}
+type PrivacyInfo = HashMap<Type, PrivacyInfoInner>;
+
+type PrivacyInfoInnerType = HashMap<String, HashSet<String>>;
```
bim9262, line 120: Using a hashset here may not be the best, maybe use a HashMap<String, usize> to be able to give a count perhaps you'd see

``` mic_icon { sink -> [Google Chrome input]}```
or 
``` mic_icon { sink -> [Google Chrome input (x2)]}```

Comment (MaxVerevkin) [reply src/blocks/privacy.rs:120]: Sounds like a good idea

File: cspell.yaml [hunk 11]
```diff
@@ -73,12 +75,14 @@ words:
...
   - iowait
   - ipapi
+  - itertools
   - iwctl
   - kbdd
   - kbddbus
   - kibi
   - kmon
   - libc
+  - libdbus
```
MaxVerevkin, line 85: Why was "libdbus" added? I don't think it is ever mentioned anywhere else.

bim9262: It's left over from when I was building pipewire from src. I'll remove it now.

MaxVerevkin (approved):

bim9262:
Just resolved merge conflict in `.github/workflows/pre-commit.yml`

Issue #1964, opened 2023-11-01
packages: add `brew` and `snap` backends


zheltikov:
Hello!

For systems where the `snap` and `brew` package managers are installed (apart from `apt`) it would be useful to have blocks that show available updates for those package managers too. The preferred configuration for these blocks would be practically identical the config of the `apt` block, excluding the block name obviously.

To list available updates for `snap`, you can run: `$ snap refresh --list`
To list available updates for `brew`, you can run: `$ brew outdated`

Some preferred icons for each block would be:
- `snap`: "🐦" or "📦"
- `brew`: "🍺"

Thanks!

MaxVerevkin:
I wonder if it would be a good idea to merge all package blocks (apt, pacman, ...) into one since they all have a very similar config and structure.

zheltikov:
Probably we should create a new "updates" or "packages" block, with the ability to configure "backends" for it.
Configuration would be like so:
```toml
[[block]]
block = "updates"
format = " $icon $count.eng(w:1) "
format_up_to_date = ""
package_manager_backends = ["apt", "pacman", "brew", "snap"]
```
@MaxVerevkin, thoughts on this approach?

IshanGrover2004:
> Probably we should create a new "updates" or "packages" block, with the ability to configure "backends" for it. Configuration would be like so:
> 
> ```toml
> [[block]]
> block = "updates"
> format = " $icon $count.eng(w:1) "
> format_up_to_date = ""
> package_manager_backends = ["apt", "pacman", "brew", "snap"]
> ```

I guess this will make some massive change in code structure. Is this OK? @MaxVerevkin 

I am thinking to break down to two PR: 1 for snap and brew and another PR maybe if you approve the above then I'll try that in another PR
Any other suggestion?

MaxVerevkin:
> I am thinking to break down to two PR: 1 for snap and brew and another PR maybe if you approve the above then I'll try that in another PR

It may make sense to do the latter first to reduce duplicated work, but this isn't required ofcourse.

> I guess this will make some massive change in code structure.

Sure, that's okay. And this doesn't have to be a breaking change for users. I thought about having `blocks/packages/{apt.rs,pacman.rs,...}` (containing only package manager specific code) and then reusing them in current blocks, which would become deprecated.

IshanGrover2004:
[...]

PR #1988, opened 2024-01-19
Combine `apt`, `pacman` & `dnf` block to `packages` block

### LINKED ISSUE #1958: packages: add `ignore_updates_regex`
We already have `warning_updates_regex` and `critical_updates_regex`. A new option can be added to filter the list of outdated packages.

- [x] Apt (#1967)
- [ ] Dnf
- [ ] Pacman

IshanGrover2004:
Worked on idea proposed in Issue #1964 

#### Before 
Individual blocks for checking updates
```toml
[[block]]
block = "apt"
interval = 1800
format = "$icon $count updates available"
format_singular = " $icon One update available "
format_up_to_date = " $icon system up to date "
```

#### After 
Unified `apt`, `pacman`, `aur`, `dnf` blocks in `packages` block and deprecates the old block
```toml
[[block]]
block = "packages"
package_manager = ["apt", "pacman", "aur","dnf"]
interval = 1800
format = "$icon $apt + $pacman + $aur + $dnf = $total updates available"
format_singular = " $icon One update available "
format_up_to_date = " $icon system up to date "
```

I have changed the structure of the code by adding a [Backend](https://github.com/IshanGrover2004/i3status-rust/blob/d2f3037be3f23f7fb1b2c8c43b759ffd590d3c47/src/blocks/packages.rs#L240) trait which we can implement in all struct for every package manager. I did try to keep the code very generic which made easy to add any package manager in future

closes #1958


MaxVerevkin:
Let's not delete the old blocks for now, just deprecate.

IshanGrover2004:
> Let's not delete the old blocks for now, just deprecate.

Done @MaxVerevkin 

Any other thing to do?

src/blocks.rs
```diff
@@ -137,6 +137,8 @@ macro_rules! define_blocks {
 define_blocks!(
     amd_gpu,
     apt,
+    pacman,
```
bim9262, line 140: This list should be alphabetical so that it's easy to find the blocks.

src/blocks/packages/apt.rs
```diff
@@ -0,0 +1,179 @@
+//! Pending updates available for your Debian/Ubuntu based system
```
bim9262, line 1: We don't usually put block specific usage notes in the submodule (apt/pacman), just in the packages' docs

src/blocks/packages/apt.rs
```diff
@@ -0,0 +1,179 @@
...
+            {
+                cnt += 1;
+            }
+        }
+
+        Ok(cnt)
+    }
+}
+
+async fn is_phased_update(config_path: &str, package_line: &str) -> Result<bool> {
```
bim9262, line 158: I'd put this in an `impl Apt` and then you'll still have access to self instead of passing around the config_path

src/blocks/packages/pacman.rs
```diff
@@ -0,0 +1,276 @@
+//! Pending updates available on pacman or an AUR helper.
```
bim9262, line 1: See previous note about docs.

bim9262 (commented):
It also seems that there's a lot of duplicate code between these and the deprecated blocks (which I don't see noted anywhere). I'd try to reuse this new code by importing from these new modules. That way if there's a bugfix/improvement made to one block it's made for both (although it may not be worth the effort depending on how long we plan to do the deprecation for).

src/blocks.rs
IshanGrover2004: Done

src/blocks/packages/pacman.rs
IshanGrover2004: I can clear the other stuff like configuration, example and add them in `packages.rs` file but some of the info like pacman has like hooking and some config setup for apt which user need to know and that is very specific so ig can't put it in package.rs file

IshanGrover2004: So if you could tell where i can put that and clear the documentation of submodule

bim9262: I'd use https://github.com/greshake/i3status-rust/blob/master/src/blocks/weather.rs and https://github.com/bim9262/i3status-rust/blob/privacy/src/blocks/privacy.rs as exmples where service/driver specific configurations are described in the block doc.

IshanGrover2004: Done

src/blocks/packages/apt.rs
IshanGrover2004: Done

bim9262:
I'd suggest the following changes to note the deprecated blocks (but maybe @MaxVerevkin has some other recommendations. I can open this in another PR if you want this to rebase on the change)

```diff
diff --git a/src/blocks.rs b/src/blocks.rs
index 836e65f7..894a0cc5 100644
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -45,11 +45,14 @@ use crate::{BoxedFuture, Request, RequestCmd};
 
 macro_rules! define_blocks {
     {
-        $( $(#[cfg(feature = $feat: literal)])? $block: ident $(,)? )*
+        $( $(#[cfg(feature = $feat: literal)])?
+        $(#[deprecated($($dep_k: ident = $dep_v: literal),+)])?
+        $block: ident $(,)? )*
     } => {
         $(
             $(#[cfg(feature = $feat)])?
             $(#[cfg_attr(docsrs, doc(cfg(feature = $feat)))])?
+            $(#[deprecated($($dep_k = $dep_v),+)])?
             pub mod $block;
         )*
 
@@ -58,6 +61,7 @@ macro_rules! define_blocks {
             $(
                 $(#[cfg(feature = $feat)])?
                 #[allow(non_camel_case_types)]
+                #[allow(deprecated)]
                 $block($block::Config),
             )*
             Err(&'static str, Error),
@@ -68,6 +72,7 @@ macro_rules! define_blocks {
                 match self {
                     $(
                         $(#[cfg(feature = $feat)])?
+                        #[allow(deprecated)]
                         Self::$block { .. } => stringify!($block),
                     )*
                     Self::Err(name, _err) => name,
@@ -78,6 +83,7 @@ macro_rules! define_blocks {
                 match self {
                     $(
                         $(#[cfg(feature = $feat)])?
+                        #[allow(deprecated)]
                         Self::$block(config) => futures.push(async move {
                             while let Err(err) = $block::run(&config, &api).await {
                                 if api.set_error(err).is_err() {
@@ -114,6 +120,7 @@ macro_rules! define_blocks {
                 match block_name {
                     $(
                         $(#[cfg(feature = $feat)])?
+                        #[allow(deprecated)]
                         stringify!($block) => match $block::Config::deserialize(table) {
                             Ok(config) => Ok(BlockConfig::$block(config)),
                             Err(err) => Ok(BlockConfig::Err(stringify!($block), crate::errors::Error::new(err.to_string()))),
@@ -136,6 +143,10 @@ macro_rules! define_blocks {
 
 define_blocks!(
     amd_gpu,
+    #[deprecated(
+        since = "0.32.4",
+        note = "The block has been deprecated in favor of the the packages block"
+    )]
     apt,
     backlight,
     battery,
@@ -163,6 +174,10 @@ define_blocks!(
     notmuch,
     nvidia_gpu,
     packages,
+    #[deprecated(
+        since = "0.32.4",
+        note = "The block has been deprecated in favor of the the packages block"
+    )]
     pacman,
     pomodoro,
     rofication,
```

IshanGrover2004:
> diff --git a/src/blocks.rs b/src/blocks.rs
> _ _ _ _ 
> -   $( $(#[cfg(feature = $feat: literal)])? $block: ident $(,)? )*
> +  $( $(#[cfg(feature = $feat: literal)])?
> +  $(#[deprecated($($dep_k: ident = $dep_v: literal),+)])?
> +  $block: ident $(,)? )*
> _ _ _ _

Do I need to do these changes and send commit?

bim9262:
I'd wait to see if @MaxVerevkin has any feedback/suggestions first

MaxVerevkin:
Deprecation patch looks good to me. Just note that the next version is 0.33. And hopefully in the future we will not need to annotate everything with `#[allow(deprecated)]` as apparently warnings inside the defining crate are a bug - https://github.com/rust-lang/rust/issues/47219. By the way, allow(deprecated) in `BlockConfig::name()` is not necessary.

src/blocks/packages.rs
```diff
@@ -0,0 +1,370 @@
...
+        let (mut apt_count, mut pacman_count, mut aur_count) = (0, 0, 0);
+        let mut critical_vec = vec![false];
+        let mut warning_vec = vec![false];
+
+        // Iterate over the all package manager listed in Config
+        for package_manager in config.package_manager.clone() {
+            match package_manager {
+                PackageManager::Apt => {
+                    let mut apt = Apt::new();
+                    apt.ignore_updates_regex = ignore_updates_regex.clone();
```
MaxVerevkin, line 277: I'm a bit confused why you have this `match` when you could've saved the list of package managers as a `Vec<Box<dyn Backend>>` (you already construct `Box<dyn Backend>` above, but then discard them).

src/blocks/packages.rs
```diff
@@ -0,0 +1,370 @@
...
+        select! {
+            _ = sleep(config.interval.0) => (),
+            _ = api.wait_for_update_request() => (),
+        }
+    }
+}
+
+#[async_trait]
+trait Backend {
+    async fn setup(&mut self) -> Result<()>;
```
MaxVerevkin, line 361: There is no reason for `setup()` to be part of this trait. Do the necessary setup work in each backend's constructor.

IshanGrover2004: Done

IshanGrover2004:
I did the changes proposed
Anyother work todo here?

src/blocks/packages.rs
```diff
@@ -0,0 +1,356 @@
...
+            PackageManager::Aur => Box::new(Aur::new()) as Box<dyn Backend>,
+        };
+
+        package_manager_vec.push(backend);
+    }
+
+    loop {
+        let (mut apt_count, mut pacman_count, mut aur_count) = (0, 0, 0);
+        let mut critical_vec = vec![false];
+        let mut warning_vec = vec![false];
```
MaxVerevkin, line 271: No need to store a vector, just have a boolean an update it like `critical |= critical_updates_regex.....`

src/blocks/packages.rs
```diff
@@ -0,0 +1,356 @@
...
+            PackageManager::Apt => Box::new(Apt::new().await?) as Box<dyn Backend>,
+            PackageManager::Pacman => Box::new(Pacman::new().await?) as Box<dyn Backend>,
+            PackageManager::Aur => Box::new(Aur::new()) as Box<dyn Backend>,
+        };
+
+        package_manager_vec.push(backend);
+    }
+
+    loop {
+        let (mut apt_count, mut pacman_count, mut aur_count) = (0, 0, 0);
```
MaxVerevkin, line 269: I think it's better to have a single place with driver-specific code in the block and that is when creating the drivers. Here you can have a `HashMap<&'static str, u32>`, so that then you don't have to `match` on the exact type of each driver. But `Backend::package_manager()` (which can be shortened to just `Backend::name()`) will have to return `&'static str`.

MaxVerevkin:
> Anyother work todo here?

Update `pacman.rs` and `apt.rs` to reuse the same code.

bim9262:
We should probably include `dnf` in this as well as that block already exists, whereas the other examples in #1964 don't exist as their own bocks yet anyway (`brew` and `snap`)

IshanGrover2004:
> Update `pacman.rs` and `apt.rs` to reuse the same code.

Do you mean to use code of `blocks/packages/pacman.rs` in `blocks/pacman.rs`

src/blocks/packages.rs
IshanGrover2004: How about if we have HashMap of (backend,update_count) at the starting here
```rust
    let mut package_manager_vec = Vec::new();

    for &package_manager in config.package_manager.iter() {
        let backend = match package_manager {
            PackageManager::Apt => Box::new(Apt::new().await?) as Box<dyn Backend>,
            PackageManager::Pacman => Box::new(Pacman::new().await?) as Box<dyn Backend>,
            PackageManager::Aur => Box::new(Aur::new()) as Box<dyn Backend>,
        };

        package_manager_vec.push(backend);
    }
```
so that we have hashmap which contains what we need till the end. and we can get the name of backend by `Backend::package_manager()`

MaxVerevkin: What I hand in mind is something like

```rust
for package_manager in &package_manager_vec {
    let updates = package_manager.get_updates_list().await?;
    values.insert(
        package_manager.name(),
        Value::number(package_manager.get_update_count(&updates).await?),
    );
    warning |= warning_updates_regex
        .as_ref()
        .is_some_and(|regex| package_manager.has_matching_update(&updates, regex));
    critical |= critical_updates_regex
        .as_ref()
        .is_some_and(|regex| package_manager.has_matching_update(&updates, regex));
}
```

src/blocks/packages.rs
```diff
@@ -0,0 +1,351 @@
...
+            let updates = package_manager.get_updates_list().await?;
+
+            match package_manager.package_manager() {
+                PackageManager::Apt => {
+                    let apt = Apt {
+                        // Config file not needed in counting updates
+                        config_file: String::new(),
+                        ignore_phased_updates: config.ignore_phased_updates,
+                        ignore_updates_regex: ignore_updates_regex.clone(),
+                    };
```
MaxVerevkin, line 284: - Why not pass these parameters once when creating the backend?
- `ignore_updates_regex` is not apt-specific and should be applied to all backends.

MaxVerevkin:
> Do you mean to use code of `blocks/packages/pacman.rs` in `blocks/pacman.rs` and same for apt

Yes.

> or Just have to use one in the packages.rs

Yes.

src/blocks/packages.rs
IshanGrover2004: > Why not pass these parameters once when creating the backend?

Done 

> `ignore_updates_regex` is not apt-specific and should be applied to all backends.

Pacman and Aur doesn't need them anywhere while checking updates or counting them
That's why the Pacman & Aur struct don't have them as variant

MaxVerevkin: Neither does apt. This is a generic user setting to filter the updates they don't care about. See #1958.

src/blocks/packages.rs
IshanGrover2004: Done

src/blocks/packages.rs
IshanGrover2004: Does` _ignore_updates_regex` allowed because that parameter is never read

MaxVerevkin: Not sure what you are referring to. Don't store it in every driver, just apply it when iterating over package managers if available.

IshanGrover2004:
Everything done you said @MaxVerevkin @bim9262

src/blocks/packages.rs
MaxVerevkin: > Don't store it in every driver, just apply it when iterating over package managers if available.

src/blocks/packages.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+}
+
+#[async_trait]
+pub trait Backend {
+    fn name(&self) -> &str;
+
+    async fn get_updates_list(&self) -> Result<String>;
+
+    async fn get_update_count(&self, updates: &str) -> Result<usize>;
```
MaxVerevkin, line 338: I realize that this is similar to how pacman and apt used to work, but since we are rewriting this, maybe let's change `get_updates_list` to return a list of strings (with backend-specific filtering applied, such as phased updates in apt). This way we don't need `get_update_count` at all. This will also make `ignore_updates_regex` option easier to implement in backend-agnostic way.

MaxVerevkin:
As mentioned by @bim9262, it's better to convert `dnf` too in the same PR.

src/blocks/packages.rs
IshanGrover2004:
```rust
Box::new(Pacman::new(ignore_updates_regex.clone()).await?) as Box<dyn Backend>
```
You mean here just send `ignore_updates_regex` through `Pacman::new()` but don't store it as a variant or do anything else

IshanGrover2004:
> As mentioned by @bim9262, it's better to convert `dnf` too in the same PR.

Can you provide more info on it. Because I don't get it what exactly to do

src/blocks/packages.rs
IshanGrover2004: I'll try to do

src/blocks/packages.rs
MaxVerevkin: No, backends don't have to know about `ignore_updates_regex` at all. When you call `get_updates_list()`, check if `ignore_updates_regex` is set, and it if is, filter the list.

src/blocks/packages.rs
IshanGrover2004: I am sorry, but I am not able to do this

MaxVerevkin: May I ask why?

src/blocks/packages.rs
IshanGrover2004: Do you need something like this
```rust
if let Some(_) = ignore_updates_regex {
    let updates_count = package_manager
        .get_update_count(&updates, ignore_updates_regex)
        .await?;
}
```

And then `get_count_update` will filter accordingly

src/blocks/packages.rs
IshanGrover2004: > May I ask why?

Aah, I am not able to think how can i do that
I did try something but i think that's wrong so I dropped the idea

src/blocks/packages.rs
MaxVerevkin: No, I was thinking about something like

```rust
let updates = package_manager.get_updates().await?;
if let Some(re) = ignore_updates_regex {
	updates.retain(|u| !re.is_match(u));
}
```

Here `.get_updates()` returns a `Vec<String>` though, so each driver should convert a string to a list of lines, applying filtering if necessary.

src/blocks/packages.rs
IshanGrover2004: Maybe, I'll just directly commit, push the code

src/blocks/packages.rs
IshanGrover2004: Oh okk

src/blocks/packages.rs
IshanGrover2004: Did it

src/blocks/packages.rs
IshanGrover2004: Done

src/blocks/packages.rs
```diff
@@ -0,0 +1,374 @@
...
+
+#[async_trait]
+pub trait Backend {
+    fn name(&self) -> &str;
+
+    async fn get_updates_list(&self) -> Result<Vec<String>>;
+
+    fn has_matching_update(&self, updates: &[String], regex: &Regex) -> bool {
+        updates.iter().any(|line| regex.is_match(line))
+    }
```
MaxVerevkin, line 373: Since none of the backends override this function, it doesn't use `self`, and I don't think any will in the future, let's move it out of the trait.

src/blocks/packages/apt.rs
```diff
@@ -0,0 +1,133 @@
...
+                let is_phased_update =
+                    async { self.is_phased_update(update_line).await.unwrap_or(false) };
+
+                Some(update_line.to_string()).filter(|_| {
+                    !self.ignore_phased_updates
+                        || !tokio::task::block_in_place(|| {
+                            tokio::runtime::Runtime::new()
+                                .unwrap()
+                                .block_on(is_phased_update)
+                        })
```
MaxVerevkin, line 126: Spawning a new tokio runtime for each update is not great. Since you can't use `.await` in a closure, you should just use a regular for loop and `.push()` updates into the vector.

IshanGrover2004: I predicted you may say this :smile:

src/blocks/packages.rs
```diff
@@ -0,0 +1,374 @@
...
+            }
+            PackageManager::Pacman => Box::new(Pacman::new().await?) as Box<dyn Backend>,
+            PackageManager::Aur => {
+                Box::new(Aur::new(config.aur_command.clone().unwrap_or_default()))
+                    as Box<dyn Backend>
+            }
+            PackageManager::Dnf => Box::new(Dnf::new()) as Box<dyn Backend>,
+        };
+
+        package_manager_vec.push(backend);
```
MaxVerevkin, line 298: I think if you inline the `backend` variable and write something like `.push(match ... { });`, you will not have to write ` as Box<dyn Backend>` everywhere, and it will be more compact.

src/blocks/packages.rs
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    loop {
+        let mut package_manager_map: HashMap<&str, u32> =
+            [("apt", 0), ("pacman", 0), ("aur", 0), ("dnf", 0)].into();
```
MaxVerevkin, line 303: As I've written before, I would prefer to keep backend-specific code to the minimum. Start with an empty hash map and insert as you go. And to not have two hash maps (this one and the one with values to be displayed), you can insert directly into the map which you will pass to `widget.set_values(...)`).

IshanGrover2004: I did this because of below code, It may panic if "apt" doesn't exist..that's why I thought to set default 0
```rust
"apt" => Value::number(package_manager_map["apt"]),
```

src/blocks/packages/apt.rs
IshanGrover2004: Done

src/blocks/packages.rs
IshanGrover2004: Done

src/blocks/packages.rs
IshanGrover2004: Done

src/blocks/packages.rs
IshanGrover2004: How can i get inner value from `Value::number(num)`

MaxVerevkin: Why? Here is an example what you can do https://github.com/greshake/i3status-rust/pull/1988#discussion_r1460760865

IshanGrover2004: > Here is an example what you can do https://github.com/greshake/i3status-rust/pull/1988#discussion_r1460760865

Yah, I got that but I didn't got how do i get total_count from a `HashMap<Cow<'static, str>, Value>`

IshanGrover2004: Done btw

MaxVerevkin: Why not just

```rust
package_manager_map.insert(package_manager.name(), Value::number(updates_count));
total += updates_cont;
```

IshanGrover2004: Oh sorry, I am dumb. I thought to go hard way

IshanGrover2004:
Any other thing left to change @MaxVerevkin

src/blocks/pacman.rs
```diff
@@ -247,13 +218,11 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
...
-            Watched::Both(aur_command) => {
-                let (pacman_updates, aur_updates) = tokio::try_join!(
-                    get_pacman_available_updates(),
-                    get_aur_available_updates(aur_command)
-                )?;
-                let pacman_count = get_update_count(&pacman_updates);
-                let aur_count = get_update_count(&aur_updates);
+            Watched::Both(_) => {
+                let pacman_updates = pacman_backend.get_updates_list().await?;
+                let aur_updates = aur_backend.get_updates_list().await?;
```
MaxVerevkin, line 223: Please restore the `try_join!` thing to get the updates in parallel. This is also something that we should probably do in the new block too.

IshanGrover2004: Done

IshanGrover2004:
This PR is ready to merge i guess if no more review left @MaxVerevkin

c3Ls1US:
Hey, great work for this feature and I agree with unifying the package manager blocks.

Though, have you already explored using libalpm for pacman's database functionality? [I believe there are Rust bindings for it](https://github.com/archlinux/alpm.rs) and am just curious why you decided not to go this way.

MaxVerevkin:
Please add a deprecation warning for `dnf` too.

IshanGrover2004:
> Please add a deprecation warning for `dnf` too.

Done :heavy_check_mark:

MaxVerevkin (approved):
LGTM!

Just some recommendations regarding the docs:
- The example section may be made a bit less verbose, for example by not including `critical_updates_regex` in every example.
- "Default" column for `package_manager` should mention that it is automatically derived from the format templates, but can be used to influence the `$total` value.

IshanGrover2004:
hey @MaxVerevkin ,
Do the `$both` works here if `$both` value is not set in code in block/packages
```rust
format = " $icon $pacman + $aur = $both updates available "
```

If not work, should i remove `$both` from here

MaxVerevkin:
> hey @MaxVerevkin ,
> Do the `$both` works here if `$both` value is not set in code in block/packages
> ```rust
> format = " $icon $pacman + $aur = $both updates available "
> ```
> 
> If not work, should i remove `$both` from [here](https://github.com/IshanGrover2004/i3status-rust/blob/1964-list-update-packages/src/blocks/packages.rs#L125)

Oh, yes, missed this one. This should be `$total` instead of `$both`.

IshanGrover2004:
> Oh, yes, missed this one. This should be `$total` instead of `$both`.

Done :heavy_check_mark:

MaxVerevkin (approved):

ammgws (approved):

MaxVerevkin:
Thanks!

PR #1993, opened 2024-01-30
default theme to "plain" even if [theme] section is absent

### LINKED ISSUE #1992: Plain theme never shows colors if theme is not explicitly set in configuration
Try this as a minimal configuration that should show colored output:

```
[[block]]
block = "load"
warning = "0"
```

It shows uncolored. Now add a section for theme:

```
[theme]

[[block]]
block = "load"
warning = "0"
```

The bar now shows a red symbol. The first version should also set plain theme and show colored symbols.

MaxVerevkin:
Also switch from container-level `serde(default)` to field-level `serde(default)` in some places to avoid unnecessary file reads and deserialization, because container-level attribute creates a default instance of the entire struct and thus calls `default()` even for specified fields. In particullar, this change avoids reading `plain.toml` if theme is set.

Fixes #1992

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -10,35 +10,60 @@ use crate::formatting::config::Config as FormatConfig;
 use crate::icons::{Icon, Icons};
 use crate::themes::{Theme, ThemeOverrides, ThemeUserConfig};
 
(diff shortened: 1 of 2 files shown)


bim9262 (approved):
lgtm

PR #2005, opened 2024-02-20
Adjust phase of time block

### LINKED ISSUE #1991: Improve time display for time update intervals other than 1 second
I want to reproduce i3status' behavior where the time update interval is 5 seconds, and the displayed time always snaps to a five second multiple, ie. 12:30:05, 12:30:10, and so on. On i3status-rust, you can set the interval, but then the rounding will be random, you get 12:30:07, 12:30:12 etc. for 5 second interval.

Another problem with the current system is that if you have a time display that doesn't show the seconds and set the update interval to 60 seconds, the clock display will have a random error since the actual time can change at any time during the 60 second interval window and the visible clock won't update.

A way to accomplish this is to adjust clock component's phase at startup so that it's first update snaps to the :00 time of the next minute. Now if the interval is a divisor of 60, the clock will keep showing a clean sequence of 00, 05, 10, 15 or 00, 10, 20, 30. If the interval is 60, the clock will be updated at the exact time the actual time changes.

skoriop:
Closes #1991 

Instead of snapping to the start (`:00`) of the next minute, it snaps to the nearest multiple of the interval value. 

For example, if the interval value is `5` and the config is initialized at `:17`, the sequence shown will be `:17`, `:20`, `:25` and so on.

### DIFF
--- a/src/blocks/time.rs
+++ b/src/blocks/time.rs
@@ -101,14 +101,6 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
     );
     timer.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
 
-    let interval_seconds = config.interval.seconds();
-    if interval_seconds <= 60 {
-        let phase = Utc::now().second() as u64 % interval_seconds;
-        if phase != 0 {
-            timer.reset_after(Duration::from_secs(interval_seconds - phase));
-        }
-    }
-
     loop {
         let mut widget = Widget::new().with_format(format.clone());
 
@@ -119,8 +111,16 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
 
         api.set_widget(widget)?;
 
+        let tick = async {
+            let phase = Utc::now().second() as u64 % config.interval.seconds();
+            if phase != 0 {
+                timer.reset_after(Duration::from_secs(config.interval.seconds() - phase));
+            }
+            timer.tick().await;
+        };
+
         tokio::select! {
-            _ = timer.tick() => (),
+            _ = tick => (),
             _ = api.wait_for_update_request() => (),
             Some(action) = actions.recv() => match action.as_ref() {
                 "next_timezone" => {
--- a/src/blocks/time.rs
+++ b/src/blocks/time.rs
@@ -49,7 +49,7 @@
 //! # Icons Used
 //! - `time`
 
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
Thanks for the patch! Currently this does not handle suspend/resume situation. I think to fix this the timer's `set_missed_tick_behavior` should be set to `MissedTickBehavior::Skip`, at which point it's better to create a timer from duration instead of `config.interval.timer()` shortcut.

MaxVerevkin:
Hm, that didn't work :(

skoriop:
My bad, I think I misunderstood your initial comment. Of the approaches I tried, the only one that seems to work consistently is to move the phase adjusting logic into the `loop` block.

MaxVerevkin:
No, no, you did exactly what I though. It's strange it didn't work, not sure why. Adjusting the phase in the loop is good too. Maybe we don't even need a timer then, and can just call `sleep`?

src/blocks/time.rs
```diff
@@ -107,8 +111,16 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
 
         api.set_widget(widget)?;
 
+        let tick = async {
```
bim9262, line 114: Why make this an async block instead of just:

```rust
        let phase = now.second() as u64 % config.interval.seconds();
        if phase != 0 {
            timer.reset_after(Duration::from_secs(config.interval.seconds() - phase));
        }
```

skoriop: Yeah, that makes sense.

skoriop:
> Maybe we don't even need a timer then, and can just call `sleep`?

I think using `interval` would be better here since it measures the time since the last tick while `sleep` can be affected by the time spent in between calls (I couldn't tell the difference when using either, but it could become significant later if more code is added)

src/blocks/time.rs
```diff
@@ -107,6 +111,11 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
 
         api.set_widget(widget)?;
 
+        let phase = Utc::now().second() as u64 % config.interval.seconds();
```
MaxVerevkin, line 114: `Utc::now()` is already called a few lines above, let's reuse that value.

skoriop: Done, I've also reused the `config.interval.seconds()` value

MaxVerevkin (approved):

PR #2008, opened 2024-02-21
Add duration formatter

bim9262:
TODO: document duration formatter.

Any thoughts on the names of the arguments, the format, or the defaults?

The formatted can be used by the battery, uptime, tea_timer and pomodoro blocks.

### DIFF
--- a/src/blocks/tea_timer.rs
+++ b/src/blocks/tea_timer.rs
@@ -4,18 +4,21 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>\" $icon {$minutes:$seconds \|}\"</code>
+//! `format` | A string to customise the output of this block. See below for available placeholders. | <code>\" $icon {$time.duration(hms:true) \|}\"</code>
 //! `increment` | The numbers of seconds to add each time the block is clicked. | 30
 //! `done_cmd` | A command to run in `sh` when timer finishes. | None
 //!
-//! Placeholder      | Value                                                          | Type   | Unit
-//! -----------------|----------------------------------------------------------------|--------|---------------
-//! `icon`           | A static icon                                                  | Icon   | -
-//! `hours`          | The hours remaining on the timer                               | Text   | h
-//! `minutes`        | The minutes remaining on the timer                             | Text   | mn
-//! `seconds`        | The seconds remaining on the timer                             | Text   | s
+//! Placeholder            | Value                                                          | Type     | Unit
+//! -----------------------|----------------------------------------------------------------|----------|---------------
+//! `icon`                 | A static icon                                                  | Icon     | -
+//! `time`                 | The time remaining on the timer                                | Duration | -
+//! `hours` *DEPRECATED*   | The hours remaining on the timer                               | Text     | h
+//! `minutes` *DEPRECATED* | The minutes remaining on the timer                             | Text     | mn
+//! `seconds` *DEPRECATED* | The seconds remaining on the timer                             | Text     | s
 //!
-//! `hours`, `minutes`, and `seconds` are unset when the timer is inactive.
+//! `time`, `hours`, `minutes`, and `seconds` are unset when the timer is inactive.
+//!
+//! `hours`, `minutes`, and `seconds` have been deprecated in favor of `time`.
 //!
 //! Action      | Default button
 //! ------------|---------------
@@ -37,13 +40,14 @@
 
 use super::prelude::*;
 use crate::subprocess::spawn_shell;
-use chrono::{Duration, Utc};
+
+use std::time::{Duration, Instant};
 
 #[derive(Deserialize, Debug, SmartDefault)]
 #[serde(deny_unknown_fields, default)]
 pub struct Config {
     pub format: FormatConfig,
-    pub increment: Option<i64>,
+    pub increment: Option<u64>,
     pub done_cmd: Option<String>,
 }
 
@@ -59,17 +63,20 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
     let interval: Seconds = 1.into();
     let mut timer = interval.timer();
 
-    let format = config.format.with_default(" $icon {$minutes:$seconds |}")?;
+    let format = config
+        .format
+        .with_default(" $icon {$time.duration(hms:true) |}")?;
 
-    let increment =
-        Duration::try_seconds(config.increment.unwrap_or(30)).error("invalid increment value")?;
-    let mut timer_end = Utc::now();
+    let increment = Duration::from_secs(config.increment.unwrap_or(30));
+    let mut timer_end = Instant::now();
 
     let mut timer_was_active = false;
 
     loop {
-        let remaining_time = timer_end - Utc::now();
-        let is_timer_active = remaining_time > Duration::zero();
+        let mut widget = Widget::new().with_format(format.clone());
+
+        let remaining_time = timer_end - Instant::now();
+        let is_timer_active = !remaining_time.is_zero();
 
         if !is_timer_active && timer_was_active {
             if let Some(cmd) = &config.done_cmd {
@@ -78,24 +85,30 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
         }
         timer_was_active = is_timer_active;
 
-        let (hours, minutes, seconds) = if is_timer_active {
-            (
-                remaining_time.num_hours(),
-                remaining_time.num_minutes() % 60,
-                remaining_time.num_seconds() % 60,
-            )
-        } else {
-            (0, 0, 0)
-        };
-
-        let mut widget = Widget::new().with_format(format.clone());
-
-        widget.set_values(map!(
+        let mut values = map!(
             "icon" => Value::icon("tea"),
-            [if is_timer_active] "hours" => Value::text(format!("{hours:02}")),
-            [if is_timer_active] "minutes" => Value::text(format!("{minutes:02}")),
-            [if is_timer_active] "seconds" => Value::text(format!("{seconds:02}")),
-        ));
+        );
+
+        if is_timer_active {
+            values.insert("time".into(), Value::duration(remaining_time));
+            let mut seconds = remaining_time.as_secs();
+
+            if format.contains_key("hours") {
+                let hours = seconds / 3_600;
+                values.insert("hours".into(), Value::text(format!("{hours:02}")));
+                seconds %= 3_600;
+            }
+
+            if format.contains_key("minutes") {
+                let minutes = seconds / 60;
+                values.insert("minutes".into(), Value::text(format!("{minutes:02}")));
+                seconds %= 60;
+            }
+
+            values.insert("seconds".into(), Value::text(format!("{seconds:02}")));
+        }
+
+        widget.set_values(values);
 
         api.set_widget(widget)?;
 
@@ -103,7 +116,7 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
             _ = timer.tick(), if is_timer_active => (),
             _ = api.wait_for_update_request() => (),
             Some(action) = actions.recv() => {
-                let now = Utc::now();
+                let now = Instant::now();
                 match action.as_ref() {
                     "increment" if is_timer_active => timer_end += increment,
                     "increment" => timer_end = now + increment,
--- a/src/formatting.rs
+++ b/src/formatting.rs
@@ -80,15 +80,15 @@
 //! ## `duration`/`dur` - Format durations
 //!
 //! Argument                     | Description                                                                                      |Default value
-//! -----------------------------|--------------------------------------------------------------------------------------------------|------------------------------------------------------
-//! `hms`                        | Should the format be hours:minutes:seconds.milliseconds                                          | `false`
-//! `max_unit`                   | The largest unit to display the duration with (see below for the list of all possible units)     | hms ? `h` : `y`
-//! `min_unit`                   | The smallest unit to display the duration with (see below for the list of all possible units)    | `s`
-//! `units`                      | The number of units to display                                                                   | min(# of units between `max_unit` and `min_unit``, 2)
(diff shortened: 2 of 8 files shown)


src/formatting/formatter/duration.rs
```diff
@@ -0,0 +1,192 @@
...
+                    );
+                }
+                "round_up" => {
+                    round_up = arg
+                        .val
+                        .parse()
+                        .ok()
+                        .error("round_up must be true or false")?;
+                }
+                "prefix_space" => {
```
bim9262, line 65: This is borrowed from the eng formatter, but the units here aren't prefixes

MaxVerevkin:
It is hard to judge/review until there is at least one block using this. The good thing is that blocks can be migrated in a non-breaking way, for example `tea_timer` can still have the old (and eventually deprecated and then removed) hours/minutes/seconds, in addition to the new placeholder.

By the way, I think it is safe to delete `fix.rs` for now.

bim9262:
Added two examples, uptime and tea_timer. Still need to doc the formatter...

src/blocks/tea_timer.rs
```diff
@@ -77,32 +83,41 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
...
-        } else {
-            (0, 0, 0)
-        };
+        let mut values = map!(
+            "icon" => Value::icon("tea"),
+        );
 
-        let mut widget = Widget::new().with_format(format.clone());
+        if is_timer_active {
+            widget.state = State::Info;
```
MaxVerevkin, line 91: The state change seems unrelated

MaxVerevkin:
Do I understand correctly that currently there is no way to use `mm:ss` format? I think this can be achieved by allowing to replace the space character between units with `:` and adding an option to hide units. Maybe we can even add a shortcut parameter to set all these values, because `.dur(units:6,show_leading_units_if_zero:false,separator:':',show_units:false)` is too much.

By the way, maybe `show_leading_units_if_zero` should default to `false`? I'm not %100 sure though.

src/formatting/formatter/duration.rs
```diff
@@ -0,0 +1,183 @@
...
+                        // Zip up the values of time with the matching units of time
+                        .zip(UNITS[self.max_unit_index..=self.min_unit_index].iter())
+                        // Skip wile the value is zero, unless we want to display the leading units of time with value of zero.
+                        // For example we want to have a minimum unit of seconds but to always show two values we could have:
+                        // " 0m 15s"
+                        .skip_while(|((i, value), _)| {
+                            **value == 0
+                                && (!self.show_leading_units_if_zero
+                                    || *i + self.max_unit_index
+                                        != self.min_unit_index - self.units + 1)
```
MaxVerevkin, line 163:
```suggestion
                                        != self.min_unit_index + 1 - self.units)
```

because

```toml
[[block]]
block = "tea_timer"
format = " $time.dur(units:6) | click to start "
```

causes a panic:

```
thread 'main' panicked at src/formatting/formatter/duration.rs:163:44:
attempt to subtract with overflow
```

bim9262:
> Do I understand correctly that currently there is no way to use `mm:ss` format? I think this can be achieved by allowing to replace the space character between units with `:` and adding an option to hide units. Maybe we can even add a shortcut parameter to set all these values, because `.dur(units:6,show_leading_units_if_zero:false,separator:':',show_units:false)` is too much.

That's right. If you wanted to do [hh]:mm:ss you'd want to do something like `.dur(max_unit:h,min_unit:s,show_leading_units_if_zero:false,separator:':',show_units:false)` otherwise after you get to a day+ you'd end up with a format like `dd:hh:mm:ss`, which probably isn't a good format.

The most flexible solution would be to allow an explicit format to be specified if we want to allow "sub-formats" like:
```.dur(format:'{$d |}{$h:|}{$m:|}'{$s:|}',show_leading_units_if_zero:false)```

or using the allow argument mentioned in https://github.com/greshake/i3status-rust/issues/1957:
```.dur(format:'{$d.eng(allow:1..) |}{$h.eng(allow:1..):|}{$m.eng(allow:1..):|}'{$s.eng(allow:1..):|}')```

> By the way, maybe `show_leading_units_if_zero` should default to `false`? I'm not %100 sure though.

I set `show_leading_units_if_zero` to true, because it means that say you have a countdown timer you you would go from `" 1m  0s"` to `" 0m 59s"` to keep a consistent width.

File: src/blocks/tea_timer.rs (hunk 2 above)
Comment (bim9262) [reply src/blocks/tea_timer.rs:91]: That's true. I was mirroring something I did in the pomodoro block (which also needed this formatter).
I can remove it from this PR if you'd like.

**Comment by bim9262** [comment]
Another option would be to  use [strftime](https://docs.rs/chrono/latest/chrono/format/strftime/index.html) style formats.

**Comment by bim9262** [comment]
Perhaps adding an `hms` option which sets the max_arg to hours and sets the separator to : and hides the units would be enough instead of adding nested formatters or a new parser.

**Comment by MaxVerevkin** [comment]
> The most flexible solution would be to allow an explicit format to be specified if we want to allow "sub-formats" like:
> `.dur(format:'{$d |}{$h:|}{$m:|}'{$s:|}',show_leading_units_if_zero:false)`

That's... too much formatting :smile: 

> That's right. If you wanted to do [hh]:mm:ss you'd want to do something like `.dur(max_unit:h,min_unit:s,show_leading_units_if_zero:false,separator:':',show_units:false)` otherwise after you get to a day+ you'd end up with a format like `dd:hh:mm:ss`, which probably isn't a good format.
>
> ...
>
> Perhaps adding an `hms` option which sets the max_arg to hours and sets the separator to : and hides the units would be enough instead of adding nested formatters or a new parser.

`.dur(hms:true)` seems pretty good.

> or using the allow argument mentioned in #1957:
> `.dur(format:'{$d.eng(allow:1..) |}{$h.eng(allow:1..):|}{$m.eng(allow:1..):|}'{$s.eng(allow:1..):|}')`

Again, nested format strings feel weird. But maybe... maybe we can add support for some sort of "fields", like `$time.h.eng(w:2):$time.s.eng(w:2)`? Not sure if it makes sense.

**Comment by bim9262** [comment]
> > Perhaps adding an `hms` option which sets the max_arg to hours and sets the separator to : and hides the units would be enough instead of adding nested formatters or a new parser.
> 
> `.dur(hms:true)` seems pretty good.
> 
> > or using the allow argument mentioned in #1957:
> > `.dur(format:'{$d.eng(allow:1..) |}{$h.eng(allow:1..):|}{$m.eng(allow:1..):|}'{$s.eng(allow:1..):|}')`
> 
> Again, nested format strings feel weird. But maybe... maybe we can add support for some sort of "fields", like `$time.h.eng(w:2):$time.s.eng(w:2)`? Not sure if it makes sense.

I'm not sure that the field concept would be generally useful (outside of the duration format) and much more verbose so I'm leaning towards the `hms` option.

**Comment by bim9262** [comment]
I think the behaviour of `round_up`  should be modified from `Round up to the nearest min_unit`  to `Round up to the nearest minimum displayed unit`

**Comment by bim9262** [comment]
I guess we should decide if hms should be enabled by default for blocks that used that format previously or if it is better to use the unambiguous default format. (2 units of hms are either hh:mm or mm::ss, but unless you know the interval/or it's a small interval and you can see the seconds changing)

**Comment by bim9262** [comment]
Once we think the settings/defaults make sense I'll add some unit tests as well.

**Comment by bim9262** [comment]
@MaxVerevkin I know this is a lot to review, do you want me to move the splitting up of the formatters into modules into its own PR?

**Comment by MaxVerevkin** [comment]
> do you want me to move the splitting up of the formatters into modules into its own PR?

That would be nice :)

**Comment by bim9262** [comment]
@MaxVerevkin, I rebased on master now that the formatter split and allow empty strings PRs have merged.
Any changes to the argument names or default values?
Should we make the default of the tea timer block use hms by default, since that's the default format now?

**Comment by MaxVerevkin** [comment]
> Should we make the default of the tea timer block use hms by default, since that's the default format now?

Yes, let's try to keep the defaults as they are.

I'll try to take a closer look tomorrow.

File: src/blocks/tea_timer.rs (hunk 2 above)
Comment (MaxVerevkin) [reply src/blocks/tea_timer.rs:93]: Let's remove this for now, maybe add it later in a new PR.

**Comment by MaxVerevkin** [comment]
The [types](https://greshake.github.io/i3status-rust/i3status_rs/formatting/index.html#types) section of the docs needs to be updated.

File: src/formatting.rs [hunk 4]
```diff
@@ -73,6 +73,20 @@
...
 //! `locale` or `l`        | Locale to apply when formatting the time                                                                  | System locale
 //!
+//!
+//! ## `duration`/`dur` - Format durations
+//!
+//! Argument                     | Description                                                                                      |Default value
+//! -----------------------------|--------------------------------------------------------------------------------------------------|---------------------------------------------
+//! `hms`                        | Should the format be hours:minutes:seconds.milliseconds                                          | `false`
+//! `max_unit`                   | The largest unit to display the duration with                                                    | hms ? `h` : `y`
+//! `min_unit`                   | The smallest unit to display the duration with                                                   | `s`
```
MaxVerevkin, line 83: I think it would be very useful to have a list of all possible units. For example below the table, with a note "see below for the list of all possible units".

MaxVerevkin (approved):
Looks good :+1:

MaxVerevkin:
@bim9262 Is this ready to be merged?

bim9262:
I still wanted to add some tests

bim9262:
@MaxVerevkin I've added some tests :)

src/formatting/formatter/duration.rs
```diff
@@ -0,0 +1,540 @@
...
+    macro_rules! fmt {
+        ($name:ident) => {{
+            fmt!($name,)
+        }};
+        ($name:ident, $($key:ident : $value:tt),* $(,)?) => {
+            new_formatter(stringify!($name), &[
+                $( Arg { key: stringify!($key), val: stringify!($value) } ),*
+            ])
+        };
+    }
```
MaxVerevkin, line 266: `eng.rs` has a similar macro for tests, let's share them?

MaxVerevkin (commented):
This currently changes the way `$time` is rendered in battery block, because `hms` is not the default and the behavior is preserved only if explicit `.str()` is used. I think there are two things we can do:

- Make `Value::Duration` hold the default value for `hms`.
- Do not touch `time` and instead deprecate it and add `remaining`/`duration`/something.

bim9262:
> * Make `Value::Duration` hold the default value for `hms`.

Can you explain what you mean? Hold the string value like I did with `time`

> * Do not touch `time` and instead deprecate it and add `remaining`/`duration`/something.

Got it, I can add another value instead changing the type of time.

src/formatting/formatter/duration.rs
bim9262: Created #2054, will rebase after that's merged

MaxVerevkin:
> > * Make `Value::Duration` hold the default value for `hms`.
> 
> Can you explain what you mean? Hold the string value like I did with `time`

Something like `Value::duration(dur).with_default_hms()`, `DurationFormatter::hms` would need to be changed from `bool` to `Option<bool>`. Not sure if this is a good idea though.

bim9262:
> Something like `Value::duration(dur).with_default_hms()`, 

The format for `time` in the battery block isn't even the default hms format, it doesn't display seconds, just hh:mm.

>`DurationFormatter::hms` would need to be changed from `bool` to `Option<bool>`. Not sure if this is a good idea though.

I had thought about doing something like:

```rs
#[derive(Debug, Default)]
pub struct DurationFormatter(Option<DurationFormatterInner>);

#[derive(Debug, Default)]
pub struct DurationFormatterInner {
    hms: bool,
    max_unit_index: usize,
    min_unit_index: usize,
    units: usize,
    round_up: bool,
    unit_has_space: bool,
    pad_with: PadWith,
    show_leading_units_if_zero: bool,
}
 ```
but I dislike having different defaults for different blocks.

So I think I'll follow the approach: 

>Do not touch time and instead deprecate it and add remaining/duration/something.

**Comment by bim9262** [comment]
@MaxVerevkin anything else needed here or are we good to merge?

File: src/formatting.rs [hunk 6]
```diff
@@ -74,6 +76,30 @@
...
+//! -----------------------------|--------------------------------------------------------------------------------------------------|------------------------------------------------------
+//! `hms`                        | Should the format be hours:minutes:seconds.milliseconds                                          | `false`
+//! `max_unit`                   | The largest unit to display the duration with (see below for the list of all possible units)     | hms ? `h` : `y`
+//! `min_unit`                   | The smallest unit to display the duration with (see below for the list of all possible units)    | `s`
+//! `units`                      | The number of units to display                                                                   | min(# of units between `max_unit` and `min_unit``, 2)
+//! `round_up`                   | Round up to the nearest minimum displayed unit                                                   | `true`
+//! `unit_space`                 | Should there be a space between the value and unit symbol (not allowed when `hms:true`)                | `false`
```
MaxVerevkin, line 89:
```suggestion
//! `unit_space`                 | Should there be a space between the value and unit symbol (not allowed when `hms:true`)          | `false`
```

src/formatting.rs
```diff
@@ -74,6 +76,30 @@
...
+//! `pad_with`                   | The character that is used to pad the numbers                                                    | hms ? `0` : ` ` (a space)
+//! `show_leading_units_if_zero` | If fewer than `units` are non-zero should leading numbers that have a value of zero be shown     | `true`
```
MaxVerevkin, line 91: I think something shorter would be nice :) Maybe `leading_zeroes`?

MaxVerevkin: I think you forgot to update the docs.

bim9262:
@MaxVerevkin, I made the suggested changes

src/formatting.rs
bim9262: Ah, because I forgot to hit save before commiting

MaxVerevkin (approved):
LGTM! Again, sorry for late review.

bim9262:
> LGTM! Again, sorry for late review.

No worries, I've also been late on the reviews too...

Issue #2006, opened 2024-02-20
"v4l" driver for Privacy block  needs more documentation or a new function to update webcam devices


tkapias:
## Issue

I use the new [Privacy](https://greshake.github.io/i3status-rust/i3status_rs/blocks/privacy/) block, which is a really nice addition, **but at first I got an error from the `v4l` driver saying that the path `/dev/v4l/by-id` did not exist.**

I found out that it's because I don't use a usb webcam, but virtual devices created with v4l2loopback connecting to android devices or a software.
By default, there is no udev rules to create a symlink in `/dev/v4l/by-id` for virtual devices and so the driver will not find them.

## Solution

But I followed this [issue from v4l2loopback](https://github.com/umlaeute/v4l2loopback/issues/96) to **create a custom one, and then it works fine.**

I suggest to add this information in the documentation.

On another hand, I didn't have issues to discover those webcams with most other software, so I think that reading `/dev/v4l/by-id` is not the most common technique for listing devices. Perhaps another method should be written for `update_devices()` function in [v4l.rs](https://github.com/greshake/i3status-rust/blob/master/src/blocks/privacy/v4l.rs).

## Extra

As I use [Scrcpy](https://github.com/Genymobile/scrcpy) to connect my main webcam, I opened an [issue](https://github.com/Genymobile/scrcpy/issues/4689) with them for documentation.

bim9262:
Perhaps it would be better to just scan for `/dev/video*`... I'll have to play around with that.

bim9262:
@tkapias can you test out the linked pr?

PR #2009, opened 2024-02-23
Change device scan method for v4l privacy driver

bim9262:
@tkapias can you test this out with and without the workaround that you mentioned in #2006 in place?
@freswa can you test this out too?


freswa:
lgtm

MaxVerevkin (approved):

PR #2018, opened 2024-03-14
blocks/memory: add zram, zswap support

luteran42:
To make zswap work, you need at least Linux kernel version 5.19 or higher.

### DIFF
--- a/src/blocks/memory.rs
+++ b/src/blocks/memory.rs
@@ -141,7 +141,7 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
         // same logic as htop
         let used_diff = mem_free + buffers + pagecache + reclaimable;
         let mem_used = if mem_total >= used_diff {
-            mem_total - used_diff + shmem
+            mem_total - used_diff
         } else {
             mem_total - mem_free
         };
--- a/src/blocks/memory.rs
+++ b/src/blocks/memory.rs
@@ -34,6 +34,13 @@
 //! `swap_free_percents`      | as above but as a percentage of total memory                                    | Number | Percents
 //! `swap_used`               | Swap used                                                                       | Number | Bytes
 //! `swap_used_percents`      | as above but as a percentage of total memory                                    | Number | Percents
+//! `zram_compressed`         | Compressed zram memory usage                                                    | Number | Bytes
+//! `zram_decompressed`       | Decompressed zram memory usage                                                  | Number | Bytes
+//! 'zram_comp_ratio'         | Ratio of the decompressed/compressed zram memory                                | Number | -
+//! `zswap_compressed`        | Compressed zswap memory usage (>=Linux 5.19)                                    | Number | Bytes
+//! `zswap_decompressed`      | Decompressed zswap memory usage (>=Linux 5.19)                                  | Number | Bytes
+//! `zswap_decompressed_percents` | as above but as a percentage of total zswap memory  (>=Linux 5.19)          | Number | Percents
+//! 'zswap_comp_ratio'        | Ratio of the decompressed/compressed zswap memory (>=Linux 5.19)                | Number | -
 //!
 //! Action          | Description                               | Default button
 //! ----------------|-------------------------------------------|---------------
@@ -57,7 +64,7 @@
 
 use std::cmp::min;
 use std::str::FromStr;
-use tokio::fs::File;
+use tokio::fs::{read_dir, File};
 use tokio::io::{AsyncBufReadExt, BufReader};
 
 use super::prelude::*;
@@ -134,7 +141,7 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
         // same logic as htop
         let used_diff = mem_free + buffers + pagecache + reclaimable;
         let mem_used = if mem_total >= used_diff {
-            mem_total - used_diff
+            mem_total - used_diff + shmem
         } else {
             mem_total - mem_free
         };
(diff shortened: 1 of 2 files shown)


src/blocks/memory.rs
```diff
@@ -271,12 +313,50 @@ impl Memstate {
...
 
             line.clear();
         }
 
+        // For ZRAM
+        for i in 0.. {
+            let zram_file_path = format!("/sys/block/zram{}/mm_stat", i);
+            match File::open(&zram_file_path)
+                .await
+                .error("/sys/block/zram[0-9]/mm_stat does not exist")
```
bim9262, line 329: This error is not used since it's just thrown away, but instead of using the match, we can keep the code stucture flatter by doing this:

```rust
            let Ok(file) = File::open(&zram_file_path).await else {
                break;
            };
```

src/blocks/memory.rs
```diff
@@ -271,12 +313,50 @@ impl Memstate {
...
+            {
+                Ok(file) => {
+                    let mut zram_file = BufReader::new(file);
+                    let mut line = String::new();
```
bim9262, line 333: From https://www.kernel.org/doc/Documentation/blockdev/zram.txt:

```
File /sys/block/zram<id>/mm_stat

The stat file represents device's mm statistics. It consists of a single
line of text and contains the following stats separated by whitespace:
 orig_data_size   uncompressed size of data stored in this disk.
		  This excludes same-element-filled pages (same_pages) since
		  no memory is allocated for them.
                  Unit: bytes
 compr_data_size  compressed size of data stored in this disk
 mem_used_total   the amount of memory allocated for this disk. This
                  includes allocator fragmentation and metadata overhead,
                  allocated for this disk. So, allocator space efficiency
                  can be calculated using compr_data_size and this statistic.
                  Unit: bytes
 mem_limit        the maximum amount of memory ZRAM can use to store
                  the compressed data
 mem_used_max     the maximum amount of memory zram have consumed to
                  store the data
 same_pages       the number of same element filled pages written to this disk.
                  No memory is allocated for such pages.
 pages_compacted  the number of pages freed during compaction
 huge_pages	  the number of incompressible pages
 ```
Since there's only one line in the file then there's no need to iterate over the lines.

I'd probably also remove the error on trying to read the file, since it's possible it was hot unloaded between the open and actually reading the file

```rust
            if zram_file.read_to_string(&mut line).await.is_err() {
                continue;
            };
```

src/blocks/memory.rs
```diff
@@ -271,12 +313,47 @@ impl Memstate {
...
+                "Zswap:" => mem_state.zswap_compressed = val,
+                "Zswapped:" => mem_state.zswap_decompressed = val,
                 _ => (),
             }
 
             line.clear();
         }
 
+        // For ZRAM
+        if let Ok(entries) = std::fs::read_dir("/sys/block/") {
```
bim9262, line 325: This can be done with tokio like this:

```rust
        let mut entries = read_dir("/sys/block/")
            .await
            .error("Could not read /sys/block")?;
        while let Some(entry) = entries
            .next_entry()
            .await
            .error("Could not get next file /sys/block")?
        {
            if let Some(name) = entry.file_name().to_str() {
                if name.starts_with("zram") {
                    let Ok(file) = File::open(entry.path()).await else {
                        break;
                    };
                    ...
```

 I don't think it matters for `/sys` and `/dev`
 
 Not being able to read from `/sys` would be an error that should be raised.

bim9262: Whoops, good catch!

src/blocks/memory.rs
```diff
@@ -271,12 +313,52 @@ impl Memstate {
...
+        while let Some(entry) = entries
+            .next_entry()
+            .await
+            .error("Could not get next file /sys/block")?
+        {
+            if let Some(name) = entry.file_name().to_str() {
+                if name.starts_with("zram") {
+                    let zram_file_path = entry.path().join("mm_stat");
+                    let Ok(file) = File::open(zram_file_path).await else {
+                        break;
```
MaxVerevkin, line 337: Is the `break` instead of `continue` intentional?

src/blocks/memory.rs
```diff
@@ -271,12 +313,52 @@ impl Memstate {
...
+        // For ZRAM
+        let mut entries = read_dir("/sys/block/")
+            .await
+            .error("Could not read /sys/block")?;
+        while let Some(entry) = entries
+            .next_entry()
+            .await
+            .error("Could not get next file /sys/block")?
+        {
+            if let Some(name) = entry.file_name().to_str() {
```
MaxVerevkin, line 333: To make it slightly less nested: `let Some(name) = entry.file_name().to_str() else { continue };`

bim9262 (approved):
Besides for the change regarding shared memory `shmem` (which I'm just unsure of), everything looks good to me. @MaxVerevkin thoughts?

src/blocks/memory.rs
```diff
@@ -134,7 +141,7 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
         // same logic as htop
         let used_diff = mem_free + buffers + pagecache + reclaimable;
         let mem_used = if mem_total >= used_diff {
-            mem_total - used_diff
+            mem_total - used_diff + shmem
```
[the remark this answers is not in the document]
ammgws: According to https://github.com/greshake/i3status-rust/commit/3b750e618a3d6db8a75fce2f3979db0059905267 this was already meant to match htop

Is this a regression?

Issue #2003, opened 2024-02-17
Pomodoro: Alternative number format: arabic / decimal


dorsiflexion:
Can we have an alternative number format for the pomodoro count in the pomodoro block, please?

Currently, the pomodoro count looks like this, e.g. for 11 pomodors: `||||||||||`
This number format is pretty hard to count, because you need your mouse cursor to not miscount the bars (pipes or whatever they are). It would be so much easier and waste less space if it was just an arabic decimal number: `11`

I have a config option in mind – a separate boolean would probably be the simplest way. Doing it via format is likely not the way to go as it is applicable to this one block only.

Oh, and if you don't believe me yet: I bet you didn't notice that I wrote “11 pomodoros” and just put 10 bars there. Yeah, it's just not easily countable.

PS: Thanks for your work on this software. 💖

PR #2020, opened 2024-03-20
pomodoro: Overhaul block

bim9262:
Changes:

* Set block state based on pomodoro state
* Set icon based on pomodoro state
* Allow pomodos to be counted with arabic numeral or tally marks #2003 

The previous config will still be valid with the current config since there are new formats for the various states.

Ideally this will be merged after the format duration is added so that we don't have an additional inflection point in the config.

### DIFF
--- a/src/blocks/pomodoro.rs
+++ b/src/blocks/pomodoro.rs
@@ -15,16 +15,21 @@
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `format` | A string to customise the output of this block. | <code>\" $icon{ $message&vert;} \"</code>
+//! `format` | The format used when in idle, prompt, or notify states | <code>\" $icon{ $message\|} \"</code>
+//! `pomodoro_format` | The format used when the pomodoro is running or paused | <code>\" $icon $status_icon{ $completed_pomodoros.tally()\|} $time_remaining.duration(hms:true) \"</code>
+//! `break_format` |The format used when the pomodoro is during the break | <code>\" $icon $status_icon Break: $time_remaining.duration(hms:true) \"</code>
 //! `message` | Message when timer expires | `"Pomodoro over! Take a break!"`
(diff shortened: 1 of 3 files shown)


bim9262:
With #2032 we won't need to expose two values for the current pomodoro number.

aadilshabier:
Hey, is there any update on this PR? I wanted to the timers to the configuration, but saw that a PR was already in place.

bim9262:
Once #2032 is merged we can review this change @aadilshabier

src/blocks/pomodoro.rs
```diff
@@ -75,21 +84,70 @@ pub struct Config {
...
+        use PomodoroState::*;
+        match self {
+            Idle | PomodoroPaused => State::Idle,
+            Prompt => State::Warning,
+            Notify => State::Good,
+            Break | PomodoroRunning => State::Info,
+        }
+    }
+
+    fn get_status_icon(&self) -> Option<Cow<'static, str>> {
```
MaxVerevkin, line 107: Couldn't this be just an `&'static str` instead of cow?

bim9262: Fixed

bim9262:
@MaxVerevkin are we ready to merge this?

MaxVerevkin:
Since we are introducing separate `*_format` options, maybe we can replace `format` + `message` + `break_message` with `pomodoro_over_format` and `break_over_format` (names are just examples)?

bim9262:
`format` is used when there aren't any pomodoros are active, there's a prompt, or a pomodoro/break over message is being displaced.

I think we need to keep `format`, but `message`/`break_message` could be replaced with new formats for those states.

bim9262:
Although by adding removing the `message`/`break_message`, we don't have anything to send to `notify_cmd`

ammgws (approved):

ammgws:
I don't think we need to block this any longer

PR #2021, opened 2024-03-20
Escape " in code blocks, replace &vert; with \|

bim9262:
Escaping the `"`s in `<code>` blocks prevents the `"` from being rendered as `“` or `”` (left and right double quotation marks).

Standardized using `\|` instead of `&vert`


MaxVerevkin (approved):
I remember using `&vert;` because `\|` didn't work for some reason. Now it seems to work though. Maybe I tried it outside `<code></code>`...

PR #2025, opened 2024-03-26
pulseaudio,pipewire: use tokio::sync::Notify instead of channels, remove spinlock

MaxVerevkin:


MaxVerevkin:
By the way, is `wait_until_ready` even necessary? If I comment out the line `client.wait_until_ready().await;` nothing bad happens.

bim9262:
> By the way, is `wait_until_ready` (src/blocks/privacy/pipewire.rs:173) even necessary? If I comment out the line `client.wait_until_ready().await;` nothing bad happens.

This is there so that we don't spam that the block with updates as the nodes/links are loaded

src/blocks/privacy/pipewire.rs
```diff
@@ -151,20 +156,29 @@ impl Client {
...
 
     async fn wait_until_ready(&self) {
-        while !*self.ready.lock().unwrap() {
-            sleep(Duration::from_millis(1)).await;
+        // Call `.notified()` _before_ checking `ready` to avoid a race condition:
+        // 1. We check `ready` and it is `false`.
+        // 2. `ready` is set to `true` and subscribers are notified by another thread.
+        // 3. We subscribe but are never notified.
+        let notify = self.ready_notify.notified();
+        atomic::fence(atomic::Ordering::SeqCst);
```
bim9262, line 179: Can you explain why the fence is required?

My understanding is that the fence would allow for the load to have `Relaxed` ordering.

Also why `SeqCst` instead of `AcqRel`?

MaxVerevkin: > Can you explain why the fence is required?

To make sure the atomic load in `.notified()` happens before our `ready.load()`. That is if understand them correctly :) Oh, and I forgot a fence when notifying that roundtrip is done.

> My understanding is that the fence would allow for the load to have Relaxed ordering.
>
> Also why SeqCst instead Release on the store and Acquire on the load?

Admittedly I'm not very familiar with memory orderings, so I've just used the strongest one. And `.notified()` also uses `SeqCst`.

MaxVerevkin:
> This is there so that we don't spam that the block with updates as the nodes/links are loaded

I tried to test it, and without `wait_until_ready()` the block is notified just once. This is probably not deterministic, and it could happen that it is notified multiple times.

I think this can be implemented using an idle callback, which is enabled only when something happens. Then blocks are notified from the that callback, which then disables itself. I try implementing this a bit later.

src/blocks/privacy/pipewire.rs
MaxVerevkin: I removed this

src/blocks/privacy/pipewire.rs
```diff
@@ -97,75 +100,65 @@ impl Client {
...
             })
             .register();
 
-        main_loop.run();
-    }
-
-    fn add_event_listener(&self, tx: UnboundedSender<()>) {
-        self.event_listeners.lock().unwrap().push(tx);
+        loop {
+            main_loop.loop_().iterate(Duration::from_secs(60 * 60 * 24));
```
bim9262, line 145: Why duration of a day instead of no timeout?
https://pipewire.pages.freedesktop.org/pipewire-rs/pipewire/loop_/struct.Loop.html: 

pub fn [iterate](https://pipewire.pages.freedesktop.org/pipewire-rs/pipewire/loop_/struct.Loop.html#method.iterate)(&self, timeout: [Duration](https://doc.rust-lang.org/1.71.1/core/time/struct.Duration.html)) -> [i32](https://doc.rust-lang.org/1.71.1/std/primitive.i32.html)
Perform one iteration of the loop.

An optional timeout can be provided. 0 for no timeout, -1 for infinite timeout.

This function will block up to the provided timeout and then dispatch the fds with activity. The number of dispatched fds is returned.

src/blocks/privacy/pipewire.rs
```diff
@@ -97,75 +100,65 @@ impl Client {
...
+            if updated.get() {
```
bim9262, line 146: This works nicely :)

src/blocks/privacy/pipewire.rs
MaxVerevkin: https://gitlab.freedesktop.org/pipewire/pipewire-rs/-/issues/100

bim9262: My understanding is that setting 0 for the timeout is the optional timeout. I imagine it's done like this since there's no `Option` in C so it just takes an int.

MaxVerevkin: No timeout means "nonblocking", but we need "infinitely blocking", which is achieved in C by passing -1. `Duration` cannot represent negative time intervals.

PR #2027, opened 2024-03-27
Allow empty string as an arg value

bim9262:


src/formatting/formatter/eng.rs
```diff
@@ -91,10 +91,13 @@ impl EngFormatter {
                         .error("force_prefix must be true or false")?;
                 }
                 "pad_with" => {
-                    pad_with = arg
-                        .val
-                        .parse()
-                        .error("pad_with must be a single character")?;
+                    pad_with = if arg.val.is_empty() {
+                        '\u{200B}' // zero-width space
```
MaxVerevkin, line 95: I think turning `pad_with` into an `Option<char>` would be better and more explicit.

bim9262: How about keeping the value as a `Cow<'static, str>` or `String`, but checking that the length is <2?

That would allow us to not to need to put in another branch here:

```rust
                let mut retval = match self.width as i32 - digits {
                    i32::MIN..=0 => format!("{sign}{}", val.round()),
                    1 => format!("{}{sign}{}", self.pad_with, val.round() as i64),
                    rest => format!("{sign}{val:.*}", rest as usize - 1),
                };
```

src/formatting/formatter/eng.rs
```diff
@@ -91,10 +93,13 @@ impl EngFormatter {
                         .error("force_prefix must be true or false")?;
                 }
                 "pad_with" => {
-                    pad_with = arg
-                        .val
-                        .parse()
-                        .error("pad_with must be a single character")?;
+                    if arg.val.len() < 2 {
```
MaxVerevkin, line 96: We may want to count chars or even graphemes instead of bytes.

PR #2028, opened 2024-03-27
extend weather block to show sunset and sunrise

sMailund:
ref #715 
image

Opted to include into existing weather block instead of creating standalone block.
Implemented for both OpenWeatherMap and MetNo providers

### DIFF
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -82,7 +82,7 @@
 //! `wind{,_{favg,fmin,fmax,ffin}}`              | Wind speed                                                                    | Number   | -
 //! `wind_kmh{,_{favg,fmin,fmax,ffin}}`          | Wind speed. The wind speed in km/h                                            | Number   | -
 //! `direction{,_{favg,fmin,fmax,ffin}}`         | Wind direction, e.g. "NE"                                                     | Text     | -
-//! `sunrise`                                    | Time of sunrise                                                                | DateTime | -
+//! `sunrise`                                    | Time of sunrise                                                               | DateTime | -
 //! `sunset`                                     | Time of sunset                                                                | DateTime | -
 //!
 //! You can use the suffixes noted above to get the following:
@@ -145,7 +145,8 @@ use std::fmt;
 use std::sync::{Arc, Mutex};
 use std::time::Instant;
 
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, Datelike, Utc};
+use sunrise::{SolarDay, SolarEvent};
 
 use crate::formatting::Format;
 
@@ -183,7 +184,6 @@ trait WeatherProvider {
         &self,
         autolocated_location: Option<&Coordinates>,
         need_forecast: bool,
-        need_sunrise_and_sunset: bool,
     ) -> Result<WeatherResult>;
 }
 
@@ -271,8 +271,8 @@ struct WeatherResult {
     location: String,
     current_weather: WeatherMoment,
     forecast: Option<Forecast>,
-    sunrise: Option<DateTime<Utc>>,
-    sunset: Option<DateTime<Utc>>,
+    sunrise: DateTime<Utc>,
+    sunset: DateTime<Utc>,
 }
 
 impl WeatherResult {
@@ -289,6 +289,8 @@ impl WeatherResult {
             "wind" => Value::number(self.current_weather.wind),
             "wind_kmh" => Value::number(self.current_weather.wind_kmh),
             "direction" => Value::text(convert_wind_direction(self.current_weather.wind_direction).into()),
+            "sunrise" => Value::datetime(self.sunrise, None),
+            "sunset" => Value::datetime(self.sunset, None),
         };
 
         if let Some(forecast) = self.forecast {
@@ -320,13 +322,6 @@ impl WeatherResult {
             }
         }
 
-        if let Some(sunset) = self.sunset {
-            values.insert("sunset".into(), Value::datetime(sunset, None));
-        }
-
-        if let Some(sunrise) = self.sunrise {
-            values.insert("sunrise".into(), Value::datetime(sunrise, None));
-        }
         values
     }
 }
@@ -465,7 +460,6 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
 
     let autolocate_interval = config.autolocate_interval.unwrap_or(config.interval);
     let need_forecast = need_forecast(&format, format_alt.as_ref());
-    let need_sunrise_and_sunset = need_sunrise_and_sunset(&format, format_alt.as_ref());
 
     let mut timer = config.interval.timer();
 
@@ -477,8 +471,7 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
             None
         };
 
-        let fetch =
-            || provider.get_weather(location.as_ref(), need_forecast, need_sunrise_and_sunset);
+        let fetch = || provider.get_weather(location.as_ref(), need_forecast);
         let data = fetch.retry(ExponentialBuilder::default()).await?;
         let data_values = data.into_values();
 
@@ -527,11 +520,21 @@ fn need_forecast(format: &Format, format_alt: Option<&Format>) -> bool {
     has_forecast_key(format) || format_alt.is_some_and(has_forecast_key)
 }
 
-fn need_sunrise_and_sunset(format: &Format, format_alt: Option<&Format>) -> bool {
-    fn has_sunrise_or_sunset_keys(format: &Format) -> bool {
-        format.contains_key("sunrise") || format.contains_key("sunset")
-    }
-    has_sunrise_or_sunset_keys(format) || format_alt.is_some_and(has_sunrise_or_sunset_keys)
+fn calculate_sunrise_sunset(
+    lat: f64,
+    lon: f64,
+    altitude: Option<f64>,
+) -> Result<(DateTime<Utc>, DateTime<Utc>)> {
+    let date = Utc::now();
+    let solar_day = SolarDay::new(lat, lon, date.year(), date.month(), date.day())
+        .with_altitude(altitude.unwrap_or_default());
+
+    Ok((
+        DateTime::<Utc>::from_timestamp(solar_day.event_time(SolarEvent::Sunrise), 0)
+            .error("Unable to convert timestamp to DateTime")?,
+        DateTime::<Utc>::from_timestamp(solar_day.event_time(SolarEvent::Sunset), 0)
+            .error("Unable to convert timestamp to DateTime")?,
+    ))
 }
 
 #[derive(Debug, Deserialize, Clone, Copy, PartialEq, Eq, SmartDefault)]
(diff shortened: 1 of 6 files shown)


src/blocks/weather/open_weather_map.rs
```diff
@@ -386,14 +389,25 @@ impl WeatherProvider for Service<'_> {
             })
         };
 
+        let sunrise = unix_to_datetime(current_data.sys.sunrise);
```
bim9262, line 392: I'd recommend using (with nsecs = 0): 

pub const fn from_timestamp(secs: i64, nsecs: u32) -> Option<Self>

Makes a new DateTime<Utc> from the number of non-leap seconds since January 1, 1970 0:00:00 UTC (aka “UNIX timestamp”) and the number of nanoseconds since the last whole non-leap second.

src/blocks/weather.rs
```diff
@@ -236,6 +252,8 @@ struct WeatherResult {
     location: String,
     current_weather: WeatherMoment,
     forecast: Option<Forecast>,
+    sunrise: DateTime<Utc>,
```
bim9262, line 255: Let's make sunrise/sunset optional so that if they aren't requested in the format string we don't have to fetch them

src/blocks/weather/met_no.rs
```diff
@@ -283,11 +300,41 @@ impl WeatherProvider for Service<'_> {
             })
         };
 
+        let current_date = chrono::Local::now().format("%Y-%m-%d").to_string();
+
+        let sun_query_string: HashMap<&str, String> = map! {
+            "lat" => &lat,
+            "lon" => &lon,
+            "date" => & current_date
```
bim9262, line 308: offset should be passed too, otherwise the timezone is assumed to be UTC

src/blocks/weather/open_weather_map.rs
```diff
@@ -386,10 +388,15 @@ impl WeatherProvider for Service<'_> {
             })
         };
 
+        let sunrise = Some(DateTime::<Utc>::from_timestamp(current_data.sys.sunrise, 0).unwrap());
```
bim9262, line 391: You don't need to unwrap the option to then wrap it in a Some.

src/blocks/weather/met_no.rs
```diff
@@ -160,8 +178,8 @@ impl WeatherProvider for Service<'_> {
             .error("No location given")?;
 
         let querystr: HashMap<&str, String> = map! {
-            "lat" => lat,
-            "lon" => lon,
+            "lat" => &lat,
```
bim9262, line 181: https://api.met.no/weatherapi/sunrise/3.0/documentation

>* **offset** (optional), timezone offset, on the format +HH:MM or -HH:MM
[Timezone offset](https://api.met.no/weatherapi/sunrise/3.0/documentation#___top)
The optional offset parameter is the difference between local time and UTC, as defined in ISO 8601. This is used mainly for returning timestamps in local time, but is also used in some instances to calculate the correct date period. If omitted, all times will be given in UTC.
>
>Presently timezones for the offset parameter range from -12:00 to +15:00, although this may change in the future. For GMT, only +00:00 is allowed. You can see a list of all current timezones on https://en.wikipedia.org/wiki/List_of_UTC_time_offsets.

src/blocks/weather/open_weather_map.rs
MaxVerevkin: Well, this does ensure that it is `Some`... But it would be much better to use `.error("...")?`, of course.

src/blocks/weather/met_no.rs
sMailund: I am not sure if this is necessary. After some testing, it seems like the status bar correctly renders the time in the correct timezone out of the box. It makes more sense for me to receive the sunset and sunrise times in GMT when it will be adjusted to the correct timezone when displayed anyways. What do you think, is it OK as it is?

image

image

(sorry late reply, been very busy lately)

bim9262: No worries. It looks like when the date is omitted I am getting the sunrise/sunset for tomorrow instead of today. You're right that it doesn't seem to matter if the offset is included. 

The example I was using is https://api.met.no/weatherapi/sunrise/3.0/sun?lat=40.43&lon=-74.01&date=2024-05-13 (on 2024-05-13) I see the times for 2024-05-14. I think that the default date is the date in GMT

src/blocks/weather.rs
```diff
@@ -290,6 +309,18 @@ impl WeatherResult {
 
             });
         }
+
+        if let Some(sunset) = self.sunset {
+            values.extend(map! {
```
bim9262, line 314: There's a new way to extend the map without creating a new temporary HashMap.

```rs
map! { @extend values
    "sunset" => Value::datetime(sunset, None)
}
```

or since you are just adding a single value you can use insert:

```rs
values.insert("sunset".into(), Value::datetime(sunset, None));
```

As things are right now both of these extends are missing a `;` which is causing the pre-commit checks to fail.

src/blocks/weather/open_weather_map.rs
```diff
@@ -388,8 +388,8 @@ impl WeatherProvider for Service<'_> {
             })
         };
 
-        let sunrise = Some(DateTime::<Utc>::from_timestamp(current_data.sys.sunrise, 0).unwrap());
-        let sunset = Some(DateTime::<Utc>::from_timestamp(current_data.sys.sunset, 0).unwrap());
+        let sunrise = DateTime::<Utc>::from_timestamp(current_data.sys.sunrise, 0);
```
bim9262, line 391: @MaxVerevkin suggested doing this, so that if there is an error it will be surfaced.

```rs
let sunrise = Some(DateTime::<Utc>::from_timestamp(current_data.sys.sunrise, 0).error("Unable to convert timestamp to DateTime")?);
```

src/blocks/weather/met_no.rs
```diff
@@ -283,11 +301,51 @@ impl WeatherProvider for Service<'_> {
             })
         };
 
+        let (sunset, sunrise) = match need_sunrise_and_sunset {
+            true => {
+                let current_date = chrono::Local::now().format("%Y-%m-%d").to_string();
```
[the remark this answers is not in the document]
sMailund: I'm guessing using the default value provided by met.no is probably the best bet instead of trying to set it myself as I did previously:
image

bim9262:
I did a little POC to get the date based on the lat/long so that we can pass that to metno.

Add `tz-search` as a dependency

```rust
use chrono::Utc;
use chrono_tz::Tz;

fn main() {
    let (lat, long) = (30., -97.);
    let tz: Tz = tz_search::lookup(lat, long).unwrap().parse().unwrap();
    let date = Utc::now().with_timezone(&tz).format("%Y-%m-%d");
    println!("{}", date);
}
```

Obviously we'd want to swap out those `unwrap`s for `error("")?`s

src/blocks/weather/met_no.rs
```diff
@@ -283,11 +301,48 @@ impl WeatherProvider for Service<'_> {
...
+                let sun_data: SunResponse = REQWEST_CLIENT
+                    .get(SUN_URL)
+                    .query(&sun_query_string)
+                    .header(reqwest::header::CONTENT_TYPE, "application/json")
+                    .send()
+                    .await
+                    .error("Forecast request failed")?
+                    .json()
+                    .await
+                    .error("Forecast request failed")?;
```
MaxVerevkin, line 320: I wonder how far off  the results are to what we can compute ourselves (https://docs.rs/sunrise/)

bim9262: https://api.met.no/weatherapi/sunrise/3.0/sun?lat=40.43&lon=-74.01&date=2024-09-07

```json
{
  "copyright": "MET Norway",
  "licenseURL": "https://api.met.no/license_data.html",
  "type": "Feature",
  "geometry": {
    "type": "Point",
    "coordinates": [-74, 40.4]
  },
  "when": {
    "interval": [
      "2024-09-07T04:53:00Z",
      "2024-09-08T04:56:00Z"
    ]
  },
  "properties": {
    "body": "Sun",
    "sunrise": {
      "time": "2024-09-07T10:29+00:00",
      "azimuth": 81.67
    },
    "sunset": {
      "time": "2024-09-07T23:17+00:00",
      "azimuth": 278.07
    },
    "solarnoon": {
      "time": "2024-09-07T16:53+00:00",
      "disc_centre_elevation": 55.29,
      "visible": true
    },
    "solarmidnight": {
      "time": "2024-09-07T04:53+00:00",
      "disc_centre_elevation": -43.72,
      "visible": false
    }
  }
}
```

```rust
use chrono::{Datelike, TimeZone, Utc};
use chrono_tz::Tz;
use sunrise::sunrise_sunset;

fn main() {
    let (lat, long) = (40.43, -74.01);
    let tz: Tz = tz_search::lookup(lat, long).unwrap().parse().unwrap();
    let date = Utc::now().with_timezone(&tz);
    let (sunrise, sunset) = sunrise_sunset(lat, long, date.year(), date.month(), date.day());
    println!(
        "{}\n{}\n{}\n{}",
        date.format("%Y-%m-%d"),
        tz,
        Utc.timestamp_opt(sunrise, 0).unwrap(),
        Utc.timestamp_opt(sunset, 0).unwrap()
    );
}
```

```
2024-09-07
America/New_York
2024-09-07 10:29:42 UTC
2024-09-07 23:18:23 UTC
```

The results are quite close (for the one random lat/long that I tested)!

MaxVerevkin: So we can save an API call _and_ it is simpler? Nice

bim9262:
@sMailund would you like to continue work on this? If not I can take it over.

sMailund:
@bim9262 been meaning to get back to this for a while, but not able to get the time for it! Feel free to take over if you want to!

bim9262:
@sMailund, I squashed your commits, and rebased them onto master + added one new commit to add the sunset to the nws provider + rearrange a bit for the rebase. Feel free to check it out/test

src/blocks/weather.rs
```diff
@@ -499,6 +521,22 @@ fn need_forecast(format: &Format, format_alt: Option<&Format>) -> bool {
     has_forecast_key(format) || format_alt.is_some_and(has_forecast_key)
 }
 
+fn calculate_sunrise_sunset(lat: f64, lon: f64) -> Result<(DateTime<Utc>, DateTime<Utc>)> {
+    let tz: Tz = tz_search::lookup(lat, lon)
+        .or_error(|| format!("Unable to find tz for lat/long {}, {}", lat, lon))?
+        .parse()
+        .error("Unable to parse tz")?;
+    let date = Utc::now().with_timezone(&tz);
+    let (sunrise, sunset) = sunrise_sunset(lat, lon, date.year(), date.month(), date.day());
```
MaxVerevkin, line 530: Is it really necessary to perform a timezone lookup? Firstly, are you sure that year/month/day are supposed to be in local time (it is not documented unfortunately)? If it is, is the difference between adjacent days that noticeable?

bim9262: Actually, I suspect your right since the value is being passed into `Utc.with_ymd_and_hms` here.
Let me change that to use the current Utc date.

PR #2031, opened 2024-03-30
privacy(pipewire): simplify notify logic

MaxVerevkin:
This works because the info from the driver is fetched on each drawing iteration anyway, so notifying only waiting blocks is enough.


bim9262:
This doesn't appear to work when I test out the block.

MaxVerevkin:
It worked in my limited testing, but now I see that it will not work if update happens after the fetch but before the call to `wait_for_change`. Eh :(

MaxVerevkin:
Oh, I have an idea how to fix it.

PR #2032, opened 2024-03-31
Add tally formatter

bim9262:
I also considered calling this the rods formatter since the counting rods aren't really tally marks... Thoughts?

### DIFF
--- a/src/formatting/formatter/tally.rs
+++ b/src/formatting/formatter/tally.rs
@@ -1,0 +1,582 @@
+use std::str::FromStr;
+
+use crate::formatting::unit::Unit;
+
+use super::*;
+
+#[derive(Debug)]
+enum Style {
+    ChineseCountingRods,
+    ChineseTally,
+    WesternTally,
+    WesternTallyUngrouped,
+}
+
+impl FromStr for Style {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self> {
+        match s {
+            "chinese_counting_rods" | "ccr" => Ok(Style::ChineseCountingRods),
+            "chinese_tally" | "ct" => Ok(Style::ChineseTally),
+            "western_tally" | "wt" => Ok(Style::WesternTally),
+            "western_tally_ungrouped" | "wtu" => Ok(Style::WesternTallyUngrouped),
+            x => Err(Error::new(format!("Unknown Style: '{x}'"))),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct TallyFormatter {
+    style: Style,
+}
+
+impl TallyFormatter {
+    pub(super) fn from_args(args: &[Arg]) -> Result<Self> {
+        let mut style = Style::WesternTally;
+        for arg in args {
+            match arg.key {
+                "style" | "s" => {
+                    style = arg.val.parse()?;
+                }
+                other => {
+                    return Err(Error::new(format!(
+                        "Unknown argument for 'tally': '{other}'"
+                    )));
+                }
+            }
+        }
+        Ok(Self { style })
+    }
+}
+
+const HORIZONTAL_CHINESE_COUNTING_RODS_CHARS: [char; 10] =
+    ['〇', '𝍠', '𝍡', '𝍢', '𝍣', '𝍤', '𝍥', '𝍦', '𝍧', '𝍨'];
+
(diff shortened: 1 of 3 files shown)


src/formatting/formatter/tally.rs
```diff
@@ -0,0 +1,583 @@
...
+
+const VERTICAL_CHINESE_COUNTING_RODS_CHARS: [char; 10] =
+    ['〇', '𝍩', '𝍪', '𝍫', '𝍬', '𝍭', '𝍮', '𝍯', '𝍰', '𝍱'];
+
+const CHINESE_TALLY_CHARS: [char; 5] = ['𝍲', '𝍳', '𝍴', '𝍵', '𝍶'];
+
+impl Formatter for TallyFormatter {
+    fn format(&self, val: &Value, _config: &SharedConfig) -> Result<String, FormatError> {
+        match val {
+            Value::Number { val, .. } => {
```
bim9262, line 62: Should we match to unit None or support unit conversions?

src/formatting/formatter/tally.rs
```diff
@@ -0,0 +1,583 @@
...
+    WesternTally,
+    WesternTallyUngrouped,
+}
+
+impl FromStr for Style {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self> {
+        match s {
+            "chinese_counting_rods" => Ok(Style::ChineseCountingRods),
```
bim9262, line 18: Maybe some short names should be chosen too.

bim9262: How about abbreviating them to `ccr`, `ct`, `wt`, and `wtu`?

MaxVerevkin: Yeah, sure.

bim9262:
Will rebase after #2054 is merged.

src/formatting/formatter/tally.rs
```diff
@@ -0,0 +1,575 @@
...
+                        for _ in 0..val {
+                            result.push('𝍷');
+                        }
+                    }
+                }
+                Ok(result)
+            }
+            other => Err(FormatError::IncompatibleFormatter {
+                ty: other.type_name(),
+                fmt: "rods",
```
bim9262, line 125: Update to format name...

src/formatting/formatter/tally.rs
bim9262: @MaxVerevkin what do you think about the units? If we do support units should we just add a argument `repr`|`representation` defaulting to "arabic" with these additional formats?

MaxVerevkin: I think it makes sense to only support `None` unit. Then if someone provides a use-case for supporting other units, we can do something about it.

bim9262:
@MaxVerevkin are we happy with this change too?
I'd like to finish this up then follow up on #2020

MaxVerevkin:
Sorry for such a late reply! Everything looks good to me, I think. It would be nice to also add [this type of tally marks](https://upload.wikimedia.org/wikipedia/commons/d/d8/Tally_marks_2.svg), but this is not a blocker and can be implemented in any time in the future.

MaxVerevkin (approved):

bim9262:
Thanks for the review! @MaxVerevkin I haven't seen that style of talky marks, what are they called? Do they have Unicode characters?

MaxVerevkin:
[The wiki page](https://en.wikipedia.org/wiki/Tally_marks) describes them as

> Tally marks used in [France](https://en.wikipedia.org/wiki/France), [Portugal](https://en.wikipedia.org/wiki/Portugal), [Spain](https://en.wikipedia.org/wiki/Spain), and their former colonies, including [Latin America](https://en.wikipedia.org/wiki/Latin_America). 1 to 5 and so on. These are most commonly used for registering scores in [card games](https://en.wikipedia.org/wiki/Card_game), like [Truco](https://en.wikipedia.org/wiki/Truco).

I can add that it is sort of used in Russia, not popular but people understand it.

---

> Do they have Unicode characters?

Good question. Apparently not :( Even though [the unicode proposal for tally symbols](https://www.unicode.org/L2/L2016/16065-tally-marks.pdf) has a mention of them, where they are referred to as "a South American tally scheme".

PR #2033, opened 2024-04-02
Fix device_id in kdeconnect block

ammgws:
With 2 devices connected to `kdeconnectd` and specifying `device_id` in i3rs config, it currently does the opposite of what is expected and ignores the `device_id` when looking for devices :laughing:


MaxVerevkin:
I confuse the meaning of the predicate in `retain` too often xD

PR #2040, opened 2024-04-07
map! macro: add @extend mode to extend an existing map

MaxVerevkin:
This avoids creating a temporary hashmap.


bim9262:
Looks good! I was surprised by how few blocks actually extended a map.

PR #2041, opened 2024-04-10
blocks/sound: add format_alt

### LINKED ISSUE #2036: Implement format_alt for sound block
I was looking at how to show the audio out/in device, but thought that would be more interesting to have the sound block be able to alternate between volume level/scroll and device used.

For that, we need a format_alt for the audio block.

heitorPB:
Fix #2036


bim9262:
Looks good, thanks :)

PR #2043, opened 2024-04-13
formatting(eng): add `range` parameter

### LINKED ISSUE #1957: feature_request(memory): treat `swap_used` zero value as `false` in conditions
## Problem

[Formatting doc](https://docs.rs/i3status-rs/latest/i3status_rs/formatting/index.html#handling-missing-placeholders-and-incorrect-types) says (if I understand it correctly) that we can write conditions like `Percentage: $percentage|nothing here`.

I would like to make `memory` block hidden when `swap_used` is zero, so I'm trying `format = "$swap_used|nothing"` but it shows "0B".

## Possible solutions

1. Maybe hack `swap_used` so zero value will fallback to category "missing placeholder"?
2. Add formatter option for type Number, like `ignore_zero` or `ignore_threshold`. This is more general approach.

MaxVerevkin:
Closes #1957

## Example

Show the cpu block only if utilization is above 10%:

```toml
[[block]]
block = "cpu"
format = " $icon $utilization.eng(range:10..) |"
```

## Unresolved Questions

Support disjoint ranges? Would it be useful? How? Maybe by allowing multiple `range` arguments.

## Todo

- [ ] Add examles to the blocks that would benefit from it.
- [x] Support half-ranges. Like `10..`.

### DIFF
--- a/src/blocks/disk_space.rs
+++ b/src/blocks/disk_space.rs
@@ -27,7 +27,7 @@
 //! ----------------|-------------------------------------------|---------------
 //! `toggle_format` | Toggles between `format` and `format_alt` | Left
 //!
-//! # Example
+//! # Examples
 //!
 //! ```toml
 //! [[block]]
@@ -48,6 +48,14 @@
 //! [[block.click]]
 //! button = "right"
 //! update = true
+//! ```
+//!
+//! Show the block only if less than 10GB is available:
+//!
+//! ```toml
+//! [[block]]
+//! block = "disk_space"
+//! format = " $free.eng(range:..10e9) |"
 //! ```
 //!
 //! # Icons Used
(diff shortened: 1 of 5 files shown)


bim9262:
For the partial ranges we can check to see if start/stop are empty strings and if so use f64's min/max values

bim9262:
I think `range(..25), range(75..)` or `range(..25,75..)` would be ok. I think I prefer the two range parameters as it's easier to read. I also agree that could be added in another PR

MaxVerevkin:
> Show the block only if less than 10GB is available:
>
> ```toml
> [[block]]
> block = "disk_space"
> format = " $free.eng(range:..10e9) |"
> ```

Uhh. Do we want to support prefixes in ranges? Like `range:..10Gi` and `range:..10M`.

bim9262:
> > Show the block only if less than 10GB is available:
> >
> > ```toml
> > [[block]]
> > block = "disk_space"
> > format = " $free.eng(range:..10e9) |"
> > ```
> 
> Uhh. Do we want to support prefixes in ranges? Like `range:..10Gi` and `range:..10M`.

That seems intuitive, could be done in another PR though as it wouldn't be breaking change

PR #2044, opened 2024-04-14
Support x:<name> syntax to fetch color definition from .Xresources

jkorinth:
Adds support to retrieve simple color definitions from `~/.Xresources` with a new prefix in the color definitions in theme overrides, e.g.,
```
[theme.overrides]
idle-bg = "x:background"
```
would look for a line like
```
*background: #feedda
```
in `~/.Xresources`.

### DIFF
--- a/src/themes/color.rs
+++ b/src/themes/color.rs
@@ -216,6 +216,12 @@ impl FromStr for Color {
             let v = components.next().or_error(err_msg)??;
             let a = components.next().unwrap_or(Ok(100.))?;
             Color::Hsva(Hsva::new(h, s / 100., v / 100., (a / 100. * 255.) as u8))
+        } else if color.starts_with("x:") {
+            let name = color.split_at(2).1;
+            super::xresources::get_color(name)?
+                .or_error(|| format!("color '{name}' not defined in ~/.Xresources"))?
+                .parse()
+                .or_error(|| format!("invalid color definition '{name}'"))?
         } else {
             let err_msg = || format!("'{color}' is not a valid RGBA color");
             let rgb = color.get(1..7).or_error(err_msg)?;
--- a/src/themes/xresources.rs
+++ b/src/themes/xresources.rs
@@ -1,0 +1,95 @@
+use log::debug;
+use regex::Regex;
+
+use std::collections::HashMap;
+use std::sync::LazyLock;
+
+use crate::errors::*;
+
+#[cfg(not(test))]
+use std::{env, path::PathBuf};
+
+#[cfg(not(test))]
+fn read_xresources() -> std::io::Result<String> {
+    use std::io::{Error, ErrorKind};
+    let home =
+        env::var("HOME").map_err(|_| Error::new(ErrorKind::Other, "HOME env var was not set"))?;
+    let xresources = PathBuf::from(home + "/.Xresources");
+    debug!(".Xresources @ {:?}", xresources);
+    std::fs::read_to_string(xresources)
+}
+
+#[cfg(test)]
+use tests::read_xresources;
+
+static COLOR_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"^\s*\*(?<name>[^: ]+)\s*:\s*(?<color>#[a-f0-9]{6,8}).*$").unwrap()
+});
+
+static COLORS: LazyLock<Result<HashMap<String, String>, Error>> = LazyLock::new(|| {
+    let content = read_xresources().error("could not read .Xresources")?;
+    debug!(".Xresources content:\n{}", content);
+    Ok(HashMap::from_iter(content.lines().filter_map(|line| {
+        COLOR_REGEX
+            .captures(line)
+            .map(|caps| (caps["name"].to_string(), caps["color"].to_string()))
+    })))
+});
+
+pub fn get_color(name: &str) -> Result<Option<&String>, Error> {
+    COLORS
+        .as_ref()
+        .map(|cmap| cmap.get(name))
+        .map_err(Clone::clone)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::Result;
+
+    #[allow(clippy::unnecessary_wraps)]
+    pub(crate) fn read_xresources() -> Result<String> {
+        static XRESOURCES: &str = "\
+        ! this is a comment\n\
+        \n\
+        *color4 : #feedda\n\
+    \n\
+        *background: #ee33aa99\n\
+        ";
+        Ok(XRESOURCES.to_string())
+    }
+
+    #[test]
+    fn test_reading_colors() {
+        let colors = COLORS.as_ref().unwrap();
+        assert_eq!(colors.get("color4"), Some(&"#feedda".to_string()));
+        assert_eq!(colors.get("background"), Some(&"#ee33aa99".to_string()));
+        assert_eq!(2, colors.len());
+    }
+
+    #[test]
+    fn test_deserializing_xcolors() {
+        use super::super::color::*;
+        let mut parsed_color = "x:color4".parse::<Color>().unwrap();
+        assert_eq!(
+            parsed_color,
+            Color::Rgba(Rgba {
+                r: 254,
+                g: 237,
+                b: 218,
+                a: 255
+            })
+        );
+        parsed_color = "x:background".parse::<Color>().unwrap();
+        assert_eq!(
+            parsed_color,
+            Color::Rgba(Rgba {
+                r: 238,
+                g: 51,
+                b: 170,
+                a: 153,
+            })
+        );
+    }
+}
--- a/cspell.yaml
+++ b/cspell.yaml
@@ -31,9 +31,11 @@ words:
   - bugz
   - busctl
   - caldav
+  - ccache
   - chrono
   - clippy
   - CLOEXEC
+  - cmap
   - conn
   - consts
   - conv
@@ -182,15 +184,16 @@ words:
   - wofi
   - wttr
   - xclip
+  - xcolors
   - xesam
   - xkbswitch
   - XKCD
   - xrandr
+  - xresources
   - xtask
   - zbus
+  - zram
+  - zswap
+  - Zswapped
   - zvariant
   - zwlr
-  - zswap
-  - zram
-  - Zswapped
-  - ccache
--- a/doc/themes.md
+++ b/doc/themes.md
@@ -124,7 +124,11 @@ cpu_boost_off = "OFF"
 
 # Available theme overrides
 
-All `bg` and `fg` overrides are html hex color codes like `#000000` or `#789ABC`. A fourth byte for alpha (like `#acbdef42`) works on some systems. `00` is transparent, `FF` is opaque.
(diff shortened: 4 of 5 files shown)


MaxVerevkin:
This looks simple enough, so why not.

This feature should be documented [here](https://github.com/greshake/i3status-rust/blob/master/doc/themes.md#available-theme-overrides).

To make testing easier, I would prefer to have a dedicated function called `extract_xresources_colors(content: &str) -> HashMap<String, String>`. Then in tests, you will not have to mock `read_xresources` function, but instead just `assert!` the contents of the returned hashmap.

src/themes/xresources.rs
```diff
@@ -0,0 +1,71 @@
+#![allow(dead_code)]
+use log::*;
+use once_cell::sync::Lazy;
+use regex::Regex;
+use std::{collections::HashMap, env, path::PathBuf};
+
+#[cfg(not(test))]
+fn read_xresources() -> std::io::Result<String> {
+    let home = env::var("HOME").expect("HOME env var was not set?!");
```
MaxVerevkin, line 9: Please return an error instead of panicking in case HOME is not set.

src/themes/xresources.rs
```diff
@@ -0,0 +1,71 @@
...
+
+pub static COLORS: Lazy<HashMap<String, String>> = Lazy::new(|| {
+    let home = env::var("HOME").expect("HOME env var was not set?!");
+    let xresources = PathBuf::from(home + "/.Xresources");
+    debug!(".Xresources @ {:?}", xresources);
+    if let Ok(content) = read_xresources() {
+        debug!(".Xresources content:\n{}", content);
+        return HashMap::from_iter(
+            content
+                .split("\n")
```
MaxVerevkin, line 29:
```suggestion
                .lines()
```

jkorinth:
Updated with the suggestions. There's no documentation for the 'hsv:` prefix, but I think that should go into a separate PR.

src/themes/xresources.rs
```diff
@@ -0,0 +1,75 @@
...
+
+#[cfg(test)]
+use tests::read_xresources;
+
+static COLOR_REGEX: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"^\s*\*(?<name>[^: ]+)\s*:\s*(?<color>#[a-f0-9]{6}).*$").unwrap());
+
+pub static COLORS: Lazy<Result<HashMap<String, String>, env::VarError>> = Lazy::new(|| {
+    let home = env::var("HOME")?;
+    let xresources = PathBuf::from(home + "/.Xresources");
```
MaxVerevkin, line 23: It seems that this variable is not used outside of debugging.

src/themes/xresources.rs
```diff
@@ -0,0 +1,75 @@
...
+                .map(|line| {
+                    COLOR_REGEX
+                        .captures(line)
+                        .map(|caps| (caps["name"].to_string(), caps["color"].to_string()))
+                })
+                .flatten(),
+        ));
+    }
+    warn!(".Xresources not found");
+    Ok(HashMap::new())
```
MaxVerevkin, line 39: I think it is better to just return an error from this closure if reading fails. This way the actual error message would be displayed in the bar, instead of "color '{name}' is not defined in .Xresources", which may become hard to debug.

jkorinth: Yes, I see your point. Have a look at the new version, is this what you meant?

jkorinth:
Btw: `errors` deserves an overhaul, too, I think. Check out `anyhow` and `thiserror`, they eliminate most of the boiler plate.

MaxVerevkin:
> Btw: `errors` deserves an overhaul, too, I think. Check out `anyhow` and `thiserror`, they eliminate most of the boiler plate.

I know about these crates and we use `thiserror` already, in same cases. Creating an error enum for each block does not seem practical. We are not using `anyhow` because in the past `Error` struct used to contain a `block: String` field. It no longer does (we now have a separate `BlockError` for this, which uses `thiserror`) and I think we can switch to `anyhow` now.

src/themes/xresources.rs
MaxVerevkin: Yes, basically. Although I think having the error type be `Error` would be better than `String`. You will just have to write `.map_err(Clone::clone)` when accessing it.

src/themes/xresources.rs
```diff
@@ -0,0 +1,81 @@
...
+        assert!(colors.contains_key("color4"));
+        assert!(colors.contains_key("background"));
+        assert_eq!(2, colors.len());
+    }
+
+    #[test]
+    fn test_deserializing_xcolors() {
+        use super::super::color::*;
+        println!("COLORS = {:?}", COLORS.as_ref().unwrap().keys());
+        let _: Color = "x:background".parse().expect("can parse background");
```
MaxVerevkin, line 79: Again, the parsed value should be `assert_eq!`ed, and there is no reason to print to stdout.

src/themes/xresources.rs
```diff
@@ -0,0 +1,103 @@
+#![allow(dead_code)]
```
MaxVerevkin, line 1:
```suggestion
```

src/themes/xresources.rs
```diff
@@ -0,0 +1,103 @@
...
+        let colors = COLORS.as_ref().unwrap();
+        assert_eq!(colors.get("color4"), Some(&"#feedda".to_string()));
+        assert_eq!(colors.get("background"), Some(&"#ee33aa".to_string()));
+        assert_eq!(2, colors.len());
+    }
+
+    #[test]
+    fn test_deserializing_xcolors() {
+        use super::super::color::*;
+        println!("COLORS = {:?}", COLORS.as_ref().unwrap().keys());
```
MaxVerevkin, line 75:
```suggestion
```

src/themes/xresources.rs
```diff
@@ -0,0 +1,103 @@
...
+        let mut parsed_color = "x:color4".parse::<Color>();
+        assert!(parsed_color.is_ok());
```
MaxVerevkin, line 77: In tests it's better to `unwrap()`: it is shorter and will print the why parsing failed if something goes wrong.

src/themes/xresources.rs
jkorinth: Ah! Yes, this looks much cleaner, tbh.

src/themes/color.rs
```diff
@@ -216,6 +216,21 @@ impl FromStr for Color {
             let v = components.next().or_error(err_msg)??;
             let a = components.next().unwrap_or(Ok(100.))?;
             Color::Hsva(Hsva::new(h, s / 100., v / 100., (a / 100. * 255.) as u8))
+        } else if color.starts_with("x:") {
+            let name = color.split_at(2).1;
+            let hex = super::xresources::get_color(name)?
+                .ok_or_else(|| {
+                    Self::Err::new(format!("color {} not defined in ~/.Xresources", name))
+                })?;
```
MaxVerevkin, line 224:
```suggestion
            let hex = super::xresources::get_color(name)?
                .or_error(|| format!("color {name} not defined in ~/.Xresources"))?;
```

src/themes/color.rs
```diff
@@ -216,6 +216,21 @@ impl FromStr for Color {
...
+            let err_msg = || format!("'{name}' def '{hex}' cannot be parsed as RGB");
+            let rgb = hex
+                .get(1..7)
+                .map(|rgb| u32::from_str_radix(rgb, 16).ok())
+                .flatten();
+            let a = u32::from_str_radix(hex.get(7..9).unwrap_or("FF"), 16).ok();
+            rgb.map(|rgb| a.map(|a| Color::Rgba(Rgba::from_hex((rgb << 8) + a))))
+                .flatten()
+                .or_error(err_msg)?
```
MaxVerevkin, line 233: Can't we just return `hex.parse()`?

MaxVerevkin: Or `hex.parse().or_error(|| format!("{name} definition is invalid"))?`

jkorinth:
> LGTM. Just needs the CI to be passing.

How do you tell cspell to ignore words?

MaxVerevkin:
> How do you tell cspell to ignore words?

Edit `cspell.yaml`

src/themes/color.rs
MaxVerevkin: But `hex.parse()` would work with alpha too?

jkorinth: Yea, but not without. 😏

MaxVerevkin: I don't get it. Could you show an example of .Xresources for which it wouldn't work?

jkorinth: `#ffeeddaa` parses as R: 255, G: 238, B: 221, A: 170.
`#ffeedd` parses as R:0, G: 255, B:238, A: 221.

MaxVerevkin: Wait, your regex strips the alpha.

jkorinth:
> > How do you tell cspell to ignore words?
> 
> 
> 
> Edit `cspell.yaml`

Wait - cspell spellchecks not only the docs, but also the code itself?! And not just strings inside, but the full code? 😂

src/themes/color.rs
jkorinth: Oh, true.So, I could add an alpha group and default it to ff. Then `hex.parse()` would work, right?

MaxVerevkin: It already works. I don't know how useful alpha channel in .Xresources is, so IDK.

MaxVerevkin:
> > > How do you tell cspell to ignore words?
> > 
> > 
> > Edit `cspell.yaml`
> 
> Wait - cspell spellchecks not only the docs, but also the code itself?! And not just strings inside, but the full code? 😂

Yea, it is sometimes annoying, but now we have less typos in variable names, etc.

src/themes/color.rs
jkorinth: Have a look at the latest version, I think this is better.

src/themes/xresources.rs
```diff
@@ -0,0 +1,103 @@
...
+    let xresources = PathBuf::from(home + "/.Xresources");
+    debug!(".Xresources @ {:?}", xresources);
+    std::fs::read_to_string(xresources)
+}
+
+#[cfg(test)]
+use tests::read_xresources;
+
+static COLOR_REGEX: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(r"^\s*\*(?<name>[^: ]+)\s*:\s*#(?<color>[a-f0-9]{6})(?<alpha>[a-f0-9]{2})?.*$")
```
MaxVerevkin, line 24:
```suggestion
    Regex::new(r"^\s*\*(?<name>[^: ]+)\s*:\s*#(?<color>[a-f0-9]{6,8}).*$")
```

You don't need to explicitly match the alpha value, if you call `hex.parse()`, the code that parses normal hex colors from normal themes will handle everything.

MaxVerevkin:
@jkorinth I've pushed a commit simplifying some stuff, can you review and test it?

jkorinth:
Sorry, forgot about this. LGTM!

bim9262:
Fixed cspell merge error and use `std::sync::LazyLock` instead of `once_cell::sync::Lazy`

PR #2056, opened 2024-05-16
Add Calendar block

luca-iachini:
This PR adds a Calendar block that displays upcoming events.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 async-trait = "0.1"
 backon = { version = "1.2", default-features = false, features = ["tokio-sleep"] }
+base64 = { version = "0.22.1", optional = true }
 calibright = { version = "0.1.6", features = ["watch"] }
 chrono = { version = "0.4", default-features = false, features = ["clock", "unstable-locales"] }
 chrono-tz = { version = "0.9", features = ["serde"] }
@@ -43,6 +44,7 @@ futures = { version = "0.3", default-features = false }
 glob = { version = "0.3.1", optional = true }
 hyper = "0.14"
 iana-time-zone = "0.1.60"
+icalendar = { version = "0.16.1", features = ["chrono-tz"], optional = true }
 icu_calendar = { version = "1.3.0", optional = true }
 icu_datetime = { version = "1.3.0", optional = true }
 icu_locid = { version = "1.3.0", optional = true }
@@ -58,7 +60,10 @@ neli-wifi = { version = "0.6", features = ["async"] }
 nix = { version = "0.29", features = ["fs", "process"] }
 nom = "7.1.2"
 notmuch = { version = "0.8", optional = true }
+oauth2 = { version = "4.4.2", optional = true }
+once_cell = "1"
 pipewire = { version = "0.8", default-features = false, optional = true }
+quick-xml = { version = "0.31.0", features = ["serialize"], optional = true }
 regex = "1.5"
 reqwest = { version = "0.11", features = ["json"] }
 sensors = "0.2.2"
@@ -72,14 +77,10 @@ swayipc-async = "2.0"
 thiserror = "1.0"
 toml = { version = "0.8", features = ["preserve_order"] }
 unicode-segmentation = "1.10.1"
+url = { version = "2.5.0", optional = true }
 wayrs-client = { version = "1.0", features = ["tokio"] }
 wayrs-protocols = { version = "0.14", features = ["wlr-foreign-toplevel-management-unstable-v1"] }
 zbus = { version = "4", default-features = false, features = ["tokio"] }
-oauth2 = { version = "4.4.2", optional = true }
-quick-xml = { version = "0.31.0", features = ["serialize"], optional = true }
-icalendar = { version = "0.16.1", features = ["chrono-tz"], optional = true }
-url = { version = "2.5.0", optional = true }
-base64 = { version = "0.22.1", optional = true }
 
 [dependencies.tokio]
 version = "1.12"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,13 +16,12 @@ members = [".", "xtask"]
 resolver = "2"
 
 [features]
-default = ["pulseaudio","calendar"]
+default = ["pulseaudio"]
 pulseaudio = ["libpulse-binding"]
 pipewire = ["dep:pipewire"]
 notmuch = ["dep:notmuch"]
 maildir = ["dep:maildir", "glob"]
 icu_calendar = ["dep:icu_datetime", "dep:icu_calendar", "dep:icu_locid"]
-calendar = ["dep:oauth2", "dep:base64", "dep:icalendar", "dep:quick-xml"]
 debug_borders = []                # Make widgets' borders visible
 
 [package.metadata.docs.rs]
@@ -32,7 +31,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 async-trait = "0.1"
 backon = { version = "1.2", default-features = false, features = ["tokio-sleep"] }
-base64 = { version = "0.22.1", optional = true }
+base64 = { version = "0.22.1" }
 calibright = { version = "0.1.6", features = ["watch"] }
 chrono = { version = "0.4", default-features = false, features = ["clock", "unstable-locales"] }
 chrono-tz = { version = "0.9", features = ["serde"] }
@@ -44,7 +43,7 @@ futures = { version = "0.3", default-features = false }
 glob = { version = "0.3.1", optional = true }
 hyper = "0.14"
 iana-time-zone = "0.1.60"
-icalendar = { version = "0.16.2", features = ["chrono-tz"], optional = true }
+icalendar = { version = "0.16.2", features = ["chrono-tz"] }
 icu_calendar = { version = "1.3.0", optional = true }
 icu_datetime = { version = "1.3.0", optional = true }
 icu_locid = { version = "1.3.0", optional = true }
@@ -60,10 +59,10 @@ neli-wifi = { version = "0.6", features = ["async"] }
 nix = { version = "0.29", features = ["fs", "process"] }
 nom = "7.1.2"
 notmuch = { version = "0.8", optional = true }
-oauth2 = { version = "4.4.2", optional = true }
+oauth2 = { version = "4.4.2" }
 once_cell = "1"
 pipewire = { version = "0.8", default-features = false, optional = true }
-quick-xml = { version = "0.31.0", features = ["serialize"], optional = true }
+quick-xml = { version = "0.31.0", features = ["serialize"] }
 regex = "1.5"
 reqwest = { version = "0.11", features = ["json"] }
 sensors = "0.2.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ wayrs-client = { version = "1.0", features = ["tokio"] }
 wayrs-protocols = { version = "0.14", features = ["wlr-foreign-toplevel-management-unstable-v1"] }
 zbus = { version = "4", default-features = false, features = ["tokio"] }
 oauth2 = { version = "4.4.2", optional = true }
-quick-xml = { version = "0.31.0", features = ["serde", "serialize"], optional = true }
+quick-xml = { version = "0.31.0", features = ["serialize"], optional = true }
 icalendar = { version = "0.16.1", features = ["chrono-tz"], optional = true }
 url = { version = "2.5.0", optional = true }
 base64 = { version = "0.22.1", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,10 +22,11 @@ pipewire = ["dep:pipewire"]
 notmuch = ["dep:notmuch"]
 maildir = ["dep:maildir", "glob"]
 icu_calendar = ["dep:icu_datetime", "dep:icu_calendar", "dep:icu_locid"]
+calendar = ["dep:oauth2", "dep:base64", "dep:icalendar", "dep:quick-xml", "dep:url"]
 debug_borders = []                # Make widgets' borders visible
 
 [package.metadata.docs.rs]
-features = ["maildir", "notmuch"]
+features = ["maildir", "notmuch", "calendar"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
@@ -74,6 +75,11 @@ unicode-segmentation = "1.10.1"
 wayrs-client = { version = "1.0", features = ["tokio"] }
 wayrs-protocols = { version = "0.14", features = ["wlr-foreign-toplevel-management-unstable-v1"] }
 zbus = { version = "4", default-features = false, features = ["tokio"] }
+oauth2 = { version = "4.4.2", optional = true }
+quick-xml = { version = "0.31.0", features = ["serde", "serialize"], optional = true }
+icalendar = { version = "0.16.1", features = ["chrono-tz"], optional = true }
+url = { version = "2.5.0", optional = true }
+base64 = { version = "0.22.1", optional = true }
 
 [dependencies.tokio]
 version = "1.12"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ futures = { version = "0.3", default-features = false }
 glob = { version = "0.3.1", optional = true }
 hyper = "0.14"
 iana-time-zone = "0.1.60"
-icalendar = { version = "0.16.1", features = ["chrono-tz"], optional = true }
+icalendar = { version = "0.16.2", features = ["chrono-tz"], optional = true }
 icu_calendar = { version = "1.3.0", optional = true }
 icu_datetime = { version = "1.3.0", optional = true }
 icu_locid = { version = "1.3.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ pipewire = ["dep:pipewire"]
 notmuch = ["dep:notmuch"]
 maildir = ["dep:maildir", "glob"]
 icu_calendar = ["dep:icu_datetime", "dep:icu_calendar", "dep:icu_locid"]
-calendar = ["dep:oauth2", "dep:base64", "dep:icalendar", "dep:quick-xml", "dep:url"]
+calendar = ["dep:oauth2", "dep:base64", "dep:icalendar", "dep:quick-xml"]
 debug_borders = []                # Make widgets' borders visible
 
 [package.metadata.docs.rs]
@@ -77,7 +77,6 @@ swayipc-async = "2.0"
 thiserror = "1.0"
 toml = { version = "0.8", features = ["preserve_order"] }
 unicode-segmentation = "1.10.1"
-url = { version = "2.5.0", optional = true }
 wayrs-client = { version = "1.0", features = ["tokio"] }
 wayrs-protocols = { version = "0.14", features = ["wlr-foreign-toplevel-management-unstable-v1"] }
 zbus = { version = "4", default-features = false, features = ["tokio"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ icu_calendar = ["dep:icu_datetime", "dep:icu_calendar", "dep:icu_locid"]
 debug_borders = []                # Make widgets' borders visible
 
 [package.metadata.docs.rs]
-features = ["calendar", "maildir", "notmuch"]
+features = ["maildir", "notmuch"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
@@ -60,7 +60,6 @@ nix = { version = "0.29", features = ["fs", "process"] }
 nom = "7.1.2"
 notmuch = { version = "0.8", optional = true }
 oauth2 = { version = "4.4.2" }
-once_cell = "1"
 pipewire = { version = "0.8", default-features = false, optional = true }
 quick-xml = { version = "0.31.0", features = ["serialize"] }
 regex = "1.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ calendar = ["dep:oauth2", "dep:base64", "dep:icalendar", "dep:quick-xml", "dep:u
 debug_borders = []                # Make widgets' borders visible
 
 [package.metadata.docs.rs]
-features = ["maildir", "notmuch", "calendar"]
+features = ["calendar", "maildir", "notmuch"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ members = [".", "xtask"]
 resolver = "2"
 
 [features]
-default = ["pulseaudio"]
+default = ["pulseaudio","calendar"]
 pulseaudio = ["libpulse-binding"]
 pipewire = ["dep:pipewire"]
 notmuch = ["dep:notmuch"]
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -202,6 +202,8 @@ define_blocks!(
     watson,
     weather,
     xrandr,
+    #[cfg(feature = "calendar")]
+    calendar,
 );
 
 /// An error which originates from a block
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -152,6 +152,8 @@ define_blocks!(
     backlight,
     battery,
     bluetooth,
+    #[cfg(feature = "calendar")]
+    calendar,
     cpu,
     custom,
     custom_dbus,
@@ -202,8 +204,6 @@ define_blocks!(
     watson,
     weather,
     xrandr,
-    #[cfg(feature = "calendar")]
-    calendar,
 );
 
 /// An error which originates from a block
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -152,7 +152,6 @@ define_blocks!(
     backlight,
     battery,
     bluetooth,
-    #[cfg(feature = "calendar")]
     calendar,
     cpu,
     custom,
--- a/src/blocks/calendar.rs
+++ b/src/blocks/calendar.rs
@@ -245,6 +245,8 @@ pub struct Config {
     pub redirect_format: FormatConfig,
     #[default(60.into())]
     pub interval: Seconds,
+    #[default(10.into())]
+    pub alternate_overlapping_events_every: Seconds,
     #[default(48)]
     pub events_within_hours: u32,
     pub source: Vec<SourceConfig>,
@@ -252,6 +254,12 @@ pub struct Config {
     pub warning_threshold: u32,
     #[default("xdg-open".into())]
     pub browser_cmd: ShellString,
+}
+
+#[derive(Eq, PartialEq)]
+enum WidgetStatus {
+    AlternateEvents,
+    FetchSources,
 }
 
 pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
@@ -285,10 +293,16 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
 
     let mut timer = config.interval.timer();
 
+    let mut alternate_events_timer = config.alternate_overlapping_events_every.timer();
+
     let mut actions = api.get_actions()?;
 
     let events_within = Duration::try_hours(config.events_within_hours.into())
         .error("Invalid events within hours configuration")?;
+
+    let mut widget_status = WidgetStatus::FetchSources;
+
+    let mut next_events = vec![].into_iter().cycle().peekable(); 
 
     loop {
         let mut widget = Widget::new().with_format(no_events_format.clone());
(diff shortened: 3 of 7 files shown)


bim9262:
I'd suggest making this a calendar block and letting google be one of several providers, or perhaps use CalDAV by default since most calendars (including Google) support that.

luca-iachini:
Hey @bim9262, thanks for the suggestion. I did not know the CalDav protocol. I'll try to push an implementation for that.

src/blocks/calendar.rs
```diff
@@ -0,0 +1,385 @@
...
+//! warning_threshold = 600
+//! browser_cmd = "firefox"
+//! [block.auth]
+//! type = "basic"
+//! username = "your_username"
+//! password = "your_password"
+//! ```
+//! Note: The `username` and `password` can also be provided by setting the environment variables `I3RS_CALENDAR_AUTH_USERNAME` and `I3RS_CALENDAR_PASSWORD` respectively.
+//!
+//! ## OAuth2 Authentication (Google Calendar)
```
bim9262, line 67: Should add a description about how to create the app and that you have to enable to caldav api (https://console.cloud.google.com/apis/library/caldav.googleapis.com)

src/blocks/calendar/caldav.rs
```diff
@@ -0,0 +1,372 @@
...
+#[derive(Debug, Deserialize, Clone, Eq, PartialEq)]
+#[serde(rename = "multistatus")]
+struct Multistatus {
+    #[serde(rename = "response", default)]
+    responses: Vec<Response>,
+}
+
+#[derive(Debug, Deserialize, Clone, Eq, PartialEq)]
+struct Response {
+    #[serde(rename = "href")]
```
bim9262, line 180: noop

src/blocks/calendar/caldav.rs
```diff
@@ -0,0 +1,372 @@
...
+            .into_iter()
+            .filter(|p| p.status.contains("200"))
+            .flat_map(|p| p.prop.values.into_iter())
+            .collect()
+    }
+}
+
+#[derive(Debug, Deserialize, Clone, Eq, PartialEq)]
+struct Propstat {
+    #[serde(rename = "status")]
```
bim9262, line 198: noop

src/blocks/calendar/caldav.rs
```diff
@@ -0,0 +1,372 @@
...
+    status: String,
+    #[serde(rename = "prop")]
```
bim9262, line 200: noop

src/blocks/calendar/caldav.rs
```diff
@@ -0,0 +1,372 @@
...
+#[serde(rename_all = "kebab-case")]
+enum ResourceType {
+    Calendar,
+    #[serde(other)]
+    Unsupported,
+}
+
+#[derive(Debug, Deserialize, Clone, PartialEq, Eq)]
+struct SupportedCalendarComponentSet {
+    #[serde(rename = "comp")]
```
bim9262, line 249: noop

src/blocks/calendar.rs
```diff
@@ -0,0 +1,385 @@
...
+pub struct OAuth2Config {
+    pub client_id: Option<String>,
+    pub client_secret: Option<String>,
+    pub auth_url: String,
+    pub token_url: String,
+    #[default("~/.config/i3status-rust/calendar.auth_token".into())]
+    pub auth_token: ShellString,
+    #[default(8080)]
+    pub redirect_port: u16,
+    pub scopes: Vec<String>,
```
bim9262, line 143: This can be `Vec<Scope>`

src/blocks.rs
```diff
@@ -201,6 +201,8 @@ define_blocks!(
     watson,
     weather,
     xrandr,
+    #[cfg(feature = "calendar")]
```
bim9262, line 204: Would prefer to keep the list alphabetical

Cargo.toml
```diff
@@ -74,6 +75,11 @@ unicode-segmentation = "1.10.1"
 wayrs-client = { version = "1.0", features = ["tokio"] }
 wayrs-protocols = { version = "0.13", features = ["wlr-foreign-toplevel-management-unstable-v1"] }
 zbus = { version = "4", default-features = false, features = ["tokio"] }
+oauth2 = { version = "4.4.2", optional = true }
+quick-xml = { version = "0.31.0", features = ["serde", "serialize"], optional = true }
```
bim9262, line 79: "serde" is redundant as "serialize" enables it.

bim9262 (commented):
It would be nice to be able to leverage https://crates.io/crates/caldav-utils, which seems to be the most used async caldav library, but it's missing oauth.

src/blocks/calendar.rs
```diff
@@ -0,0 +1,385 @@
...
+//! browser_cmd = "firefox"
+//! [block.auth]
+//! type = "oauth2"
+//! client_id = "your_client_id"
+//! client_secret = "your_client_secret"
+//! auth_url = "https://accounts.google.com/o/oauth2/auth"
+//! token_url = "https://oauth2.googleapis.com/token"
+//! auth_token = "~/.config/i3status-rust/calendar.auth_token"
+//! redirect_port = 8080
+//! scopes = ["https://www.googleapis.com/auth/calendar", "https://www.googleapis.com/auth/calendar.events"]
```
bim9262, line 90: BTW, I would have thought that the readonly versions of the scopes would be valid. However that doesn't seem to be the case. I opened an issue about it: https://issuetracker.google.com/issues/342730758

luca-iachini:
> It would be nice to be able to leverage https://crates.io/crates/caldav-utils, which seems to be the most used async caldav library, but it's missing oauth.

That would be nice. Unfortunately, I checked all the Rust clients to see if they support OAuth2 or have a way to intercept HTTP client errors in a usable way.

src/blocks/calendar.rs
```diff
@@ -0,0 +1,396 @@
...
+        .filter(|c| config.calendars.is_empty() || config.calendars.contains(&c.name))
+        .collect();
+    let mut events: Vec<Event> = vec![];
+    for calendar in calendars {
+        let calendar_events: Vec<_> = client
+            .events(
+                &calendar,
+                Utc::now(),
+                Utc::now()
+                    + Duration::try_days(config.events_within_hours.try_into().unwrap()).unwrap(),
```
bim9262, line 362: This should be `try_hours`.  I'd make `events_within_hours` `u32` to constrain it to positive numbers that will fit within the i64 that `try_hours` takes.

bim9262 (commented):
There's a lot of unwraps that should be handled too.

BTW good work on this, your contribution is appreciated :)

src/blocks/calendar.rs
```diff
@@ -0,0 +1,396 @@
...
+//! Key | Values | Default
+//! ----|--------|--------
+//! `next_event_format` | A string to customize the output of this block when there is a next event in the calendar. See below for available placeholders. | <code>\" $icon $start.datetime(f:'%a %H:%M') $summary \"</code>
+//! `ongoing_event_format` | A string to customize the output of this block when an event is ongoing. | <code>\" $icon $summary (ends at $end.datetime(f:'%H:%M')) \"</code>
+//! `no_events_format` | A string to customize the output of this block when there are no events | <code>\" $icon \"</code>
+//! `interval` | Update interval in seconds | `60`
+//! `url` | CalDav calendar server URL | N/A
+//! `auth` | Authentication configuration (unauthenticated, basic, or oauth2) | `unauthenticated`
+//! `calendars` | List of calendar names to monitor. If empty, all calendars will be fetched. | `[]`
+//! `events_within_hours` | Number of hours to look for events in the future | `24`
```
bim9262, line 16: The default here is 24 hours, but it actually defaults to 48 hours.

src/blocks/calendar.rs
```diff
@@ -0,0 +1,396 @@
...
+    let mut actions = api.get_actions()?;
+
+    loop {
+        let mut widget = Widget::new().with_format(no_events_format.clone());
+        widget.set_values(map! {
+            "icon" => Value::icon("calendar"),
+        });
+
+        let mut next_event = None;
+        let mut retries = 0;
```
bim9262, line 213: I'd probably use a for loop (same thing for your other retry loop). Also I think you want to handle the case when the AskUser failed on the first attempt. Something like:

```diff
diff --git a/src/blocks/calendar.rs b/src/blocks/calendar.rs
index efb5aa114..8a6220e69 100644
--- a/src/blocks/calendar.rs
+++ b/src/blocks/calendar.rs
@@ -210,8 +210,7 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
         });
 
         let mut next_event = None;
-        let mut retries = 0;
-        while retries <= 1 {
+       for retries in 0..=1{
             let next_event_result = get_next_event(config, &mut client).await;
             match next_event_result {
                 Ok(event) => {
@@ -223,12 +222,7 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
                         let authorization =
                             client.authorize().await.error("Authorization failed")?;
                         match &authorization {
-                            auth::Authorize::Completed => {
-                                return Err(Error::new(
-                                    "Authorization failed. Check your configurations",
-                                ))
-                            }
-                            auth::Authorize::AskUser(AuthorizeUrl { url, .. }) => {
+                            auth::Authorize::AskUser(AuthorizeUrl { url, .. }) if retries == 0 => {
                                 widget.set_format(redirect_format.clone());
                                 api.set_widget(widget.clone())?;
                                 open_browser(config, url).await?;
@@ -237,11 +231,15 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
                                     .await
                                     .error("Ask user failed")?;
                             }
+                            _ => {
+                                return Err(Error::new(
+                                    "Authorization failed. Check your configurations",
+                                ))
+                            }
                         }
                     }
                 }
             };
-            retries += 1;
         }
         if let Some(event) = next_event.clone() {
             if let (Some(start_date), Some(end_date)) = (event.end_at, event.end_at) {
```

src/blocks/calendar.rs
```diff
@@ -0,0 +1,385 @@
...
+//! [[block]]
+//! block = "calendar"
+//! next_event_format = " $icon $start.datetime(f:'%a %H:%M') $summary "
+//! ongoing_event_format = " $icon $summary (ends at $end.datetime(f:'%H:%M')) "
+//! no_events_format = " $icon no events "
+//! interval = 30
+//! url = "https://caldav.example.com/calendar/"
+//! calendars = [ "Holidays" ]
+//! events_within_hours = 48
+//! ignore_all_day_events = true
```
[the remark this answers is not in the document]
bim9262: That's not to say that it isn't a good option to have.

src/blocks/calendar.rs
```diff
@@ -0,0 +1,396 @@
...
+        let mut retries = 0;
+        while retries <= 1 {
+            let next_event_result = get_next_event(config, &mut client).await;
+            match next_event_result {
+                Ok(event) => {
+                    next_event = event;
+                    break;
+                }
+                Err(err) => {
+                    if let CalendarError::AuthRequired = err {
```
bim9262, line 222: You should surface the other errors that could have happended for example `CalendarError::Http` could be raised for using the wrong url and you get a 404 error.

```diff
diff --git a/src/blocks/calendar.rs b/src/blocks/calendar.rs
index efb5aa114..6b12f7b43 100644
--- a/src/blocks/calendar.rs
+++ b/src/blocks/calendar.rs
@@ -131,6 +131,7 @@ use self::caldav::Event;
 use super::prelude::*;
 
 use std::path::Path;
+use std::sync::Arc;
 
 use caldav::CalDavClient;
 
@@ -218,8 +219,8 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
                     next_event = event;
                     break;
                 }
-                Err(err) => {
-                    if let CalendarError::AuthRequired = err {
+                Err(err) => match err {
+                    CalendarError::AuthRequired => {
                         let authorization =
                             client.authorize().await.error("Authorization failed")?;
                         match &authorization {
@@ -239,7 +240,13 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
                             }
                         }
                     }
-                }
+                    e => {
+                        return Err(Error {
+                            message: None,
+                            cause: Some(Arc::new(e)),
+                        })
+                    }
+                },
             };
             retries += 1;
         }
```

luca-iachini:
> There's a lot of unwraps that should be handled too.

Ah yep, I forgot to properly handle those cases :sweat_smile: .

> BTW good work on this, your contribution is appreciated :)

Thanks :smile:

Cargo.toml
```diff
@@ -74,6 +75,11 @@ unicode-segmentation = "1.10.1"
 wayrs-client = { version = "1.0", features = ["tokio"] }
 wayrs-protocols = { version = "0.13", features = ["wlr-foreign-toplevel-management-unstable-v1"] }
 zbus = { version = "4", default-features = false, features = ["tokio"] }
+oauth2 = { version = "4.4.2", optional = true }
```
bim9262, line 78: let's keep this sorted too

bim9262 (commented):
Still seems like a lot of `unwrap`s.
You'll need to rebase on master to fix the merge error.

MaxVerevkin:
`expect` is not much better than `unwrap`, it still panics. Please use `error`.

luca-iachini:
> expect is not much better than unwrap, it still panics. Please use error.

I know both panics. I changed to `expect` the unwraps that cannot happen by design. Such as a `url` parsing from a string is hard-coded, HTTP method from a static string, .... Maybe, there is some that should be a proper error instead.

Cargo.toml
```diff
@@ -72,6 +77,7 @@ swayipc-async = "2.0"
 thiserror = "1.0"
 toml = { version = "0.8", features = ["preserve_order"] }
 unicode-segmentation = "1.10.1"
+url = { version = "2.5.0", optional = true }
```
MaxVerevkin, line 80: `oauth2` reexports the `url` crate, so we don't have to depend on it directly.

src/blocks/calendar/auth.rs
```diff
@@ -0,0 +1,303 @@
...
+impl TokenStore {
+    pub fn new(path: &Path) -> Self {
+        Self {
+            path: path.into(),
+            token: None,
+        }
+    }
+
+    pub fn store(&mut self, token: OAuth2TokenResponse) -> Result<(), CalendarError> {
+        let file = File::create(&self.path).unwrap();
```
bim9262, line 286: These two unwraps can fail. Also I think we should use the tokio::fs version.

src/blocks/calendar/auth.rs
```diff
@@ -0,0 +1,303 @@
...
+use oauth2::{
+    AuthUrl, AuthorizationCode, ClientId, ClientSecret, CsrfToken, EmptyExtraTokenFields,
+    PkceCodeChallenge, PkceCodeVerifier, RedirectUrl, RefreshToken, Scope, StandardTokenResponse,
+    TokenResponse, TokenUrl,
+};
+use reqwest::header::{HeaderMap, HeaderValue, AUTHORIZATION};
+use reqwest::Url;
+use std::fs::File;
+use std::io::{BufRead, BufReader, BufWriter, Write};
+use std::net::TcpListener;
```
bim9262, line 13: See tokio::fs, tokio::io, and tokio::net

src/blocks/calendar.rs
```diff
@@ -0,0 +1,401 @@
...
+        }
+
+        api.set_widget(widget)?;
+        loop {
+            select! {
+              _ = timer.tick() => break,
+              _ = api.wait_for_update_request() => break,
+              Some(action) = actions.recv() => match action.as_ref() {
+                    "open_link" => {
+                            if let Some(url) = next_event.clone().and_then(|e| e.url) {
```
bim9262, line 287: We can remove a clone here:

```diff
diff --git a/src/blocks/calendar.rs b/src/blocks/calendar.rs
index 1be0760cf..547ef6094 100644
--- a/src/blocks/calendar.rs
+++ b/src/blocks/calendar.rs
@@ -284,11 +284,11 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
               _ = api.wait_for_update_request() => break,
               Some(action) = actions.recv() => match action.as_ref() {
                     "open_link" => {
-                            if let Some(url) = next_event.clone().and_then(|e| e.url) {
-                                if let Ok(url) = Url::parse(&url) {
-                                    open_browser(config, &url).await?;
-                                }
+                        if let Some(Event { url: Some(url), .. }) = &next_event {
+                            if let Ok(url) = Url::parse(&url) {
+                                open_browser(config, &url).await?;
                             }
+                        }
                     }
                     _ => ()
                 }
```

src/blocks/calendar/caldav.rs
bim9262, line 1: We can clean up the derives here, also `ResponseCalendarError` isn't used anywhere:

```diff
diff --git a/src/blocks/calendar/caldav.rs b/src/blocks/calendar/caldav.rs
index 7c5c939aa..e2ebdcd6b 100644
--- a/src/blocks/calendar/caldav.rs
+++ b/src/blocks/calendar/caldav.rs
@@ -24,7 +24,7 @@ pub struct Event {
     pub end_at: Option<DateTime<Utc>>,
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug)]
+#[derive(Serialize, Deserialize, Debug)]
 pub struct Calendar {
     pub url: Url,
     pub name: String,
@@ -153,14 +153,14 @@ impl CalDavClient {
     }
 }
 
-#[derive(Debug, Deserialize, Clone, Eq, PartialEq)]
+#[derive(Debug, Deserialize)]
 #[serde(rename = "multistatus")]
 struct Multistatus {
     #[serde(rename = "response", default)]
     responses: Vec<Response>,
 }
 
-#[derive(Debug, Deserialize, Clone, Eq, PartialEq)]
+#[derive(Debug, Deserialize)]
 struct Response {
     href: String,
     #[serde(rename = "propstat", default)]
@@ -177,19 +177,19 @@ impl Response {
     }
 }
 
-#[derive(Debug, Deserialize, Clone, Eq, PartialEq)]
+#[derive(Debug, Deserialize)]
 struct Propstat {
     status: String,
     prop: Prop,
 }
 
-#[derive(Debug, Deserialize, Clone, Eq, PartialEq)]
+#[derive(Debug, Deserialize)]
 struct Prop {
     #[serde(rename = "$value")]
     pub values: Vec<PropValue>,
 }
 
-#[derive(Debug, Deserialize, Clone, Eq, PartialEq)]
+#[derive(Debug, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 enum PropValue {
     CurrentUserPrincipal(HrefProperty),
@@ -202,12 +202,12 @@ enum PropValue {
     CalendarData(String),
 }
 
-#[derive(Debug, Deserialize, Clone, Eq, PartialEq)]
+#[derive(Debug, Deserialize)]
 pub struct HrefProperty {
     href: String,
 }
 
-#[derive(Debug, Deserialize, Clone, PartialEq, Eq)]
+#[derive(Debug, Deserialize, PartialEq)]
 struct ResourceTypes {
     #[serde(rename = "$value")]
     pub values: Vec<ResourceType>,
@@ -218,7 +218,7 @@ impl ResourceTypes {
         self.values.contains(&ResourceType::Calendar)
     }
 }
-#[derive(Debug, Deserialize, Clone, PartialEq, Eq)]
+#[derive(Debug, Deserialize, PartialEq)]
 #[serde(rename_all = "kebab-case")]
 enum ResourceType {
     Calendar,
@@ -226,7 +226,7 @@ enum ResourceType {
     Unsupported,
 }
 
-#[derive(Debug, Deserialize, Clone, PartialEq, Eq)]
+#[derive(Debug, Deserialize)]
 struct SupportedCalendarComponentSet {
     comp: Option<Comp>,
 }
@@ -236,19 +236,12 @@ impl SupportedCalendarComponentSet {
     }
 }
 
-#[derive(Debug, Deserialize, Clone, PartialEq, Eq)]
+#[derive(Debug, Deserialize)]
 struct Comp {
     #[serde(rename = "@name", default)]
     name: String,
 }
 
-#[derive(Debug, Deserialize, Clone, PartialEq, Eq)]
-struct ResponseCalendarError {
-    code: String,
-    #[serde(rename = "internalReason")]
-    internal_reason: String,
-}
-
 fn parse_href(multi_status: Multistatus, base_url: Url) -> Result<Url, CalendarError> {
     let props = multi_status
         .responses
```

src/blocks/calendar/auth.rs
luca-iachini: yep, I missed those. Thx

cfsmp3:
Can you have more than one calendar source in the same block? 
That would be my specific use case.

bim9262:
That should be doable. @luca-iachini if you want an example of how to do that check out the privacy block's drivers and see how the drivers are polled together

luca-iachini:
> Can you have more than one calendar source in the same block? That would be my specific use case.

Hey @cfsmp3, can you expand a bit on your use case? Why are you looking to have events from two distinct servers on the same block? Why using two calendar blocks doesn't work for you?

> That should be doable. @luca-iachini if you want an example of how to do that, check out the privacy block's drivers and see how the drivers are polled together.

@bim9262 It is probably doable, but there are caveats to consider in order to make this work properly and avoid overcomplicating the solution:

- How do we display conflicting events between the two sources? (alternating between sources, give priorities, ...)
- How do you handle errors from both clients effectively?
- How do you manage the authentication flows?
- Allow multiple sources or just 2?

 using two separate blocks is the easiest way to go.

IMHO, I would commit to close this PR and provide an initial solution and then iterate to improve it. WDYT?

BTW, I would like to work on displaying conflicting events in an effective way, such as alternating events that are scheduled in overlapping periods, rather than supporting multiple sources in the same block.

bim9262:
I think the logic for conflicts for one or more calendars would be the same. You could have overlapping events on a single calendar too.

You would have auth per calendar. The reson to do it now, add multiple calendars from the get go, is that it means users won't have to change their configs when multi calendar support is added.

The try_join_all macro will return an exception if any of the calendars fail to load.

bim9262:
@MaxVerevkin what are your thoughts on the calendar feature flag? I think that it's probably not required since the dependencies are pretty light and don't require system libraries.

luca-iachini:
> I think the logic for conflicts for one or more calendars would be the same. You could have overlapping events on a single calendar too.
> You would have auth per calendar. The reson to do it now, add multiple calendars from the get go, is that it means users won't have to change their configs when multi calendar support is added.
> The try_join_all macro will return an exception if any of the calendars fail to load.

Certainly, here's a refined version:

@bim9262 Yes, I know how to do it. My questions were not for technical help but to highlight that adding more sources requires considering additional aspects, which increases complexity. If we keep adding features, we won't reach a conclusion. In my opinion, we should agree on a minimum viable product (MVP) that is stable and meets most users' needs. Since I can only work on this in my free time, my focus is to create a stable solution and push to close this PR. This will allow you and the community to take over and add any features you like.
What is currently implemented meets my needs. I can continue to work on it but at my own pace. Since two blocks or sync the external calendar  in one of the two server is a fair solution, I don't see a point in supporting multiple sources from the start.

bim9262:
This should allow us to keep the same config for when multiple source are supported, but not actually implement how multiple sources are handled.

```diff
diff --git a/src/blocks/calendar.rs b/src/blocks/calendar.rs
index 5f8c3243e..12a09377d 100644
--- a/src/blocks/calendar.rs
+++ b/src/blocks/calendar.rs
@@ -11,12 +11,20 @@
 //! `no_events_format` | A string to customize the output of this block when there are no events | <code>\" $icon \"</code>
 //! `redirect_format` | A string to customize the output of this block when the authorization is asked | <code>\" $icon Check your web browser \"</code>
 //! `interval` | Update interval in seconds | `60`
+//! `source` | Array of sources to pull calendars from | `[]`
+//! `warning_threshold` | Warning threshold in seconds for the upcoming event | `300`
+//! `browser_cmd` | Command to open event details in a browser. The block passes the HTML link as an argument | `"xdg-open"`
+//!
+//! # Source Configuration
+//!
+//! Key | Values | Default
+//! ----|--------|--------
 //! `url` | CalDav calendar server URL | N/A
 //! `auth` | Authentication configuration (unauthenticated, basic, or oauth2) | `unauthenticated`
 //! `calendars` | List of calendar names to monitor. If empty, all calendars will be fetched. | `[]`
 //! `events_within_hours` | Number of hours to look for events in the future | `48`
-//! `warning_threshold` | Warning threshold in seconds for the upcoming event | `300`
-//! `browser_cmd` | Command to open event details in a browser. The block passes the HTML link as an argument | `"xdg-open"`
+//!
+//! Note: Currently only one source is supported
 //!
 //! Action          | Description                               | Default button
 //! ----------------|-------------------------------------------|---------------
@@ -33,12 +41,13 @@
 //! ongoing_event_format = " $icon $summary (ends at $end.datetime(f:'%H:%M')) "
 //! no_events_format = " $icon no events "
 //! interval = 30
+//! warning_threshold = 600
+//! browser_cmd = "firefox"
+//! [[block.source]]
 //! url = "https://caldav.example.com/calendar/"
 //! calendars = ["user/calendar"]
 //! events_within_hours = 48
-//! warning_threshold = 600
-//! browser_cmd = "firefox"
-//! [block.auth]
+//! [block.source.auth]
 //! type = "unauthenticated"
 //! ```
 //!
@@ -51,12 +60,13 @@
 //! ongoing_event_format = " $icon $summary (ends at $end.datetime(f:'%H:%M')) "
 //! no_events_format = " $icon no events "
 //! interval = 30
+//! warning_threshold = 600
+//! browser_cmd = "firefox"
+//! [[block.source]]
 //! url = "https://caldav.example.com/calendar/"
 //! calendars = [ "Holidays" ]
 //! events_within_hours = 48
-//! warning_threshold = 600
-//! browser_cmd = "firefox"
-//! [block.auth]
+//! [block.source.auth]
 //! type = "basic"
 //! username = "your_username"
 //! password = "your_password"
@@ -86,12 +96,13 @@
 //! ongoing_event_format = " $icon $summary (ends at $end.datetime(f:'%H:%M')) "
 //! no_events_format = " $icon no events "
 //! interval = 30
+//! warning_threshold = 600
+//! browser_cmd = "firefox"
+//! [[block.source]]
 //! url = "https://apidata.googleusercontent.com/caldav/v2/"
 //! calendars = ["primary"]
 //! events_within_hours = 48
-//! warning_threshold = 600
-//! browser_cmd = "firefox"
-//! [block.auth]
+//! [block.source.auth]
 //! type = "oauth2"
 //! client_id = "your_client_id"
 //! client_secret = "your_client_secret"
@@ -165,6 +176,16 @@ pub enum AuthConfig {
     OAuth2(OAuth2Config),
 }
 
+#[derive(Deserialize, Debug, SmartDefault)]
+#[serde(deny_unknown_fields, default)]
+pub struct Source {
+    pub url: String,
+    pub auth: AuthConfig,
+    pub calendars: Vec<String>,
+    #[default(48)]
+    pub events_within_hours: u32,
+}
+
 #[derive(Deserialize, Debug, SmartDefault)]
 #[serde(deny_unknown_fields, default)]
 pub struct Config {
@@ -174,11 +195,7 @@ pub struct Config {
     pub redirect_format: FormatConfig,
     #[default(60.into())]
     pub interval: Seconds,
-    pub url: String,
-    pub auth: AuthConfig,
-    pub calendars: Vec<String>,
-    #[default(48)]
-    pub events_within_hours: u32,
+    pub source: Vec<Source>,
     #[default(300)]
     pub warning_threshold: u32,
     #[default("xdg-open".into())]
@@ -199,13 +216,26 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
 
     api.set_default_actions(&[(MouseButton::Left, None, "open_link")])?;
 
-    let mut client = caldav_client(config).await?;
+    let source = match config.source.len() {
+        0 => return Err(Error::new("A calendar source must be supplied")),
+        1 => config
+            .source
+            .first()
+            .expect("There must be a first entry since the length is 1"),
+        _ => {
+            return Err(Error::new(
+                "Currently only one calendar source is supported",
+            ))
+        }
+    };
+
+    let mut client = caldav_client(source).await?;
 
     let mut timer = config.interval.timer();
 
     let mut actions = api.get_actions()?;
 
-    let events_within = Duration::try_hours(config.events_within_hours.into())
+    let events_within = Duration::try_hours(source.events_within_hours.into())
         .error("Invalid events within hours configuration")?;
 
     loop {
@@ -216,7 +246,7 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
 
         let mut next_event = None;
         for retries in 0..=1 {
-            let next_event_result = get_next_event(config, &mut client, events_within).await;
+            let next_event_result = get_next_event(source, &mut client, events_within).await;
             match next_event_result {
                 Ok(event) => {
                     next_event = event;
@@ -297,8 +327,8 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
     }
 }
 
-async fn caldav_client(config: &Config) -> Result<caldav::CalDavClient> {
-    let auth = match &config.auth {
+async fn caldav_client(source: &Source) -> Result<caldav::CalDavClient> {
+    let auth = match &source.auth {
         AuthConfig::Unauthenticated => auth::Auth::Unauthenticated,
         AuthConfig::Basic(BasicAuthConfig { username, password }) => {
             let username = username
@@ -338,13 +368,13 @@ async fn caldav_client(config: &Config) -> Result<caldav::CalDavClient> {
         }
     };
     Ok(CalDavClient::new(
-        Url::parse(&config.url).error("Invalid CalDav server url")?,
+        Url::parse(&source.url).error("Invalid CalDav server url")?,
         auth,
     ))
 }
 
 async fn get_next_event(
-    config: &Config,
+    source: &Source,
     client: &mut CalDavClient,
     within: Duration,
 ) -> Result<Option<caldav::Event>, CalendarError> {
@@ -352,7 +382,7 @@ async fn get_next_event(
         .calendars()
         .await?
         .into_iter()
-        .filter(|c| config.calendars.is_empty() || config.calendars.contains(&c.name))
+        .filter(|c| source.calendars.is_empty() || source.calendars.contains(&c.name))
         .collect();
     let mut events: Vec<Event> = vec![];
     for calendar in calendars {
```

src/blocks/calendar.rs
```diff
@@ -0,0 +1,611 @@
...
+                        .to_utc(),
+                    Utc::now() + within,
+                )
+                .await?
+                .into_iter()
+                .filter(|e| {
+                    let not_started = e.start_at.is_some_and(|d| d > Utc::now());
+                    let is_ongoing = e.start_at.is_some_and(|d| d < Utc::now())
+                        && e.end_at.is_some_and(|d| d > Utc::now());
+                    not_started || is_ongoing
```
bim9262, line 515: Shouldn't all of the events received be not_started or ongoing?

bim9262 (commented):
When events are scheduled too close to now I don't see them show up, but if I put them later enough today the do show up. I'll need to spend some more time troubleshooting this behavior.

MaxVerevkin:
> @MaxVerevkin what are your thoughts on the calendar feature flag? I think that it's probably not required since the dependencies are pretty light and don't require system libraries.

If there are no runtime dependencies, and if the binary size is not affected significantly (unlike `icu_calendar`), then lets not put it behind a feature flag. (We also may reconsider if `maildir` has to be behind a feature flag too).

cfsmp3:
> Hey @cfsmp3, can you expand a bit on your use case? Why are you looking to have events from two distinct servers on the same block? Why using two calendar blocks doesn't work for you?

I have more than one Google account, and I'd like to see the next event considering all calendars (not separately for each calendar).

src/blocks/calendar.rs
luca-iachini: The block retrieves events from the start of the day to make the ongoing visible after `Utc::now`.  It is not a perfect solution, because you can have events that last multiple days and started days before. 
I usually use this block to be warned about my next meeting of the day and I want to know if there is an ongoing meeting for which I missed the warning. 
If it is an event that lasts days, I should already know about it so I do not care if I do not see it as an ongoing event in the block.

luca-iachini:
@bim9262 I couldn't replicate the bug you found regarding upcoming events. However, I did update the `icalendar-rs` dependency to fix a parser bug.
@MaxVerevkin I've also removed the feature flag `calendar`. Let me know if this works for you.

src/blocks/calendar.rs
```diff
@@ -0,0 +1,610 @@
...
+            .into_iter()
+            .filter(|c| self.config.calendars.is_empty() || self.config.calendars.contains(&c.name))
+            .collect();
+        let mut events: Vec<Event> = vec![];
+        for calendar in calendars {
+            let calendar_events: Vec<_> = self
+                .client
+                .events(
+                    &calendar,
+                    Utc::now()
```
bim9262, line 498: I finally was able to chase down why my events weren't showing up. This should be `Local::now()` instead of `Utc::now()` so that it's the local date part that were using.

Cargo.toml
```diff
@@ -25,12 +25,13 @@ icu_calendar = ["dep:icu_datetime", "dep:icu_calendar", "dep:icu_locid"]
 debug_borders = []                # Make widgets' borders visible
 
 [package.metadata.docs.rs]
-features = ["maildir", "notmuch"]
+features = ["calendar", "maildir", "notmuch"]
```
bim9262, line 28: You ended up removing the calendar as a feature.

Cargo.toml
```diff
@@ -57,7 +59,10 @@ neli-wifi = { version = "0.6", features = ["async"] }
 nix = { version = "0.29", features = ["fs", "process"] }
 nom = "7.1.2"
 notmuch = { version = "0.8", optional = true }
+oauth2 = { version = "4.4.2" }
+once_cell = "1"
```
bim9262, line 63: You accidentally reintroduced a dependency we removed from the project.

bim9262 (commented):
Just 2 little changes to Cargo.toml, and we're set. I've tested it out and everything looks good to me.

bim9262:
Looks like Cargo.lock didn't get updated to reflect once_cell being removed.

bim9262:
Thanks @luca-iachini !!

PR #2060, opened 2024-06-04
block[battery]: give priority to charging over empty

### LINKED ISSUE #2059: Revert empty battery state to charging when plugged in
Hi, I recently migrated to `i3status-rust`. Thanks for this awesome tool!

## Issue

Running `i3status-rust` on a laptop with the `upower` driver, I mistakenly left my laptop on overnight and noted that the battery level was around 5% the next day. This showed a critical color and empty icon/status for the battery module. After plugging in power, I expected to see the plugged-in icon. However, because the battery level was below the `empty_threshold`, the plugged in icon did not show despite the battery being charged, up until the battery level surpassed the `empty_threshold`.

## Feature request

It would be really good to override the empty battery state once power is plugged in. IMO, the state should change from empty to charging, regardless of the current battery level.

## Workaround

In the meantime, I am disabling the empty state altogether by setting `empty_threshold = -1`.

## Specs

`i3status-rs 0.33.1` on Arch Linux

bim9262:
Fixes #2059

### DIFF
--- a/src/blocks/battery.rs
+++ b/src/blocks/battery.rs
@@ -138,7 +138,9 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
         if let Some(info) = &mut info {
             if info.capacity >= config.full_threshold {
                 info.status = BatteryStatus::Full;
-            } else if info.capacity <= config.empty_threshold {
+            } else if info.capacity <= config.empty_threshold
+                && info.status != BatteryStatus::Charging
+            {
                 info.status = BatteryStatus::Empty;
             }
         }


src/blocks/battery.rs
```diff
@@ -138,7 +138,9 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
         if let Some(info) = &mut info {
             if info.capacity >= config.full_threshold {
                 info.status = BatteryStatus::Full;
-            } else if info.capacity <= config.empty_threshold {
+            } else if info.capacity <= config.empty_threshold
+                && !matches!(info.status, BatteryStatus::Charging)
```
MaxVerevkin, line 142: You can just compare the values directly, no need to use `matches!`.

PR #2061, opened 2024-06-09
block[weather]: Add support for the US National Weather Service

Cognoscan:
I prefer using the US National Weather Service to get my forecast information in the US, especially as it means I don't need an API key like OpenWeatherMap requires. Thanks to the previous work that was done to support met.no, it was quite easy to add in support for NWS as an additional service option. Would it be alright to merge this support in?

### DIFF
--- a/src/blocks/weather/nws.rs
+++ b/src/blocks/weather/nws.rs
@@ -254,8 +254,8 @@ fn combine_forecasts(data: &[ForecastAggregate], fin: WeatherMoment) -> Forecast
         if val.wind > max.wind {
             max.wind_direction = val.wind_direction;
             max.wind = val.wind;
+            max.wind_kmh = val.wind_kmh;
         }
-        max.wind_kmh = max.wind_kmh.max(val.wind_kmh);
 
         // Min
         min.temp = min.temp.min(val.temp);
@@ -264,8 +264,8 @@ fn combine_forecasts(data: &[ForecastAggregate], fin: WeatherMoment) -> Forecast
         if val.wind < min.wind {
             min.wind_direction = val.wind_direction;
             min.wind = val.wind;
+            min.wind_kmh = val.wind_kmh;
         }
-        min.wind_kmh = min.wind_kmh.min(val.wind_kmh);
     }
 
     let count = data.len() as f64;
--- a/src/blocks/weather/nws.rs
+++ b/src/blocks/weather/nws.rs
@@ -169,7 +169,7 @@ impl ApiForecast {
         }
     }
 
-    fn apparant_temp(&self) -> f64 {
+    fn apparent_temp(&self) -> f64 {
         let temp = if self.temperature.unit_code.ends_with("degC") {
             self.temperature.value
         } else {
@@ -188,7 +188,7 @@ impl ApiForecast {
             weather,
             weather_verbose: self.short_forecast.clone(),
             temp: self.temperature.value,
-            apparent: self.apparant_temp(),
+            apparent: self.apparent_temp(),
             humidity: self.relative_humidity.value,
             wind: self.wind_speed.value,
             wind_kmh: self.wind_kmh(),
@@ -199,7 +199,7 @@ impl ApiForecast {
     fn to_aggregate(&self) -> ForecastAggregate {
         ForecastAggregate {
             temp: self.temperature.value,
-            apparent: self.apparant_temp(),
+            apparent: self.apparent_temp(),
             humidity: self.relative_humidity.value,
             wind: self.wind_speed.value,
             wind_kmh: self.wind_kmh(),
--- a/src/blocks/weather/nws.rs
+++ b/src/blocks/weather/nws.rs
@@ -208,7 +208,7 @@ impl ApiForecast {
     }
 }
 
-fn aggregate_avg(data: &[ForecastAggregate]) -> ForecastAggregate {
+fn combine_forecasts(data: &[ForecastAggregate], fin: WeatherMoment) -> Forecast {
     let mut temp = 0.0;
     let mut apparent = 0.0;
     let mut humidity = 0.0;
@@ -216,7 +216,24 @@ fn aggregate_avg(data: &[ForecastAggregate]) -> ForecastAggregate {
     let mut wind_east = 0.0;
     let mut wind_kmh_north = 0.0;
     let mut wind_kmh_east = 0.0;
+    let mut max = ForecastAggregate {
+        temp: -1000.0,
+        apparent: -1000.0,
+        humidity: 0.0,
+        wind: 0.0,
+        wind_kmh: 0.0,
+        wind_direction: Some(0.0),
+    };
+    let mut min = ForecastAggregate {
+        temp: 1000.0,
+        apparent: 1000.0,
+        humidity: 100.0,
+        wind: 1000.0,
+        wind_kmh: 1000.0,
+        wind_direction: Some(0.0),
+    };
     for val in data {
+        // Summations for averaging
         temp += val.temp;
         apparent += val.apparent;
         humidity += val.humidity;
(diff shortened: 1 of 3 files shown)


src/blocks/weather/nws.rs
```diff
@@ -0,0 +1,391 @@
...
+        wind_east += val.wind * sin;
+        wind_kmh_north += val.wind_kmh * cos;
+        wind_kmh_east += val.wind_kmh * sin;
+    }
+
+    let count = data.len() as f64;
+    ForecastAggregate {
+        temp,
+        apparent,
+        humidity,
```
bim9262, line 238: Temp, apparent, and humidity should be divided by count.

src/blocks/weather/nws.rs
```diff
@@ -0,0 +1,391 @@
...
+    let mut humidity = 0.0;
+    let mut wind_north = 0.0;
+    let mut wind_east = 0.0;
+    let mut wind_kmh_north = 0.0;
+    let mut wind_kmh_east = 0.0;
+    for val in data {
+        temp += val.temp;
+        apparent += val.apparent;
+        humidity += val.humidity;
+        let (sin, cos) = val
```
bim9262, line 223: See `average_wind`

bim9262 (commented):
I think that it would be better to not have the three `aggregate_*` fbs so that you don't need to iterate over the same data three times.

src/blocks/weather/nws.rs
Cognoscan: Total mistake on my part; fixed.

src/blocks/weather/nws.rs
Cognoscan: Reasonable enough, I merged them together into a single function that iterates over the `ForecastAggregate` data once for all items, returning a completed `Forecast`.

bim9262 (commented):
This is looking pretty good :)

src/blocks/weather/nws.rs
```diff
@@ -0,0 +1,380 @@
...
+        let humidity = self.relative_humidity.value;
+        let wind_speed = self.wind_kmh();
+        australian_apparent_temp(temp, humidity, wind_speed)
+    }
+
+    fn to_moment(&self) -> WeatherMoment {
+        let icon = short_forecast_to_icon(&self.short_forecast, !self.is_daytime);
+        let weather = Self::icon_to_word(icon);
+        WeatherMoment {
+            icon: short_forecast_to_icon(&self.short_forecast, !self.is_daytime),
```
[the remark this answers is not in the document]
Cognoscan: Good point! And I don't even need to switch to a reference since `WeatherIcon` is `Copy`. Extra call has been dropped.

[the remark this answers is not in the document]
bim9262: It's better to not copy if it's not necessary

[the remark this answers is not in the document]
Cognoscan: I'm not sure I understand why. `WeatherIcon` is 2 bytes in size and can get passed in directly as an argument in registers, versus following a reference. I certainly get wanting to avoid `clone`, but `copy`?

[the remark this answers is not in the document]
bim9262: Sorry, I think you're right. 😅 I hadn't looked again at what a `WeatherIcon` and for some reason in my head though it was larger.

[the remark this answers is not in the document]
Cognoscan: Totally fair, I certainly had to double-check it myself too.

src/blocks/weather/nws.rs
```diff
@@ -0,0 +1,380 @@
...
+            max.wind_direction = val.wind_direction;
+        }
+        max.wind = max.wind.max(val.wind);
+        max.wind_kmh = max.wind_kmh.max(val.wind_kmh);
+
+        // Min
+        min.temp = min.temp.min(val.temp);
+        min.apparent = min.apparent.min(val.apparent);
+        min.humidity = min.humidity.min(val.humidity);
+        if val.wind > min.wind {
```
bim9262, line 264: Should be <

Also the wind speeds can be set inside the if statement, without using the min/max functions.

Cognoscan: Fixed

src/blocks/weather/nws.rs
```diff
@@ -0,0 +1,380 @@
...
+    temperature: ApiValue,
+    relative_humidity: ApiValue,
+    wind_speed: ApiValue,
+    wind_direction: String,
+    short_forecast: String,
+}
+
+impl ApiForecast {
+    fn wind_direction(&self) -> f64 {
+        let dir = match self.wind_direction.as_str() {
```
bim9262, line 130: I'd make this return a `Some(f64)` so that "N" is `Some(0.0)` vs "" being `None`

src/blocks/weather/nws.rs
```diff
@@ -0,0 +1,380 @@
...
+
+fn combine_forecasts(data: &[ForecastAggregate], fin: WeatherMoment) -> Forecast {
+    let mut temp = 0.0;
+    let mut apparent = 0.0;
+    let mut humidity = 0.0;
+    let mut wind_north = 0.0;
+    let mut wind_east = 0.0;
+    let mut wind_kmh_north = 0.0;
+    let mut wind_kmh_east = 0.0;
+    let mut max = ForecastAggregate {
```
bim9262, line 219: I'd use `f64::MIN` for the `f64`s and `None` for the `wind_direction`

bim9262 (commented):
Sorry for the long delay...
Just a few minor tweaks I'd recommend.

src/blocks/weather/nws.rs
bim9262: (and `f64:MAX` for the min)

src/blocks/weather/nws.rs
Cognoscan: Done. I changed it so that "N" will return `Some(0.0)` and any unrecognized string will return `None`.

src/blocks/weather/nws.rs
Cognoscan: Done, the minimums all default to `f64::MAX` and the maximums all default to `f64::MIN`. Both start with `None` for `wind_direction`.

src/blocks/weather/nws.rs
```diff
@@ -0,0 +1,381 @@
...
+        wind_direction: None,
+    };
+    for val in data {
+        // Summations for averaging
+        temp += val.temp;
+        apparent += val.apparent;
+        humidity += val.humidity;
+        let (sin, cos) = val
+            .wind_direction
+            .unwrap_or_default()
```
bim9262, line 243: Ah and instead of treating None as north just don't use that reading to contribute to the average

Cognoscan: Got it, I updated the wind average calculation to match up with what the `average_wind` function does in the `weather.rs` code. For what it's worth, I do still want to keep this `combine_forecasts` function even though there's stuff like `average_wind(...)`, as it seems like a general function that merges together `ForecastAggregate`s would be more reusable down the line? Once I stop having so many little mistakes in it, anyway.

bim9262: Yes, I was actually thinking of (in a later PR) moving your new function to combine the forecast up the weather module and using it for the other weather backends.

I was thinking too about how we can make is so that we don't have to precompute the apparent weather and wind kmh for each forecast used to create the average/min/max

bim9262:
Thanks for your work on this :)

src/blocks/weather/nws.rs
Cognoscan: So, I was wondering about what the approach should be for measurement units as well. Wind is the only one that has two potentially different units at the same time, and confusingly to me, there's actually *three* different units in use in the code: m/s, km/h, and mph. This means I have at least one more tweak to make - my code calls the `australian_apparent_temp` function with a value in km/h, when it should actually be in m/s.

Apparent temperature also results in a measurement in degrees Celsius, and I should probably be converting that back into a temperature in degrees Fahrenheit *if* that's the chosen measurement unit. OpenWeatherMap seems to keep the feels-like temperature in Fahrenheit, so I'll need to do that to be consistent.

I'm not sure it matters all that much if I compute the wind speed twice, as it's a pretty infrequent calculation, but it might be good if there weren't two wind speed measurements in `WeatherAggregate` and `WeatherMoment` to begin with. My proposed change (probably not in this pull request) would be to add a `units: UnitSystem` field to `WeatherResult`, ditch the `wind_kmh` fields, and patch the display code to check the unit system and convert to km/h when someone uses the `wind_kmh` format keys.

Cognoscan: Whoops, I partially misunderstood the issue. It seems everyone uses m/s for wind when in metric units, and km/h is an alternate view of that. I still think the fix of patching the display code and lifting `wind_kmh` out of the weather data structures would work though.

Cognoscan: Does met.no have any support for imperial units at all? I thought it was fixed to being metric only.

bim9262: Sorry, right you are 😅

PR #2064, opened 2024-06-18
Support for Sway's scratchpad

qehgt:
It closes one item from #715


src/blocks/scratchpad.rs
```diff
@@ -0,0 +1,92 @@
+//! Scratchpad indicator
+//!
+//! # Configuration
+//!
+//! Key | Values | Default
+//! ----|--------|--------
+//! `format`          | A string to customise the output of this block      | `" $icon $count "`
+//! `hide_when_empty` | Hides the block when scratchpad contains no windows | `true`
```
MaxVerevkin, line 8: Technically, `hide_when_empty` is not needed, because the same effect can be achieved using

```toml
format = " $count.range(1..) |"
```

src/blocks/scratchpad.rs
```diff
@@ -0,0 +1,92 @@
...
+    loop {
+        let mut widget = Widget::new().with_format(format.clone());
+
+        let root_node = connection_for_tree
+            .get_tree()
+            .await
+            .error("could not get windows tree")?;
+        let count = count_scratchpad_windows(&root_node);
+
+        widget.state = State::Idle;
```
MaxVerevkin, line 67: The default widget state is idle, so this is not necessary.

qehgt: The redundant line was removed

src/blocks/scratchpad.rs
qehgt: Good to know, the code was updated

src/blocks/scratchpad.rs
```diff
@@ -0,0 +1,86 @@
...
+#[derive(Deserialize, Debug, SmartDefault)]
+#[serde(deny_unknown_fields, default)]
+pub struct Config {
+    pub format: FormatConfig,
+}
+
+fn count_scratchpad_windows(node: &Node) -> usize {
+    node.find_as_ref(|n| n.name == Some("__i3_scratch".to_string()))
+        .map(|node| node.floating_nodes.len())
+        .unwrap_or(0)
```
MaxVerevkin, line 36: To not allocate a new string on each iteration:

```suggestion
    node.find_as_ref(|n| n.name.as_deref() == Some("__i3_scratch"))
        .map_or(0, |node| node.floating_nodes.len())
```

jalvesaq:
Thank you! I'm starting to use scratchpad block, and it works.

Note the markdown `toml` block is unfinished in `scratchpad.rs`. You have to add the three backticks:

```diff
diff --git a/src/blocks/scratchpad.rs b/src/blocks/scratchpad.rs
index 2862c7b5..93610542 100644
--- a/src/blocks/scratchpad.rs
+++ b/src/blocks/scratchpad.rs
@@ -16,6 +16,7 @@
 //! ```toml
 //! [[block]]
 //! block = "scratchpad"
+//! ```
 //!
 //! # Icons Used
 //! - `scratchpad`
```

MaxVerevkin (approved):
Sorry for the late reply, everything looks good to me 👍

PR #2068, opened 2024-07-10
block[weather]: Remove redundant code

bim9262:
Moved `combine_forecasts` that @Cognoscan just added from nws to be used by all of the backends (now `Forecast::new`).

Added `ForecastAggregateSegment` which has the same members as `ForecastAggregate`, but makes all of the fields Optional since met.no does not promise that they're included in the response.

### DIFF
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -179,14 +179,25 @@ pub enum WeatherService {
     Nws(nws::Config),
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Default)]
 enum WeatherIcon {
-    Clear { is_night: bool },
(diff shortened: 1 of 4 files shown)


src/blocks/weather.rs
```diff
@@ -179,14 +179,25 @@ pub enum WeatherService {
     Nws(nws::Config),
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, SmartDefault)]
```
MaxVerevkin, line 182: Regular `Default` is capable of doing this, IIRC.

bim9262:
Replaced `SmartDefault` with `Default`.
Rebased to fix merge conflict.

PR #2069, opened 2024-07-10
block[weather]: Fix nws apparent temperature

bim9262:
australian_apparent_temp requires m/s not km/h.
convert temperature back to fahrenheit if original temperature is in fahrenheit

@Cognoscan does this capture the changes that you mentioned on your PR?

### DIFF
--- a/src/blocks/weather/nws.rs
+++ b/src/blocks/weather/nws.rs
@@ -177,8 +177,13 @@ impl ApiForecast {
             (self.temperature.value - 32.0) * 5.0 / 9.0
         };
         let humidity = self.relative_humidity.value;
-        let wind_speed = self.wind_kmh();
-        australian_apparent_temp(temp, humidity, wind_speed)
+        let wind_speed = self.wind_kmh() / 3.6;
+        let apparent = australian_apparent_temp(temp, humidity, wind_speed);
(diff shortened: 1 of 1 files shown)


Cognoscan:
Almost. There's one other possible issue, which is that other services in using metric units use m/s for wind, whereas NWS will return km/h when metric units are requested. I'm not sure if that matters or not.

bim9262:
Good call, I think it would be best to keep the speeds in standard units regardless of backend

src/blocks/weather/nws.rs
```diff
@@ -177,8 +185,13 @@ impl ApiForecast {
             (self.temperature.value - 32.0) * 5.0 / 9.0
         };
         let humidity = self.relative_humidity.value;
-        let wind_speed = self.wind_kmh();
-        australian_apparent_temp(temp, humidity, wind_speed)
+        let wind_speed = self.wind_kmh() / 3.6;
```
MaxVerevkin, line 188: Isn't this just `= self.wind_speed();`?

bim9262: Wind speed could be mph or m/s

MaxVerevkin: But wind_speed() is the function which you've added which always returns m/s?

PR #2078, opened 2024-08-14
Fix Bluetooth block not working when device has no icon

### LINKED ISSUE #2077: Bluetooth block not working when device has no icon
Bluetooth device "icon" property can be null for some devices (Asus Pen for example).  
When a device has no icon, the Bluetooth block will fail with the reason "failed to fetch device info" as the `icon` property is expected to be not null.

```
[2024-08-14T18:05:47Z DEBUG bluetooth] Found device with path OwnedObjectPath(ObjectPath("/org/bluez/hci0/dev_64_03_7F_27_4C_5A"))
[2024-08-14T18:05:47Z DEBUG bluetooth] failed to fetch device info, assuming device or bluez disappeared
```

Blendman974:
The icon can be null for some bluetooth devices, so ignore errors while fetching the icon and default it to "bluetooth".

Fix #2077

### DIFF
--- a/src/blocks/bluetooth.rs
+++ b/src/blocks/bluetooth.rs
@@ -300,25 +300,27 @@ impl DeviceMonitor {
     async fn get_device_info(&mut self) -> Option<DeviceInfo> {
         let device = self.device.as_ref()?;
 
-        let Ok((connected, icon, name)) = tokio::try_join!(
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
Thanks!

Issue #2088, opened 2024-10-03
Panic when stdin is closed


Jeidnx:
There is a panic with a very confusing error message when the stdin is closed.

How to reproduce:
```sh
echo "" | i3status-rs /some/valid/config
```

Here `events_stream.next()` returns None when stdin is closed, but this is never checked.
https://github.com/greshake/i3status-rust/blob/e18835c6a182295cd15b13b8200ba462d2097f97/src/lib.rs#L386

I couldn't figure out a way to handle this gracefully. The only approach that worked for me was to exit the program when stdin is closed.

ammgws:
Out of interest how are you actually encountering this in practice?

Jeidnx:
I recently started using https://github.com/mahkoh/jay and encountered this issue. I thought it would save someone else the two hours of troubleshooting it took me to find the problem. I'm still not sure why this is occurring with jay, and jay has some other problems with the i3bar protocol as well, but I'm planning on addressing those.

NextWork123:
> I recently started using https://github.com/mahkoh/jay and encountered this issue. I thought it would save someone else the two hours of troubleshooting it took me to find the problem. I'm still not sure why this is occurring with jay, and jay has some other problems with the i3bar protocol as well, but I'm planning on addressing those.

i have the same issue :/ it panics

Jeidnx:
@NextWork123 Just to be clear that is an issue with jay. Jay currently doesn't have support for i3bar click events (afaict) and the parser isn't implemented properly. I didn't bother to create an issue there because i stopped using it. If you want it fixed, you probably should.

MaxVerevkin:
Should be fixed now!

PR #2089, opened 2024-10-05
Error handling when stdin is closed

Jeidnx:
As described in #2088 

I would have liked to handle this gracefully but couldn't find a way to do that without changing large parts of the event loop.


MaxVerevkin:
I think I found a more graceful solution :)

PR #2093, opened 2024-10-20
Reduce amount of blocking in blocks

bim9262:


src/util.rs
```diff
@@ -7,6 +7,10 @@ use tokio::process::Command;
 
 use crate::errors::*;
 
+pub async fn file_exists(path: impl AsRef<Path>) -> bool {
+    tokio::fs::metadata(path).await.is_ok()
```
MaxVerevkin, line 11: Is there a reason to choose this over `tokio::fs::try_exists`?

bim9262: This matches the implementation that it was replacing `exists`: https://doc.rust-lang.org/nightly/src/std/path.rs.html#3030-3032

Perhaps all of the `exists` should be `try_exists` (either async or sync, spending on the context though)?

ammgws:
Let's take this for now

PR #2106, opened 2024-11-17
Update quick-xml and thiserror

MaxVerevkin:


MaxVerevkin:
I have not tested the calendar block, which depends on quick-xml.

bim9262 (approved):
I tested out the calendar block, and everything still seems ok.

PR #2108, opened 2024-11-19
block[custom]: detect file delete/move events

### LINKED ISSUE #2107: Watching a directory doesn't detect file removals
Watching a directory with `watch_files = ["/path/to/some/directory"]` detects new files (create) but not file removals (delete).

To reproduce, create a `~/testing` directory and add the following block to the config:

```
[[block]]
  block = "custom"
  watch_files = ["~/testing/"]
  interval = "once"
  command = """
    if [[ -e ~/testing/foo ]]
    then echo yes
    else echo no
    fi
    """
```

Restart `i3status-rust` and:

1. Touch a file in that directory: `touch ~/testing/foo`.
2. Wait for the statusbar to update, the block should display "yes".
3. Run `rm ~/testing/foo`.
4. The status won't update.

Stebalien:
fixes #2107

### DIFF
--- a/src/blocks/custom.rs
+++ b/src/blocks/custom.rs
@@ -173,7 +173,13 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
             for file in files {
                 let file = file.expand()?;
                 watches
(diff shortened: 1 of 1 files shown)


bim9262:
Looks good to me. Thanks @Stebalien!

PR #2114, opened 2024-12-09
formatting(eng): add show parameter

### LINKED ISSUE #2113: maildir block: add configuration "hide_if_total_is_zero"
It would be both consistent<sup>1</sup> and useful for maildir notifications to only appear when mail is present. I have not seen the source, but it should *successfully* return nothing if `status = 0` and `hide_if_total_is_zero = true`.

<sup>1</sup> The github block has an identical configuration.

MaxVerevkin:
Fixes #2113

Example usage (will show the icon but only if $status >= 1):

```toml
[[block]]
block = "maildir"
format = " $icon $status.eng(range:1..,show:false)|"
```


bim9262 (approved):
looks good!

PR #2116, opened 2024-12-10
remove deprecated apt, pacman and dnf blocks

MaxVerevkin:
In #1988 these blocks were replaced by the packages block. They are deprecated since v0.33 (2024-02-19).


ammgws (approved):

bim9262 (approved):
lgtm too!

PR #2119, opened 2024-12-17
block[disk_iostats]: new block

emersion:
Reads disk stats from sysfs.

TODO:

- [x] Write docs
- [x] Maybe `io_stats` would be a better name? Or `block_device_stats`? Not sure.

See: https://docs.kernel.org/admin-guide/iostats.html

### DIFF
--- a/src/blocks/disk_iostats.rs
+++ b/src/blocks/disk_iostats.rs
@@ -1,0 +1,177 @@
+//! Disk I/O statistics
+//!
+//! # Configuration
+//!
+//! Key | Values | Default
+//! ----|--------|--------
+//! `device` | Block device name to monitor (as specified in `/dev/`) | If not set, device will be automatically selected every `interval`
+//! `format` | A string to customise the output of this block. See below for available placeholders. | `" $icon $speed_read.eng(prefix:K) $speed_write.eng(prefix:K) "`
+//! `interval` | Update interval in seconds | `2`
+//! `missing_format` | Same as `format` but for when the device is missing | `" × "`
+//!
+//! Placeholder | Value | Type   | Unit
+//! ------------|-------|--------|-------
+//! `icon` | A static icon | Icon | -
+//! `device` | The name of device | Text | -
+//! `speed_read` | Read speed | Number | Bytes per second
+//! `speed_write` | Write speed | Number | Bytes per second
+//!
+//! # Examples
+//!
+//! ```toml
+//! [[block]]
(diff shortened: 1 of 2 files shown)


emersion:
Does any of the maintainers have an opinion on this PR? Is the idea good? Is the name fine? Is the approach taken good enough?

MaxVerevkin:
Looks good! Didn't have enough time to explicitly write this sooner, waiting for the docs :)

emersion:
Added docs!

src/blocks/disk_stats.rs
```diff
@@ -0,0 +1,168 @@
+//! Disk I/O statistics
+//!
+//! # Configuration
+//!
+//! Key | Values | Default
+//! ----|--------|--------
+//! `device` | Block device name to monitor (as specified in `/dev/`) | If not set, device will be automatically selected every `interval`
+//! `format` | A string to customise the output of this block. See below for available placeholders. | `" $icon $speed_read.eng(prefix:K) $speed_write.eng(prefix:K) "`
+//! `interval` | Update interval in seconds | `2`
```
MaxVerevkin, line 9: `missing_format` is missing :smile:

emersion: Oops, fixed!

ammgws:
Naming nit - perhaps `disk_iostats` is better (sincec one might think disk space is a "stat" that should be in `disk_stats`)

emersion:
Sure, renamed

ammgws (approved):

src/blocks/disk_iostats.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+async fn read_stats(device: &str) -> Result<Stats> {
+    let raw = read_file(Path::new(BLOCK_DEVICES_PATH).join(device).join("stat"))
+        .await
+        .error("Failed to read stat file")?;
+    let fields: Vec<&str> = raw.split_whitespace().collect();
+    Ok(Stats {
+        sectors_read: fields[2].parse().error("Failed to parse sectors read")?,
+        sectors_written: fields[6].parse().error("Failed to parse sectors written")?,
```
MaxVerevkin, line 156: Let's use `fields.get(n).error("bad stat file")?` in case for some reason there are not enough fields

emersion: Done

MaxVerevkin (approved):
Thanks!

PR #2120, opened 2024-12-22
Femboy theme :3

scd31:
it's pink and cute :3 

image


bim9262:
@scd31, I don't know if this is necessarily a good fit as I don't think there's a common femboy aesthetic. For example others may want femboy to match https://www.color-hex.com/color-palette/1030161 or be more like https://github.com/Kouromo/astolfo-vscode-theme

The other themes are common themes that are used by many other applications for example: https://github.com/morhetz/gruvbox and https://draculatheme.com/

scd31:
Aw okay :c

Would you be open to it if it were renamed to e.g. bubblegum?

bim9262:
It's not the name rather that it's not a standard theme. Here is a non-exhaustive list of themes with their hex values. https://gogh-co.github.io/Gogh/. I'd see if you can get one of those pallets to work, perhaps Japanesque, Base2Tone Suburb, Base4Tone Classic Q, Base4Tone Modern N, or Website?

scd31:
I just wanted it to be pink, but if you don't think it would be a good addition to the project then I understand!

PR #2122, opened 2025-01-03
Convert gen-screenshots from fish to bash

bim9262:
Added in check to see if the screenshot already exist, and skips it if so.


ammgws (approved):

ammgws:
Does it still work as expected? It's been years since last generated instead I think

bim9262:
Yes, the script still works. However depending on the fonts you have you may need to tweak the theme's separator

For example this is one of the generated themes (solarized-dark) on my computer without any tweaking.
solarized-dark

bim9262:
That issue of the the theme separators looking off is not due to a change in the script though.

Issue #2084, opened 2024-09-07
Formatting: treat `key` as a shortcut for `key:true`


MaxVerevkin:
For example `.dur(hms)` instead of `.dur(hms:true)`.

bim9262:
I like the idea, but I think we'd want to only do that, if the parameter is a boolean. Otherwise we'd get odd errors for args like `arg_takes_number`, `arg_takes_string`

MaxVerevkin:
Funny thing is that the comments in `parse.rs` suggest this was the intention all along

https://github.com/greshake/i3status-rust/blob/a2704aedebf8818628977cde99a5a1886e63e09f/src/formatting/parse.rs#L114-L116

PR #2126, opened 2025-01-10
Formatting: treat key as a shortcut for key:true

bim9262:
Fixes: #2084

### DIFF
--- a/src/formatting/formatter/duration.rs
+++ b/src/formatting/formatter/duration.rs
@@ -51,39 +51,27 @@ impl DurationFormatter {
         for arg in args {
             match arg.key {
                 "hms" => {
-                    hms = arg.val.parse().ok().error("hms must be true or false")?;
+                    hms = arg.parse_value()?;
                 }
                 "max_unit" => {
-                    max_unit = Some(arg.val);
+                    max_unit = Some(arg.val.error("max_unit must be specified")?);
                 }
                 "min_unit" => {
-                    min_unit = arg.val;
+                    min_unit = arg.val.error("min_unit must be specified")?;
                 }
                 "units" => {
-                    units = Some(
-                        arg.val
-                            .parse()
-                            .ok()
-                            .error("units must be a positive integer")?,
-                    );
+                    units = Some(arg.parse_value()?);
                 }
                 "round_up" => {
-                    round_up = arg
-                        .val
-                        .parse()
-                        .ok()
-                        .error("round_up must be true or false")?;
+                    round_up = arg.parse_value()?;
                 }
                 "unit_space" => {
-                    unit_has_space = arg
-                        .val
-                        .parse()
-                        .ok()
-                        .error("unit_space must be true or false")?;
+                    unit_has_space = arg.parse_value()?;
                 }
                 "pad_with" => {
-                    if arg.val.graphemes(true).count() < 2 {
-                        pad_with = Some(Cow::Owned(arg.val.into()));
+                    let pad_with_str = arg.val.error("pad_with must be specified")?;
+                    if pad_with_str.graphemes(true).count() < 2 {
+                        pad_with = Some(Cow::Owned(pad_with_str.into()));
                     } else {
                         return Err(Error::new(
                             "pad_with must be an empty string or a single character",
@@ -91,7 +79,7 @@ impl DurationFormatter {
                     };
                 }
                 "leading_zeroes" => {
-                    leading_zeroes = arg.val.parse().ok().error("units must be true or false")?;
+                    leading_zeroes = arg.parse_value()?;
                 }
 
                 _ => return Err(Error::new(format!("Unexpected argument {:?}", arg.key))),
--- a/src/formatting/parse.rs
+++ b/src/formatting/parse.rs
@@ -1,3 +1,5 @@
+use std::{any::TypeId, str::FromStr};
+
 use nom::{
     branch::alt,
     bytes::complete::{escaped_transform, tag, take_while, take_while1},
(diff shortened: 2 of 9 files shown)


src/formatting/parse.rs
```diff
@@ -94,11 +94,17 @@ fn arg1(i: &str) -> IResult<&str, &str, PError> {
...
 
 // `key:val`
+// `abc`
 fn parse_arg(i: &str) -> IResult<&str, Arg, PError> {
-    map(
-        separated_pair(alphanum1, cut(char(':')), cut(arg1)),
-        |(key, val)| Arg { key, val },
-    )(i)
+    alt((
+        map(separated_pair(alphanum1, char(':'), arg1), |(key, val)| {
```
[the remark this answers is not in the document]
bim9262: cut transforms an Err::Error (recoverable) to Err::Failure (unrecoverable)

This commits the parse result, preventing alternative branch paths like with nom::branch::alt.

https://docs.rs/nom/latest/nom/combinator/fn.cut.html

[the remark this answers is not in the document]
MaxVerevkin: I've read the doc description, if I understand it correctly, `cut` is needed exactly because `alt` is used here. So that `key:<invalid>` does not get parsed as `key` + `:<invalid>`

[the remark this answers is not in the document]
bim9262: Thanks for the catch and explanation. I added in another test for `parse_arg` to verify what I had was incorrect, I missed it when testing just `parse_args` - besides having the wrong input to that function before.

src/formatting/formatter/duration.rs
```diff
@@ -51,47 +51,56 @@ impl DurationFormatter {
...
                 "unit_space" => {
-                    unit_has_space = arg
-                        .val
-                        .parse()
-                        .ok()
-                        .error("unit_space must be true or false")?;
+                    unit_has_space = match arg.val {
+                        Some(val) => val.parse().ok().error("unit_space must be true or false")?,
+                        None => true,
+                    };
```
MaxVerevkin, line 84: I think it would be convenient to have an `fn Arg::parse_value<T: FromStr + 'static>(&self) -> Result<T>`  which would 1. call `.parse()` 2. automatically create an error message 3. check if `TypeId::of::<T>() == TypeId::of::<bool>()` and return `true` in that case.

MaxVerevkin: Then this and many other instances of this pattern would become `unit_has_space = arg.parse_value()?,`

bim9262: I came up with:

```rust
impl Arg<'_> {
    pub fn parse_value<T: FromStr + 'static>(&self) -> Result<T> {
        if TypeId::of::<T>() == TypeId::of::<bool>() && self.val.is_none() {
            "true"
                .parse()
                .map_err(|_| Error::new(format!("invalid value for argument '{}'", self.key)))
        } else {
            self.val
                .ok_or_else(|| Error::new(format!("missing value for argument '{}'", self.key)))
                .and_then(|val| {
                    val.parse().map_err(|_| {
                        Error::new(format!("invalid value for argument '{}'", self.key))
                    })
                })
        }
    }
}
```

I wanted to do something like:

```rust
impl Arg<'_> {
    pub fn parse_value<T: FromStr + 'static>(&self) -> Result<T> {
        if TypeId::of::<T>() == TypeId::of::<bool>() && self.val.is_none() {
            Ok(true)
        } else {
            self.val
                .ok_or_else(|| Error::new(format!("missing value for argument '{}'", self.key)))
                .and_then(|val| {
                    val.parse().map_err(|_| {
                        Error::new(format!("invalid value for argument '{}'", self.key))
                    })
                })
        }
    }
    ```
    
    but got this error, even though it seems like it should be ok since the type of T is bool:
    
    ```
    error[E0308]: mismatched types
   --> src/formatting/parse.rs:24:16
    |
22  |     pub fn parse_value<T: FromStr + 'static>(&self) -> Result<T> {
    |                        - expected this type parameter
23  |         if TypeId::of::<T>() == TypeId::of::<bool>() && self.val.is_none() {
24  |             Ok(true)
    |             -- ^^^^ expected type parameter `T`, found `bool`
    |             |
    |             arguments to this enum variant are incorrect
    |
    = note: expected type parameter `T`
                         found type `bool`
help: the type constructed contains `bool` due to the type of the argument passed
   --> src/formatting/parse.rs:24:13
    |
24  |             Ok(true)
    |             ^^^----^
    |                |
    |                this argument influences the type of `Ok`
    ```

Comment (MaxVerevkin) [reply src/formatting/formatter/duration.rs:84]: Yeah, returning `true` would not work because Rust's type system does not understand that that branch is taken only if T is bool and it does not perform any type narrowing.

Also, to not lose error context:

```rust
impl Arg<'_> {
    pub fn parse_value<T>(&self) -> Result<T>
    where
        T: FromStr + 'static,
        T::Err: StdError + Send + Sync + 'static,
    {
        if TypeId::of::<T>() == TypeId::of::<bool>() && self.val.is_none() {
            Ok("true".parse().expect("'true' is valid bool"))
        } else {
            self.val
                .or_error(|| format!("missing value for argument '{}'", self.key))?
                .parse()
                .or_error(|| format!("invalid value for argument '{}'", self.key))
        }
    }
}
```

MaxVerevkin:
One last thing, this should be mentioned _somewhere_ in the docs.

bim9262:
> One last thing, this should be mentioned _somewhere_ in the docs.

I added a note in `formatting.rs`

MaxVerevkin (approved):

Issue #1395, opened 2022-01-14
Keyboard block doesn't catch layout changes


cfsmp3:
For me, at least.

dbus-monitor shows this event for me - totally different info:

```
signal time=1642201033.914655 sender=:1.85 -> destination=(null destination) serial=1073 path=/org/freedesktop/portal/desktop; interface=org.freedesktop.impl.portal.Settings; member=SettingChanged
   string "org.gnome.desktop.input-sources"
   string "mru-sources"
   variant       array [
         struct {
            string "xkb"
            string "es"
         }
         struct {
            string "xkb"
            string "us"
         }
      ]
signal time=1642201033.914770 sender=:1.84 -> destination=(null destination) serial=4586 path=/org/freedesktop/portal/desktop; interface=org.freedesktop.portal.Settings; member=SettingChanged
   string "org.gnome.desktop.input-sources"
   string "mru-sources"
   variant       array [
         struct {
            string "xkb"
            string "es"
         }
         struct {
            string "xkb"
            string "us"
         }
      ]
```

MaxVerevkin:
Which desktop environment do you use? Which `driver` have you set in your config?

cfsmp3:
> Do you use `sway` or `i3`? Which `driver` have you set in your config? How do you change your layout?

i3
- kbddbus doesn't work ("is kbdddaemon bus running?")
- localebus -> Doesn't work (that path just doesn't get updated)

I haven't tried the other two as polling here is out of the question :-) 

I change the layout by pressing Super + Alt + Backspace , I haven't looked into what that triggers (I was just casually looking into this block)

MaxVerevkin:
> kbddbus doesn't work ("is kbdddaemon bus running?")

So, do you run https://github.com/qnikst/kbdd?

cfsmp3:
> > kbddbus doesn't work ("is kbdddaemon bus running?")
> 
> So, do you run https://github.com/qnikst/kbdd?

No, I don't need to have per-window setting. I was just mentioning that I tried the driver, I didn't expect it to work :-)

BTW I tried 
`` 
setxkbmap us
```
And it doesn't seem to trigger any D-Bus notification (I might be filtering wrong).

MaxVerevkin:
AFAIK there is no standard way to monitor kbd layout on X11.

I imagine you can create a script that somehow toggles the layout and sends a signal to `i3status-rs` to update. In this case you can use `setxkbmap` driver with a large interval.

[...]

PR #2133, opened 2025-02-06
keyboard_layout: Add xkb events support

amiremohamadi:
Fixes: #1395.

TLDR: catch the keyboard layout events on x11.


amiremohamadi:
fixed the pipeline.

amiremohamadi:
Would appreciate a review. Thanks!
cc: @bim9262

src/blocks/keyboard_layout/xkb_event.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    let layout = prefix
+        .split(|&c| c == b'+')
+        .skip(1) // layout names start from index 1
+        .nth(index)
+        .error("Index out of range")?
+        .to_vec();
+
+    String::from_utf8(layout).map_err(|e| Error::new(e.to_string()))
+}
```
MaxVerevkin, line 37: A non-allocating version:

```suggestion
fn parse_layout(buf: &[u8], index: usize) -> Result<&str> {
    let colon_i = buf.iter().position(|c| *c == b':').unwrap_or(buf.len());

    let layout = buf[..colon_i]
        .split(|&c| c == b'+')
        .skip(1) // layout names start from index 1
        .nth(index)
        .context("Index out of range")?;

    std::str::from_utf8(layout).context("non utf8 layout")
}
```

amiremohamadi: perfect! applied the changes.

MaxVerevkin:
Does this make `setxkbmap` and possibly other drivers obsolete?

amiremohamadi:
yeah, `xkbswitch` and `setxkbmap` can be replaced with this one -- as it's more resource-friendly and optimized. but we can still keep them for backward compatibility

Cargo.toml
```diff
@@ -79,6 +79,7 @@ unicode-segmentation = "1.10.1"
 wayrs-client = { version = "1.0", features = ["tokio"] }
 wayrs-protocols = { version = "0.14", features = ["wlr-foreign-toplevel-management-unstable-v1"] }
 zbus = { version = "5", default-features = false, features = ["tokio"] }
+x11rb-async = { version = "0.13.1", features = ["xkb", "allow-unsafe-code"] }
```
MaxVerevkin, line 82: I _think_ the `allow-unsafe-code` feature is not used here, needlessly adds a dependency on `lidxcb` and can be removed?

amiremohamadi: oops! right!

bim9262:
> yeah, `xkbswitch` and `setxkbmap` can be replaced with this one -- as it's more resource-friendly and optimized. but we can still keep them for backward compatibility

This driver could replace those others then, but keep the old driver names and point them to your me driver, and update the docs to reflect that.

amiremohamadi:
@bim9262 actually the old drivers require and interval and the new one doesn't. so just aliasing the names would break their functionality in the old configurations. am i right?

MaxVerevkin:
Isn't async event listening strictly 'better' than interval polling? So the interval can just be ignored.

bim9262:
If none of the drivers require interval make a comment that it's depreciated and will be removed in v0.35.0.

amiremohamadi:
applied the changes. let me know if everything is ok.

MaxVerevkin:
I think we should also deprecate the driver aliases immediately and remove them together with the interval option.

amiremohamadi:
i agree on deprecating them at this point if both of you agree.

@MaxVerevkin 
@bim9262

MaxVerevkin:
Sorry if I wasn't clear, I meant _deprecating_ them now and removing _later_.

amiremohamadi:
ah, got it! you mean marking them as *deprecated* just in the docs?

MaxVerevkin:
Yes

bim9262:
Sorry, I'd suggested getting rid of them now, but still accepting the old driver names as aliases for the new driver.

src/blocks/keyboard_layout/xkb_event.rs
```diff
@@ -0,0 +1,141 @@
...
+    let colon_i = buf.iter().position(|c| *c == b':').unwrap_or(buf.len());
+    let layout = buf[..colon_i]
+        .split(|&c| c == b'+')
+        .skip(1) // layout names start from index 1
+        .nth(index)
+        .error("Index out of range")?;
+    std::str::from_utf8(layout).error("non utf8 layout")
+}
+
+async fn get_layout(connection: &RustConnection) -> Result<String, Box<dyn std::error::Error>> {
```
MaxVerevkin, line 33: Why not `Result<String>`? I would also prefer giving a context description like

```rust
    let xkb_state = connection
        .xkb_get_state(ID::USE_CORE_KBD.into())
        .await
        .error("xkb_get_state failed")?
        .reply()
        .await
        .error("xkb_get_state reply failed")?;
```

src/blocks/keyboard_layout/xkb_event.rs
```diff
@@ -0,0 +1,141 @@
...
+                MapPart::default(),
+                MapPart::default(),
+                &SelectEventsAux::new(),
+            )
+            .await
+            .error("Failed to select events")?;
+
+        let cur_layout = get_layout(&connection)
+            .await
+            .map_err(|_| Error::new("Failed to get current layout".to_string()))?;
```
MaxVerevkin, line 108:
```suggestion
            .error("Failed to get current layout")?;
```

src/blocks/keyboard_layout.rs
```diff
@@ -1,49 +1,36 @@
...
+//!
 //! Which of these methods is appropriate will depend on your system setup.
 //!
 //! # Configuration
 //!
 //! Key | Values | Default
 //! ----|--------|--------
-//! `driver` | One of `"setxkbmap"`, `"xkbswitch"`, `"localebus"`, `"kbddbus"` or `"sway"`, depending on your system. | `"setxkbmap"`
-//! `interval` | Update interval, in seconds. Only used by the `"setxkbmap"` driver. | `60`
+//! `driver` | One of `"xkbevent"`, `"localebus"`, `"kbddbus"` or `"sway"`, depending on your system. | `"xkbevent"`
```
bim9262, line 19: I'd leave `"setxkbmap"`, `"xkbswitch"`, listed even though they're depricated.

src/blocks/keyboard_layout.rs
```diff
@@ -1,49 +1,36 @@
 //! Keyboard layout indicator
 //!
 //! Four drivers are available:
```
bim9262, line 3: Looks like there are 6 now

src/blocks/keyboard_layout/xkb_event.rs
```diff
@@ -0,0 +1,151 @@
...
+                MapPart::default(),
+                MapPart::default(),
+                &SelectEventsAux::new(),
+            )
+            .await
+            .error("Failed to select events")?;
+
+        let cur_layout = get_layout(&connection)
+            .await
+            .error("Failed to get current layout")?;
```
MaxVerevkin, line 118: I think there is no need to store current layout in the struct and this line can be moved down to `get_info` implementation. This would also simplify `wait_for_change`.

amiremohamadi: so you mean moving the whole logic of `wait_for_change` to `get_info`? because calling `get_layout` function has a cost and we only wanna call it for specific events (`xkb::StatePart::GROUP_STATE`)

MaxVerevkin: `get_info` is only called when `wait_for_change` returns and on the first render. So it is okay to perform costly operations in it.

MaxVerevkin: So instead of calling `get_layout` from two locations we can call it from only one.

MaxVerevkin:
Since we don't have polling drivers anymore, `_ = api.wait_for_update_request() => (),` [line](https://github.com/greshake/i3status-rust/blob/b5a60d6f77eb9883d02d1edae26dd48610c9a42e/src/blocks/keyboard_layout.rs#L164) (and the entire `select!`) can be removed.

Issue #2135, opened 2025-02-10
packages: add `xbps` support


thewalkingforest:
Hello!

Xbps is the package manger for [Void Linux](https://voidlinux.org).  Having support for it built into the bar instead of relying on a script would be nice.  It would go into the `packages` block with `apt`, `pacman`, etc.

Get updates command: `$ xbps-install -Mun`
option meanings:
  - `M`: Repo data is stored in memory, so root privileges are not needed
  - `u`: Update packages
  - `n`: Set dry-run mode

PR #2136, opened 2025-02-10
Add xbps package manager to `packages` block

thewalkingforest:
Implements #2135 and updates docs to match.


bim9262 (commented):
`xbps` needs to be added to `cspell.yaml`

bim9262 (approved):
lgtm

PR #2137, opened 2025-02-12
Add global geolocator

### LINKED ISSUE #2075: Add support for IP2Location.io API
It would be good to have an alternative or secondary IP geolocation API such as IP2Location.io.

The [free API plan](https://www.ip2location.io/pricing) from IP2Location.io allows users to query up to 30,000 IP addresses per month.

If using the API without an API key, there is a limit of 500 queries per day.

bim9262:
Resolves #2075

TODO
- [x]  Add documentation

### DIFF
--- a/src/blocks.rs
+++ b/src/blocks.rs
@@ -40,6 +40,7 @@ use std::time::Duration;
 
 use crate::click::MouseButton;
 use crate::errors::*;
+use crate::geolocator::{Geolocator, IPAddressInfo};
 use crate::widget::Widget;
 use crate::{BoxedFuture, Request, RequestCmd};
 
@@ -208,6 +209,7 @@ pub struct CommonApi {
     pub(crate) update_request: Arc<Notify>,
     pub(crate) request_sender: mpsc::UnboundedSender<Request>,
     pub(crate) error_interval: Duration,
+    pub(crate) geolocator: Arc<Geolocator>,
 }
 
 impl CommonApi {
@@ -266,5 +268,18 @@ impl CommonApi {
 
     pub async fn wait_for_update_request(&self) {
         self.update_request.notified().await;
+    }
+
+    fn locator_name(&self) -> Cow<'static, str> {
+        self.geolocator.name()
+    }
+
+    /// No-op if last API call was made in the last `interval` seconds.
+    pub async fn find_ip_location(
+        &self,
+        client: &reqwest::Client,
+        interval: Duration,
+    ) -> Result<IPAddressInfo> {
+        self.geolocator.find_ip_location(client, interval).await
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,6 +13,7 @@ pub mod config;
 pub mod errors;
 pub mod escape;
 pub mod formatting;
+pub mod geolocator;
 pub mod icons;
 mod netlink;
 pub mod protocol;
@@ -268,6 +269,7 @@ impl BarState {
             update_request: update_request.clone(),
             request_sender: self.request_sender.clone(),
             error_interval: Duration::from_secs(block_config.common.error_interval),
+            geolocator: self.config.geolocator.clone(),
         };
 
         let error_format = block_config
--- a/README.md
+++ b/README.md
@@ -59,12 +59,14 @@ short_error_message | The short error message, if available
 - [Blocks](https://docs.rs/i3status-rs/latest/i3status_rs/blocks/index.html)
 - [Formatting](https://docs.rs/i3status-rs/latest/i3status_rs/formatting/index.html)
 - [Themes and Icons](https://github.com/greshake/i3status-rust/blob/v0.32.0/doc/themes.md)
+- [Geolocator](https://docs.rs/i3status-rs/latest/i3status_rs/geolocator/index.html)
 
 #### Master
 
 - [Blocks](https://greshake.github.io/i3status-rust/i3status_rs/blocks/index.html)
 - [Formatting](https://greshake.github.io/i3status-rust/i3status_rs/formatting/index.html)
 - [Themes and Icons](doc/themes.md)
+- [Geolocator](https://greshake.github.io/i3status-rust/i3status_rs/geolocator/index.html)
 
 ## Integrate it into i3/sway
 
--- a/src/blocks/external_ip.rs
+++ b/src/blocks/external_ip.rs
@@ -65,8 +65,6 @@ use zbus::MatchRule;
 use super::prelude::*;
 use crate::util::{country_flag_from_iso_code, new_system_dbus_connection};
 
-const API_ENDPOINT: &str = "https://ipapi.co/json/";
-
 #[derive(Deserialize, Debug, SmartDefault)]
 #[serde(deny_unknown_fields, default)]
 pub struct Config {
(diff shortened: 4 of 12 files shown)


src/locator.rs
```diff
@@ -0,0 +1,143 @@
...
+    {
+        let guard = LAST_AUTOLOCATE.lock().unwrap();
+        if let Some(cached) = &*guard {
+            if cached.timestamp.elapsed() < interval {
+                return Ok(cached.location.clone());
+            }
+        }
+    }
+
+    let locator = LOCATOR.lock().unwrap().clone();
```
bim9262, line 131: perhaps another kind of mutex/rwlock could be used across the await so that it doesn't have to be cloned...

src/locator.rs
```diff
@@ -0,0 +1,143 @@
...
+
+use async_trait::async_trait;
+use serde::Deserialize;
+use smart_default::SmartDefault;
+
+mod ip2location;
+mod ipapi;
+
+static LAST_AUTOLOCATE: Mutex<Option<AutolocateResult>> = Mutex::new(None);
+static LOCATOR: LazyLock<Mutex<Locator>> = LazyLock::new(|| Mutex::new(Locator::default()));
```
bim9262, line 15: instead of having a global locator, perhaps it could live in another existing construct like api?

MaxVerevkin:
We don't need `async_trait` here :)

```diff
diff --git a/src/locator.rs b/src/locator.rs
index 9aa6b9773..f5ebd654a 100644
--- a/src/locator.rs
+++ b/src/locator.rs
@@ -1,10 +1,10 @@
 use crate::errors::{Error, ErrorContext, Result, StdError};
 use std::borrow::Cow;
 use std::fmt;
+use std::future::Future;
 use std::sync::{Arc, LazyLock, Mutex};
 use std::time::{Duration, Instant};
 
-use async_trait::async_trait;
 use serde::Deserialize;
 use smart_default::SmartDefault;
 
@@ -82,18 +82,16 @@ pub enum LocatorDriver {
     Ip2Location,
 }
 
-#[async_trait]
 pub trait Backend {
     fn name(&self) -> Cow<'static, str>;
 
-    async fn get_info(
+    fn get_info(
         &self,
         client: &reqwest::Client,
         api_key: &Option<String>,
-    ) -> Result<IPAddressInfo>;
+    ) -> impl Future<Output = Result<IPAddressInfo>>;
 }
 
-#[async_trait]
 impl Backend for LocatorDriver {
     fn name(&self) -> Cow<'static, str> {
         match self {
diff --git a/src/locator/ip2location.rs b/src/locator/ip2location.rs
index e4b998c83..b7019075f 100644
--- a/src/locator/ip2location.rs
+++ b/src/locator/ip2location.rs
@@ -185,7 +185,6 @@ impl StdError for ApiError {}
 #[derive(Deserialize, Debug, Default, Clone)]
 pub struct Ip2Location;
 
-#[async_trait]
 impl Backend for Ip2Location {
     fn name(&self) -> Cow<'static, str> {
         Cow::Borrowed("ip2location.io")
diff --git a/src/locator/ipapi.rs b/src/locator/ipapi.rs
index 898b4bf70..7e9bbfe02 100644
--- a/src/locator/ipapi.rs
+++ b/src/locator/ipapi.rs
@@ -90,7 +90,6 @@ impl StdError for ApiError {}
 #[derive(Deserialize, Debug, Default, Clone)]
 pub struct Ipapi;
 
-#[async_trait]
 impl Backend for Ipapi {
     fn name(&self) -> Cow<'static, str> {
         Cow::Borrowed("ipapi.co")
```

src/locator.rs
MaxVerevkin: I'm okay with it being global

src/locator.rs
MaxVerevkin: If we store `Locator::api_key` as an `Arc<str>`, this clone will be cheap.

src/locator.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub org: Option<String>,
+}
+
+#[derive(Deserialize, Debug, Default, Clone)]
+#[serde(default, deny_unknown_fields)]
+pub struct Locator {
+    pub driver: LocatorDriver,
+    #[serde(default = "getenv_api_key")]
+    api_key: Option<String>,
+}
```
MaxVerevkin, line 71: Since not all drivers need a key, maybe `WeatherService`-approach would be better, where each driver has its own set of options?

src/locator.rs
bim9262: Wouldn't it need to be `Arc<Option<String>>`?

MaxVerevkin: or `Option<Arc<str>>`. Or `Locator` can be wrapped in `Arc`.

src/locator.rs
MaxVerevkin: On a second thought, storing `Locator` or `Arc<Locator>` in api may be slightly better, `set_global_locator` and the fact that it can be called from anywhere feels a bit weird.

bim9262: Moved it over to the `CommonApi`

src/locator/ip2location.rs
```diff
@@ -0,0 +1,238 @@
+use super::*;
+
+const IP_API_URL: &str = "https://api.ip2location.io/";
+pub(super) const API_KEY_ENV: &str = "IP2LOCATION_API_KEY";
+
+#[derive(Deserialize, Debug, Default, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct Config {
+    #[serde(default = "getenv_api_key")]
+    pub api_key: Arc<Option<String>>,
```
MaxVerevkin, line 10: Since `Locator` is already wrapped in `Arc`, this is not necessary.

src/blocks.rs
```diff
@@ -207,6 +208,8 @@ pub struct CommonApi {
     pub(crate) update_request: Arc<Notify>,
     pub(crate) request_sender: mpsc::UnboundedSender<Request>,
     pub(crate) error_interval: Duration,
+    pub(crate) locator: Arc<Locator>,
+    pub(crate) last_autolocate: Arc<Mutex<Option<AutolocateResult>>>,
```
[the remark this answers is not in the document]
bim9262: Inside of the enun?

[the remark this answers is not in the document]
MaxVerevkin: More like

```rust
pub struct Locator {
    backend: LocatorBackend,
    last_autolocate: Mutex<Option<AutolocateResult>>,
}

#[derive(Deserialize, Debug, SmartDefault, Clone)]
#[serde(tag = "name", rename_all = "lowercase", deny_unknown_fields)]
pub enum LocatorBackend {
    #[default]
    Ipapi(ipapi::Config),
    Ip2Location(ip2location::Config),
}
```

Also I don't think `trait Backend` is actually needed as a trait?

bim9262:
fixed merge conflict

src/lib.rs
```diff
@@ -268,6 +270,7 @@ impl BarState {
             update_request: update_request.clone(),
             request_sender: self.request_sender.clone(),
             error_interval: Duration::from_secs(block_config.common.error_interval),
+            geolocator: Arc::new(Geolocator::new(self.config.geolocator.clone())),
```
MaxVerevkin, line 273: This creates a new instance of geolocator per each block, which prevents one block to share the cache with another block, if I understood the code correctly.

bim9262: Thanks, that was a good catch! I have updated the PR so that the `Config` holds the `Arc<Geolocator>` and then the reference is cloned into the `CommonApi`

MaxVerevkin (approved):

PR #2138, opened 2025-02-17
Remove debounced dependency

bim9262:


src/blocks/privacy/v4l.rs
```diff
@@ -149,7 +145,13 @@ impl PrivacyMonitor for Monitor<'_> {
                         break;
                     }
                 },
-                _ = self.updates.next() => break,
+                _ = self.stream.next() => {
```
bim9262, line 148: Any thoughts on adding our own stream extensions that could include this pattern (currently used in three blocks)?

```rust
impl<T: ?Sized> StreamExt2 for T where T: Stream {}

pub(crate) trait StreamExt2: Stream {
    async fn consume_for(&mut self, duration: Duration)
    where
        Self: Unpin,
    {
        let _ = self.next().await;
        let _ = tokio::time::timeout(duration, async {
            loop {
                let _ = self.next().await;
            }
        })
        .await;
    }
}
```

bim9262: @MaxVerevkin if you like the idea, but have a better function name, I'm not too fond of  `consume_for`.
`wait_for_next_then_discard_for_duration` seems a bit too verbose.

MaxVerevkin: I think your implementation is not cancel-safe, which may be problem. Perhaps something like (this does not have a timeout and expects successive items to be available immediately, not sure if it is good enough. on a plus side it does not have a delay :smile: )

```rust
pub trait StreamExtDebounced: futures::StreamExt {
    fn next_debounced(&mut self) -> impl Future<Output = Option<Self::Item>>;
}

impl<T: futures::StreamExt + Unpin> StreamExtDebounced for T {
    async fn next_debounced(&mut self) -> Option<Self::Item> {
        let mut result = self.next().await?;
        let mut noop_ctx = std::task::Context::from_waker(futures::task::noop_waker_ref());
        loop {
            match self.poll_next_unpin(&mut noop_ctx) {
                std::task::Poll::Ready(Some(x)) => result = x,
                std::task::Poll::Ready(None) | std::task::Poll::Pending => return Some(result),
            }
        }
    }
}

```

MaxVerevkin: `std::task::Waker::noop()` is now conveniently stable :)

bim9262: Will wait until #2143 is merged so that `use std::future::Future;` doesn't need to be imported too.

PR #2143, opened 2025-02-21
Bump edition to 2024

bim9262:


MaxVerevkin:
Interesting that

- Formatting changed so much across editions (just didn't expect that)
- `trait Future` was added to the prelude, but neither rustc nor clippy catched that we can remove manual import now (`src/lib.rs:29`)

Issue #2100, opened 2024-11-09
[Feature] Support CloudFlare Warp in VPN block


Bad3r:
It would be great if the VPN block could indicate if [CloudFlare Warp](https://developers.cloudflare.com/cloudflare-one/connections/connect-devices/warp/) is connected.

Bad3r:
my bad attempt using AI to do this since I don't know rust:
```rs
use std::process::Stdio;
use tokio::process::Command;

use crate::blocks::prelude::*;

use super::{Driver, Status};

pub struct WarpDriver;

impl WarpDriver {
    pub async fn new() -> WarpDriver {
        WarpDriver
    }

    async fn run_network_command(arg: &str) -> Result<()> {
        Command::new("warp-cli")
            .args([arg])
            .stdin(Stdio::null())
            .stdout(Stdio::null())
            .spawn()
            .error(format!("Problem running warp-cli command: {arg}"))?
            .wait()
            .await
            .error(format!("Problem running warp-cli command: {arg}"))?;
        Ok(())
    }

    async fn find_line(stdout: &str, needle: &str) -> Option<String> {
        stdout
            .lines()
            .find(|s| s.contains(needle))
            .map(|s| s.to_owned())
    }
}

#[async_trait]
impl Driver for WarpDriver {
    async fn get_status(&self) -> Result<Status> {
        let stdout = Command::new("warp-cli")
            .args(["status"])
            .output()
            .await
            .error("Problem running warp-cli command")?
            .stdout;

        let stdout = String::from_utf8(stdout).error("warp-cli produced non-UTF8 output")?;
        let line_status = Self::find_line(&stdout, "Status:").await;

        if line_status.is_none() {
            return Ok(Status::Error);
        }
        let line_status = line_status.unwrap();

        if line_status.ends_with("Disconnected") {
            return Ok(Status::Disconnected);
        } else if line_status.ends_with("Connected") {
            return Ok(Status::Connected { country: String::from("Unknown"), country_flag: String::from("🌐") });
        }
        Ok(Status::Error)
    }

    async fn toggle_connection(&self, status: &Status) -> Result<()> {
        match status {
            Status::Connected { .. } => Self::run_network_command("disconnect").await?,
            Status::Disconnected => Self::run_network_command("connect").await?,
            Status::Error => (),
        }
        Ok(())
    }
}

```

issues and limitations:
1. warp does not show up in network interfaces i.e it doesn't use something like tun0
[...]

PR #2149, opened 2025-03-21
[feat] Support Cloudflare Warp in `vpn` block

vats004:
Added support for Cloudflare Warp in the vpn block.
Should close [issue 2100](https://github.com/greshake/i3status-rust/issues/2100)


vats004:
It does not state country/server info so those fields are left blank
as warp does not provide that(I think)

vats004:
Hello @bim9262 , @MaxVerevkin 
Could you please review this

bim9262:
I finally got around to setting up warp and testing this out :see_no_evil:

Bad3r:
thank you so much @bim9262. Better late than never :D

Issue #1851, opened 2023-04-03
$essid for Ethernet connection


Atreyagaurav:
I recently shifted from i3status, so far it has a lot more features but I found some features missing.

The wireless block on i3status supports the id of the ethernet as well with variable `%essid`.

This is a config I had for i3status.
```
wireless _first_ {
        format_up = "E: (%quality at %essid) %ip"
        format_down = "E: down"
}
```

I don't see any similar variable, and `$ssid` fails when connected to Ethernet, Maybe we can make `$ssid` give Ethernet id by default when it's on Ethernet like the `icon`?

vats004:
Is this issue still open?

vats004:
Hello @Atreyagaurav 
Do you mean `connection name` assigned by `NetworkManager` by ethernet-id
because ssid is wifi-specific

PR #2150, opened 2025-03-22
[feat] Display `connection-name` when connected to ethernet

vatsalkeshav:
This pr adds functionality to get and display the `connection_name` for wired connection as set by `NetworkManager` as `ethernet_id` along with its icon.
Should solve [issue 1851](https://github.com/greshake/i3status-rust/issues/1851)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1126,7 +1126,6 @@ dependencies = [
  "wayrs-protocols",
  "x11rb-async",
  "zbus",
- "zbus_macros",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1126,6 +1126,7 @@ dependencies = [
  "wayrs-protocols",
  "x11rb-async",
  "zbus",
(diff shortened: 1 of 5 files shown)


vats004:
Implementation of MAC address as `ethernet_id` could also be done if that'd help

bim9262 (changes requested):
Not all systems use `NetworkManager`. This change will cause an error to be thrown on those systems.

vatsalkeshav:
Hi,
(post rebase)I've fixed the NetworkManager d-bus calls to gracefully return none on systems without NM

Also added a systemd-networkd fallback in case NM is not there - which tries NM first, then queries `org.freedesktop.network1` via d bus( also with a `Describe` method fallback for older systemd <250 where the `NetworkFil`e property isn't exposed) - also returns none in case that's present neither

_Tested both paths on an orbstack ubuntu 22.04 VM running on mac m1 (it has systemd 249, systemd-networkd managing eth0_

Issue #2117, opened 2024-12-12
net block: get_default_interface in netlink.rs can not detect wwan0 


cjacker:
My thinkpad has a xmm7360 4g module, after it connected to internet, i3status-rust can not detect it as default interface.

After adding some println! to netlink.rs and get the output as:

```
All interfaces:
lo, index:1
enp0s31f6, index:2
wlan0, index:4
wwan0, index:15

get_default_interface: 4
```

`route -n`:
```
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         0.0.0.0         0.0.0.0         U     1000   0        0 wwan0
0.0.0.0         192.168.2.1     0.0.0.0         UG    3004   0        0 wlan0
192.168.2.0     0.0.0.0         255.255.255.0   U     3004   0        0 wlan0
```
 `ip route show default`:
```
default dev wwan0 proto static metric 1000
default via 192.168.2.1 dev wlan0 proto dhcp src 192.168.2.148 metric 3004
```
`ip route get 8.8.8.8`:
```
8.8.8.8 dev wwan0 src 10.225.190.236 uid 1000
```

ammgws:
Should be resolved with https://github.com/greshake/i3status-rust/pull/2151

PR #2151, opened 2025-03-23
[fix] Updated the way default interfaces are chosen in `netlink.rs`

vatsalkeshav:
This pr changes the way the default route and corresponding default interface are chosen

possible problem in [issue 2117](https://github.com/greshake/i3status-rust/issues/2117)
1. default route is set by looking for any attrtibure that indicates the route being default route(by looking for Rta::Gateway)
2. and simply took the first matching route’s interface index and the corresponding interface as default

What this fix does :
1. Now, every default route(routes where rtm_dst_len == 0) is looked at 
2. and its interface index and a “metric” value is looked at 
3. whichever default route has the smallest metric, the interfave at its interface index is taken  as the default interface

This should ensure that if multiple interfaces have “default routes,” the one with the lowest metric is chosen

### DIFF
--- a/src/netlink.rs
+++ b/src/netlink.rs
@@ -351,27 +351,36 @@ async fn get_default_interface(
     ))
     .await?;
 
-    let mut default_index = 0;
+    let mut best_index = 0;
+    let mut best_metric = u32::MAX;
 
     recv_until_done!(sock, msg: Rtmsg => {
         if msg.rtm_type != Rtn::Unicast {
             continue;
         }
+        // Only check default routes (rtm_dst_len == 0)
+        if msg.rtm_dst_len != 0 {
+            continue;
+        }
+
         let mut index = None;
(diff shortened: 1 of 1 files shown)


Stebalien:
@greshake sorry to bug you, but is there any chance this could get reviewed?

In my case, this fixes my VPN block:

```toml
[[block]]
  block = "net"
  device = ".*-wg-.*"
  interval = 5
  format = "{ $ip.str(w:0)$icon $device |}"
  inactive_format = ""
  missing_format = ""
```

Given multiple VPN devices (where all but one are down):

1. With this patch, i3status-rs prioritizes my active VPN.
2. Without this patch, i3status-rs picks the first VPN, then displays nothing because it has no IP address.

MaxVerevkin:
Sorry for such a late response! This looks good to me, thank you!

Issue #2145, opened 2025-03-06
add option to overrides colors based on the sway config


akrram09:
i found it a hassle to change colors on sway and then found out you need to do the same on i3status-rs it would be amazing if you could add option in override to follow sway bar config part

`bar {
    position top
    icon_theme "Papirus-Dark"
    status_command i3status-rs "/home/akram/.config/i3status-rust/config.toml"

    colors {
        statusline $text
        background $border
	focused_workspace $border #89b4fa #1e1e2e 
        inactive_workspace $border $border $intext
    }
}
`
statusline for text and background for background

PR #2154, opened 2025-04-03
feat(theme): add sway bar color integration

him-cyber:
## Summary

This PR introduces optional integration with the user's Sway configuration to automatically apply bar colors (background and foreground) from the `~/.config/sway/config` file. Fix for the open issue #2145

## Motivation

As a Wayland-native window manager, Sway allows defining custom bar colors. This enhancement bridges the theme configuration in i3status-rust with Sway’s color settings, improving visual consistency and user experience for Sway users.

## Implementation

- Introduced a `sway_integration` section in `config.toml`
- Parsed Sway bar colors (`background`, `statusline`) from the local Sway config
- Applied parsed colors to the theme's `idle_bg` and `idle_fg` fields
- Added necessary Rust glue code with minimal footprint

## Config Usage

To enable this, add to your `config.toml`:

```toml
[sway_integration]
use_sway_bar_colors = true

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -11,6 +11,15 @@ use crate::geolocator::Geolocator;
 use crate::icons::{Icon, Icons};
 use crate::themes::{Theme, ThemeOverrides, ThemeUserConfig};
 
+use dirs;               //used in implementing parsing logic for the sway override function
+
+
+// === Optional sway integration ===
+#[derive(Deserialize, Debug)]
+pub struct SwayIntegration {
+    pub use_sway_bar_colors: Option<bool>,
+}
+
 #[derive(Deserialize, Debug)]
 #[serde(deny_unknown_fields)]
 pub struct Config {
@@ -36,6 +45,9 @@ pub struct Config {
     #[serde(default)]
     #[serde(rename = "block")]
     pub blocks: Vec<BlockConfigEntry>,
+
+    #[serde(default)]
+    pub sway_integration: Option<SwayIntegration>, 
 }
 
 #[derive(Deserialize, Debug, Clone)]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by GreyXor** [comment]
Hello, thanks for this PR.
I'm not sure that can work with `included` config like mine : https://gitlab.com/greyxor/dotfiles/-/blob/main/.config/sway/config
Any thoughts ?

**Comment by bim9262** [comment]
Instead of trying to read the config from disk, I'd recommend using `swayipc_async` to get the values.

Here's a toy program that reads the background and statusline colors:

```rust
use swayipc_async::Connection;

#[tokio::main]
async fn main() {
    let mut connection = Connection::new().await.unwrap();
    if let Some(bar_id) = connection.get_bar_ids().await.unwrap().first() {
        let bar_config = connection.get_bar_config(bar_id).await.unwrap();
        println!(
            "Bar config for {}: background {:?} statusline {:?}",
            bar_id, bar_config.colors.background, bar_config.colors.statusline
        );
    }
}
```

Here's another extension of this that finds the bar id for the current i3status-rust that's running:

```rust
use std::{os::unix::process::parent_id, path::Path};
use swayipc_async::Connection;
use tokio::io::AsyncReadExt as _;

pub async fn read_file(path: impl AsRef<Path>) -> std::io::Result<String> {
    let mut file = tokio::fs::File::open(path).await?;
    let mut content = String::new();
    file.read_to_string(&mut content).await?;
    Ok(content.trim_end().to_string())
}

#[tokio::main]
async fn main() {
    let mut connection = Connection::new().await.unwrap();
    eprintln!("Parent process ID: {}", parent_id());
    let ppid_stat = read_file(
        Path::new("/proc")
            .join(parent_id().to_string())
            .join("stat"),
    )
    .await
    .unwrap();
    let grandppid = ppid_stat.split(' ').nth(3).unwrap();
    eprintln!("Grandparent process ID: {}", grandppid);
    let grandppid_cmdline = read_file(Path::new("/proc").join(grandppid).join("cmdline"))
        .await
        .unwrap();
    let grandppid_cmdline_parts: Vec<_> = grandppid_cmdline.split('\0').collect();
    if let Some(bar_id) = grandppid_cmdline_parts
        .iter()
        .position(|&s| s == "-b" || s == "--bar_id")
        .and_then(|pos| grandppid_cmdline_parts.get(pos + 1))
    {
        let bar_config = connection.get_bar_config(&bar_id).await.unwrap();
        eprintln!(
            "Bar config for {}: background {:?} statusline {:?}",
            bar_id, bar_config.colors.background, bar_config.colors.statusline
        );
    }

    println!("{{\"version\": 1, \"click_events\": true}}\n[");
    loop {
        println!("[],")
    }
}
```

This will be able to get the arguments that are used to start bar:

(excerpt from `pstree -ap`)
```sh
  |           |-swaybar,2147460 -b bar-1
  |           |   |-sh,2147520 -c /tmp/sway_test/target/debug/sway_test 2> /tmp/l
  |           |   |   `-sway_test,2147521
  ```
  
  and the sway config:
  
  ```lua
 set $background #ffdddd
bar {
    position top
    strip_workspace_numbers yes
    height $bar_height
    status_padding 0

    # When the status_command prints a new line to stdout, swaybar updates.
    status_command /tmp/sway_test/target/debug/sway_test 2> /tmp/l

    colors {
            background $background
            statusline #222222
            separator #666666

    #                          border  backgr. text
            focused_workspace #0088CC #0088CC #ffffff
            active_workspace #333333 #333333 #ffffff
            inactive_workspace #333333 #333333 #888888
            urgent_workspace #2f343a #900000 #ffffff
    }
}
```

bim9262:
@him-cyber I updated the PR to make sure that we actually check if `use_sway_bar_colors` is true as well as updating the logic to find the colors for the specific bar that's running using swayipc.

bim9262:
What are your thoughts on pulling in the `separator` color from the config to override `separator_fg` too?

Issue #1654, opened 2022-11-03
Disk space on `btrfs` file systems


Kabouik:
As a new `btrfs` user, I am myself confused with how much space is used, allocated, or free on my system, and from the looks of a rapid web search, it seems many users are similarly confused. This appears to be because it's actually [not a trivial question on `btrfs` file systems](https://btrfs.wiki.kernel.org/index.php/FAQ#How_much_free_space_do_I_have.3F) and base GNU/Linux tools like `du` and `df` can get confused and give erroneous figures on such systems. However, that last link to the `btrfs` wiki does provide command line tools to more accurately assess disk usage. Would it be possible to make the [disk-space](https://github.com/greshake/i3status-rust/blob/v0.22.0/doc/blocks.md#disk-space) block compatible with them, perhaps by adding an option to indicate the file system and use adequate commands to calculate used or free space?

REALERvolker1:
I can't really do this myself but I would be happy to see this :D

lbatalha:
I am currently doing this with a custom block, would be nice to have maybe something like a "backend" option in the disk-space block to use btrfs utils to get the info.

My block looks like:
```toml
[[block]]
block = "custom"
command = "btrfs fi usage / 2>/dev/null | sed -n '7p' | cut -f 4"
format = " $text.pango-str() "
interval = 64
```

Which gets the "Used" stat.
The manpage (`man 8 btrfs-filesystem`) has detailed information on all the stats returned.

Kabouik:
Thanks for that custom block example! Could you elaborate on what `$text.pango-str()` does, and what are its dependencies? It doesn't work on my system, I suppose it may need some packages I have not installed.

I can make the block work when just removing the whole "format" line, but I'd rather customize it a bit to add the unit, or some other symbols.

lbatalha:
Hmm you are right, the format line does nothing here, it must have been an artefact of when I recently had to redo my config due to i3status-rs' breaking changes.

As for the units, they are included because btrfs outputs that, you can customize them in the command via `-k`, `-m`, and others, by default it picks the unit automatically.
Looks like this for me:
![image](https://user-images.githubusercontent.com/5883822/233681753-fc56ee06-e296-47e6-8777-884222d7c845.png)

The program output is something like:
```
❯ btrfs fi usage / 2>/dev/null
Overall:
    Device size:		  1.82TiB
    Device allocated:		764.06GiB
    Device unallocated:		  1.07TiB
[...]

PR #2159, opened 2025-05-18
disk_space: Support btrfs backend

bpeetz:
Btrfs is too smart for the statvfs based backend (i.e., only counting blocks leads to wrong numbers).

For example, a btrfs disk with a lot of de-duplicated blocks (via the copy on write mechanism) might have a drastically over-reported disk usage.

The btrfs backend is currently implemented by parsing the output of the `btrfs filesystem usage --raw` command. This is suboptimal, as this now relies on the command output not changing.

Vendoring the algorithm used internally by the `btrfs` command does not seem to be a reasonable alternative, considering that the code[1] is rather complex, low level and would require semi-constant maintenance. Additionally, the c code would need bindings to be usable from rust.

I assume, that the `btrfs` command output will stay rather similar in the future, as a lot of tools rely on directly parsing it (see the various scripts in the issue, this commit fixes).

[1]: https://github.com/kdave/btrfs-progs/blob/eeab081e9d9fbdf4583122ed1caedf541383cf2d/cmds/filesystem-usage.c#L442

Fixes: #1654


ammgws:
Thanks!

PR #2160, opened 2025-05-19
memory: Avoid estimating available memory, use kernel estimate instead

bpeetz:
`free`, `btop` and other tools (like the `sysinfo` rust library) use `mem_total - mem_avail` to calculate the used memory. As explained in the already linked [kernel commit][1], `mem_avail` is an estimate of the available memory based on the current way the kernel handles memory.

The previous logic for calculating used memory tried to estimate this based on `mem_free`, `buffers`, `pagecache` and `reclaimable`. Unfortunately, as the kernel commit message predicts, this user space estimate bitrots, as the kernel memory usage patterns change.

Thus, we now simply use the kernel estimate `mem_avail` as we know that that is in-sync with the relevant kernel internals.

[1]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773


ammgws:
Sorry for the delay! So apparently when I wrote that logic, it was done based on the code in `htop`. However since around that same time other tools switched to use MemAvailable (https://gitlab.com/procps-ng/procps/-/commit/2184e90d2e7cdb582f9a5b706b47015e56707e4d) and htop seems to be an outlier now: https://github.com/htop-dev/htop/issues/1307

PR #2172, opened 2025-06-30
Add support for systemd user services

### LINKED ISSUE #1798: Feature request - service status for user services
I initialize a specific service with ```systemctl start --user <service>```. I can't however use the  ```service_status``` blocks to see their status. Some extra attribute such as user=true would be useful to distinguish between root and user systemd services.

Stebalien:
This patch adds support for displaying the status of user services (when the new `user` field is `true`).

fixes #1798


bim9262:
I thought we'd tried this a few years ago and that it hadn't worked as expected. I can't find my notes about it though.
I tested out the change and it looks good on my end.

PR #2176, opened 2025-07-18
Updated icu4x deps to 2.x

bim9262:
`icu_locid` has been replaced with `icu_locale_core`.
Removed direct dependency on `icu_calendar`.
Added `precision` argument to `datetime` formatter to specify the precision of the datetime.


bim9262:
Any thoughts on having an argument on the datetime formatter that's only accepted when using ICU datetimes?

bim9262:
@ammgws 
Any thoughts on having an argument on the datetime formatter that's only accepted when using ICU datetimes?

bim9262:
Separated the icu4x version 2 update from the added `precession` argument.

Issue #2175, opened 2025-07-11
calendar: Error with nextcloud caldav : duplicate field `comp`


matclab:
The *calendar* block raised an error : Error cause: duplicate field `comp`.

kkoralsky:
@matclab I made PRs that fixes the same issue from my caldav server (from https://github.com/Alinto/sogo) You may want to test if this would work also for you

matclab:
> [@matclab](https://github.com/matclab) I made PRs that fixes the same issue from my caldav server (from https://github.com/Alinto/sogo) You may want to test if this would work also for you

@kkoralsky thank you very much ! It solves the error for me…

PR #2178, opened 2025-07-20
blocks[calendar]: parse multiple comp within support-calendar-componet-set

kkoralsky:
fixes calendar retrieval if caldav server replies with multiple `<comp ..>` entities eg.

```xml
<?xml version="1.0" encoding="utf-8"?>
<D:multistatus xmlns:D="DAV:" xmlns:a="urn:ietf:params:xml:ns:caldav">
  <D:response>
    <D:href>/SOGo/dav/___redacted___/Calendar/</D:href>
    <D:propstat>
      <D:status>HTTP/1.1 200 OK</D:status>
      <D:prop>
        <D:displayname>Calendar</D:displayname>
        <D:resourcetype>
          <D:collection/>
        </D:resourcetype>
        <n1:supported-calendar-component-set xmlns:n1="urn:ietf:params:xml:ns:caldav" xmlns:D="DAV:">
          <n1:comp name="VEVENT"/>
          <n1:comp name="VTODO"/>
        </n1:supported-calendar-component-set>
      </D:prop>
    </D:propstat>
  </D:response>
  <D:response>
....
```

this is the case for sogo caldav service and apparently for nexcloud accoring to: #2175

### DIFF
--- a/src/blocks/calendar/caldav.rs
+++ b/src/blocks/calendar/caldav.rs
@@ -228,11 +228,12 @@ enum ResourceType {
 
 #[derive(Debug, Deserialize)]
 struct SupportedCalendarComponentSet {
-    comp: Option<Comp>,
+    #[serde(rename = "$value", default)]
+    pub values: Vec<Comp>,
 }
 impl SupportedCalendarComponentSet {
     fn supports_events(&self) -> bool {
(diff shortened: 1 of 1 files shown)


bim9262:
Sorry it took so long to review. Thanks for your contribution!

PR #2182, opened 2025-07-29
Allow one or more format to be used

### LINKED ISSUE #2163: `toggle_format` not avaible for `time` module.
In the normal case, I would like to have the short time by default. But on left click, have the ability to see the verbose time. This seems like not possible in the current version.

bim9262:
Currently this has only been applied to the time block, but any block that doesn't have formats for specific states/conditions could use this do allow one or more format, while still being able to continue using `format` and `format_alt`. (This could work as a fix to #2163)

One drawback about this is that I was unable to figure out how to be able to keep the `deny_unknown_fields` on the `Config`

This will allow configs to be specified in any of these ways:

```toml
[[block]]
block = "time"
[[block.click]]
button = "left"
action = "next_format"
[[block.click]]
button = "right"
action = "prev_format"

[[block]]
block = "time"
format = " $icon $timestamp.datetime(f:'%a %Y-%m-%d %R %z') "
[[block.click]]
button = "left"
action = "next_format"
[[block.click]]
button = "right"
action = "prev_format"

[[block]]
block = "time"
format = " $icon $timestamp.datetime(f:'%a %Y-%m-%d %R %z') "
format_alt = " $icon $timestamp.datetime(f:long, l:en-u-ca-hebrew-hc-h23)"
[[block.click]]
button = "left"
action = "next_format"
[[block.click]]
button = "right"
action = "prev_format"

[[block]]
block = "time"
format = [" $icon $timestamp.datetime(f:'%a %Y-%m-%d %R %z') ", " $icon $timestamp.datetime(f:long, l:en-u-ca-hebrew-hc-h23)"]
[[block.click]]
button = "left"
action = "next_format"
[[block.click]]
button = "right"
action = "prev_format"

[[block]]
block = "time"
[[block.format]]
full = " $icon $timestamp.datetime(f:'%a %Y-%m-%d %R %z') "
[[block.format]]
full = " $icon $timestamp.datetime(f:long, l:en-u-ca-hebrew-hc-h23)"
[[block.click]]
button = "left"
action = "next_format"
[[block.click]]
button = "right"
action = "prev_format"

[[block]]
block = "time"
testing = true
[block.format]
full = " $icon $timestamp.datetime(f:'%a %Y-%m-%d %R %z') "
[[block.click]]
button = "left"
action = "next_format"
[[block.click]]
button = "right"
action = "prev_format"
```

### DIFF
--- a/src/blocks/amd_gpu.rs
+++ b/src/blocks/amd_gpu.rs
@@ -5,8 +5,7 @@
 //! Key | Values | Default
 //! ----|--------|--------
 //! `device` | The device in `/sys/class/drm/` to read from. | Any AMD card
-//! `format` | A string to customise the output of this block. See below for available placeholders. | `" $icon $utilization "`
-//! `format_alt` | If set, block will switch between `format` and `format_alt` on every click | `None`
+//! `format` | A [MultiFormat][MaybeMultiFormatConfig] string to customise the output of this block. See below for available placeholders. | `[" $icon $utilization "]`
 //! `interval` | Update interval in seconds | `5`
 //!
 //! Placeholder          | Value                               | Type   | Unit
@@ -19,7 +18,9 @@
 //!
 //! Action          | Description                               | Default button
 //! ----------------|-------------------------------------------|---------------
-//! `toggle_format` | Toggles between `format` and `format_alt` | Left
+//! `toggle_format` **DEPRECATED** | Toggles between `format` and `format_alt` | -
+//! `next_format`  | Switches to the next format in the list     | Left
+//! `prev_format`  | Switches to the previous format in the list | Right
 //!
 //! # Example
 //!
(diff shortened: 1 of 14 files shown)


bim9262:
> One drawback about this is that I was unable to figure out how to be able to keep the `deny_unknown_fields` on the `Config`

I fount that we can have `deny_unknown_fields`  on the `MaybeMultiConfig` instead.

bim9262:
@MaxVerevkin any thoughts on this approach to allow one or more format (either as how formats are currently configured, or as a list of format configs)?

bim9262:
@ammgws, any concerns with this approach? I only added support to the time block, but it could be added to other blocks. For blocks that already support `format_alt`, I'd suggest deprecating the action `toggle_format` for `next_format` and `prev_format`.

ammgws:
On phone now but just to double check before going further, are the examples in your parent comment still the same?

bim9262:
@ammgws , yes the example config works as before (well except the last time block, which fails on purpose)

The change I made the other day was to keep the behavior that the first format keeps the `default_full`, `default_short` for the first format instead of filling it with `""`. This would only matter in the case that you wanted to set only the first format's short/full format. This is consistent with the behavior today.

Your question did raise an issue for me, what if a block does specify a default for `format_alt`? Today the only block that does this is the privacy block, all others' `format_alt` default to `""`. I think that adding `with_default_formats` to `MaybeMultiConfig` would probably be the most ergonomic option.

bim9262:
I've added support in two other blocks as examples: `privacy` and `weather`

ammgws:
Sorry I keep forgetting to come back to this except when out on phone

I think this is fine though

bim9262:
Verify MaybeMultiConfig is not given an empty list of configs while deserializing

The `toggle_format` action has been deprecated in favor of `next_format` and `prev_format`

Implemented for the following blocks:

* `amd_gpu`
* `cpu`
* `disk_space`
* `memory`
* `music`
  * Note: `prev_format` action is not tied to a button by default as right clicking is already assigned to `next_player`.
* `net`
* `privacy`
* `sound`
  * Note: `prev_format` action is not tied to a button by default as right clicking is already assigned to `toggle_mute`.
* `temperature`
* `time`
  * Note: neither `next_format` nor `prev_format` are tied to buttons by default as left and right clicking are already assigned to `next_timezone` and `prev_timezone` respectively.
* `weather`

PR #2190, opened 2025-08-17
Treat oneshot systemd services as "active" when "activating"

Stebalien:
Oneshot systemd services transition directly from activating to dead/inactive and never show up as "active". This change treats these services as "active" when "activating" because that's likely what the user wants. Otherwise, backup tasks, etc. never show up as active services in i3status-rs.

### DIFF
--- a/src/blocks/service_status.rs
+++ b/src/blocks/service_status.rs
@@ -108,6 +108,7 @@ trait Driver {
 
 struct SystemdDriver {
     proxy: UnitProxy<'static>,
+    service_proxy: ServiceProxy<'static>,
     active_state_changed: PropertyStream<'static, String>,
 }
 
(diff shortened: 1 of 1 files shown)


src/blocks/service_status.rs
```diff
@@ -141,27 +142,49 @@ impl SystemdDriver {
...
+            .error("Could not get active_state")?;
+
+        Ok(match &*active_state {
+            "active" => true,
+            "activating" => {
+                let service_type = self
+                    .service_proxy
+                    .type_()
+                    .await
+                    .error("Could not get service type")?;
```
Stebalien, line 182: I read this property every time in case the underlying service changes.

Stebalien:
From the `systemd.service` man page (do you don't have to go looking for it):

> Behavior of oneshot is similar to simple; however, the service
   manager will consider the unit up after the main process exits.
   It will then start follow-up units.  RemainAfterExit= is
   particularly useful for this type of service.  Type=oneshot is
   the implied default if neither Type= nor ExecStart= are
   specified. Note that if this option is used without
   RemainAfterExit= the service will never enter "active" unit
   state, but will directly transition from "activating" to
   "deactivating" or "dead", since no process is configured that
   shall run continuously. In particular this means that after a
   service of this type ran (and which has RemainAfterExit= not
   set) it will not show up as started afterwards, but as dead.

bim9262:
Can you rebase your pr? That should fix the cargo-fmt error you got.

Stebalien:
@bim9262 done!

src/blocks/service_status.rs
bim9262: Under what circumstance would the `service_type` switch from `"oneshot"` to something else?

Stebalien: The user changes the service file and calls `systemctl daemon-reload`.

Stebalien: This is obviously a pretty weird edge case, but I'm imagining the following scenario:

1. User is developing a unit file.
2. They've wired up i3status-rs to monitor it.
3. They change it to a oneshot unit.
4. Suddenly it stops showing up in i3status-rs whenever they restart it.

(also, it's easy enough to just check every time)

bim9262:
looks good, thanks!

Stebalien:
Thanks!

PR #2193, opened 2025-08-19
icons: update material codepoints

bim9262:


ammgws:
@bim9262 Shall we go ahead with this?

and do we need to remove/update this in the docs or is this PR separate to that issue? https://github.com/greshake/i3status-rust/blob/45586330364885cf52cb8ceaad1af7f92d55d2a8/doc/themes.md?plain=1#L78

bim9262:
@ammgws  I'm ok to update the codepoints. I'm not sure about that comment though. I imagine it would depend on your distro and if material fonts are available out of the box on that distro?

Issue #2197, opened 2025-09-05
"Failed to render full text error: Icon not found" but the error shows the Icon correctly


shtrom:
I have this custom example, showing a Firefox icon from the Unicode private use area.

```toml
[[block]]
block = "custom"
command = """
echo '{
          "icon": "󰈹",
          "text": "test"
  }'
"""
json = true
interval = 300
```

It fails to render:

<img width="170" height="43" alt="Image" src="https://github.com/user-attachments/assets/1656c090-c7ce-4425-9f52-4ef29e736b55" />

But the error message actually renders the icon correctly:

<img width="418" height="44" alt="Image" src="https://github.com/user-attachments/assets/6d97d5b7-8918-4ec9-afe7-9d278e69aa57" />

This is probably because the font supports it, but that should be a non-failing fallback.

Stebalien:
I think the icon is supposed to be an icon name. That custom block should probably just be:

```toml
[[block]]
block = "custom"
command = """
echo '{
          "text": "󰈹 test"
  }'
"""
json = true
interval = 300
```

bim9262:
Or more simply:

```toml
[[block]]
block = "custom"
command = "echo '󰈹 test'"
interval = "once"
```

shtrom:
> I think the icon is supposed to be an icon name. That custom block should probably just be

🤦🏼 yeah, that works. Thanks!

PR #2198, opened 2025-09-05
Use icon if mapping not found

shtrom:
fixes: #2197

---

This is my first Rust PR ever. Critiques welcome (:


shtrom:
Unneeded, just put the character in the `text` rather than the `icon`.

PR #2205, opened 2025-10-04
notmuch: use a file watch and auto-detect the database location

Stebalien:
This PR brings two (breaking) improvements to the notmuch block:

1. The first change watches the database lock instead of relying on an update interval. From my testing, this only triggers when the database was opened for writing. NOTE: I'm not doing this to get email notifications faster, I just don't want to query the notmuch database every 10s.
2. The second change improves the notmuch block's configuration. Specifically, it renames `maildir` to `database` and makes the field optional, while also adding a new optional `profile` option. This should bring the block in-line with the `notmuch` CLI tool (the same defaults and environment variables will be used). In most cases, this means the user can simply omit any database configuration.

Both of these changes are breaking as they change the config. I'm happy to split them into multiple PRs and/or add a better migration path if desired.


Stebalien:
TBH, I'd also be in favor of removing this block given that:

1. It requires the notmuch library (and will therefore almost certainly require that the user compile from source).
2. It can easily be implemented with a custom block:

   ```toml
   [[block]]
     block = "custom"
     watch_files = ["~/.local/share/notmuch/default/xapian/flintlock"]
     interval = "once"
     format = "$icon $text "
     command = "notmuch count --output=threads tag:inbox and tag:unread"
    ```

(I just figured I'd improve it as long as it exists)

PR #2216, opened 2025-12-13
Disable landlock for pacman 7.1

freswa:
Not disabling sandboxing leads to:
```
error: restricting filesystem access failed because the Landlock ruleset could not be applied: Operation not permitted
error: switching to sandbox user 'alpm' failed!
error: failed to synchronize all databases (failed to retrieve some files)
```

### DIFF
--- a/src/blocks/packages/pacman.rs
+++ b/src/blocks/packages/pacman.rs
@@ -85,6 +85,7 @@ impl Backend for Pacman {
                 "--".as_ref(),
                 "pacman".as_ref(),
                 "-Sy".as_ref(),
+                "--disable-sandbox-filesystem".as_ref(),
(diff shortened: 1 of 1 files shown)


ammgws:
Since we use fakeroot then we perhaps we need to wait for this: https://gitlab.archlinux.org/pacman/pacman/-/commit/2df7c407e852543e5c9e135f4bda5ce2d3a21b70 which is what `checkupdates` will use as per 
https://gitlab.archlinux.org/pacman/pacman-contrib/-/commit/ea3ed064f0cc071f843e18741e74138b6421959e

freswa:
@ammgws Great catch. The original commit worked as well, but just disabling filesystem sandboxing is even better. Tested and pushed to this branch.

PR #2219, opened 2025-12-16
Fix failures that are causing pre-commit to fail

bim9262:


src/main.rs
```diff
@@ -65,6 +65,7 @@ fn main() {
 }
 
 /// Restart in-place
+#[allow(unreachable_code)]
```
bim9262: @ammgws can you confirm if you think this is the best way to work around the errors:

```rust
error: unreachable expression
  --> src/main.rs:89:5
   |
88 |     nix::unistd::execvp(&exe, &arg).unwrap();
   |     ---------------------------------------- any code following this expression is unreachable
89 |     unreachable!();
   |     ^^^^^^^^^^^^^^ unreachable expression
   |
note: this expression has type `std::convert::Infallible`, which is uninhabited
  --> src/main.rs:88:5
   |
88 |     nix::unistd::execvp(&exe, &arg).unwrap();
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: `-D unreachable-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unreachable_code)]`
   ```

PR #2221, opened 2025-12-29
issue-1964: Add brew and snap package manager backends.

zheltikov:
- packages: Add support for the Homebrew Package Manager.
- packages: Add support for Snap packages.
- packages: Update and clean-up documentation.
- packages: Refactor package manager initialization and registration.

Resolves #1964 

@MaxVerevkin @bim9262 @ammgws

PR #2222, opened 2025-12-30
issue-2209: Add support for the Flatpak package manager.

### LINKED ISSUE #2209: Add Support for Flatpak in `packages` Block
It'd be nice to get a counter for the Flatpak updates in the `packages` block.

zheltikov:
- packages: Add Flatpak package manager backend.

Resolves #2209 .
Depends on #2221 being merged first.

@MaxVerevkin @bim9262 @ammgws

### DIFF
--- a/src/blocks/packages.rs
+++ b/src/blocks/packages.rs
@@ -175,7 +175,7 @@
 //! [[block.click]]
 //! # shows dmenu with cached available updates. Any dmenu alternative should also work.
 //! button = "left"
-//! cmd = "flatpak remote-ls --updates --no-header --columns=ref | tail -n +2 | rofi -dmenu"
+//! cmd = "flatpak remote-ls --updates --columns=ref | rofi -dmenu"
 //! ```
 //!
 //! Pacman-only config:
--- a/src/blocks/packages/flatpak.rs
+++ b/src/blocks/packages/flatpak.rs
@@ -27,7 +27,7 @@ impl Backend for Flatpak {
 
         let stdout = Command::new("flatpak")
             .env("LC_ALL", "C")
-            .args(["remote-ls", "--updates", "--no-header", "--columns=ref"])
+            .args(["remote-ls", "--updates", "--columns=ref"])
             .output()
             .await
             .error("Failed to run `flatpak remote-ls`")?


bim9262:
@zheltikov can you rebase your change?

zheltikov:
Rebased, @bim9262

src/blocks/packages/flatpak.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+#[async_trait]
+impl Backend for Flatpak {
+    fn name(&self) -> Cow<'static, str> {
+        "flatpak".into()
+    }
+
+    async fn get_updates_list(&self) -> Result<Vec<String>> {
+        Command::new("flatpak")
```
bim9262: Is this necessary? I thought that `flatpak remote-ls` alone would list the updates?
Under what circumstance do you need to run the `flatpak update --appstream -y` first?

src/blocks/packages/flatpak.rs
```diff
@@ -0,0 +1,44 @@
...
+            .env("LC_ALL", "C")
+            .args(["update", "--appstream", "-y"])
+            .output()
+            .await
+            .error("Failed to run `flatpak update`")?;
+
+        let stdout = Command::new("flatpak")
+            .env("LC_ALL", "C")
+            .args(["remote-ls", "--updates", "--no-header", "--columns=ref"])
```
bim9262: When I run this I get the following error:

```
> LC_ALL=C flatpak remote-ls --updates --no-header --columns=ref
error: Unknown option --no-header

> flatpak --version
Flatpak 1.16.2
```

src/blocks/packages.rs
```diff
@@ -158,6 +160,24 @@
...
+//! interval = 1800
+//! error_interval = 300
+//! max_retries = 5
+//! format = " $icon $flatpak.eng(w:1) updates available "
+//! format_singular = " $icon One update available "
+//! format_up_to_date = " $icon system up to date "
+//! [[block.click]]
+//! # shows dmenu with cached available updates. Any dmenu alternative should also work.
+//! button = "left"
+//! cmd = "flatpak remote-ls --updates --no-header --columns=ref | tail -n +2 | rofi -dmenu"
```
bim9262: With the --no-header (if it worked) you wouldn't need the tail. Without --no-header, you only need to skip the first line, not the first 2.

src/blocks/packages/flatpak.rs
zheltikov: As per the manual:
`remote-ls` lists the whole contents of a remote repository.
`update --appstream` updates the info for that all remotes.

So we need to update the local info before listing updates.

src/blocks/packages/flatpak.rs
zheltikov: Upon further investigation, it turns out that `flatpak` does not print headers for non-tty outputs. So we can just omit stripping the header line entirely.

zheltikov: Fixed

bim9262:
This looks good now! I'm able to see the correct count of packages available for update :smile:

PR #2224, opened 2026-01-06
Fix headphones indicator detection on PipeWire/WirePlumber

mokhnatchuk:
## Problem
The headphones indicator icon fails to appear on Linux systems using PipeWire/WirePlumber when `form_factor` is reported as `Unknown` or other unexpected values.

## Root Cause
The current logic only checks `active_port` as a fallback when `form_factor` is `None`. However, on many PipeWire setups, `form_factor` returns `Some("Unknown")` rather than `None`, causing the fallback to never trigger even when `active_port` correctly reports "analog-output-headphones".

## Solution
Extended the fallback logic to check `active_port` for all `form_factor` values that don't explicitly match known headphone types. This covers both `None` and unexpected values like `Unknown`.

## Changes
- Unified match arms to use wildcard pattern `_` for fallback
- Changed substring check from "headphones" to "headphone" to match both singular and plural port names
- Added `.as_ref()` to prevent moving `active_port` value

## Testing
Tested on PipeWire 1.4.9 with WirePlumber where `form_factor` reports as `Unknown` and `active_port` as "analog-output-headphones".

### DIFF
--- a/src/blocks/sound.rs
+++ b/src/blocks/sound.rs
@@ -160,11 +160,11 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
                 Some("headset") | Some("headphone") | Some("hands-free") | Some("portable") => true,
                 // Per discussion at
                 // https://github.com/greshake/i3status-rust/pull/1363#issuecomment-1046095869,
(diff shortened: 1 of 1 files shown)


bim9262:
Looks good to me! Thanks :smiley:

PR #2226, opened 2026-01-07
Expand recurring caldav events

### LINKED ISSUE #2186: calendar: recurring event are not taken into account
When an event is recurring, only the first occurrence is taken into account by the block.

bim9262:
Fixes #2186

Depends on implementation on `icalendar` dependency:

- https://github.com/hoodie/icalendar/issues/88
- https://github.com/hoodie/icalendar/pull/128

FYI @matclab @zheltikov

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,12 +27,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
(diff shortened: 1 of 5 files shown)


zheltikov:
Tested with Google Calendar, seems to work fine!

bim9262:
@ammgws What are your thoughts on using a version of icalendar tied to the commit of a PR?
I believe we've done it in the past, however I only recall seeing that for other crates managed by other maintainers of this repo.

ammgws:
Yeah... I wouldn't really want to go there unless it is something really critical.

bim9262:
Vendored `get_recurrence` from https://github.com/hoodie/icalendar/pull/128.
Updated logic so that if there is a recurrence more often than daily, they will be rendered correctly.

PR #2227, opened 2026-01-07
Draft: New Graph Formatting

### LINKED ISSUE #1555: v0.30: reimplement "Net graph formatting"
See #1457

The result could be something like `format = "$graph_down.graph(10)" # a graph 10 characters wide`. The repetition of the word "graph" might seem weird, so I also imagine something like this working: `$speed_down.graph(5) $speed_down` - "Number" could have an associated "history" or "graph" field (with _some_ maximum length), so it could be formatted with both `.eng()` and `.graph()`.

zheltikov:
Resolves #1555 .

### Changelog

- Add new `ValueInner::Numbers` variant for collections of values. Supports two formatting modes:
  - `EngFormatter`: (default) just renders the last value of the set like a regular `Value::Number`.
  - `BarFormatter`: if vertical, renders a barchart with the last `width` values from the set; if horizontal, renders the last value as a "meter" bar.

- Refactor `EngFormatter` and `BarFormatter` to implement formatting of new `Value::Numbers` variant. Add unit tests with examples for `BarFormatter`.
- Refactor `cpu` and `net` blocks to use the new `BarFormatter` functionality.

### Open questions

1. The current approach is not too flexible: it does not allow to "smartly" detect the `width` or `max_value` of the bar charts. This has to be manually be set in the user's `config.toml` file. 
  1.1. `max_value`: can be automatically detected in the BarFormatter on the fly, while rendering the barchart. But then, the manual `max_value` user formatting argument has to be made optional in some cases. When? 
  1.2. `width`: should be exposed somehow to the blocks by the formatter, and vice-versa? In case of the `net` block, we should only accumulate as much data as we would render. In case of the `cpu` block, it has to be enforced the other way around: number of cores from the block to the formatter. 
  1.3. the examples and solutions proposed above would create some confusion both code-wise (block-logic and formatting logic are now aware of each-other and coupled) and usage-wise (the user's formatting arguments in their config file should be followed or ignored in some cases).

---

@MaxVerevkin , @bim9262 , thoughts on the current implementation and the open questions?

### DIFF
--- a/src/formatting/formatter/bar.rs
+++ b/src/formatting/formatter/bar.rs
@@ -55,14 +55,14 @@ impl BarFormatter {
     #[inline]
     fn format_single_vertical(&self, val: f64, max_value: f64) -> char {
         let val = self.norm(val, max_value);
-        VERTICAL_BAR_CHARS[(val * 8.0) as usize]
+        VERTICAL_BAR_CHARS[(val * 8.0).ceil() as usize]
     }
 
     #[inline]
     fn format_horizontal_cell(&self, val: f64, max_value: f64, i: usize) -> char {
         let val = self.norm(val, max_value);
         let chars_to_fill = val * self.width as f64;
-        HORIZONTAL_BAR_CHARS[((chars_to_fill - i as f64).clamp(0.0, 1.0) * 8.0) as usize]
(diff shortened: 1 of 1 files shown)


bim9262:
I haven't had a chance to fully review this, but I worked on this last year too. Maybe it can help :man_shrugging: 

https://github.com/greshake/i3status-rust/compare/master...bim9262:i3status-rust:bar_graph?expand=1

Issue #2229, opened 2026-01-09
Regex bug in `xresources.rs`


pretentiousUsername:
It seems like `i3status-rs` can only read colors from `~/.Xresources` only if the letters in an HTML color code are all lowercase. I've tested this with two `.Xresources` files and the same, *very* minimal theme.

<details>
<summary> Config file </summary>

This is ripped straight from [the original `pywal` repository](https://github.com/dylanaraps/pywal/blob/master/pywal/templates/colors.Xresources); the only changes I made were removing the cosmetic formatting, since I wasn't sure if that caused any issues.

At first I thought the issue was that the regex wouldn't recognize colors with multiple spaces between their name and the colorcode—hence the changes I made here—but that turned out to be wrong.

```
! X colors.
! Generated by 'wal'
*foreground: {foreground}
*background: {background}
*.foreground: {foreground}
*.background: {background}
emacs*foreground: {foreground}
emacs*background: {background}
URxvt*foreground: {foreground}
XTerm*foreground: {foreground}
UXTerm*foreground: {foreground}
URxvt*background: {background.alpha}
XTerm*background: {background}
UXTerm*background: {background}
URxvt*cursorColor: {cursor}
XTerm*cursorColor: {cursor}
UXTerm*cursorColor: {cursor}
URxvt*borderColor: {background.alpha}

! Colors 0-15.
*color0: {color0}
*color1: {color1}
*color2: {color2}
*color3: {color3}
*color4: {color4}
*color5: {color5}
*color6: {color6}
*color7: {color7}
*color8: {color8}
*color9: {color9}
*color10: {color10}
*color11: {color11}
*color12: {color12}
*color13: {color13}
*color14: {color14}
*color15: {color15}
*.color0: {color0}
*.color1: {color1}
*.color2: {color2}
*.color3: {color3}
*.color4: {color4}
*.color5: {color5}
*.color6: {color6}
*.color7: {color7}
*.color8: {color8}
*.color9: {color9}
*.color10: {color10}
*.color11: {color11}
*.color12: {color12}
*.color13: {color13}
*.color14: {color14}
*.color15: {color15}

! Black color that will not be affected by bold highlighting.
*.color66: {color0}
*color66: {color0}

! Xclock colors.
XClock*foreground: {foreground}
XClock*background: {background}
XClock*majorColor: rgba:{color15.xrgba}
XClock*minorColor: rgba:{color15.xrgba}
XClock*hourColor: rgba:{color15.xrgba}
[...]

PR #2230, opened 2026-01-09
change `COLOR_REGEX`

pretentiousUsername:
Changed `COLOR_REGEX` to allow for it to read both upper-, and lowercase letters in HTML colorcodes in a `.Xresources` file.

I opened [an issue](https://github.com/greshake/i3status-rust/issues/2229) (#2229) about this, but I figured that it would be best to also propose a change for the matter.

### DIFF
--- a/src/themes/xresources.rs
+++ b/src/themes/xresources.rs
@@ -20,7 +20,6 @@ fn read_xresources() -> std::io::Result<String> {
 #[cfg(test)]
 use tests::read_xresources;
 
-
 static COLOR_REGEX: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r"^\s*\*(?<name>[^: ]+)\s*:\s*(?<color>#[A-Fa-f0-9]{6,8}).*$").unwrap()
 });


ammgws:
I don't use xresources and couldn't find any examples that use uppercase letters in the colour code and couldn't find manpage to double check either but if uppercase is valid and working for you then that is proof enough I guess..

ammgws:
Can you remove the extraneous newline?

bim9262:
@ammgws I found https://man.archlinux.org/man/X.7#COLOR_NAMES and it looks like #RRGGBB is one of several deprecated formats for specifying a color. I can't tell about the case, though. We allow 6-8 characters though, when it looks like it should be 3, 6, 9, or 12 bytes. However I think we can only use 6 characters (and that we shouldn't ready 8 characters (RRGGBBAA) as alpha is not allowed and it could be a substring of #RRRGGGBBB or #RRRRGGGGBBBB.

We could allow converting up #RGB to #R0G0B0 as X would do though.

bim9262:
Thanks!

PR #2232, opened 2026-01-22
Add history_count variable and show_all action for notify block

antoyo:
I'm not sure how to implement `history_count` for SwayNC.
Does anyone have an idea of how to implement it?

### DIFF
--- a/src/blocks/notify.rs
+++ b/src/blocks/notify.rs
@@ -9,16 +9,20 @@
 //! `driver` | Which notifications daemon is running. Available drivers are: `"dunst"` and `"swaync"` | `"dunst"`
 //! `format` | A string to customise the output of this block. See below for available placeholders. | `" $icon "`
 //!
-//! Placeholder                               | Value                                      | Type   | Unit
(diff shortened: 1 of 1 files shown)


bim9262:
Hi @antoyo , thanks for your PR :smile: 

dunst and swaync operate pretty differently. 
While both pop up a notification, that after several seconds clear, what happens to those messages after differs by notification manager.

In dunst, messages are moved to the history buffer (20 by default).
While swaync displays all of the history in a notification center (those messages have a little x button on them to actually clear them) as well as there being a clear all button.

I'd just impliment history_count for swaync like:

```rust
async fn history_count(&self) -> Result<u32> {
    self.notification_count().await
}
```

and add documentation that for swaync `show` and `show_all` have the same behavior as well as there not being any distinctions between `notification_count` and `history_count`.

Also add `(omitted if 0)` to the documentation for `history_count`.

antoyo:
Thanks for your review!
I addressed it and implemented `history_count` for SwayNC in the way you mentioned.

bim9262:
Thanks for your contribution @antoyo!

PR #2233, opened 2026-01-24
taskwarrior: Fix high CPU usage with taskwarrior v3

cfsmp3:
Taskwarrior v3 updates some internal status files (specifically, SQLite's shared memory file - taskchampion.sqlite3-shm) even on read, which causes a feedback loop with inotify - each time inotify notifies us of a change we run taskwarrior to update our task list count, which in turn causes a file change, which triggers another inotify notification and so on.

This gets us to high CPU usage which leads to a noisy fan I can't tolerate.

### DIFF
--- a/src/blocks/taskwarrior.rs
+++ b/src/blocks/taskwarrior.rs
@@ -147,15 +147,32 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
 
         api.set_widget(widget)?;
 
-        select! {
-            _ = sleep(config.interval.0) =>(),
-            _ = updates.next() => (),
-            _ = api.wait_for_update_request() => (),
-            Some(action) = actions.recv() => match action.as_ref() {
-                "next_filter" => {
-                    filter = filters.next().unwrap();
+        loop {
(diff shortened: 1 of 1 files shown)


bim9262:
Could we just change the default `data_location` from `"~/.task"` to `"~/.task/taskchampion.sqlite3`"?

cfsmp3:
> Could we just change the default `data_location` from `"~/.task"` to `"~/.task/taskchampion.sqlite3`"?

That wouldn't work on old versions of taskwarrior though. There's valid reasons not to upgrade taskwarrior from v2 to v3 - it's a breaking change and people do have scripts / tools /apps / whatever that depend on v2 or the v2 storage, so I think supporting both is reasonable.

Also, but this is a bit beyond the point, watching a single file has some caveats - if the inode changes (which is quite common if the next content is just written to a temp file and then moved over the final path), inotify will stop sending notifications (unless a new watch is set).

bim9262:
Thanks @cfsmp3 , I'm not familiar with taskwarrior, let alone older versions :sweat_smile:.

PR #2235, opened 2026-01-25
Fix format

antoyo:
It seems like my last PR #2232 broke the formatting.
We didn't run the CI, so I guess that's why we didn't notice.

Also, somehow, `cargo fmt` doesn't fix the formatting for me. I had to make these changes manually from the CI output.
Am I doing anything wrong?

### DIFF
--- a/src/blocks/notify.rs
+++ b/src/blocks/notify.rs
@@ -95,8 +95,11 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
     };
 
     loop {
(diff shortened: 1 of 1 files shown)


cfsmp3:
> It seems like my last PR #2232 broke the formatting. We didn't run the CI, so I guess that's why we didn't notice.
> 
> Also, somehow, `cargo fmt` doesn't fix the formatting for me. I had to make these changes manually from the CI output. Am I doing anything wrong?

You're probably using an old version of cargo.

PR #2236, opened 2026-01-26
fix(alsa): Don't leak alsactl process

bim9262:
This bug was found when alsactl was repeatedly failing due to a missing device causing a number of alsactl processes to be running.

This PR uses tokio's kill_on_drop and an added the process handle to a field in the device.

@exrok can you double-check this PR? (Even when I use the alsa driver to set the volume, `alsactl monitor` doesn't output anything on my system).


exrok:
Yep that looks good and should fix the issue, thanks.

PR #2238, opened 2026-01-31
block[vpn]: add driver for tailscale

trnila:
- `country` and `country_flag` is optional
- added new `tailscale` driver
- `Connected` and `Disconnected` state is having optional `profile` that is used to show selected tailnet in tailscale


bim9262:
Rebased change and added requisite fixes for warp driver

PR #2243, opened 2026-02-13
Change the default error_fullscreen_format

### LINKED ISSUE #1898: Change the default `error_fullscreen_format`
Some errors are way too long to fit into the width of the status bar. Perhaps a better default for `error_fullscreen_format` would be something like

```toml
[error_fullscreen_format]
full = " $full_error_message "
short = " $full_error_message.str(w:170, rot_interval:0.2) "
```

ayushrawat-10:
-> Changes the default `error_fullscreen_format` to use auto-scrolling text, now long error messages don't overflow the status bar.

-> The new default uses `.str(w:170,rot_interval:0.2)` (as mentioned) which limits the displayed width to 170 characters and scrolls the text every 0.2 seconds.

Fixes #1898


bim9262:
This looks good to me. Thanks!

PR #2246, opened 2026-02-20
Use native github doc deploy action

### LINKED ISSUE #2244: [for maintainers] change github pages deploy action
Currently this thirdparty action is used: JamesIves/github-pages-deploy-action

However looks like we can use "native" action from GitHub itself instead, see https://github.com/thomvaill/log4brains/issues/133

Note: if we decide not to change anything then we should at least pin the current action via commit hash instead of tag

I think only @greshake has access to the Pages setting

bim9262:
Resolves #2244

@greshake if you could go update the settings like this to use Github Actions as the soruce:

image

And then merge this I beleive the action will be able to deploy the pages.

You should be able to see what the build looks like on my fork here: https://github.com/bim9262/i3status-rust/actions/runs/22244468055

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -9,8 +9,8 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  default:
-
+  # Build job
+  build:
     runs-on: ubuntu-latest
     container:
       image: archlinux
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -25,7 +25,30 @@ jobs:
       run: cargo doc --all-features --no-deps
     - name: Create index.html
       run: echo "<meta http-equiv=\"Refresh\" content=\"0; url='/i3status-rust/i3status_rs'\" />" > target/doc/index.html
-    - name: Deploy 🚀
-      uses: JamesIves/github-pages-deploy-action@v4
+    - name: Upload static files as artifact
+      id: deployment
```
ammgws: If this `id` is not referenced anywhere in this build job then we can remove this line

bim9262:
@greshake could you make the changes to the settings (notes in the first comment) and then merge this PR?

Thanks!

bim9262:
@greshake , could you take a look at updating the settings to enable this?

Thanks!

PR #2248, opened 2026-02-26
block[speedtest]: Use networkquality-rs

### LINKED ISSUE #2245: speedtest is discontinued
The referenced https://github.com/sivel/speedtest-cli utility became a public archive recently.

Since it seems discontinued/unmaintained, it might be a good idea to drop support for it?

bim9262:
The speedtest block was using [speedtest-cli](https://github.com/sivel/speedtest-cli) which has been become archived and is no longer maintained.

This commit replaces speedtest-cli with [networkquality-rs](https://github.com/cloudflare/networkquality-rs).

Note that this requires `cmake` to be installed.
I had to switch over to `rustls` as not doing so causes linking errors.
The new version of `reqwest` moved to `rustls`, so we'll move in that direction anyways.

The nq crates are not published on crates.io so they are being checked out directly from github.

Resolves #2245

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,12 +61,16 @@ neli-wifi = { version = "0.6", features = ["async"] }
 nix = { version = "0.29", features = ["fs", "process"] }
 nom = "8.0.0"
 notmuch = { version = "0.8", optional = true }
-oauth2 = { version = "5.0.0", default-features = false, features = ["reqwest"] }
+nq-core = { git = "https://github.com/cloudflare/networkquality-rs.git", rev = "cbad8dc970ca0d3f68a50d76c78a253317cffd22" }
+nq-latency = { git = "https://github.com/cloudflare/networkquality-rs.git", rev = "cbad8dc970ca0d3f68a50d76c78a253317cffd22" }
+nq-rpm = { git = "https://github.com/cloudflare/networkquality-rs.git", rev = "cbad8dc970ca0d3f68a50d76c78a253317cffd22" }
+nq-tokio-network = { git = "https://github.com/cloudflare/networkquality-rs.git", rev = "cbad8dc970ca0d3f68a50d76c78a253317cffd22" }
 num-traits = "0.2"
+oauth2 = { version = "5.0.0", default-features = false, features = ["reqwest"] }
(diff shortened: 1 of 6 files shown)


bim9262:
@snaeil feel free to provide your feedback!

snaeil:
@bim9262 I am not yet familiar with the repo but overall looks good to me. Since the commit message is _almost_ [conventional ](https://www.conventionalcommits.org/) I would how you deal with breaking changes here and whether there's a need to make this clear somewhere

bim9262:
@snaeil , I assume you're refering to the addition of `speedtest_args` to `Config`.
We'll note that in `NEWS.md` that the block depends on network `networkquality-rs` instead of `speedtest-cli`.
We don't usually note changes to the config specifically, unless they're breaking from the perspective of a user of the binary.

bim9262:
I've reached out in https://github.com/cloudflare/networkquality-rs/issues/38 to see if it's possible for the crates to be published to crates.io instead of pulling directly from github.

bim9262:
@ammgws, I know that we generally don't like having git repos as a source for a crate. As this this crate won't be used as another source by another crate do you feel that changes the calculus?

Another idea would be to fork the repo, then rename the crates to be prefixed by `i3status- `, and then publish those crates.

ammgws:
We've done it before but for projects maintained by us. In this case however, is there a rush to move on from speedtest-cli crate even if it's archived? It still works, right? Maybe by the time it actually becomes an issue then cloudflare has published crates for their repo :shrug:

bim9262:
I guess it's only a problem in that if there is a CVE for any of the dependencies  for `speedtest-cli` there won't be any update to `speedtest-cli` itself.

Cargo.toml
```diff
@@ -61,12 +61,13 @@ neli-wifi = { version = "0.6", features = ["async"] }
 nix = { version = "0.29", features = ["fs", "process"] }
 nom = "8.0.0"
 notmuch = { version = "0.8", optional = true }
+cf-mach = { git = "https://github.com/cloudflare/networkquality-rs.git", rev = "a139c893c95119d875e8e60d6d781a743d9facbf" }
```
ammgws: should be able to use the crate now :)

PR #2254, opened 2026-03-13
blocks::packages: add zypper support

cfconrad:
Use `zypper list-upgrades` to find upgradable packages on openSUSE


bim9262:
Looks good! Thanks :)

PR #2255, opened 2026-03-14
block[sound]: Restore pulseauido as the default

### LINKED ISSUE #2253: Sound blocks broken in 0.36
Since update from 0.35 to 0.36:

- the mic icon is always on in the `sound` block, `source` device
- it is blocked at 0%
- volume is always on
- it is blocked at 0% in the status bar (but is actually at the last set volume that can't be changed anymore)

Here is my config for info:

```toml
[[block]]
block = "sound"
step_width = 2
device_kind = "source"
format = "  $icon  {$volume.eng(w:2)|}"
[block.theme_overrides]
idle_bg = "#204060"
warning_bg = "#204060"
warning_fg = "#a6b7bf"

[[block]]
block = "sound"
step_width = 2
format = "  $icon  {$volume.eng(w:2)|}  "
[[block.click]]
button = "left"
cmd = "pavucontrol"
[block.theme_overrides]
warning_bg = "#204060"
warning_fg = "#a6b7bf"
```

OS: ArchLinux 6.19.6-arch1-1
Sound software: pavucontrol 1:6.2-1
WM: i3-wm 4.25.1-1

bim9262:
Additionally handle pipwire thread failure

Partially resolves #2253

### DIFF
--- a/src/blocks/privacy/pipewire.rs
+++ b/src/blocks/privacy/pipewire.rs
@@ -2,7 +2,7 @@ use itertools::Itertools as _;
 use tokio::sync::mpsc::{UnboundedReceiver, unbounded_channel};
 
 use super::*;
-use crate::pipewire::{CLIENT, EventKind, Link, Node};
+use crate::pipewire::{CLIENT, EventKind, Link, Node, PIPEWIRE_CONNECTION_ERROR_MSG};
 
 #[derive(Deserialize, Debug, SmartDefault)]
 #[serde(rename_all = "lowercase", deny_unknown_fields, default)]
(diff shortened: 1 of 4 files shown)


bim9262:
@prosoitos feel free to test out this PR, which should restore the behavior that you saw in 0.35 where you didn't need to set the driver explicitly.

@fwieber feel free to check this out too, I imagine that when the driver isn't set (which uses the auto driver) that the sound block will work. However if you want to be able to use the pipewire driver more work/testing will be required. The pipewire driver wasn't tested with bluetooth devices before.

bim9262:
@ammgws do you have any thoughts on which driver is selected first by the auto driver?

I'm of two minds.

1. Pipewire is the new audio stack and should take presidence over pulseaudo, if you've built with the pipewire feature.
2. However, just because you are using the status bar with the pipewire feature, you may still not have migrated to pipewire. (For example Arch has pipewire and pulseaudio enabled).

With out reordering the pipewire and pulseaudo drivers within the auto driver, but with the other changes in this PR to detect if the pipewire thread crashes, pipewire will fail to connect and when the sound block retires pulseaudo would be used.

ammgws:
I'd do whatever keeps best compat for now. On phone so can't check properly but if pulse was default then leave it. Users who went out there way to build only with pipewire can make sure to update their block configuration too and help fix remaining pipewire issues with Bluetooth etc.

PR #2263, opened 2026-04-06
One more Font Awesome battery icon

GreyXor:
Hello,

Font Awesome 6 bat icons progress miss this one : https://fontawesome.com/icons/battery-low

### DIFF
--- a/files/icons/awesome6.toml
+++ b/files/icons/awesome6.toml
@@ -10,6 +10,7 @@ bat_charging = "\uf1e6"
 bat_not_available = "\uf244"
 bat = [
     "\uf244",
+    "\ue0b1",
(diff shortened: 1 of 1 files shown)


bim9262:
We don't have any pro glyfs in our themes

GreyXor:
> We don't have any pro glyfs in our themes

ah ok. Would that make sense to have a new  `fontawesome7pro.toml` theme ?

bim9262:
I think that would be ok

GreyXor:
Alright thanks, I'll think about it.

GreyXor:
https://github.com/greshake/i3status-rust/pull/2267

PR #2264, opened 2026-04-06
One more Font Awesome volume icon

GreyXor:
Hello,

Font Awesome 6 vol icons progress miss this one : https://fontawesome.com/icons/volume

image

### DIFF
--- a/files/icons/awesome5.toml
+++ b/files/icons/awesome5.toml
@@ -71,9 +71,10 @@ unknown = "\uf128"
 update = "\uf062"
 uptime = "\uf2f2"
 volume = [
-	"\uf026",
-	"\uf027",
-	"\uf028",
+    "\uf026",
+    "\uf027",
+    "\uf6a8",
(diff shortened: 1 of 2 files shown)


bim9262 (changes requested):
This was as added in font awesome 5. Can you add it to that font file too?

GreyXor:
good catch, added to fa5 too

bim9262:
Hey @GreyXor, sorry I went back to double check `f6a8` (volume), but realized that it's actually a pro font, until v7.
I'd accept this in a new awesome7 font config, but we can't add them to the free fa5/6.

GreyXor:
> Hey @GreyXor, sorry I went back to double check `f6a8` (volume), but realized that it's actually a pro font, until v7. I'd accept this in a new awesome7 font config, but we can't add them to the free fa5/6.

Yeah, make sense. so I think we can now close this PR. Thanks for review

PR #2266, opened 2026-04-07
block[weather]: fix place name geocoding for OpenWeatherMap

cdbrkfxrpt:
Using `query_pairs_mut().append_pair()` to build the geo URL percent-encodes the comma in place names like "London,UK", turning it into "London%2CUK" and causing the request to 404. Build the query string directly instead to preserve the comma as-is.

While here, move `get_location_query` out of `impl Service` into a free function and resolve the API key from `config` internally rather than threading it through as a parameter. Add a trailing slash to `GEO_URL` so that `Url::join` works correctly with relative path segments. Add an ignored integration test for place name resolution.

### DIFF
--- a/src/blocks/weather/open_weather_map.rs
+++ b/src/blocks/weather/open_weather_map.rs
@@ -3,7 +3,7 @@ use chrono::{DateTime, Utc};
 use reqwest::Url;
 use serde::{Deserializer, de};
 
-pub(super) const GEO_URL: &str = "https://api.openweathermap.org/geo/1.0";
+pub(super) const GEO_URL: &str = "https://api.openweathermap.org/geo/1.0/";
 pub(super) const CURRENT_URL: &str = "https://api.openweathermap.org/data/2.5/weather";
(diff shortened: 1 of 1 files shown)


src/blocks/weather/open_weather_map.rs
```diff
@@ -1,9 +1,11 @@
+// Copyright 2026 bmc::labs GmbH. All rights reserved.
```
ammgws: Why do we need this?

cdbrkfxrpt: we don't, that's an IDE autoinsert, sorry

cdbrkfxrpt: removed

ammgws:
> While here, move get_location_query out of impl Service into a free function and resolve the API key from config internally rather than threading it through as a parameter.

Can you please move this into a separate commit?

cdbrkfxrpt:
> > While here, move get_location_query out of impl Service into a free function and resolve the API key from config internally rather than threading it through as a parameter.
> 
> Can you please move this into a separate commit?

done

ammgws (approved):

PR #2267, opened 2026-04-21
New Font Awesome 7 Pro icons theme

GreyXor:
Hello,

This is a new Font Awesome 7 Pro icon theme that allows us to use icons from the Pro collection.

The set may include icons from both the Free and Pro collections. It’s still a work in progress, as I’m continuing to expand it with additional icons, particularly from the Pro collection, which offers improved visual options compared to the Free set.

For each completed icon, you’ll find its Font Awesome URL included in the comments, giving you a sense of the current progress.

Context: https://github.com/greshake/i3status-rust/pull/2263#issuecomment-4258304637

### DIFF
--- a/files/icons/awesome7pro.toml
+++ b/files/icons/awesome7pro.toml
@@ -1,0 +1,112 @@
+# Font Awesome 7 Pro (https://fontawesome.com/)
+#
+# The icons used here come from both the Font Awesome 7 "Free" and "Pro" collections.
+#
+# - Free icons: https://fontawesome.com/search?ic=free-collection
+# - Pro icons: https://fontawesome.com/search?ic=pro-collection
+
+backlight = [
+    "\ue0ca", #  https://fontawesome.com/icons/brightness-low
+    "\ue0c9", #  https://fontawesome.com/icons/brightness
+    "\ue28f", #  https://fontawesome.com/icons/sun-bright
+]
+bat_charging = "\uf376" #  https://fontawesome.com/icons/battery-bolt
+bat_not_available = "\uf377" #  https://fontawesome.com/icons/battery-slash
+bat = [
+    "\uf244", #  https://fontawesome.com/icons/battery-empty
+    "\ue0b1", #  https://fontawesome.com/icons/battery-low
+    "\uf243", #  https://fontawesome.com/icons/battery-quarter
+    "\uf242", #  https://fontawesome.com/icons/battery-half
+    "\uf241", #  https://fontawesome.com/icons/battery-three-quarters
+    "\uf240", #  https://fontawesome.com/icons/battery-full
+]
+bell = "\uf0f3" #  https://fontawesome.com/icons/bell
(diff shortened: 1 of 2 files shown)


files/icons/awesome7pro.toml
```diff
@@ -0,0 +1,103 @@
...
+# - Free icons: https://fontawesome.com/search?ic=free-collection
+# - Pro icons: https://fontawesome.com/search?ic=pro-collection
+
+backlight = [
+    "\ue0ca", # https://fontawesome.com/icons/brightness-low
+    "\ue0c9", # https://fontawesome.com/icons/brightness
+    "\ue28f", # https://fontawesome.com/icons/sun-bright
+]
+bat_charging = "\uf376" # https://fontawesome.com/icons/battery-bolt
+bat_not_available = "\uf377" # https://fontawesome.com/icons/battery-slash
```
GreyXor: For `bat_not_available` i'm torn between 

- image
- image

GreyXor (commented):
In my opinion, "slash" is better for "no battery" as "exclamation" is more for "problem with battery"

GreyXor:
Re-added all icons and used better icons more adequate from "pro".

GreyXor:
Hello @bim9262,

Can I directly use UTF-8 glyphs in TOML strings, similar to the [emoji icon set](https://github.com/greshake/i3status-rust/blob/master/files/icons/emoji.toml), instead of using escape sequences like `\u...`? I already tested, it works as expected

For example:

```toml
refresh = ""
resolution = ""
scratchpad = ""
```

instead of:

```toml
refresh = "\uf021"
resolution = "\ue278"
scratchpad = "\uf2d2"
```

This approach would make it easier for users who have the appropriate font installed to immediately recognize the icons, and it avoids the need to include comments describing each glyph.


image
image

bim9262:
I'd recommend not using the icons instead of the utf-8 codes.
This makes it much more difficult for somebody without the fonts to verify the icons as well as being really annoying when working on multiple themes as you'd have to set your font for each file you're reviewing.

GreyXor:
> I'd recommend not using the icons instead of the utf-8 codes. This makes it much more difficult for somebody without the fonts to verify the icons as well as being really annoying when working on multiple themes as you'd have to set your font for each file you're reviewing.

Thanks for your input! I reverted the utf-8 codes

GreyXor:
Hi @Bim9262,
Do we need anything else for the merge? Thanks!
FYI, I've been using this font theme myself for a month.

ammgws (approved):

ammgws:
Add to docs and we are all set
I.e.
https://github.com/greshake/i3status-rust/blob/master/doc/themes.md#available-icon-sets

Squash into one commit please

GreyXor:
> Add to docs and we are all set I.e. https://github.com/greshake/i3status-rust/blob/master/doc/themes.md#available-icon-sets
> 
> Squash into one commit please

Thanks, done!

PR #2270, opened 2026-04-24
Fix mullvad vpn

### LINKED ISSUE #1882: [Feature] Add `Connecting` status to VPN block
Right now vpn block has only 3 statuses: `Connected`, `Disconnected` and `Error`.
The way drivers work now leads to "connecting" status being displayed as `Error`.
I can implement it, but I don't have a NordVPN account, so I'll need help with testing of `NordVpnDriver`.

kl:
The Mullvad status parsing was done by calling `mullvad status` and parsing this output which has changed so the country code was not extracted correctly.

This PR changes that instead parse the JSON output from `mullvad status -j`, which should be more robust.

It also adds a connecting state to the VPN block so that the error icon is not shown when the VPN is connecting. Note that I only implemented the connecting state for Mullvad, so the other VPNs will behave as they did before and show the error icon when connecting.

Closes: https://github.com/greshake/i3status-rust/issues/1882

### DIFF
--- a/src/blocks/vpn.rs
+++ b/src/blocks/vpn.rs
@@ -10,6 +10,7 @@
 //! `interval` | Update interval in seconds. | `10`
 //! `format_connected` | A string to customise the output in case the network is connected. See below for available placeholders. | `" VPN: $icon "`
 //! `format_disconnected` | A string to customise the output in case the network is disconnected. See below for available placeholders. | `" VPN: $icon "`
+//! `format_connecting` | A string to customise the output in case the network is in a connecting state. See below for available placeholders. | `" VPN: $icon "`
 //! `state_connected` | The widgets state if the vpn network is connected. | `info`
 //! `state_disconnected` | The widgets state if the vpn network is disconnected | `idle`
 //!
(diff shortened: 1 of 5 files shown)


ammgws:
Has the JSON output flag always been there or is it only in certain newer versions of mullvad cli?

kl:
It's been there since the beginning of 2025. Do you prefer adding a check to see if the flag is available and if not fall back to the old parsing?

ammgws:
The goal is not to break user's config if it is not guaranteed that all users have a version > 2025. It depends how users usually install mullvad. If it is not from a distro package manager then likely we can assume most or all users will have the latest mullvad client version. Since I do not use mullvad I am not sure - can you advise?

kl:
I think it is fine in this case as I think very few users would be on such an old version and the app would warn them that they are using an outdated version as well. The offical Mullvad Ubuntu and Fedora repos as well as the AUR package all have the latest version of Mullvad.

I also added a commit fixing an issue I found with the mullvad json status command.

ammgws:
Thanks

PR #2273, opened 2026-04-28
net block: allow defining absolute up and down bandwidth

GRBurst:
Hey there 👋,

thanks for the awesome bar, I am using it for quite some time and I happy with it :-)

Regarding, the network graph, for me, the current implementation is not very intuitive. The main reason is that it is relative and the base is constantly changing. I am usually interested in the network load in relation to my network capability, such that a bit "more" noise does not show up as spikes.

This PR adds the capability to define your network up and down speed. I am using this implementation on my site for quite some time already. It works stable and reliable on my site.

I am open for discussion, because although this is much more intuitive to me, it does not have to be the case for everyone. I still believe that for most users, it will be more intuitive. Let's discuss improvements / alternatives if you don't agree.

Here is an example:
```toml
[[block]]
block = "net"
network_speed_down = 100.0  # 100 Mbit/s down
network_speed_up = 20.0     # 20 Mbit/s up
```

The result would be:
- With 50 Mbit/s traffic:   `graph_down` shows 50% of full bar (not 100% due to spike)

### DIFF
--- a/src/blocks/net.rs
+++ b/src/blocks/net.rs
@@ -8,6 +8,8 @@
 //! ----|--------|--------
 //! `device` | Network interface to monitor (as specified in `/sys/class/net/`). Supports regex. | If not set, device will be automatically selected every `interval`
 //! `interval` | Update interval in seconds | `2`
+//! `network_speed_down` | Network speed in Mbit / s | `None`
+//! `network_speed_up` | Network speed in Mbit / s | `None`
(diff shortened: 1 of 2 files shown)


ammgws:
This sounds like it could be a nice option to have. Some users might be more interested in local network speeds so perhaps in that case it can default to the link speed?

src/blocks/net.rs
```diff
@@ -171,6 +197,12 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
                     "device" => Value::text(device.iface.name),
                 });
 
+                if speed_down / max_network_speeds.down > 0.8 || speed_up / max_network_speeds.up > 0.8 {
```
ammgws, line 200: Perhaps break this 0.8 out into a named variable

PR #2279, opened 2026-05-27
formater[datetime]: Added `precision` argument

bim9262:
Add `precision` argument to `datetime` formatter to be used with icu4x to specify the precision of the datetime.

@ammgws ,

Any thoughts on adding a argument to a formatter that's only used for `icu4x`, not `chrono`?

### DIFF
--- a/cspell.yaml
+++ b/cspell.yaml
@@ -206,6 +206,7 @@ words:
   - xresources
   - xtask
   - YMDE
+  - YMDET
   - YMDT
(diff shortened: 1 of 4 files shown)


ammgws (commented):
I don't see an issue with it - the format string spec is different between them and for chrono the equivalent of precision can be handled by the freedom of the format string.

bim9262:
Thanks @ammgws!

PR #2284, opened 2026-06-13
add manpage upload action

adamthiede:
It would be nice for this project to provide a manpage as a release artifact. Generating the manpage requires pandoc, and in Alpine linux, that's only available on two of the several architectures, so it can't be built in the regular package build step. If it were built upstream for each release that would be nice.

This is just a github action that will run on every newly published release. It will generate and upload the manpage to that release.

Thanks!

### DIFF
--- a/.github/workflows/manpage-upload.yml
+++ b/.github/workflows/manpage-upload.yml
@@ -1,0 +1,28 @@
+name: Man page release upload
+on:
+  release:
+    types: [published]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  manpage:
+    runs-on: ubuntu-latest
+    container:
+      image: archlinux:base-devel
+    permissions:
+      contents: write
+    steps:
+    - name: Get required packages and config git
+      run: |
+        pacman --noconfirm --noprogressbar --needed -Syu base-devel clang git libpipewire libpulse lm_sensors notmuch openssl pandoc github-cli
+        git config --global --add safe.directory "$GITHUB_WORKSPACE"
+    - uses: actions/checkout@v5
+    - uses: dtolnay/rust-toolchain@stable
+    - name: Generate and upload manpage
+      env:
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        cargo xtask generate-manpage
+        gh release upload -R ${{ github.repository }} --clobber ${{ github.event.release.tag_name }} ./man/i3status-rs.1


bim9262 (changes requested):
A few changes I'd like to see to make this more in line with the setup we have in the pre-commit workflow:

```diff
diff --git a/.github/workflows/manpage-upload.yml b/.github/workflows/manpage-upload.yml
index e2554e44b..cfb953739 100644
--- a/.github/workflows/manpage-upload.yml
+++ b/.github/workflows/manpage-upload.yml
@@ -11,18 +11,18 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: archlinux:base-devel
-    permissions: write-all
+    permissions:
+      contents: write
     steps:
-    - name: Install packages
-      run: pacman --noconfirm --noprogressbar --needed -Syu clang git libpipewire libpulse lm_sensors notmuch openssl pandoc github-cli
-    - uses: dtolnay/rust-toolchain@stable
+    - name: Get required packages and config git
+      run: |
+        pacman --noconfirm --noprogressbar -Syu clang git libpipewire libpulse lm_sensors notmuch openssl pandoc github-cli
+        git config --global --add safe.directory "$GITHUB_WORKSPACE"
     - uses: actions/checkout@v5
+    - uses: dtolnay/rust-toolchain@stable
     - name: Generate and upload manpage
       env:
         GH_TOKEN: ${{ github.token }}
       run: |
-        git config --global safe.directory '*'
-        git config --global safe.directory '/__w/i3status-rust/i3status-rust'
         cargo xtask generate-manpage
-        gh repo set-default ${{ github.repository }}
-        gh release upload --clobber ${{ github.event.release.tag_name }} ./man/i3status-rs.1
+        gh release upload -R ${{ github.repository }} --clobber ${{ github.event.release.tag_name }} ./man/i3status-rs.1
```

adamthiede:
Thank you for the fast response. I've updated per your suggestions and tested on my fork, and it works.

bim9262:
@adamthiede , you're welcome!

I was wondering what your rationale for not having the `--add` for the `safe.directory` is and why you've added the `--needed` to the pacman install. I ask so that I can decide if they're changes we should make to our other workflows.

Thanks!

adamthiede:
For not adding `--add` to git: my ignorance. It's probably nice to have. I wasn't aware $GITHUB_WORKSPACE was passed correctly into containers, either. So I'm glad that it worked.

For adding `--needed` to pacman: from the manpage it says "Do not reinstall the targets that are already up-to-date." Just shortens the install if already-installed packages are specified. I also used the base-devel image so that base-devel should already be installed and it won't waste extra time reinstalling.

bim9262:
I agree that not having `base-devel` in the list of install items and using the base image with them already installed is a good choice. None of the other items should be installed by default, but I guess adding the `--needed` doesn't hurt.

adamthiede:
Yes, I think it's good to keep everything required still in the list (and skip it with --needed if applicable) so you know exactly what packages are required for the CI.

I think this is ready now. Next release I will pull the manpage from the release, so alpine can have a properly packaged i3status-rust. :)

bim9262:
Thanks @adamthiede!

PR #2285, opened 2026-06-13
freebsd build

### LINKED ISSUE #2278: Fails to compile
OS: FreeBSD
rustc: 1.95.0

failure: https://gist.github.com/hasufell/f8b1ad1b3a7f9d5df69923682850de6e

bim9262:
@hasufell

I ended up making an attempt to make i3status-rust work on FreeBSD.

The two main differences that this introduced is that it doesn't seem like FreeBSD allows for the custom signals used on linux and some of the blocks/dependencies aren't compatible with FreeBSD.

Note that if a block failed, I disabled it entirely instead of going through each one and determining if it was only a certain driver/backend/service that needed to be disabled on non-linux systems. It's also possible that I wasn't able to build some of the features like notmuch or pipewire because I am missing a system library. If that's the case we can update the `pkg install` line to not need to fence off as many features.

@ammgws, any thoughts on adding this in and allowing folks to improve what's available to non-linux systems down the line? I wouldn't think that there's any non `target_os = "linux"` users today that this would impact.

### DIFF
--- a/.github/workflows/pre-commit-freebsd.yml
+++ b/.github/workflows/pre-commit-freebsd.yml
@@ -1,0 +1,25 @@
+name: pre-commit-freebsd
+
+on: [push, pull_request]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        hook: [
+                cargo-fmt,
+                cargo-clippy,
+                cargo-test,
+              ]
+    steps:
+    - uses: actions/checkout@v6
+    - name: Test in FreeBSD
+      uses: vmactions/freebsd-vm@v1
+      with:
+          run: |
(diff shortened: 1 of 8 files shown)


ammgws:
The only worry I have is if no one steps up to help maintain it but if it's not that much burden for us to have those flags then maybe its ok for now. Are you running freebsd daily or was it just for testing this?

bim9262:
My thought is that this should be little maintenance. If you create a new block and it fails the CI you can just throw the linux only flag on it. If a FreeBSD user wants the block they can work on making it compatible with both (again has to pass the CI). I am not a daily user of FreeBSD and therefore may have over excluded some blocks that would work, but just need the right packages to be installed in the CI.

bim9262:
@hasufell feel free to test this out!

PR #2288, opened 2026-07-14
hueshift: Add support for xsct

blueshiftlabs:
The original `sct` program is not available through many package managers. Its replacement, `xsct`, is more widely available. Since `sct` and `xsct` have (mostly*) compatible command lines, we can use the same `HueShiftDriver` for both by simply telling it to run `sct` or `xsct` respectively.

[*] There is a minor difference with the reset behavior - for `sct`, running it with an empty command line resets the color temperature, while for `xsct`, doing so prints usage. Running `[x]sct 0` resets the color temperature for both, so we use that for the `reset` function.

### DIFF
--- a/cspell.yaml
+++ b/cspell.yaml
@@ -204,6 +204,7 @@ words:
   - XKCD
   - xrandr
   - xresources
+  - xsct
   - xtask
   - YMDE
   - YMDET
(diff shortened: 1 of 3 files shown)


ammgws (approved):

ammgws:
Need to fixup the lints

blueshiftlabs:
Done, PTAL.

PR #2289, opened 2026-07-14
Rework launching for user-specified commands.

blueshiftlabs:
For user-specified commands, such as in the `custom` block, the `cmd`
parameter of `[[block.click]]`, and the `if_command` parameter of
`[[block]]`, make two changes for consistency:

- Always use the user's $SHELL, instead of `sh`, if it's configured.
- Pass an environment variable `I3STATUS_RS_PID` with the PID of the launching `i3status-rs` process. This allows custom commands to send signals to the bar without needing to use `pkill`, which may affect other bars running on the system.

### DIFF
--- a/cspell.yaml
+++ b/cspell.yaml
@@ -204,6 +204,7 @@ words:
   - XKCD
   - xrandr
   - xresources
+  - xsct
   - xtask
   - YMDE
   - YMDET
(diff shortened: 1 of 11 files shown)


blueshiftlabs:
Note: this PR has already been rebased onto #2288 because that PR fixes an unrelated `cargo clippy` issue - I'll update this PR after that one is submitted.

PR #2290, opened 2026-07-18
fix: use lowercase k as the SI unit prefix for kilo

jrosdahl:
The SI unit prefix for kilo is lowercase "k" (see <https://en.wikipedia.org/wiki/Metric_prefix>), so adjust parsing and formatting accordingly. For backward compatibility, uppercase "K" still works for parsing.


bim9262:
Thanks @jrosdahl !

PR #2293, opened 2026-07-29
Add ical support

esemeniuc:
## Summary

- Add direct, read-only ICS feed support to the calendar block with `type = "ics"`, keeping CalDAV as the default.
- Share and harden iCalendar parsing for recurrence overrides, RFC 5545 durations and DST behavior, malformed events, and expansion limits.
- Protect secret feed URLs in errors and enforce a streamed 32 MiB response limit.
- Document Google Calendar secret iCal setup without requiring Google Cloud or OAuth.

Tested on current arch with i3 4.25.1
image

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 ### New Blocks and Features
 
+* `calendar`: added support for direct iCalendar (`.ics`) feeds
 * `packages`: added support for `zypper` (#2254)
 
 ### Bug Fixes and Improvements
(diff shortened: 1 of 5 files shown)


esemeniuc:
bump @bim9262, been running this for a while now

bim9262:
@esemeniuc, I tried to review this, but too much changed in one go to wrap my head around. Can you break the PR into a series of commits for each change? I'm also not sure I follow why what the changes to the recurrence overides and the RFC 5545 durations and DST stuff is handled here and not in the icalendar crate. If it's not doing something right it's best to implement upstream instead of here.

PR #2295, opened 2026-07-31
fix: prevent external_ip block from hammering the geolocation service

cfsmp3:
## Problem

The `external_ip` block could end up rate limited by the geolocation service, and once limited it kept requesting fast enough that the limit never lifted. Several issues compounded:

1. **Busy loop when the signal stream ends.** If the D-Bus connection died (or `with_network_manager = false`, which used `stream::empty()`), polling the finished stream resolved immediately, turning the update loop into an unthrottled busy loop of API requests.
2. **Rate limit responses were retried like transient errors.** HTTP 429 (and ipapi.co's `"reason": "RateLimited"`) produced a generic error, so backon retried it (4 requests in ~7s), and when retries were exhausted the block's restart machinery re-ran `run()` every `error_interval` (5s default) — forever. Net effect: ~4 requests every ~12s at a server that is asking you to stop.
3. **Signal bursts on network transitions.** One VPN connect / Wi-Fi switch emits many NetworkManager `PropertiesChanged` signals over several seconds, often before connectivity is usable, causing several requests (plus failed-request retries) per transition. `next_debounced` only collapses signals already queued at poll time, not ones arriving during the in-flight HTTP request.

## Fix

- `geolocator`: new `RateLimited` error cause + `is_rate_limited()` helper; the ipapi.co backend maps HTTP 429 and the `"RateLimited"` reason to it, the ip2location.io backend maps HTTP 429.
- `external_ip`: skip retries for rate limit errors, keep showing the last known IP, and wait 10 minutes before asking again (instead of erroring out into the 5s restart loop).
- `external_ip`: chain a `pending` stream after the D-Bus stream (and use `stream::pending()` instead of `stream::empty()`) so a finished stream means "no more signals", not "wake up constantly".
- `external_ip`: after a signal wakes the loop, wait until the stream has been quiet for 2s before querying, and pass a 3s cache interval to `find_ip_location` instead of 0, so one network transition results in one request made once the new connection is likely up.

Network changes are still picked up promptly (~2s after the signal burst settles), and the periodic refresh interval is unchanged.

## Testing

- `cargo clippy --all-targets` clean
- `cargo test --lib` — 52 passed

### DIFF
--- a/src/blocks/external_ip.rs
+++ b/src/blocks/external_ip.rs
@@ -57,13 +57,30 @@
 //! periodic refresh exists to catch IP updates that don't trigger a notification,
 //! for example due to a IP refresh at the router.
 //!
+//! If the service reports rate limiting, the block keeps showing the last
+//! known IP and waits 10 minutes before asking again.
+//!
 //! Flags: They are not icons but unicode glyphs. You will need a font that
 //! includes them. Tested with: <https://www.babelstone.co.uk/Fonts/Flags.html>
 
 use zbus::MatchRule;
 
 use super::prelude::*;
+use crate::geolocator::is_rate_limited;
 use crate::util::{country_flag_from_iso_code, new_system_dbus_connection};
+
+/// Coalesces bursts of update triggers (a single network transition emits
+/// several D-Bus signals) into one API request.
+const CACHE_INTERVAL: Duration = Duration::from_secs(3);
+
+/// How long the network-change signal stream must stay quiet before the IP is
+/// re-queried; a transition keeps emitting signals for a while, often before
+/// connectivity is actually usable.
+const SETTLE_INTERVAL: Duration = Duration::from_secs(2);
+
+/// How long to wait before asking again after the service reported rate
+/// limiting; retrying sooner only prolongs the block.
+const RATE_LIMIT_INTERVAL: Duration = Duration::from_secs(600);
 
 #[derive(Deserialize, Debug, SmartDefault)]
 #[serde(deny_unknown_fields, default)]
@@ -123,9 +140,12 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
             .await
             .error("Failed to add match")?;
         let stream: zbus::MessageStream = dbus.into();
-        Box::pin(stream.map(|_| ()))
+        // If the D-Bus connection dies the stream ends; without the chained
+        // pending stream, polling it again would resolve instantly forever,
+        // turning the loop below into a busy loop of API requests.
+        Box::pin(stream.map(|_| ()).chain(futures::stream::pending()))
     } else {
-        Box::pin(futures::stream::empty())
+        Box::pin(futures::stream::pending())
     };
 
     let client = if config.use_ipv4 {
@@ -135,8 +155,23 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
     };
 
     loop {
-        let fetch_info = || api.find_ip_location(client, Duration::from_secs(0));
-        let info = fetch_info.retry(ExponentialBuilder::default()).await?;
+        let fetch_info = || api.find_ip_location(client, CACHE_INTERVAL);
+        let info = match fetch_info
+            .retry(ExponentialBuilder::default())
+            .when(|err| !is_rate_limited(err))
+            .await
+        {
+            Ok(info) => info,
+            Err(err) if is_rate_limited(&err) => {
+                // Keep displaying the last known IP and try again much later;
+                // erroring out here would make the block restart machinery
+                // re-query every `error_interval` seconds, which keeps the
+                // rate limit from ever lifting.
+                sleep(RATE_LIMIT_INTERVAL).await;
+                continue;
+            }
+            Err(err) => return Err(err),
+        };
 
         let mut values = map! {
             "ip" => Value::text(info.ip),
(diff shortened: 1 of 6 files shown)


src/geolocator/ipapi.rs
```diff
@@ -100,19 +100,32 @@ impl Ipapi {
...
+        let response: ApiResponse = response
             .json()
             .await
             .error("Failed while parsing location API result")?;
 
         if response.error {
             Err(Error {
                 message: Some("ipapi.co error".into()),
-                cause: Some(Arc::new(response.reason)),
+                cause: if response.reason.0.as_deref() == Some("RateLimited") {
```
bim9262: Is this reachable? Won't we have already returned the error when we get hit ` reqwest::StatusCode::TOO_MANY_REQUESTS`?

https://ipapi.co/api/#errors

src/blocks/external_ip.rs
```diff
@@ -135,8 +155,23 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
     };
 
     loop {
-        let fetch_info = || api.find_ip_location(client, Duration::from_secs(0));
-        let info = fetch_info.retry(ExponentialBuilder::default()).await?;
+        let fetch_info = || api.find_ip_location(client, CACHE_INTERVAL);
```
bim9262: Instead of using the `CACHE_INTERVAL`, perhaps it would be good to match what the weather block does and use `let autolocate_interval = config.autolocate_interval.unwrap_or(config.interval);`

I think that it may also make sense that instead of both blocks setting up a retry to handle that within the `find_ip_location` function and adding to the `Geolocator` when (if ever) the last rate-limit was hit so that both blocks respect the same rate limit interval. I think that exposing the rate limit interval as part of the geolocate config would be good too.

src/geolocator/ipapi.rs
cfsmp3: You're right — per their docs rate limit responses always come with HTTP 429, which is already handled above, so this branch was unreachable. Removed in c44c905.

src/blocks/external_ip.rs
cfsmp3: Done. 
Thanks for the review.
This is still a work in progress -shouldn't be a PR yet, so marked it as draft. I want to use for a few days before submitting it.

src/geolocator.rs
```diff
@@ -137,6 +206,21 @@ impl Geolocator {
...
+    /// rate limiting.
+    #[serde(default = "default_rate_limit_interval")]
+    #[default(default_rate_limit_interval())]
+    rate_limit_interval: Seconds,
+}
+
+fn default_rate_limit_interval() -> Seconds {
+    600.into()
+}
+
```
bim9262: I find this a bit easier to read

```suggestion
#[derive(Deserialize, Debug, SmartDefault)]
#[serde(default)]
pub struct GeolocatorConfig {
    #[serde(flatten)]
    backend: GeolocatorBackend,
    /// How long to wait before contacting the service again after it reported
    /// rate limiting.
    #[default(600.into())]
    rate_limit_interval: Seconds,
}
```

src/geolocator.rs
```diff
@@ -123,7 +172,27 @@ impl Geolocator {
...
+                && at.elapsed() < self.rate_limit_interval
+            {
+                return Err(Error {
+                    message: Some("geolocation service is rate limited, backing off".into()),
+                    cause: Some(Arc::new(RateLimited)),
+                });
+            }
+        }
+
+        let location = match self.backend.get_info(client).await {
```
bim9262: I'd put the retry logic here, and then remove if from both the external_ip and weather blocks

src/blocks/external_ip.rs
```diff
@@ -57,20 +58,41 @@
...
 #[derive(Deserialize, Debug, SmartDefault)]
 #[serde(deny_unknown_fields, default)]
 pub struct Config {
     pub format: FormatConfig,
     #[default(300.into())]
     pub interval: Seconds,
+    /// Unlike the weather block this defaults to 1 second, not `interval`:
+    /// picking up a fresh IP right after a network change is the whole point
+    /// of this block, so cached results must expire quickly.
+    pub autolocate_interval: Option<Seconds>,
```
bim9262: I'd make this have a default of 1 second instead of making it an option. I think that it makes reviewing the config to see what the defaults are easier.

src/blocks/external_ip.rs
```diff
@@ -135,8 +161,29 @@ pub async fn run(config: &Config, api: &CommonApi) -> Result<()> {
...
+                // so we don't wake up just before it expires); erroring out
+                // here would make the block restart machinery re-query every
+                // `error_interval` seconds, which keeps the rate limit from
+                // ever lifting.
+                sleep(api.locator_rate_limit_interval() + Duration::from_secs(1)).await;
+                continue;
+            }
+            Err(err) => return Err(err),
+        };
+        log::debug!(
```
bim9262: Can you replace these log lines and use `make_log_macro!(debug, "external_ip");` like the other blocks? It just makes the logs cleaner if we specify the target.

cfsmp3: > Can you replace these log lines and use `make_log_macro!(debug, "external_ip");` like the other blocks? It just makes the logs cleaner if we specify the target.

@bim9262 Done

bim9262: I don't see this change in the commits.

cfsmp3: > I don't see this change in the commits.

Of course I forgot to actually push. Sorry :-(

PR #2298, opened 2026-08-01
fix: use net inactive_format for disconnected interfaces

### LINKED ISSUE #1917: net inactive_format is not used when the interface is down and `$ip` placeholder not found
I'm not sure if this is intentional or not however `inactive_format` seems not to be used when the interface is down.  In my case I have a vpn which I turn on and off as needed.  It is defined as the following block:

```
[[block]]
block = "net"
device = "vpn0"
format = "$icon UP"
inactive_format = "$icon DOWN"
```

If the state of the adapter is down:

```
] > ip addr show dev vpn0
10: vpn0: <NO-CARRIER,POINTOPOINT,MULTICAST,NOARP,UP> mtu 1390 qdisc fq_codel state DOWN group default qlen 500
    link/none
    inet6 fe80::8cb5:e9db:d6a3:1d8/64 scope link stable-privacy
```

i3status-rs reports:

```
  {
    "full_text": " UP",
    "color": "#C1C1C1FF",
    "background": "#06060FFF",
    "name": "10",
    "instance": "10:",
    "separator": false,
    "separator_block_width": 0,
    "markup": "pango"
  },
```

This also seems to cause some issues with the `$ip` placeholder which reports `Failed to render full text (Cause: Plaseholder 'ip' not found)`

Everything works fine if I bring the vpn up though.

cfsmp3:
## Problem

`NetDevice::is_up()` had two independent holes:

- tun/wireguard/ppp interfaces were special-cased as always active, even when their operational state was down
- every interface with `Operstate::Up` was considered active even when it had no address

The second case occurs when an Ethernet cable still has carrier but NetworkManager disconnects the device and removes its addresses. The link remains `Up`, so the net block selected `format`; a required placeholder such as `$ip` was then missing and produced `Failed to render full text` instead of using `inactive_format`.

## Fix

Remove the VPN always-active special case and define an active interface as one that:

1. has an operational state that can carry traffic (`Up` or `Unknown`), and
2. has an IPv4 or IPv6 address.

| operstate | address | active |
|---|---:|---:|
| `Up` | yes | yes |
| `Up` | no | no (connected cable, NetworkManager disconnected) |
| `Unknown` | yes | yes (typical active tun/wireguard device) |
| `Unknown` | no | no |
| any other state | any | no |

This also keeps a down VPN inactive even if it retains a stale or link-local address.

Unit tests cover the complete state/address behavior, including both reproductions from #1917.

Fixes #1917

### DIFF
--- a/src/netlink.rs
+++ b/src/netlink.rs
@@ -105,15 +105,17 @@ impl NetDevice {
         }))
     }
 
-    /// Note: tun/wireguard/ppp interfaces are not special-cased to be always
-    /// up (as they were since #350): they usually report `Operstate::Unknown`,
-    /// which is covered by the has-an-address clause below, and the
-    /// special-casing made `inactive_format` unreachable for VPN interfaces
-    /// that are actually down (#1917).
+    /// An interface is active only when its operational state allows traffic
+    /// and it has an address. In particular, `Operstate::Up` only describes
+    /// the link state, so an Ethernet interface can remain `Up` after a
+    /// network manager disconnects it and removes its addresses (#1917).
+    ///
+    /// Tun/wireguard/ppp interfaces are not special-cased to be always up (as
+    /// they were since #350): they usually report `Operstate::Unknown`, which
+    /// is covered by the has-an-address clause below.
     pub fn is_up(&self) -> bool {
-        self.iface.operstate == Operstate::Up
-            || (self.iface.operstate == Operstate::Unknown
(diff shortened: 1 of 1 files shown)


cfsmp3:
@bim9262 @MaxVerevkin @ammgws  Can you review this, since it fixes a reported bug? https://github.com/greshake/i3status-rust/issues/1917

PR #2302, opened 2026-08-02
WIP: Replace chrono with jiff

bim9262:
`chrono` has been soft deprecated. `jiff` is the recommended replacement. https://github.com/chronotope/chrono/issues/1768


bim9262:
after all, I realized that jiff doesn't support localizing strtime :(

ammgws:
> after all, I realized that jiff doesn't support localizing strtime :(

Do they state why, or alternative other than chrono?

bim9262:
https://github.com/BurntSushi/jiff/issues/4

Looks like they suggest using icu for anything that's not English

PR #2304, opened 2026-08-08
Icon font (experimental)

cfsmp3:
A LLM generated font that contains all the icons we use, in SVG format, plus the tooling to build the actual font.

Font is usable but ugly - still, we might want to ship it, maybe a decent designer can replace worst icons?


src/icons.rs
```diff
@@ -2,6 +2,8 @@ use crate::errors::*;
 use crate::util;
 use serde::Deserialize;
 use std::collections::HashMap;
+#[cfg(test)]
+use std::collections::HashSet;
```
bim9262: Why not put this in the tests mod?

fonts/i3status-icons/i3status-icons.ttf
bim9262, line 1: I think it would be best not to commit this and generate as needed. Like the manpage-upload github workflow this could be generated as part of the release.

cfsmp3: > I think it would be best not to commit this and generate as needed. Like the manpage-upload github workflow this could be generated as part of the release.

Yes, good idea. Changed.

PR #2305, opened 2026-08-14
block[music]: fix combo doc

bim9262:
Fixed combo output to match documentation: `"$artist[sep]$title"`.

Also fixed order of ` and " in the doc.

### DIFF
--- a/src/blocks/music.rs
+++ b/src/blocks/music.rs
@@ -35,7 +35,8 @@
 //! `artist`      | Current artist | Text
 //! `title`       | Current title  | Text
 //! `url`         | Current song url | Text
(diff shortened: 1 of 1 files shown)


ammgws:
I think we should rather fix the docs instead and leave the code as-is

bim9262:
I had considered that too, but I think that artist - title is actually more usual when labeling music. Here are some arguments https://www.reddit.com/r/Music/comments/18kdey7/is_it_artist_song_or_song_artist/

Perhaps we should have a `combo` and `combo_reversed`...

ammgws:
Yeah I suggested so we don't affect existing configs.

ammgws (approved):

PR #2306, opened 2026-08-15
refactor: give blocks a declared output contract

cfsmp3:
This is preparation for the doctor feature.

The idea is that all blocks now declare what they will render (icons used, format strings, etc.) so that it can be checked by the framework itself, and inspected by an external subsystem (doctor) without running the block.

Each block module defines

```rust
prepare(config) -> Result<Arc<BlockPlan>>
```

which returns one `OutputPlan` per output variant the block can produce: an id, the effective format for that variant (defaults and inheritance already resolved), and the icon names each icon-valued placeholder is allowed to carry — `IconChoices::Fixed` for a closed set, including names that come from configuration, or `IconChoices::OpenResolvable` where the name is only known at runtime. A block that renders plain runtime text rather than a format (`menu`) declares an `OutputKind::Text` output instead.

The framework prepares the plan once and passes it to `run`. Blocks render through `OutputHandle`s taken from it: the handle installs that output's format on the widget and mints icon values, so an icon name that was never declared cannot be rendered. The check runs when a widget is published, not when a value is constructed, so an icon built outside the plan is still caught. Declaring an icon does not resolve it — resolution stays lazy, and configurations that omit icons their formats never reference keep working.

The invariant is that everything a block puts on the bar comes from a declared output:

- publishing a format or text widget without a handle of the matching kind is a contract failure;
- a plan that declares no outputs is rejected when it is built (and `BlockPlan` is deliberately not `Default`, so there is no second way to make one), so the contract cannot be satisfied vacuously;
- `Widget::set_format` and `set_text` void the contract rather than quietly keeping it, and there is no setter for the contract alone — source and handle are only ever set together;
- a widget with no source renders nothing and is exempt.

### DIFF
--- a/src/block_plan.rs
+++ b/src/block_plan.rs
@@ -1,0 +1,769 @@
+//! Block output contracts.
+//!
+//! Every standard block prepares a [`BlockPlan`] from its configuration before
+//! running. The plan declares each output variant the block can produce, the
+//! effective format for that output (defaults and inheritance already
+//! resolved), and the icon names each icon-valued placeholder can carry. A
+//! block that renders plain runtime text rather than a format declares that
+//! too, as an [`OutputKind::Text`] output.
+//!
+//! The runtime renders through [`OutputHandle`]s derived from the plan, which
+//! install the effective format and can look up the icon names the plan
+//! declares. Icons are checked in one place, when a widget is published (see
+//! [`Widget::check_contract`][crate::widget::Widget]): every icon value it
+//! carries must be one the output declared, however that value was built.
+//!
+//! The invariant is that everything a block puts on the bar comes from a
+//! declared output: publishing a format or text widget without a handle of
+//! the matching kind fails, a plan that declares nothing is rejected when it
+//! is built, and [`Widget::set_format`][crate::widget::Widget] and
+//! [`Widget::set_text`][crate::widget::Widget] void the contract rather than
+//! quietly keeping it. A widget with no source at all renders nothing and is
+//! exempt.
+//!
+//! Because the plan is prepared before the block runs, a block's icon and
+//! format surface can also be inspected without running it, and what such an
+//! inspection sees cannot drift from what the block actually renders.
+//!
+//! Declaring an icon does not resolve it. Resolution stays lazy: an icon is
+//! looked up in the icon set only when a reachable format actually renders it,
+//! so configurations that omit icons their formats never reference keep
+//! working.
+
+use std::borrow::Cow;
+use std::sync::Arc;
+
+use crate::errors::*;
+use crate::formatting::Format;
+use crate::formatting::value::{Value, ValueInner};
+use crate::widget::Widget;
+
+/// The icon the bar draws on a restartable block error, declared by
+/// [`error_plan`] and rendered by the error widget.
+pub(crate) const RESTART_ICON: &str = "refresh";
+
+/// Marks an error caused by a block breaking its own contract rather than
+/// by the user's configuration. Every contract failure is prefixed with it,
+/// so the two kinds of error can be told apart by their message.
+pub(crate) const CONTRACT_BUG: &str = "block contract bug";
+
+/// The icon names one placeholder of one output variant can carry.
+#[derive(Debug, Clone)]
+pub(crate) enum IconChoices {
+    /// A closed set, fully known once the block's configuration is parsed.
+    /// Includes configuration-derived names (e.g. `toggle`'s `icon_on`).
+    Fixed(Vec<Cow<'static, str>>),
+    /// Any icon name is permitted and resolves through the normal icon set
+    /// and override rules at render time. Reserved for deliberately dynamic
+    /// blocks (`custom`, `custom_dbus`); standard blocks declare closed sets.
+    OpenResolvable,
+}
+
+impl IconChoices {
+    pub(crate) fn fixed<I, S>(names: I) -> Self
+    where
+        I: IntoIterator<Item = S>,
+        S: Into<Cow<'static, str>>,
+    {
+        Self::Fixed(names.into_iter().map(Into::into).collect())
+    }
+
+    pub(crate) fn one<S: Into<Cow<'static, str>>>(name: S) -> Self {
+        Self::Fixed(vec![name.into()])
+    }
+
+    pub(crate) fn permits(&self, name: &str) -> bool {
+        // An empty icon name renders as empty output (a runtime no-op), so
+        // it is always permitted.
+        name.is_empty()
+            || match self {
+                Self::Fixed(names) => names.iter().any(|n| n == name),
+                Self::OpenResolvable => true,
+            }
+    }
+}
+
+/// What an output renders.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum OutputKind {
+    /// A format template: placeholders, and icons among them.
+    Format,
+    /// Plain text computed at runtime, with no placeholders and no icons
+    /// (`menu`). The text itself cannot be known before the block runs, but
+    /// declaring the output keeps the block inside the contract: it states
+    /// that this is all the block renders.
+    Text,
+}
+
+/// One output variant of a block: a named state, its effective format, and
+/// the icons each icon-valued placeholder may carry in that state.
+#[derive(Debug, Clone)]
+pub(crate) struct OutputPlan {
+    id: Cow<'static, str>,
+    kind: OutputKind,
+    format: Format,
+    icons: Vec<(&'static str, IconChoices)>,
+}
+
+impl OutputPlan {
+    pub(crate) fn new(id: impl Into<Cow<'static, str>>, format: Format) -> Self {
+        Self {
+            id: id.into(),
+            kind: OutputKind::Format,
+            format,
+            icons: Vec::new(),
+        }
+    }
+
+    /// An output that renders plain runtime text instead of a format.
+    pub(crate) fn text(id: impl Into<Cow<'static, str>>) -> Self {
+        Self {
+            id: id.into(),
+            kind: OutputKind::Text,
+            format: Format::new(
+                "".parse().expect("the empty template always parses"),
+                "".parse().expect("the empty template always parses"),
+            ),
+            icons: Vec::new(),
+        }
+    }
+
+    pub(crate) fn id(&self) -> &str {
+        &self.id
+    }
+
+    pub(crate) fn kind(&self) -> OutputKind {
+        self.kind
+    }
+
+    /// The effective format of this output. A [`OutputKind::Text`] output
+    /// renders no template, so its format is empty.
+    pub(crate) fn format(&self) -> &Format {
+        &self.format
+    }
+
+    /// Declare the icon choices for an icon-valued placeholder.
+    /// Duplicate declarations are rejected when the plan is built.
+    pub(crate) fn icon(mut self, placeholder: &'static str, choices: IconChoices) -> Self {
+        self.icons.push((placeholder, choices));
+        self
+    }
+
+    pub(crate) fn choices_for(&self, placeholder: &str) -> Option<&IconChoices> {
+        self.icons
+            .iter()
+            .find(|(p, _)| *p == placeholder)
+            .map(|(_, c)| c)
+    }
+
+    /// The icon-valued placeholders this output declares. The bar itself
+    /// renders through [`OutputHandle`] and looks placeholders up by name, so
+    /// nothing in the hot path needs this — it exists because a plan that can
+    /// only be queried by a name you already know is not inspectable. The
+    /// declaration tests are its consumer today.
+    #[cfg(test)]
+    pub(crate) fn icon_placeholders(&self) -> impl Iterator<Item = (&'static str, &IconChoices)> {
+        self.icons.iter().map(|(p, c)| (*p, c))
+    }
+
+    /// Every static `^icon_name` this output's format can render: both the
+    /// full and short templates, and every alternative branch within them.
+    ///
+    /// These are the other half of the icon surface. `net` and `speedtest`
+    /// spell their arrows and their ping icon straight into the format
+    /// rather than passing icon values, so they declare no icon placeholder
+    /// for them, and an inspector that only walked [`Self::icon_placeholders`]
+    /// would conclude the block draws no icons at all.
+    #[cfg(test)]
+    pub(crate) fn static_icons(&self) -> Vec<&str> {
+        use crate::formatting::template::{FormatTemplate, Token};
+
+        fn walk<'a>(template: &'a FormatTemplate, found: &mut Vec<&'a str>) {
+            for list in template.0.iter() {
+                for token in &list.0 {
+                    match token {
+                        Token::Icon { name } => {
+                            if !found.contains(&name.as_str()) {
+                                found.push(name);
+                            }
+                        }
+                        Token::Recursive(inner) => walk(inner, found),
+                        Token::Text(_) | Token::Placeholder { .. } => (),
+                    }
+                }
+            }
+        }
+        let mut found = Vec::new();
+        walk(&self.format.full, &mut found);
+        walk(&self.format.short, &mut found);
+        found
+    }
+}
+
+/// One [`OutputPlan`] per format the user configured, so a block that can
+/// rotate through several formats declares each of them. `declare` adds the
+/// icon choices, which are the same whichever format is on screen.
+///
+/// Use this when the block's config has a `MaybeMultiFormatConfig`: the
+/// number of outputs is not known until the config is parsed. A block with a
+/// single `format` has one output and builds it inline instead (see
+/// `docker`). This returns a `Vec` rather than a finished plan because a
+/// block can append outputs that are not format variants (see `net`).
+pub(crate) fn format_outputs(
+    formats: Vec<Format>,
+    declare: impl Fn(OutputPlan) -> OutputPlan,
+) -> Vec<OutputPlan> {
+    formats
+        .into_iter()
+        .enumerate()
+        .map(|(index, format)| declare(OutputPlan::new(format_id(index), format)))
+        .collect()
+}
+
+/// The output id of the `index`th configured format.
+pub(crate) fn format_id(index: usize) -> Cow<'static, str> {
+    match index {
+        0 => "format".into(),
+        n => format!("format{}", n + 1).into(),
+    }
+}
+
+/// Rotating over the outputs [`format_outputs`] produced, mirroring the
+/// bar's own next/prev format actions.
+pub(crate) struct FormatRotation {
+    outputs: Vec<OutputHandle>,
+    index: usize,
+}
+
+impl FormatRotation {
+    /// Collects the plan's format outputs. A block may declare further
+    /// outputs for states that are not format variants (net's `inactive`
+    /// and `missing`); those are left alone.
+    pub(crate) fn new(plan: &Arc<BlockPlan>) -> Result<Self> {
+        let outputs: Vec<OutputHandle> = (0..)
+            .map_while(|index| plan.output(&format_id(index)).ok())
+            .collect();
+        if outputs.is_empty() {
+            return Err(Error::new(format!(
+                "{CONTRACT_BUG}: no format outputs to rotate over"
+            )));
+        }
+        Ok(Self { outputs, index: 0 })
+    }
+
+    pub(crate) fn current(&self) -> &OutputHandle {
+        &self.outputs[self.index]
+    }
+
+    pub(crate) fn next(&mut self) {
+        self.index = (self.index + 1) % self.outputs.len();
+    }
+
+    pub(crate) fn prev(&mut self) {
+        self.index = (self.index + self.outputs.len() - 1) % self.outputs.len();
+    }
+}
+
+/// The full prepared contract of one block instance. The validated fields
+/// are immutable from outside this module (read-only iteration only), so
+/// construction-time uniqueness is a permanent invariant.
+/// Deliberately not `Default`: an empty plan is exactly what [`Self::new`]
+/// rejects, and a derived `Default` would hand every block a way around it.
+#[derive(Debug, Clone)]
+pub(crate) struct BlockPlan {
+    outputs: Vec<OutputPlan>,
+}
+
+impl BlockPlan {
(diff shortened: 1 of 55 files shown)


bim9262:
I started to write this comment on the `doctor` PR, but had forgotten to click comment...

It looks like there's a few ways that the plans can be created which seems a bit confusing. FWIW, I like the feel the docker example provides.

For example: 

**amdgpu**
```rs
pub(crate) fn prepare(config: &Config) -> Result<Arc<BlockPlan>> {
    // Every output renders the same value set, built unconditionally on
    // every update.
    let declare = |output: OutputPlan| output.icon("icon", IconChoices::one("gpu"));
    let formats = config.formats.with_default(" $icon $utilization ")?;
    BlockPlan::new(format_outputs(&formats, declare))
}
```

**docker**
```rs
pub(crate) fn prepare(config: &Config) -> Result<Arc<BlockPlan>> {
    BlockPlan::new(vec![
        OutputPlan::new(
            "main",
            config.format.with_default(" $icon $running.eng(w:1) ")?,
        )
        .icon("icon", IconChoices::one("docker")),
    ])
}
```

cfsmp3:
> It looks like there's a few ways that the plans can be created which seems a bit confusing. FWIW, I like the feel the docker example provides.

Depends on the block - some can generate more than one type of output.

11 blocks have formats: MaybeMultiFormatConfig, so the user configures a list of formats to rotate through and the number of outputs isn't known until the config is parsed. Those use format_outputs. The other 33 have a single format and one output, so they write it inline, which is the docker one.

bim9262:
Seems like the `MultiFormat` isn't really being used anymore and can be removed.

<details>
    <summary>Something like...</summary>

```diff
diff --git a/src/block_plan.rs b/src/block_plan.rs
index 49c06ad54..d6a6e2bd4 100644
--- a/src/block_plan.rs
+++ b/src/block_plan.rs
@@ -133,6 +133,10 @@ impl OutputPlan {
         }
     }
 
+    pub(crate) fn contains_key(&self, key: &str) -> bool {
+        self.format.contains_key(key)
+    }
+
     pub(crate) fn id(&self) -> &str {
         &self.id
     }
@@ -213,7 +217,7 @@ impl OutputPlan {
 /// `docker`). This returns a `Vec` rather than a finished plan because a
 /// block can append outputs that are not format variants (see `net`).
 pub(crate) fn format_outputs(
-    formats: &crate::formatting::MultiFormat,
+    formats: Vec<Format>,
     declare: impl Fn(OutputPlan) -> OutputPlan,
 ) -> Vec<OutputPlan> {
     formats
diff --git a/src/blocks/weather.rs b/src/blocks/weather.rs
index 0c1821fd5..f3348e9fb 100644
--- a/src/blocks/weather.rs
+++ b/src/blocks/weather.rs
@@ -152,7 +152,7 @@ use chrono::{DateTime, Utc};
 use sunrise::{SolarDay, SolarEvent};
 
 use super::prelude::*;
-use crate::formatting::{Format, MultiFormat};
+use crate::formatting::Format;
 pub(super) use crate::geolocator::IPAddressInfo;
 use crate::util::{celsius_to_fahrenheit, kmh_to_mph, kmh_to_mps};
 
@@ -482,9 +482,7 @@ pub(crate) async fn run(config: &Config, api: &CommonApi, plan: &Arc<BlockPlan>)
     let autolocate_interval = config.autolocate_interval.unwrap_or(config.interval).0;
     // The plan holds one output per configured format, so the formats the
     // user can rotate to are exactly the ones a forecast may be needed for.
-    let need_forecast = need_forecast(&MultiFormat::new(
-        plan.outputs().map(|o| o.format().clone()).collect(),
-    ));
+    let need_forecast = need_forecast(&plan.outputs().map(|o| o.format().clone()).collect());
 
     let mut timer = config.interval.timer();
 
@@ -527,7 +525,7 @@ pub(crate) async fn run(config: &Config, api: &CommonApi, plan: &Arc<BlockPlan>)
     }
 }
 
-fn need_forecast(formats: &MultiFormat) -> bool {
+fn need_forecast(formats: &Vec<Format>) -> bool {
     fn has_forecast_key(format: &Format) -> bool {
         macro_rules! format_suffix {
             ($($suffix: literal),* $(,)?) => {
@@ -792,14 +790,16 @@ mod tests {
             "service = { name = \"metno\" }\nformat = [\" $icon $temp \", \" $temp_favg \"]",
         ))
         .unwrap();
-        let formats = MultiFormat::new(plan.outputs().map(|o| o.format().clone()).collect());
-        assert!(need_forecast(&formats));
+        assert!(need_forecast(
+            &plan.outputs().map(|o| o.format().clone()).collect()
+        ));
 
         let plan = prepare(&config(
             "service = { name = \"metno\" }\nformat = [\" $icon $temp \", \" $humidity \"]",
         ))
         .unwrap();
-        let formats = MultiFormat::new(plan.outputs().map(|o| o.format().clone()).collect());
-        assert!(!need_forecast(&formats));
+        assert!(!need_forecast(
+            &plan.outputs().map(|o| o.format().clone()).collect()
+        ));
     }
 }
diff --git a/src/formatting.rs b/src/formatting.rs
index b8bf3bc3e..e90e2a85c 100644
--- a/src/formatting.rs
+++ b/src/formatting.rs
@@ -163,37 +163,6 @@ pub enum FormatError {
     Other(#[from] Error),
 }
 
-#[derive(Debug)]
-pub struct MultiFormat {
-    // The length is always at least one, so we don't need to worry about
-    // dividing by zero when doing the modulo math to calculate the
-    // previous and next formats.
-    formats: Vec<Format>,
-    index: usize,
-}
-
-impl MultiFormat {
-    pub fn new(formats: Vec<Format>) -> Self {
-        Self { formats, index: 0 }
-    }
-
-    pub fn iter(&self) -> std::slice::Iter<'_, Format> {
-        self.formats.iter()
-    }
-
-    pub fn get_format(&self) -> Format {
-        self.formats[self.index].clone()
-    }
-
-    pub fn next_format(&mut self) {
-        self.index = (self.index + 1) % self.formats.len();
-    }
-
-    pub fn prev_format(&mut self) {
-        self.index = (self.index + (self.formats.len() - 1)) % self.formats.len();
-    }
-}
-
 #[derive(Debug, Clone)]
 pub struct Format {
     full: FormatTemplate,
diff --git a/src/formatting/config.rs b/src/formatting/config.rs
index 0489ad752..cdddb5b9f 100644
--- a/src/formatting/config.rs
+++ b/src/formatting/config.rs
@@ -1,4 +1,4 @@
-use super::{Format, MultiFormat, template::FormatTemplate};
+use super::{Format, template::FormatTemplate};
 use crate::errors::*;
 use itertools::Itertools as _;
 use serde::de::{MapAccess, Visitor};
@@ -142,12 +142,12 @@ pub enum MaybeMultiConfig {
 }
 
 impl MaybeMultiConfig {
-    pub fn with_default(&self, default_full: &str) -> Result<MultiFormat> {
+    pub fn with_default(&self, default_full: &str) -> Result<Vec<Format>> {
         self.with_defaults(default_full, "")
     }
 
-    pub fn with_defaults(&self, default_full: &str, default_short: &str) -> Result<MultiFormat> {
-        Ok(MultiFormat::new(match self.clone() {
+    pub fn with_defaults(&self, default_full: &str, default_short: &str) -> Result<Vec<Format>> {
+        Ok(match self.clone() {
             MaybeMultiConfig::Multiple { configs } => configs
                 .into_iter()
                 .enumerate()
@@ -171,28 +171,26 @@ impl MaybeMultiConfig {
                 }
                 formats
             }
-        }))
+        })
     }
 
-    pub fn with_default_formats(&self, default_formats: &[Format]) -> MultiFormat {
-        MultiFormat::new(
-            match self.clone() {
-                MaybeMultiConfig::Multiple { configs } => configs,
-                MaybeMultiConfig::Split { config, config_alt } => {
-                    vec![config.unwrap_or_default(), config_alt.unwrap_or_default()]
-                }
+    pub fn with_default_formats(&self, default_formats: &[Format]) -> Vec<Format> {
+        match self.clone() {
+            MaybeMultiConfig::Multiple { configs } => configs,
+            MaybeMultiConfig::Split { config, config_alt } => {
+                vec![config.unwrap_or_default(), config_alt.unwrap_or_default()]
             }
-            .into_iter()
-            .zip_longest(default_formats)
-            .filter_map(|pair| match pair {
-                itertools::EitherOrBoth::Both(config, default_format) => {
-                    Some(config.with_default_format(default_format))
-                }
-                itertools::EitherOrBoth::Left(config) => Some(config.into()),
-                itertools::EitherOrBoth::Right(_) => None,
-            })
-            .collect(),
-        )
+        }
+        .into_iter()
+        .zip_longest(default_formats)
+        .filter_map(|pair| match pair {
+            itertools::EitherOrBoth::Both(config, default_format) => {
+                Some(config.with_default_format(default_format))
+            }
+            itertools::EitherOrBoth::Left(config) => Some(config.into()),
+            itertools::EitherOrBoth::Right(_) => None,
+        })
+        .collect()
     }
 }
 
```
</details>

cfsmp3:
> Seems like the `MultiFormat` isn't really being used anymore and can be removed.
> 
Thanks,  that makes things a bit cleaner. Done.

bim9262:
It seems like nothing prevents someone from still setting an icon as a value directly, circumventing the `check_declared`. I haven't looked into it, but do you think that it would make sense pass the `OutputHandle` to `Formatter::format` so that the check is actually enfornced?

cfsmp3:
> It seems like nothing prevents someone from still setting an icon as a value directly, circumventing the `check_declared`. I haven't looked into it, but do you think that it would make sense pass the `OutputHandle` to `Formatter::format` so that the check is actually enfornced?

 It is enforced, just not at construction. check_declared is the handle's guard when it mints a value; the enforcement is Widget::check_contract, which CommonApi::set_widget calls on every publish. It walks the widget's values and rejects any icon value the output didn't declare, however it was built. 

See an_icon_built_outside_the_plan_is_still_caught in block_plan.rs.

bim9262:
Since all widgets `check_contract` does it make sense to also `check_declared` by calling the  `OutputHandle`'s icon functions? Seems like we're double checking the same thing.

<details>
    <summary>Also looks like these <code>#[allow(dead_code)]</code> are covered by <code>#[cfg(test)]</code></summary>

```diff
diff --git a/src/block_plan.rs b/src/block_plan.rs
index 83cbf77b1..24b053a46 100644
--- a/src/block_plan.rs
+++ b/src/block_plan.rs
@@ -35,7 +35,6 @@ use std::sync::Arc;
 
 use crate::errors::*;
 use crate::formatting::Format;
-use crate::formatting::template::{FormatTemplate, Token};
 use crate::formatting::value::{Value, ValueInner};
 use crate::widget::Widget;
 
@@ -166,7 +165,7 @@ impl OutputPlan {
     /// nothing in the hot path needs this — it exists because a plan that can
     /// only be queried by a name you already know is not inspectable. The
     /// declaration tests are its consumer today.
-    #[allow(dead_code)]
+    #[cfg(test)]
     pub(crate) fn icon_placeholders(&self) -> impl Iterator<Item = (&'static str, &IconChoices)> {
         self.icons.iter().map(|(p, c)| (*p, c))
     }
@@ -179,8 +178,10 @@ impl OutputPlan {
     /// rather than passing icon values, so they declare no icon placeholder
     /// for them, and an inspector that only walked [`Self::icon_placeholders`]
     /// would conclude the block draws no icons at all.
-    #[allow(dead_code)]
+    #[cfg(test)]
     pub(crate) fn static_icons(&self) -> Vec<&str> {
+        use crate::formatting::template::{FormatTemplate, Token};
+
         fn walk<'a>(template: &'a FormatTemplate, found: &mut Vec<&'a str>) {
             for list in template.token_lists() {
                 for token in &list.0 {
@@ -196,6 +197,7 @@ impl OutputPlan {
                 }
             }
         }
+
         let mut found = Vec::new();
         walk(self.format.full_template(), &mut found);
         walk(self.format.short_template(), &mut found);
@@ -678,16 +680,6 @@ mod tests {
         assert_eq!(widget.contract().unwrap().id(), "connected");
     }
 
-    /// `BlockPlan::new` is the only way to build a plan, and it rejects the
-    /// empty one. A `Default` impl (derived or written) would be a second,
-    /// unchecked constructor producing exactly that — so assert at compile
-    /// time that there isn't one: these two implementations overlap if, and only if,
-    /// `BlockPlan: Default`.
-    #[allow(dead_code)]
-    trait OnlyBuiltByNew {}
-    impl<T: Default> OnlyBuiltByNew for T {}
-    impl OnlyBuiltByNew for BlockPlan {}
-
     #[test]
     fn a_plan_that_declares_nothing_is_rejected() {
         // Otherwise a block could satisfy the framework with an empty plan
diff --git a/src/formatting.rs b/src/formatting.rs
index e90e2a85c..77fa570cf 100644
--- a/src/formatting.rs
+++ b/src/formatting.rs
@@ -189,11 +189,13 @@ impl Format {
 
     /// The compiled full-text template, for inspecting what this format can
     /// render without rendering it (see [`crate::block_plan`]).
+    #[cfg(test)]
     pub(crate) fn full_template(&self) -> &FormatTemplate {
         &self.full
     }
 
     /// The compiled short-text template. See [`Self::full_template`].
+    #[cfg(test)]
     pub(crate) fn short_template(&self) -> &FormatTemplate {
         &self.short
     }
diff --git a/src/formatting/template.rs b/src/formatting/template.rs
index e28df60f2..d540bee67 100644
--- a/src/formatting/template.rs
+++ b/src/formatting/template.rs
@@ -35,6 +35,7 @@ impl FormatTemplate {
     /// The alternative token lists of this template, in the order `render`
     /// tries them. Traversal entry point for inspecting a template without
     /// rendering it (see [`crate::block_plan`]).
+    #[cfg(test)]
     pub(crate) fn token_lists(&self) -> &[TokenList] {
         &self.0
     }
```
</details>

bim9262:
Alternatively `Format::{full,short}}` and `FormatTemplate`'s value  could be `pub(crate)`

cfsmp3:
> Since all widgets `check_contract` does it make sense to also `check_declared` by calling the `OutputHandle`'s icon functions? Seems like we're double checking the same thing.

Thanks!, good feedback. A result from several refactors.
Fixed.

src/block_plan.rs
```diff
@@ -0,0 +1,776 @@
...
+}
+
+/// One output variant of a block: a named state, its effective format, and
+/// the icons each icon-valued placeholder may carry in that state.
+#[derive(Debug, Clone)]
+pub(crate) struct OutputPlan {
+    id: Cow<'static, str>,
+    kind: OutputKind,
+    format: Format,
+    icons: Vec<(&'static str, IconChoices)>,
```
bim9262, line 113: I'm a bit confused by why this is a Vec of placeholders and IconChoices, when IconChoices is what would hold the Vec of icon names.

src/block_plan.rs
```diff
@@ -0,0 +1,776 @@
...
+                format!(
+                    "{CONTRACT_BUG}: output '{}' does not declare exactly one \
+                     icon for '${placeholder}'",
+                    self.id()
+                )
+            })
+    }
+
+    /// The single declared icon of `placeholder`, as a value.
+    pub(crate) fn icon_value(&self, placeholder: &str) -> Result<Value> {
```
bim9262, line 383: I'm not sure that we gain anything by doing this lookup of the icons in every block (that has just the one icon value/placeholder) instead of having them just be hard coded before, every time they output.

bim9262:
If you just want to decrease the chance that the name in the plan matches up with the used value you use a const.

<details>
    <summary>Like...</summary>

```diff
diff --git a/src/blocks/amd_gpu.rs b/src/blocks/amd_gpu.rs
index 0d015fa63..274d3718e 100644
--- a/src/blocks/amd_gpu.rs
+++ b/src/blocks/amd_gpu.rs
@@ -43,6 +43,8 @@ use tokio::fs::read_dir;
 use super::prelude::*;
 use crate::util::read_file;
 
+const ICON_NAME: &str = "gpu";
+
 #[derive(Deserialize, Debug, SmartDefault)]
 #[serde(default)]
 pub struct Config {
@@ -56,7 +58,7 @@ pub struct Config {
 pub(crate) fn prepare(config: &Config) -> Result<Arc<BlockPlan>> {
     // Every output renders the same value set, built unconditionally on
     // every update.
-    let declare = |output: OutputPlan| output.icon("icon", IconChoices::one("gpu"));
+    let declare = |output: OutputPlan| output.icon("icon", IconChoices::one(ICON_NAME));
     let formats = config.formats.with_default(" $icon $utilization ")?;
     BlockPlan::new(format_outputs(formats, declare))
 }
@@ -85,7 +87,7 @@ pub(crate) async fn run(config: &Config, api: &CommonApi, plan: &Arc<BlockPlan>)
         let info = device.read_info().await?;
 
         widget.set_values(map! {
-            "icon" => output.icon_value("icon")?,
+            "icon" => Value::icon(ICON_NAME),
             "utilization" => Value::percents(info.utilization_percents),
             "vram_total" => Value::bytes(info.vram_total_bytes),
             "vram_used" => Value::bytes(info.vram_used_bytes),
```
</details>

cfsmp3:
> Alternatively `Format::{full,short}}` and `FormatTemplate`'s value could be `pub(crate)`

This is better. Implemented.

src/block_plan.rs
cfsmp3: > I'm a bit confused by why this is a Vec of placeholders and IconChoices, when IconChoices is what would hold the Vec of icon names.

IconChoices is the set one placeholder can take; the Vec is one entry per placeholder, and a format usually draws several at once — music has icon, play, next, prev, volume_icon, xrandr icon, brightness_icon, res_icon, kdeconnect icon, bat_icon, network_icon, notif_icon.

The pairing is also what the check uses: check_contract validates each value against the choices for the key it was set under, so "play" => Value::icon("music_next") is a contract bug even though music_next is declared — for next.

If you mean why Vec and not other container, because it respects insert order and for very small amounts of elements, linear scan is actually faster than a Hash.

cfsmp3:
> If you just want to decrease the chance that the name in the plan matches up with the used value you use a const.
> 
I like it, adopted.

cfsmp3:
@bim9262  Anything else here? I'd like to call this complete soon

src/block_plan.rs
```diff
@@ -0,0 +1,780 @@
...
+        // it is always permitted.
+        name.is_empty()
+            || match self {
+                Self::Fixed(names) => names.iter().any(|n| n == name),
+                Self::OpenResolvable => true,
+            }
+    }
+
+    /// The single fixed name, when the set has exactly one element.
+    pub(crate) fn single(&self) -> Option<&Cow<'static, str>> {
```
bim9262: Have a function `one` and `single` does seem a bit confusing.

src/lib.rs
```diff
@@ -191,25 +197,42 @@ impl Block {
         self.set_error_with_restartable(fullscreen, false, error);
     }
 
+    /// Report a contract failure on the framework's own error widget. It is
+    /// an i3status-rs bug in a path that has nowhere to report it — the block
+    /// is already in its error state — so it is loud in the log and fatal in
+    /// debug builds, and the bar carries on in release.
+    fn report_contract_failure(result: Result<()>) {
```
bim9262: This function is only used once. Also 8 lines of comments to explain 3 lines of code is a bit much...

src/blocks/bluetooth.rs
```diff
@@ -51,17 +51,20 @@
...
-//! - `headphones` for bluetooth devices identifying as "audio-card", "audio-headset" or "audio-headphones"
-//! - `joystick` for bluetooth devices identifying as "input-gaming"
-//! - `keyboard` for bluetooth devices identifying as "input-keyboard"
-//! - `mouse` for bluetooth devices identifying as "input-mouse"
-//! - `bluetooth` for all other devices
+//! - `headphones` (`$icon`) for bluetooth devices identifying as "audio-card", "audio-headset" or "audio-headphones"
+//! - `joystick` (`$icon`) for bluetooth devices identifying as "input-gaming"
+//! - `keyboard` (`$icon`) for bluetooth devices identifying as "input-keyboard"
+//! - `mouse` (`$icon`) for bluetooth devices identifying as "input-mouse"
+//! - `bluetooth` (`$icon`, in `disconnected_format`) when the device is unavailable
```
bim9262: This doesn't seem right. `bluetooth` is used as a fallback icon when the device type doesn't match any of the other categories.

src/blocks/battery.rs
```diff
@@ -61,16 +61,18 @@
...
+//! - `bat_charging` (`$icon`, in `charging_format`, as a progression)
+//! - `bat_not_available` (`$icon`, in `missing_format`)
 
 use regex::Regex;
 use std::convert::Infallible;
 use std::str::FromStr;
 
 use super::prelude::*;
 
+const MISSING_ICON: &str = "bat_not_available";
```
bim9262, line 76: Might as well add the `bat` and `bat_charging` icons as const values.

bim9262 (commented):
It seems like the `# Icons Used` sections of each block are not docmmented consistently. Some say what the name of the key is used for the value:

```
`github` (`$icon`)
```

while others don't, or only do if multiple keys are used.

cfsmp3:
@bim9262 I've addressed all your comments.
CI failing is unrelated - see https://github.com/greshake/i3status-rust/pull/2310

cfsmp3:
@bim9262 Any chance this can be merged soon? There's really a lot of PRs waiting for review...

bim9262:
I still had the block changes to review, but had looked at the other files.
It seems like it's not really consistent when a `const` value is used for the icon names.

Perhaps in `icons.rs` we should define all of the keys  of the default icons as keys and then the blocks can all refer to the const names defined in one place.

It makes sense to not use the const value in the tests.

cfsmp3:
> Perhaps in `icons.rs` we should define all of the keys of the default icons as keys and then the blocks can all refer to the const names defined in one place.

OK. Take a look at https://github.com/greshake/i3status-rust/pull/2315

Is that what you had in mind?

bim9262:
Yes, thank you, I just merged that in. Will you rebase this?

cfsmp3:
> Yes, thank you, I just merged that in. Will you rebase this?

Done. Tested it to make sure nothing breaks - LGTM.

ammgws (approved):

cfsmp3:
@ammgws @bim9262 could one of you merge?

Issue #1956, opened 2023-10-24
nvidia_gpu block resumes GPU from suspend on Optimus devices


GaspardCulis:
### System info

> CPU : AMD Ryzen 9 5900HS with Radeon Graphics
> GPU : NVIDIA GeForce RTX 3070 Mobile
> iGPU : AMD ATI Radeon Vega Series

### Issue

On Laptops with both an iGPU and a dedicated GPU, the dedicated GPU can be suspended to lower power draw,  and gets resumed when high performance is needed ([PRIME - Arch Wiki](https://wiki.archlinux.org/title/PRIME#PRIME_render_offload)).

I can query the GPU state with this command:
```bash
cat /sys/bus/pci/devices/0000:01:00.0/power/runtime_status
```

I noticed that the nvidia_gpu block resumes my dedicated GPU from its suspended state.

I suggest that a new block option should get added that checks if the GPU is suspended and avoids resuming it.

GaspardCulis:
Just displaying something like this instead of waking up the GPU only to get its stats:
![image](https://github.com/greshake/i3status-rust/assets/79080828/dd4013f4-436f-482a-99a2-3ffb08e192e8)

<details>
<summary>Block config</summary>

```toml
[[block]]
block = "custom"
command = "cat /sys/bus/pci/devices/0000:01:00.0/power/runtime_status"
interval = 2
format = " \uf26c   $text "
```

</details>

And when you detect that it's active, just display the stuff normally.

I tried to achieve this behavior with a special toml config but couldn't get it working as intended without just doing the whole thing (check for power state, and then retrieve all of the info) in a script.

MaxVerevkin:
Ah, okay, this sounds good. Maybe we can add `suspended_format` to customize the text, or even hide the block.

>`cat /sys/bus/pci/devices/0000:01:00.0/power/runtime_status`

Do you know a generic way to get this path?

Also, I don't have an nvidia gpu to play with, so would you be interested in implementing this?

GaspardCulis:
This `000:01:00.0` part of the path refers to the PCI slot of the NVIDIA GPU retrieved with `lspci` (`[domain:]bus:device.function`). I'll have to find a way to find this reliably. 
I have some rust projects but I'm still a beginner. I'll take a look at the code in a few days and tell you if I'm confident in implementing this.

GaspardCulis:
[...]

Issue #1354, opened 2021-11-20
Running xrandr command wakes up the dGPU


Abijeet:
The example command for [the toggle block](https://github.com/greshake/i3status-rust/blob/master/doc/blocks.md#toggle) mentions using the `xrandr` command to determine if external displays are connected. 

The problem with running the `xrandr` command again and again is that on a laptop, that [wakes up the dedicated GPU](https://gitlab.freedesktop.org/drm/amd/-/issues/1775). I think it might make sense to add a warning in the documentation about that.

ammgws:
If `cat /sys/class/drm/*HDMI*/dpms | grep On` does the exact same thing as `"xrandr | grep DP1\\ connected\\ 38 | grep -v eDP1"` then we should probably change the docs to recommend that method instead.

PR #2307, opened 2026-08-29
nvidia_gpu: warn that nvidia-smi wakes the dedicated GPU

### LINKED ISSUE #1880: Nvidia block needs a warning about `nvidia-smi`
https://github.com/greshake/i3status-rust/blob/86213c7d1b808db043cfbf51cce74b133caff0e1/src/blocks/nvidia_gpu.rs#L11

Just an FYI, it is worth putting a very visible warning somewhere in docs about `nvidia-smi` use on hybrid systems. Every time `nvidia-smi` is called it will wake the dGPU from sleep causing excess power draw.

An alternative would be to use the (limited) dbus interface of my [supergfxctl](https://gitlab.com/asus-linux/supergfxctl/). As I don't use i3 or similar myself I don't know if I will contribute this as a block - perhaps i can find someone in my discord to assist.

YatoVoid:
The nvidia_gpu block runs `nvidia-smi` on every update. On laptops with switchable graphics that call powers on the dedicated GPU and keeps it awake, which noticeably increases power draw.

Add a note to the block docs pointing this out and suggesting a larger `interval` (or not using the block) when running on battery.

Closes #1880


ammgws (approved):
I thought we had documented this already but maybe I am thinking of something else

LGTM

ammgws:
CI failure is unrelated

ammgws:
> I thought we had documented this already but maybe I am thinking of something else
> 
> LGTM

Ah I was thinking of xrandr (e.g. https://github.com/greshake/i3status-rust/issues/1354)

ammgws:
Actually it was this https://github.com/greshake/i3status-rust/issues/1956

PR #2310, opened 2026-09-07
Fix chunks_exact_to_as_chunks clippy lint

cfsmp3:
clippy 1.98 added chunks_exact_to_as_chunks, which fires on the constant chunk size in wlr_toplevel_management. CI runs clippy with -D warnings, so every build on the new toolchain fails, master included.

as_chunks::<4>() yields &[u8; 4] directly, so the per-item try_into().unwrap() goes away too.


cfsmp3:
@bim9262 This is needed now - otherwise all PRs fail.

PR #2312, opened 2026-09-13
create theme based on Sway's default workspace colors

realchoodle:
I looked at `sway/swaybar/config.c` to create a theme that matches `swaybar`'s default colors. I think it looks pretty nice, but one issue is the size of the separator `\ue0b2`; it does not seem to align with the rest of the text. Although, I think this might be an issue with `swaybar`, not `i3status-rust`.


bim9262:
Instead of pulling the original sway values out of the sway source files, I'd be interested in combining the mapping work you've done with https://github.com/greshake/i3status-rust/pull/2154. That PR pulls in the values from the actual configuration of sway.

PR #2313, opened 2026-09-15
Add QBZ doc for music block

Maxou56800:


bim9262:
This is a non-exhaustive list, but qbz is shutting down, so I don't really want to add documentation about it here.

